PR #10, opened 2023-07-17
Fix bad error message (bad content length)

joereuss12:
The error message 'bad content length' was given when the local copy of a file is larger than the remote server copy when downloading a file. The error message now reads 'Local copy of file is larger than remote copy' which I believe is much clearer. In addition, fixed the bug stating the error is retryable. It no longer states this type of error is retryable. This fixed issue #52 in htcondor/osdf-client

### DIFF
--- a/handle_http.go
+++ b/handle_http.go
@@ -478,6 +478,10 @@ func DownloadHTTP(transfer TransferDetails, dest string, token string) (int64, e
 	// Check the error real quick
 	if resp.IsComplete() {
 		if err := resp.Err(); err != nil {
+			if errors.Is(err, grab.ErrBadLength) {
+				grab.ErrBadLength = errors.New("Local copy of file is larger than remote copy")
+				err = grab.ErrBadLength
+			}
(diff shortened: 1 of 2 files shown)


handle_http.go
```diff
@@ -478,6 +478,10 @@ func DownloadHTTP(transfer TransferDetails, dest string, token string) (int64, e
 	// Check the error real quick
 	if resp.IsComplete() {
 		if err := resp.Err(); err != nil {
+			if errors.Is(err, grab.ErrBadLength) {
+				grab.ErrBadLength = errors.New("Local copy of file is larger than remote copy")
```
bbockelm, line 482: Why this assignment?  Doesn't seem necessary to override the text in the other module itself... can you keep the type but provide only `err` with the new message?

joereuss12: Fixed with new commit!

bbockelm (approved):
LGTM.

I think the lint failure is unrelated -- will override and merge.

PR #13, opened 2023-07-24
Finish implementation of the init login flow

bbockelm:
This implements the admin flow; the admin inputs a code printed out in the terminal into the browser to login then the user (now admin) can reset their password.


CannonLock (approved):
@bbockelm Logically looks good to me and it compiles. I am not to the point with GO experience that I can do a thorough code review but I want to start hacking on this so I am going to merge it.

PR #14, opened 2023-07-24
Add Namespace Management CLI Commands

CohenQU:
This commit adds several new commands to the CLI for managing namespaces. The commands added include:

- Register a new namespace
- Delete a namespace
- List all namespaces
- Get a specific namespace This commit also adds the necessary flag definitions and configurations for each command.

In addition, the commit includes the implementation of JWT-based authentication when registering a namespace. This includes handling of JSON Web Key Sets (JWKS), signing/verification of payloads, nonce generation, and HTTP request/response handling.

The code also includes functions for loading and saving signatures to/from a file, and for loading ECDSA public/private keys from PEM and JWKS formats.


cmd/namespace_registry.go
```diff
@@ -0,0 +1,253 @@
...
+	keyInBytes, err := ioutil.ReadFile(publicKeyPath)
+	if err != nil {
+		return nil, err
+	}
+	var jwks JWKS
+	err = json.Unmarshal(keyInBytes, &jwks)
+	if err != nil {
+		return nil, err
+	}
+	keyData := jwks.Keys[0]  // Assumes there's at least one key
```
bbockelm, line 58: Similar to the private key comment, I think that we should use the loader here: https://github.com/PelicanPlatform/pelican/blob/main/config/init_server_creds.go#L204

In general, let's aim to use the JWT library instead of rolling our own.

cmd/namespace_registry.go
```diff
@@ -0,0 +1,253 @@
...
+func signPayload(payload []byte, privateKey *ecdsa.PrivateKey) ([]byte, error) {
+    hash := sha256.Sum256(payload)
+    signature, err := privateKey.Sign(rand.Reader, hash[:], crypto.SHA256)  // Use crypto.SHA256 instead of the hash[:]
+    if err != nil {
+        return nil, err
+    }
+    return signature, nil
+}
+
+/*
```
bbockelm, line 81: Remove commented-out code.

cmd/namespace_registry.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		panic(err)
```
bbockelm, line 126: Avoid panic'ing within the CLI -- let's instead propagate the error upward through a return variable.

cmd/namespace_registry.go
```diff
@@ -0,0 +1,253 @@
...
+		respData = resp_to_json(resp)
+
+		if respData["status"] == "APPROVED" {
+			done = true
+		} else {
+			fmt.Printf("Waiting for approval...\n")
+			reader := bufio.NewReader(os.Stdin)
+			fmt.Print("Press Enter after verification")
+			_, _ = reader.ReadString('\n')
+			// time.Sleep(60 * time.Second)
```
bbockelm, line 171: Remove commented-out code.

cmd/namespace_registry.go
```diff
@@ -0,0 +1,253 @@
...
+	signature, err := signPayload([]byte(clientPayload), privateKey)
+	if err != nil {
+		panic(err)
+	}
+
+	// Create data for the second POST request
+	data2 := map[string]interface{}{
+		"client_nonce":      client_nonce,
+		"server_nonce":      respData["server_nonce"],
+		"pubkey":        	 map[string]string{
```
bbockelm, line 220: May be possible to replace this with a JWT library call as well.

cmd/root.go
```diff
@@ -91,4 +92,4 @@ func initConfig() {
 	if viper.GetBool("Debug") {
 		setLogging(log.DebugLevel)
 	}
-}
+}
```
bbockelm, line 95: Remove spurious whitespace change.

bbockelm (changes requested):
A few cleanup items requested; comments are inline.

The big picture item is to switch to using the JWT library that's already used in the client.  See:

https://github.com/PelicanPlatform/pelican/blob/main/origin_ui/origin.go#L148-L165

That should actually decrease the size of the PR quite a bit.

cmd/namespace_registry.go
```diff
@@ -0,0 +1,253 @@
...
+
+type JWK struct {
+	Crv string `json:"crv"`
+	Kid string `json:"kid"`
+	Kty string `json:"kty"`
+	X   string `json:"x"`
+	Y   string `json:"y"`
+}
+
+func loadPrivateKey(privateKeyPath string) (*ecdsa.PrivateKey, error) {
```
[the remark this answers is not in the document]
CohenQU: ✅

cmd/namespace_registry.go
CohenQU: ✅

cmd/namespace_registry.go
CohenQU: ✅

cmd/namespace_registry.go
CohenQU: ✅

cmd/root.go
CohenQU: ✅

cmd/namespace_registry.go
CohenQU: ✅

bbockelm (approved):
A few review comments - but given we're short on time I'll merge anyway:
1.  Better to unmarshal into `struct`s instead of `map[string]interface{}`; that will ensure the expected keys are present.
2. Elsewhere in the code, we've been using CamelCase instead of underscores.
3. I suspect there's a better way to handle the JWKS serialization besides converting it to a map.  A follow-up might be to dig through the `jwk` library to see if there's some serialization routines we can reuse from there.

PR #16, opened 2023-07-26
Have UI always use TLS

bbockelm:
- Factor out the engine launching code into a common `web_ui` module.
- Fix the generation of the host certificate to use a key compatible with Chrome (issuer key is unchanged).
- Change the Gin engine startup to always run with TLS.


jhiemstrawisc (approved):

jhiemstrawisc:
LGTM

PR #18, opened 2023-07-27
Have origin advertise to the director service

bbockelm:
This draft is the first piece of the periodic advertisement of the origin to the director service.  When complete, it'll contain:
- New APIs in the director for advertising services
- Periodic go routine in the origin code to invoke new APIs
- Bearer token generation / verification for authorizing the advertisement.

Goal is that all origins connecting to the director advertise their capabilities directly (and verified by the namespace registration service) instead of relying on OSG Topology to bootstrap the namespace.

### DIFF
--- a/cmd/director_serve.go
+++ b/cmd/director_serve.go
@@ -5,6 +5,7 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/pelicanplatform/pelican/config"
 	"github.com/pelicanplatform/pelican/director"
 	"github.com/pelicanplatform/pelican/web_ui"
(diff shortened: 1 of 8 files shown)


bbockelm:
Going to self-merge this one in order to get the federation auto-discovery pieces in place.  This will help improve the namespace URL discovery (necessary for a separate cleanup of the namespace registration).

The current state of the branch is the origin periodically advertises and the director responds on its URL -- however, the origin still needs to _use_ the token it generates as part of the API call.  Therefore, all the attempts to advertise end in an "Authorization denied".

PR #19, opened 2023-07-28
Various Director cleanups

bbockelm:
- Check HTTP response code from topology and fail as appropriate.
- Improve error messages coming from OSDF failures.  Was helpful in understanding a local misconfiguration.
- Switch to the common Gin configuration.  Enables prometheus and shares startup code with the origin.
- Add simple signal handling code so a Ctrl+C will cause the director to exit.  Useful for CLI debugging.

### DIFF
--- a/cmd/director_serve.go
+++ b/cmd/director_serve.go
@@ -1,11 +1,14 @@
 package main
 
 import (
-	"github.com/gin-gonic/gin"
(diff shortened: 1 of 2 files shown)


bbockelm:
Bypassing branch protections -- want to get this in so I can test the new director.

PR #28, opened 2023-07-31
Allow passing key directly via CL

jhiemstrawisc:
For the kubernetes deployment of the Director, it would be more convenient to pass the MaxMind key as an environment variable via the sealedSecret mechanism than trying to get the value safely into a configMap.


bbockelm (approved):
LGTM

PR #29, opened 2023-07-31
Dockerize

jhiemstrawisc:
Two related pieces in this PR:

First, the contents in the new `/images` are what I've cooked up for building a container that enters in the `entrypoint.sh` script. The idea is that a kubernetes deployment can launch the pod with an argument like "director_serve" and this will cause the entrypoint script to launch the `pelican_director_serve` daemon through supervisord on the specified port. I haven't added supervisord configurations for the namespace registry or the origin because I thought those would be best added by the people most familiar with them. 

Second, the new build+push action will trigger whenever there's a push to main or a tag that matches v[0-9]+.[0-9]+.[0-9]+. This will cause goreleaser to run and generate its release artifacts. In particular, the release binary artifact for linux/amd64 will be used later to build the container (at some point it would be good to think about other architectures, but let's walk before we run). The action should then push the container to harbor with a tag that's either "latest" in the event that the action was triggered by a regular push, or a tag matching the commit's tag if the commit has a tag. I thought this might be a nice way to handle keeping tagged stable versions and development versions of the container organized in harbor.

### DIFF
--- a/.github/workflows/publish-container.yml
+++ b/.github/workflows/publish-container.yml
@@ -1,0 +1,141 @@
+
+name: Release, Build, and Push
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+    branches:
+      - main
+  repository_dispatch:
+    types:
+      - dispatch-build
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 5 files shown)


bbockelm (approved):
Changes LGTM -- appears to be mostly based on prior OSG containers.

We may have to do a bit of follow-up post-merge to see if the image building needs tweaked.  I'm suspicious of the copying of the output binary into the `images/` directory; I agree we'll need to do something different (perhaps setup symlinks to the build directories and copy from the build directories) for other architectures.

That said, going to merge this to help move forward as @CohenQU was looking for a container image for his testing.

PR #31, opened 2023-08-01
Dockerize

jhiemstrawisc:
This *should* fix a few issues that were causing the build+push action to fail.


jhiemstrawisc:
Bypassing to test -- no real code updates, just a few minor docker build context fixes

PR #36, opened 2023-08-02
Add cl flag to make default endpoint serve caches or origins

jhiemstrawisc:
We need a way to redirect xrootd caches to origins, but as of the time of this PR, setting:
`pss.origin <director-domain>:443/api/v1.0/director/origin`
results in xrootd crashing after it's unable to parse the URL. 

This PR adds a command line option to set whether the default behavior is to serve caches or origins. Now a user can run:
`pelican director serve -p 443 -e <origin/cache>`
and the director will select the appropriate default endpoint redirection middleware.

As far as xrootd caches go, this should make it possible to set the `pss.origin` correctly

### DIFF
--- a/cmd/director.go
+++ b/cmd/director.go
@@ -41,4 +41,10 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
+
+	directorServeCmd.Flags().StringP("default-endpoint", "e", "", "Set whether the default endpoint should redirect to caches or origins")
+	err = viper.BindPFlag("DirectorDefaultEndpoint", directorServeCmd.Flags().Lookup("default-endpoint"))
+	if err != nil {
+		panic(err)
+	}
 }
--- a/cmd/director.go
+++ b/cmd/director.go
@@ -42,8 +42,8 @@ func init() {
 		panic(err)
 	}
 
-	directorServeCmd.Flags().StringP("default-endpoint", "e", "", "Set whether the default endpoint should redirect to caches or origins")
-	err = viper.BindPFlag("DirectorDefaultEndpoint", directorServeCmd.Flags().Lookup("default-endpoint"))
+	directorServeCmd.Flags().StringP("default-response", "", "", "Set whether the default endpoint should redirect clients to caches or origins")
+	err = viper.BindPFlag("Director.DefaultResponse", directorServeCmd.Flags().Lookup("default-response"))
 	if err != nil {
 		panic(err)
 	}
--- a/cmd/director_serve.go
+++ b/cmd/director_serve.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"crypto/elliptic"
+	"errors"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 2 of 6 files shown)


cmd/director.go
```diff
@@ -41,4 +41,10 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
+
+	directorServeCmd.Flags().StringP("default-endpoint", "e", "", "Set whether the default endpoint should redirect to caches or origins")
```
bbockelm, line 45: From the user perspective, it's not really changing the "default endpoint" but rather the default response.
```suggestion
	directorServeCmd.Flags().StringP("default-response", "", "", "Set whether the default endpoint should redirect clients to caches or origins")
```

cmd/director.go
```diff
@@ -41,4 +41,10 @@ func init() {
...
+	err = viper.BindPFlag("DirectorDefaultEndpoint", directorServeCmd.Flags().Lookup("default-endpoint"))
```
bbockelm, line 46: Let's use the hierarchical capabilities of viper:
```suggestion
	err = viper.BindPFlag("Director.DefaultResponse", directorServeCmd.Flags().Lookup("default-endpoint"))
```

cmd/director_serve.go
```diff
@@ -52,9 +53,14 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
...
 	}
 
-	// Use the shortcut middleware so that GET /foo/bar
-	// acts the same as GET /api/v1.0/director/object/foo/bar
-	engine.Use(director.ShortcutMiddleware())
+	// Configure the shortcut middleware to either redirect to a cache
+	// or to an origin
+	defaultEndpoint := viper.GetString("DirectorDefaultEndpoint")
+	if !(defaultEndpoint == "cache" || defaultEndpoint == "origin" || defaultEndpoint == "") {
+		return errors.New("The director's default endpoint must either be set to 'cache' or 'origin'." +
```
bbockelm, line 60: Use `fmt.Error` to create a formatted error including what string the user provided.

images/supervisord/pelican_director_serve.conf
```diff
@@ -1,5 +1,5 @@
 [program:pelican_director_serve]
-command=/pelican/osdf-client director serve -p %(ENV_OSDF_DIRECTOR_PORT)s
+command=/pelican/osdf-client director serve -p %(ENV_OSDF_DIRECTOR_PORT)s -e %(ENV_OSDF_DIRECTOR_DEFAULT_ENDPOINT)s
```
bbockelm, line 2: What happens if the environment variable is not set?

bbockelm (changes requested):
A few minor changes requested.

Instead of having the empty string for the default response (`""`) indicate the default, can you set the default in `config/resources/default.yaml`?  Probably will look like this:

```
Director:
  DefaultResponse: cache
```

bbockelm (changes requested):
Few changes requested.  See inline comments.

PR #39, opened 2023-08-14
Draft integration of namespace registry app in Pelican

jhiemstrawisc:
This is a draft for integrating Yuxiao's namespace server into the main pelican framework, with hookups to the CLI and the ability to serve the namespace registration app directly through the Pelican binary itself. There are a few important things to note about this as of the time of opening the draft PR:

1. I made a few choices about changing names of a few files that had "namespace" in the title, but we may choose to revert these changes. Personally, I think that calling the namespace server "namespace" is a bit confusing, mostly because we can register a namespace, but the thing we serve is not the namespace, but the registry itself. I tried to reflect this distinction by using the term registry.
2. When Yuxiao handed this over to me, he mentioned limitations regarding manipulation of namespace prefixes that had preceding or internal /'s. I started looking into how I could resolve these limitations, but it seems like there are limitations within gin's zero-allocation router itself that make this difficult. We could probably implement a custom regex middleware router, but that seems contrary to why we're using Gin. I was hoping to start a conversation about this before settling on a solution.
3. There's currently no security included with the command for deleting namespaces, and that's obviously not good. I was also hoping to discuss this before working on it


bbockelm:
For this item:

> When Yuxiao handed this over to me, he mentioned limitations regarding manipulation of namespace prefixes that had preceding or internal /'s

What do you mean by this?  It seems like we should always normalize the incoming data with something like the following:

```
input := "/foo//bar"
path.Clean("/" + input)
```

That way, we ensure that we only get absolute paths (`foo/bar` doesn't make sense here).

jhiemstrawisc:
The problem is in routing to handler functions. From what I can tell, Gin can take two different types of wild cards for setting up routing groups, and they are `/cli-namespaces/:prefix/suffix` and `/cli-namespaces/*wildcard`. In the first case, `prefix` will only be read up until the the first `/` after it starts, so there's no good way to route a request to `/cli-namespaces/my/namespace/issuer.jwks`  because it won't match `/cli-namespaces/:prefix/issuer.jwks`. For that path, we'd need the router to be `/cli-namespaces/:prefix1/:prefix2/issuer.jwks`.

Another headache with this is that currently when the client tries to delete the namespace `/my/namespace`, it sends a DELETE request for the resource `/cli-namespaces//my/namespace` and Gin cannot match that route by using the `:prefix` style wildcard because of the double slash. 

On the other hand, the `*wildcard` grabs everything after that point, so `/cli-namespaces/my/namespace` would match `/cli-namespaces/*wildcard`, but that prevents us from declaring any route after `/cli-namespaces/` because it will conflict with the wildcard. 

One solution might be to use some middleware that grabs everything after `/cli-namespaces` and determines the appropriate route, but then we lose some of the benefits from Gin.

cmd/namespace_registry.go
```diff
@@ -1,206 +1,49 @@
+/*
```
bbockelm, line 1: Remove the copyright statement for now.

IIUC, the copyright goes to Morgridge... to do this in bulk, I created a separate issue on the topic: https://github.com/PelicanPlatform/pelican/issues/42

cmd/namespace_registry_serve.go
```diff
@@ -0,0 +1,54 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	// "github.com/pelicanplatform/pelican/config"
```
bbockelm, line 9: Remove commented-out lines

cmd/namespace_registry_serve.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+
+func serveNamespaceRegistry( /*cmd*/ *cobra.Command /*args*/, []string) error {
+	log.Info("Initializing the namespace registry's database...")
+
+	// Initialize the registry's sqlite database
+	err := nsregistry.InitializeDB()
+	if err != nil {
+		return fmt.Errorf("Unable to initialize the namespace registry database: %q", err)
```
bbockelm, line 24: Switch to `errors.Wrapf` to wrap the error instead of converting the underlying error to a string.

cmd/namespace_registry_serve.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	// function defined in director_serve
+	err = generateTLSCertIfNeeded()
+	if err != nil {
+		return fmt.Errorf("Failed to generate TLS certificate: %q", err)
```
bbockelm, line 30: Switch to `errors.Wrapf` to wrap the error instead of converting the underlying error to a string.

config/config.go
```diff
@@ -231,6 +231,9 @@ func InitServer() error {
 		viper.SetDefault("OriginUI.PasswordFile", "/etc/pelican/origin-ui-passwd")
 		viper.SetDefault("XrootdMultiuser", true)
 		viper.SetDefault("GeoIPLocation", "/var/cache/pelican/maxmind/GeoLite2-City.mmdb")
+		viper.SetDefault("NSRegistryLocation", "/var/lib/pelican/ns-registry/registry.sqlite")
+		viper.SetDefault("OIDCClientIDFile", "/etc/pelican/ns-registry/client-id")
```
bbockelm, line 235: Can we avoid making a subdir here?  Maybe just a longer filename?

config/config.go
```diff
@@ -245,6 +248,10 @@ func InitServer() error {
 		viper.SetDefault("MacaroonsKeyFile", filepath.Join(configBase, "macaroons-secret"))
 		viper.SetDefault("IssuerKey", filepath.Join(configBase, "issuer.jwk"))
 		viper.SetDefault("OriginUI.PasswordFile", filepath.Join(configBase, "origin-ui-passwd"))
+		viper.SetDefault("GeoIPLocation", filepath.Join(configBase, "maxmind", "GeoLite2-City.mmdb"))
```
bbockelm, line 251: This is overwritten about 4 lines later.  Which is preferred?

namespace-registry/client_commands.go
```diff
@@ -0,0 +1,188 @@
...
+	if viper.GetBool("TLSSkipVerify") {
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		client = &http.Client{Transport: tr}
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil,err
```
bbockelm, line 40: Add space:
```suggestion
		return nil, err
```

namespace-registry/client_commands.go
```diff
@@ -0,0 +1,188 @@
...
+
+func NamespaceRegisterWithIdentity(privateKeyPath string, namespaceRegistryEndpoint string, prefix string) (error) {
+	data := map[string]interface{}{
+		"identity_required": "true",
+	}
+	resp, err := makeRequest(namespaceRegistryEndpoint, "POST", data)
+	if err != nil {
+		return fmt.Errorf("Failed to make request: %s\n", err)
+	}
+	respData := respToJSON(resp)
```
bbockelm, line 66: Switch to unmarshaling from a struct.  That way we can avoid some of the boilerplate type safety issues.

namespace-registry/client_commands.go
```diff
@@ -0,0 +1,188 @@
...
+
+	done := false
+	for !done {
+		data = map[string]interface{}{
+			"identity_required": "true",
+			"device_code": deviceCode,
+		}
+		resp, err = makeRequest(namespaceRegistryEndpoint, "POST", data)
+		if err != nil {
+			return fmt.Errorf("Failed to make request: %s\n", err)
```
bbockelm, line 79: Wrap error.

namespace-registry/client_commands.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+		respData = respToJSON(resp)
+		if respData["status"] == "APPROVED" {
+			done = true
+		} else {
+			fmt.Printf("Waiting for approval. Press Enter after verification.\n")
```
bbockelm, line 85: If there's no formatting to be done, I think you want `fmt.Println` instead?

namespace-registry/client_commands.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	accessToken := respData["access_token"]
+	return NamespaceRegister(privateKeyPath, namespaceRegistryEndpoint, accessToken, prefix)
+}
+
+func NamespaceRegister(privateKeyPath string, namespaceRegistryEndpoint string, accessToken string, prefix string) (error) {
+	publicKey, err := config.LoadPublicKey("", privateKeyPath)
+	if err != nil {
+		return fmt.Errorf("Failed to retrieve public key: %w\n", err)
```
bbockelm, line 97: Across this function, switch from formatting errors to wrapping them.

namespace-registry/client_commands.go
```diff
@@ -0,0 +1,188 @@
...
+	clientPayload := clientNonce + respData["server_nonce"]
+
+	// Sign the payload
+	signature, err := signPayload([]byte(clientPayload), privateKey)
+	if err != nil {
+		return fmt.Errorf("Failed to sign payload: %w\n", err)
+	}
+
+	// Create data for the second POST request
+	xBytes, _ := base64.RawURLEncoding.DecodeString(jwks["x"])
```
bbockelm, line 136: Handle errors here and the next line.

namespace-registry/client_commands.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil,err
+	}
+	defer resp.Body.Close()
+
+	body, _ := ioutil.ReadAll(resp.Body)
+	return body, nil
```
bbockelm, line 45: This function doesn't check the HTTP status code, meaning a `500`- or `403`-type response is going to be marked as successful.

namespace-registry/registry.go
```diff
@@ -0,0 +1,411 @@
...
+
+type TokenResponse struct {
+	AccessToken string `json:"access_token"`
+}
+
+/*
+Various auxiliary functions used for client-server security handshakes
+*/
+func keySignChallenge(c *gin.Context, data map[string]interface{}, action string) {
+	_, cnOk := data["client_nonce"].(string)
```
bbockelm, line 57: This boilerplate could all be avoided with a struct-based deserialization.

namespace-registry/registry.go
```diff
@@ -0,0 +1,411 @@
...
+	ctx.JSON(http.StatusOK, gin.H{
+        "server_nonce": serverNonce,
+        "client_nonce": clientNonce,
+        "server_payload": hex.EncodeToString(serverPayload),
+        "server_signature": hex.EncodeToString(serverSignature),
+    })
+}
+
+func jwksToEcdsaPublicKey(jwks map[string]interface{}) *ecdsa.PublicKey {
+	x := jwks["x"].(string)
```
bbockelm, line 143: Add back in the error handling.

namespace-registry/registry.go
```diff
@@ -0,0 +1,411 @@
...
+        } 
+    } else {
+        ctx.JSON(http.StatusMultipleChoices, gin.H{"status": "Key Sign Challenge FAILED"})
+    }
+}
+
+/*
+Handler functions called upon by the gin router
+*/
+func cliRegisterNamespace(c *gin.Context) {
```
bbockelm, line 196: Rename from `c` to `ctx`; using `c` makes it hard to search for the variable in text editors.

namespace-registry/registry.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	device_code := requestData["device_code"]
+
+	OIDCClientIDFile := viper.GetString("OIDCClientIDFile")
+	OIDCClientIDFromEnv := viper.GetString("OIDCCLIENTID")
+	if OIDCClientIDFile != "" {
+		contents, err := os.ReadFile(OIDCClientIDFile)
+		if err != nil {
+			log.Errorln(err)
```
bbockelm, line 240: Missing an error return and response.

namespace-registry/registry.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+		OIDCClientID = strings.TrimSpace(string(contents))
+	} else if OIDCClientIDFromEnv != "" {
+		OIDCClientID = OIDCClientIDFromEnv
+	} else {
+		log.Errorln("An OIDC Client Identity file must be specified in the config (OIDCClientIDFile), or the identity must be provided via the environment variable PELICAN_OIDCCLIENTID")
```
bbockelm, line 246: Missing an error return and response.

namespace-registry/registry.go
```diff
@@ -0,0 +1,411 @@
...
+
+func getOpenIDConfiguration(c *gin.Context) {
+	prefix := c.Param("prefix")
+	c.JSON(http.StatusOK, gin.H{"status": "getOpenIDConfiguration is not implemented", "prefix": prefix})
+}
+*/
+
+func RegisterNamespaceRegistry(router *gin.RouterGroup) {
+	// Establish various routes to be used by the namespace registry
+	router.POST("/cli-namespaces/registry", cliRegisterNamespace)
```
bbockelm, line 406: Let's dump the `cli-` prefix.  Affix this to the common API prefix used by other components.

bbockelm (changes requested):
Lots of minor refactoring requested -- probably simply small cleanups from the rush to get it handed off when Yuxiao left.

bbockelm (approved):
Ok - still a few rough spots here but the major ones are all recorded as separate issues.  To avoid this PR from becoming any larger, I'm going to merge now.

PR #41, opened 2023-08-18
Add http/s3 plugin to repo

jhiemstrawisc:
This moves the HTTP/S3 plugins into Pelican and installs them in the container. I haven't yet looked into how pelican serves origins with XRootD, so there's no hookup to the command line to serve an http or s3 origin yet.

### DIFF
--- a/.github/workflows/publish-container.yml
+++ b/.github/workflows/publish-container.yml
@@ -134,6 +134,7 @@ jobs:
       - name: Build and push Docker images
         uses: docker/build-push-action@v4
         with:
-          context: ./images
+          context: ./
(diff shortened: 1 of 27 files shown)


bbockelm:
Hey @jhiemstrawisc -

I think I'm going to NACK this one.  While it makes a ton of sense to have everything in the same repo for go-related items (after all, it compiles down to a single binary and we aim for the same release schedule), it doesn't make much logical sense to have an XRootD plugin in the same repo.

What does make sense is having the plugin as a standalone repo in the same `PelicanPlatform` organization.  Can we do that instead?

PR #48, opened 2023-08-24
Improve the use of the X-Pelican-* responses

bbockelm:
This:

1.  Converts the client to look for `X-Pelican-*` instead of `X-Osdf-*`.
2. Has the director always set `X-Pelican-Namespace`, even when the credential generation isn't known.
3. Avoids a panic when the `X-Pelican-Namespace` header is missing.
4. Slightly improves the debugging experience when a panic is raised.

### DIFF
--- a/director.go
+++ b/director.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	namespaces "github.com/pelicanplatform/pelican/namespaces"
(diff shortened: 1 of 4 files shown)


bbockelm:
Merging this without review -- right now, focusing on getting most of the PRs in to allow a reformat.

PR #49, opened 2023-08-25
Enable `go fmt` as one of the linters

bbockelm:
Go has a reference source code formatting - and the VS code defaults seem to like auto-formatting to it.  We might as well enable it in the linters just to keep things consistent.

Note this disables reporting all linter issues as there's a codebase-wide reformat that should be done as a separate branch once some other pending work lands.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -16,23 +16,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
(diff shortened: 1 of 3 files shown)


bbockelm:
@jhiemstrawisc - did some investigation into linters last night and found out that `go fmt` is simply an optional linter we can enable in the existing check.

I'd like to run a cross-codebase `go fmt -s -w ./...` once #39 is merged (hopefully today?).  When that's done, we can revert the second change here which causes the linter to only flag new problems during PRs.

bbockelm:
Merging just to lock this approach in place.

PR #52, opened 2023-08-29
Updated Pelican release.yaml

joereuss12:
Updated for the upcoming Pelican release. Only change needed was to upgrade Go version from 1.18 to 1.19. This GHA includes RPM and the binaries within the release tag. Below is a list of all the files within the release when a git tag is created when tested in my throw away repo:
release

Let me know if any more changes need to be made!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.18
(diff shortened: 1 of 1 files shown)


bbockelm:
@joereuss12 - please review the merge failure!  I think a different recently merged PR made the same change.

Since it's minimal, please rebase and squash once it's done.

bbockelm:
Actually, looking more closely, everything here is already incorporated -- can just go ahead and close it off!

PR #58, opened 2023-08-30
Fix YAML error in go version for GHA

bbockelm:


bbockelm:
Merging directly again -- only way to test it out is merge against main, unfortunately.

PR #59, opened 2023-08-30
Reformat code base with `go fmt`

bbockelm:
While we're in a lull of PRs, let's rip off the bandaid and get things following the standard formatting.


bbockelm:
Self-merging; only contains formatting changes.

PR #65, opened 2023-08-31
Create a single `--port` flag object

### LINKED ISSUE #63: WebPort not being set correctly by the Director
There's a command line option to set what port the director is served on: `pelican director serve -p <port>`. Under the hood, this binds the command line int to the viper variable `WebPort` that is used to run the gin engine in `web_ui/ui.go`. This was working properly at one point, but there's been some drift and now it seems like whatever value is passed for port via the command line is overwritten by the value specified in `config/resources/defaults.yaml`.

bbockelm:
When a flag is bound to the viper key (such as binding the `--port` flag object to the `WebPort` key), it's a global.  If you do this multiple times with multiple pointers, only the last one is used.

That means, if we invoke `cmd.Flags().StringP(...)` in multiple sub-commands, only one will actually set `WebPort`.

The solution for this is to create the flag once globally, then add the global object to multiple commands.  Note that I didn't use a parent-level "persistent flag" as that would also make `--port` usage strings appear for the `object` sub-command (which doesn't have an embedded webserver).

Fixes #63.

### DIFF
--- a/cmd/director.go
+++ b/cmd/director.go
@@ -36,14 +36,10 @@ func init() {
 	directorCmd.AddCommand(directorServeCmd)
 
 	// Set up flags for the command
-	directorServeCmd.Flags().StringP("port", "p", "", "Set the port to which the Director's redirect services should be bound")
-	err := viper.BindPFlag("WebPort", directorServeCmd.Flags().Lookup("port"))
-	if err != nil {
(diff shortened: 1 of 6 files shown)


jhiemstrawisc (approved):
Reviewed and tested. This does indeed fix #63.

PR #66, opened 2023-08-31
Upgrade remainder of the code to lestrrat-go v2

### LINKED ISSUE #61: Replace any usage of "github.com/lestrrat-go/jwx/jw(k,a,t)" packages with their V2 equivalent
I noticed that in `cmd/origin_api` we're using "github.com/lestrrat-go/jwx/jw(k,a,t)" to deal with jwk, jwa, and jwt objects. Elsewhere we use v2 of this library. We should replace any usage of the old library with the new.

bbockelm:
The library we use for JWKs and JWTs has both a v1 and v2 API.  As the various Pelican pieces were independently developed, some modules used V1 and others V2.  This brings up all the code to use V2, dropping V1 as a dependency.

Fixes #61.

### DIFF
--- a/config/init_server_creds.go
+++ b/config/init_server_creds.go
@@ -85,7 +85,9 @@ func LoadPublicKey(existingJWKS string, issuerKeyFile string) (*jwk.Set, error) 
 	if err != nil {
 		return nil, errors.Wrap(err, "Failed to assign key ID to public key")
 	}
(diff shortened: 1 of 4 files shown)


jhiemstrawisc (approved):
LGTM. Merging so I can incorporate some of this into one of the other issues I'm working on.

PR #75, opened 2023-08-31
Fix small bug in ns-registry auto creation

### LINKED ISSUE #67: Auto-create the NS registry database
Starting on a completely new setup currently has the following failure:

```
$ pelican registry serve -d
INFO[2023-08-31T07:59:26-05:00] Initializing the namespace registry's database... 
Error: Unable to initialize the namespace registry database: Error checking NSRegistryLocation filepath information: stat /home/bbockelm/.config/pelican/ns-registry.sqlite: no such file or directory
```

jhiemstrawisc:
We talked this morning about config variables that specify "Location" and how they should specify the actual filename. I had previously tried to implement some logic that if a file location was passed such as `/foo/bar/`, the namespace registry would auto create the registry database as `/foo/bar/ns-registry.sqlite`. Instead of fixing the bug in that logic, let's just say whatever is passed should be considered the file name. A side effect of this is that if the user configures `NSRegistryLocation = /foo/bar/`, the file will be created as `/foo/bar/.sqlite` because the code automatically adds a file extension when none is provided and in this case, the actual filename is empty.

This fixes #67

### DIFF
--- a/namespace-registry/registry-db.go
+++ b/namespace-registry/registry-db.go
@@ -141,19 +141,10 @@ func InitializeDB() error {
 		return errors.Wrap(err, "Failed to create directory for namespace registry database")
 	}
 
-	fileInfo, err := os.Stat(dbPath)
-	if err != nil {
-		return errors.Wrap(err, "Error checking NSRegistryLocation filepath information")
-	}
-
-	// Check if it's a directory
-	if fileInfo.IsDir() { // if directory, we add the filename to it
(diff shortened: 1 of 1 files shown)


bbockelm (approved):
I don't particularly love the "magic addition of the .sqlite extension" here.  If the user wants to call it foo, it should be foo -- don't surprise them with foo.sqlite.

That said, my gripe isn't part of this PR... so approving.  Just wanted to share the "philosophy" here for the future.

PR #76, opened 2023-08-31
Use lestrrat jwks instead of custom structs in NS Registry

### LINKED ISSUE #54: Reformat namespace registry app to use 3rd party JWKS package
There are some custom structs floating around in the namespace registry app that try to handle various jwk sets. We should instead replace these with the 3rd party library that's used elsewhere throughout the code base. This will also fix the fact that the namespace registry app currently stores keys that are malformed.

jhiemstrawisc:
There were missing key values in the registry database (like kty, kid, etc) because of the custom structs being used to create the jwks that got stored. By using lestrrat's jwk package, we get those values for free.

One thing to note about the PR is that while the registration pubkey verification dance works when using the client in its current form, this won't work for a jwks with multiple keys in it. I didn't try to tackle that problem in this commit because I'm not sure where those other key pairs are coming from, so I wasn't sure how load them into the key set. 

Fixes #54

### DIFF
--- a/namespace-registry/client_commands.go
+++ b/namespace-registry/client_commands.go
@@ -2,6 +2,7 @@ package nsregistry
 
 import (
 	"crypto/tls"
+
 	"github.com/pkg/errors"
 
 	"bufio"
@@ -117,13 +118,24 @@ func NamespaceRegister(privateKeyPath string, namespaceRegistryEndpoint string, 
 		return errors.Wrap(err, "Failed to retrieve public key")
 	}
 
-	if log.GetLevel() == log.DebugLevel || log.GetLevel() == log.TraceLevel {
+	/*
+	 * TODO: For now, we only allow namespace registration to occur with a single key, but
+	 *       at some point we should expose an API for adding additional pubkeys to each
+	 *       namespace. There is a similar TODO listed in registry.go, as the choices made
+	 *       there mirror the choices made here.
+	 * To enforce that we're only trying to register one key, we check the length here
+	 */
+	if (*publicKey).Len() > 1 {
+		return errors.Errorf("Only one public key can be registered in this step, but %d were provided\n", (*publicKey).Len())
+	}
+
+	if log.IsLevelEnabled(log.DebugLevel) {
 		// Let's check that we can convert to JSON and get the right thing...
 		jsonbuf, err := json.Marshal(publicKey)
 		if err != nil {
 			return errors.Wrap(err, "failed to marshal the public key into JWKS JSON")
 		}
-		log.Debugf("Constructed JWKS from loading public key: %s\n", jsonbuf)
+		log.Debugln("Constructed JWKS from loading public key:", string(jsonbuf))
 	}
 
 	privateKey, err := config.LoadPrivateKey(privateKeyPath)
--- a/namespace-registry/client_commands.go
+++ b/namespace-registry/client_commands.go
@@ -6,16 +6,15 @@ import (
 
 	"bufio"
 	"bytes"
-	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
-	"math/big"
 	"net/http"
 	"os"
 
 	"github.com/pelicanplatform/pelican/config"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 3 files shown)


namespace-registry/client_commands.go
```diff
@@ -118,9 +117,13 @@ func NamespaceRegister(privateKeyPath string, namespaceRegistryEndpoint string,
 		return errors.Wrap(err, "Failed to retrieve public key")
 	}
 
-	jwks, err := config.JWKSMap(publicKey)
-	if err != nil {
-		return errors.Wrap(err, "Failed to convert public key to JWKS")
+	if log.GetLevel() == log.DebugLevel || log.GetLevel() == log.TraceLevel {
```
bbockelm, line 120:
```suggestion
	if log.IsLevelEnabled(log.DebugLevel) {
```

namespace-registry/client_commands.go
```diff
@@ -118,9 +117,13 @@ func NamespaceRegister(privateKeyPath string, namespaceRegistryEndpoint string,
 		return errors.Wrap(err, "Failed to retrieve public key")
 	}
 
-	jwks, err := config.JWKSMap(publicKey)
```
bbockelm, line 121: Can we also remove `config.JWKSMap` at this point?

namespace-registry/client_commands.go
```diff
@@ -118,9 +117,13 @@ func NamespaceRegister(privateKeyPath string, namespaceRegistryEndpoint string,
...
-	if err != nil {
-		return errors.Wrap(err, "Failed to convert public key to JWKS")
+	if log.GetLevel() == log.DebugLevel || log.GetLevel() == log.TraceLevel {
+		// Let's check that we can convert to JSON and get the right thing...
+		jsonbuf, err := json.Marshal(publicKey)
+		if err != nil {
+			return errors.Wrap(err, "failed to marshal the public key into JWKS JSON")
+		}
+		log.Debugf("Constructed JWKS from loading public key: %s\n", jsonbuf)
```
bbockelm, line 126:
```suggestion
		log.Debugln("Constructed JWKS from loading public key:", jsonbuf)
```

namespace-registry/registry.go
```diff
@@ -231,83 +232,78 @@ func keySignChallengeInit(ctx *gin.Context, data *registrationData) error {
...
+	clientJwks, err := jwk.Parse(data.Pubkey)
+	if err != nil {
+		return errors.Wrap(err, "Couldn't parse the pubkey from the client")
 	}
 
-	clientPubkey := &ecdsa.PublicKey{
-		Curve: elliptic.P521(),
-		X:     xBigInt,
-		Y:     yBigInt,
+	if log.GetLevel() == log.DebugLevel || log.GetLevel() == log.TraceLevel {
```
bbockelm, line 242:
```suggestion
	if log.IsLevelEnabled(log.DebugLevel) {
```

namespace-registry/registry.go
```diff
@@ -231,83 +232,78 @@ func keySignChallengeInit(ctx *gin.Context, data *registrationData) error {
...
+		// Let's check that we can convert to JSON and get the right thing...
+		jsonbuf, err := json.Marshal(clientJwks)
+		if err != nil {
+			return errors.Wrap(err, "failed to marshal the client's keyset into JSON")
+		}
+		log.Debugf("Client JWKS as seen by the registry server: %s\n", jsonbuf)
```
bbockelm, line 248:
```suggestion
		log.Debugln("Client JWKS as seen by the registry server:", jsonbuf)
```

namespace-registry/registry.go
```diff
@@ -231,83 +232,78 @@ func keySignChallengeInit(ctx *gin.Context, data *registrationData) error {
...
 	}
 
-	return clientPubkey, nil
-}
+	// Now iterate through the jwks and verify signatures:
```
bbockelm, line 251: Why are we iterating here?  I assume it's because, if there are multiple public keys in the submitted JWKS, any one of them could have been used for the challenge, right?  If that's right, can you add a comment along those lines?

That somewhat defeats the purpose of only registering keys we can prove we have access to, right?  That means either:
1.  We should only allow the addition of one public key at a time, OR
2. We should extend the challenge to allow multiple signatures.

I think I lean toward (1).  We will need a way to add new public keys to (or remove public keys from) the namespace.  However, I think this is clearly a separate issue.  Once you expand the comment, can you please also create a new issue for adding/removing public keys from the namespace?

bbockelm (changes requested):
A few quite minor items -- otherwise, close to going in!

bbockelm (approved):

PR #78, opened 2023-08-31
Remove CVMFS shortcut within Pelican client

### LINKED ISSUE #62: Remove CVMFS shortcut in code
We have a special case in the Pelican code for the `/osgconnect` prefix in the OSDF, here:

https://github.com/PelicanPlatform/pelican/blob/main/main.go#L511C38-L511C50

When that prefix is used, we first try to read the file out of CVMFS.

Recently, we've been retiring this part of the namespace in favor of `/ospool`.  The storage backing that origin is being unmounted -- seems like a good time to remove this special case (and any other cleanout -- potentially, `handle_cvmfs.go` can also be removed at that point).

joereuss12:
Removing all instances of this shortcut since CVMFS is moving to be https instead. Gutted out handle_cvmfs.go as well as any other instances. Did some light tests by hand and /osgconnect transfers seem to still work without it.

Fixes #62


bbockelm (approved):
LGTM.

We may need to accept "cvmfs" as a valid flag in the CLI for backward compatibility-- but I'm somewhat dubious anyone ever set that.  So, let's keep as is and only do the backward compat if we see it's needed.

PR #79, opened 2023-09-01
Include token in redirect

### LINKED ISSUE #44: Director should keep token authorization in the URL
Currently, the director redirects clients to a cache by constructing a URL from the selected cache's hostname/port and the file path.

However, if the incoming request has a token embedded in the URL (`?authz=XYZABCD`), then we currently drop the URL parameter and, as a result, the client gets an authorization failure at the cache.

When constructing a redirect, we should keep the URL query parameters as part of the resulting URL.

bbockelm:
If a client presents a token in the original query, we should reflect the token into the redirect as well.  This allows HTTP clients to authenticate at the next cache or origin (as the HTTP spec mandates dropping the `Authorization` header on redirect).

Fixes #44

### DIFF
--- a/director/redirect.go
+++ b/director/redirect.go
@@ -2,6 +2,7 @@ package director
 
 import (
 	"fmt"
+	"net/http"
 	"net/netip"
 	"net/url"
 	"path"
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
Confirmed by hand that the change picks up both header and url query tokens and appends them to the redirect.

I agree with the choice of only adding the token to the redirect and not the link header urls.

Issue #55, opened 2023-08-29
Add authentication for namespace deletion


jhiemstrawisc:
There's currently no authentication required to delete a namespace, which is bad. Once issue #54 is finished, we should add authentication to this. The client should provide a signed jwt with a scope like "pelican.namespace_delete" that the namespace registry app can verify using its stored jwks for that origin.

PR #89, opened 2023-09-01
Add deletion authorization

jhiemstrawisc:
One thing I'd like close review of is the fact that I don't validate the token at the registry against a whitelist of issuers because I'm not sure where such a whitelist would come from. AFAIK, the issuer isn't something that gets stored in the pubkeys that are looked up at the registry.


bbockelm:
Fixes #55

bbockelm (approved):
Did some rebasing and hand merging.  While testing, I hit an annoyance with the `--namespace-url` parameter and included a commit to fix that as well.

Tests well by hand.

PR #92, opened 2023-09-02
Fix director responses

### LINKED ISSUE #68: Investigate X-Pelican-* header response from Director
@jhiemstrawisc reports the latest Director code is not providing the correct `X-Pelican-*` headers that would allow the client to generate the token.  Investigate and see what's missing.

bbockelm:
This fixes up the header generation for the director response.  Now, when I try the command from @jhiemstrawisc in #68, I get the following in the response headers:

```
...
< location: https://dtn-pas.cinc.nrp.internet2.edu:8443/ospool/PROTECTED/justin.hiemstra
< x-pelican-authorization: issuer=https://osg-htc.org/ospool
< x-pelican-namespace: namespace=/ospool/PROTECTED, require-token=true
< x-pelican-token-generation: issuer=https://osg-htc.org/ospool, max-scope-depth=4, strategy=OAuth2
```

Fixes #68

### DIFF
--- a/director.go
+++ b/director.go
@@ -63,7 +63,7 @@ func CreateNsFromDirectorResp(dirResp *http.Response, namespace *namespaces.Name
 	}
 	xPelicanNamespace := HeaderParser(pelicanNamespaceHdr[0])
 	namespace.Path = xPelicanNamespace["namespace"]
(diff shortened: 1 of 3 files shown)


jhiemstrawisc (approved):
Ah, good catch with the originNS --> nsAd bug.

PR #93, opened 2023-09-02
Add pre-commit hooks for the repo

### LINKED ISSUE #83: Add pre-commit.com workflow to repo
pre-commit.com has support for at least `go fmt` and `golangci-linter`.  This would allow commits to be linted on the client side (if the developer has installed the pre-commit.com hooks), reducing the number of linter GHA failures and "fix linter issues"-style commit message.

bbockelm:
This introduces pre-commit configurations to run golangci before any commits occur.  Goal is to decrease the number of "fixup linter" type commit messages in the repo.

Fixes #83

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,14 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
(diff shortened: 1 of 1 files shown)


jhiemstrawisc:
@bbockelm I didn't see anyone assigned for review on this, but it would be a nice feature to have at this point. I tested and made a small change so that our `parameters.yaml` doesn't cause pre-commit to fail. I'm going to go ahead and merge, since we aren't enforcing anything yet and running pre-commit is still optional (but recommended). I'll also follow up with a PR that runs pre-commit against the whole repo so that everything gets tidied at least once.

PR #94, opened 2023-09-02
Add simple verification to namespace registration

### LINKED ISSUE #53: Restrict some paths from being registered as namespace prefixes
We should disallow the registration of some namespace prefixes for security reasons. These include but may not be limited to `/api`, paths containing special characters for unix filepaths like `..` and `~/`, and perhaps special unix directory names like `/var`, `/etc`, and others.

bbockelm:
This forbids a few potentially-confusing prefixes, including:
- Any path component starting with '.'
- The root prefix '/'
- /pelican, /api, /view

Fixes #53


jhiemstrawisc (approved):
LGTM. I think the choice to move the check for existing namespaces and the prefix validation into the `verified` block is probably a good choice.

PR #95, opened 2023-09-02
Remove default user

### LINKED ISSUE #90: Fix published container so that supervisor can run as user `xrootd`
Deployments on Tiger quit serving the director. After investigating, it looks like the switch to default to user `xrootd` in the published container is causing permissions issues with `supervisord`. When the container tries to launch the entrypoint script, `supervisord -c /etc/supervisord.conf` fails because it sets the log (and other outputs) to places in `/var` that the user doesn't have permissions to write to.

I'm not sure which approach we should take -- a) get `supervisord` running as root, b) have it run as the `xrootd` user and adjust file permissions so that it can write to `/var/log/supervisord` or c) run `supervisord` with user `xrootd` and have those misc files written to someplace the user already has permissions to write to.

bbockelm:
supervisord startup was failing when this was set ... let's revert for now and revisit dropping permissions in the future.

Fixes #90

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,12 +1,12 @@
 #
 # Copyright (C) 2023, Pelican Project, Morgridge Institute for Research
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License"); you
 # may not use this file except in compliance with the License.  You may
 # obtain a copy of the License at
-# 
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
LGTM

PR #103, opened 2023-09-05
Using common transport for https based writes

joereuss12:
This commit creates one common create transport function for https based writes, rather than having duplicating code. In addition, the timeouts for these transports are created in defaults.yaml so they can be changed easily. 
Other changes include:
1. Adding TLSSkipVerify to client's director query, which is a useful feature for testing for us when using a running director locally.
2. Found a bug in namespaces.go, where NamespaceURL was used when TopologyNamespaceURL should be used instead.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -394,7 +394,7 @@ func InitClient() error {
 		}
 	}
 	for _, prefix := range prefixes {
-		if val, isSet := os.LookupEnv(prefix + "_TOPOLOGYNAMESPACEURL"); isSet {
+		if val, isSet := os.LookupEnv(prefix + "_TOPOLOGY_NAMESPACE_URL"); isSet {
 			viper.Set("TopologyNamespaceURL", val)
 			break
 		}
--- a/config/config.go
+++ b/config/config.go
@@ -393,6 +393,12 @@ func InitClient() error {
 			break
 		}
 	}
+	for _, prefix := range prefixes {
+		if val, isSet := os.LookupEnv(prefix + "_TOPOLOGYNAMESPACEURL"); isSet {
+			viper.Set("TopologyNamespaceURL", val)
+			break
+		}
+	}
 
 	// Check the environment variable STASHCP_MINIMUM_DOWNLOAD_SPEED (and all the prefix variants)
 	var downloadLimit int64 = 1024 * 100
--- a/handle_http.go
+++ b/handle_http.go
@@ -20,6 +20,7 @@ package pelican
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io"
@@ -36,7 +37,6 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
-	"crypto/tls"
 
 	grab "github.com/cavaliercoder/grab"
 	log "github.com/sirupsen/logrus"
@@ -829,8 +829,7 @@ Loop:
 
 }
 
-//transport = getTransport()
-var UploadClient = &http.Client{Transport: transport} //Global: Used by handle_http_test.go but nothing else
+var UploadClient = &http.Client{Transport: getTransport()}
 
 // Actually perform the Put request to the server
 func doPut(request *http.Request, responseChan chan<- *http.Response, errorChan chan<- error) {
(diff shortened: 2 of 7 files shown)


bbockelm:
@joereuss12 - can you take a look at the linter failures?

Also, the `TLSSkipVerify` stuff should go into the common transport setup method, not adding another conditional block.

config/config.go
```diff
@@ -393,6 +393,12 @@ func InitClient() error {
 			break
 		}
 	}
+	for _, prefix := range prefixes {
+		if val, isSet := os.LookupEnv(prefix + "_TOPOLOGYNAMESPACEURL"); isSet {
```
bbockelm, line 397: Let's follow the pattern above - prefix + `_TOPOLOGY_NAMESPACE_URL`:
```suggestion
		if val, isSet := os.LookupEnv(prefix + "_TOPOLOGY_NAMESPACE_URL"); isSet {
```

handle_http.go
```diff
@@ -438,23 +438,39 @@ func parseTransferStatus(status string) (int, string) {
...
+	idleConnTimeout := viper.GetDuration("Transport.IdleConnTimeout")
+	transportTLSHandshakeTimeout := viper.GetDuration("Transport.TLSHandshakeTimeout")
+	expectContinueTimeout := viper.GetDuration("Transport.ExpectContinueTimeout")
+	responseHeaderTimeout := viper.GetDuration("Transport.ResponseHeaderTimeout")
+
+	transportDialerTimeout := viper.GetDuration("Transport.Dialer.Timeout")
+	transportKeepAlive := viper.GetDuration("Transport.Dialer.KeepAlive")
+
+	//Set up the transport
+	transport := http.Transport {
```
bbockelm, line 453: From the documentation page for transports:

> Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used.

So let's create this as a global (using `sync.Once`; see https://github.com/PelicanPlatform/pelican/blob/main/namespace-registry/registry.go#L126-L134) and return a pointer to the global.

handle_http.go
```diff
@@ -438,23 +438,39 @@ func parseTransferStatus(status string) (int, string) {
...
+			Timeout: transportDialerTimeout,
+			KeepAlive: transportKeepAlive,
+		}).DialContext,
+		MaxIdleConns: maxIdleConns,
+		IdleConnTimeout: idleConnTimeout,
+		TLSHandshakeTimeout: transportTLSHandshakeTimeout,
+		ExpectContinueTimeout: expectContinueTimeout,
+		ResponseHeaderTimeout: responseHeaderTimeout,
+	}
+	return transport
```
bbockelm, line 465: From the linter failure, you need to return a pointer to the common transport object as a `sync.Mutex` object isn't copyable.

bbockelm (changes requested):
A few changes requested (mostly elaborating on the quick comment I had left earlier).

config/config.go
joereuss12: Fixed (making note for myself)

main_test.go
```diff
@@ -60,8 +60,8 @@ func TestGetIps(t *testing.T) {
 func TestGetToken(t *testing.T) {
 
 	// Need a namespace for token acquisition
-	defer os.Unsetenv("PELICAN_NAMESPACE_URL")
-	os.Setenv("PELICAN_NAMESPACE_URL", "https://topology.opensciencegrid.org/osdf/namespaces")
+	defer os.Unsetenv("PELICAN_TOPOLOGYNAMESPACEURL")
+	os.Setenv("PELICAN_TOPOLOGYNAMESPACEURL", "https://topology.opensciencegrid.org/osdf/namespaces")
```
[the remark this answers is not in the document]
joereuss12: Fixed

handle_http.go
joereuss12: Fixed

config/config.go
bbockelm: Yeah -- for anything new, we should use definitely use `viper.Get(...)` as we import the environment variables as well.

This is mostly for backward compatibility: `AutomaticEnv` only works with a single prefix and for namespaces we've supported a list of prefixes.

In other words, we shouldn't create any more of these -- but I'm ok with this one for now.

bbockelm (approved):

PR #104, opened 2023-09-05
Create a privkey for registry

jhiemstrawisc:
All of my local testing disguised the fact that the registry didn't create its own signing key when it's started (I just happened to have one in the right spot from previous testing).

The changes I've made will create the private key `issuer.jwk` in the default location for `IssuerKey`, and this is primarily used by the registry to handle the key signing challenge on namespace registration. Tested by hand and it works locally.

### DIFF
--- a/cmd/namespace_registry_serve.go
+++ b/cmd/namespace_registry_serve.go
@@ -19,15 +19,18 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


bbockelm (approved):
LGTM!

PR #105, opened 2023-09-05
UI updates

CannonLock:
Adds Copyright info to the Typescript pages
Redirects root to /view
Adds some basic graphs to the initial view on login with xrootd data

### DIFF
--- a/origin_ui/origin.go
+++ b/origin_ui/origin.go
@@ -118,11 +118,11 @@ func WaitUntilLogin() error {
 			fmt.Printf("\033[A\033[A\033[A\033[A")
 			fmt.Printf("\033[2K\n")
 			fmt.Printf("\033[2K\rPelican admin interface is not initialized\n\033[2KTo initialize, "+
-				"login at \033[1;34mhttps://%v:%v\033[0m with the following code:\n",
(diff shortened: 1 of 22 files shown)


bbockelm (approved):
@CannonLock - going to approve this just to get things going.

However, there's two issues:
1. We need to do a better job handling the "no data" use case.  Default to zero, spinning beachball, something better than the placeholder now.
2. We don't appear to require authorization for the splash page itself... if you go to some part of the UI _besides_ the code initialization page, it'll happily serve up the regular UI.  Instead, we should require the password to be set (via redirect to the code page) or require users to login.
   - Additionally, we should add a "logout" page.

I'm OK with both being done in a follow-up PR, however.

bbockelm:
Oh!  And forgot to mention: I did a quick rebase to fix up the lack of line endings that your editor is generating.  Please review settings for whatever editor you use ... those shouldn't occur!

CannonLock:
@bbockelm I iterated past both of those options in favor of the text which I found more informative. 

Screenshot 2023-09-06 at 8 16 25 AM (2)

CannonLock:
How do you feel about this?

PR #106, opened 2023-09-06
UI changes

CannonLock:
- Adds a redirect for non auth'd users
- Refines the UI in the first 10 minutes where data is not yet populated
- 
Screenshot 2023-09-06 at 8 34 59 AM (2)


CannonLock:
@bbockelm It would be sweet if the prometheus metrics got moved behind auth if they are going to be private. If those metric endpoints are meant to be public why not let a user see the homepage without auth?

bbockelm:
Yes - the Prometheus metrics are going to be behind the authorization long term.  It's one of the tickets for @turetske in v7.1.0.

bbockelm (approved):
Much better!  Merging.

PR #115, opened 2023-09-07
Fix bug with common transport TLSSkipVerify

joereuss12:
I found a bug with setting up TLSSkipVerify from the common transports. Looks like I put checking for that within the creation of the transport that only happens once. Therefore, if the transport is created, then it will never check again if TLSSkipVerify is set. From running my own tests by hand it seems to be working now.


joereuss12:
I set both of you as reviewers since it's just a one line change. Should be working properly now so whoever can get to it first.

jhiemstrawisc (approved):
LGTM

PR #121, opened 2023-09-07
Send token and namespace ad to director

jhiemstrawisc:
I'm leaving this in draft mode for the moment because I haven't gotten through the whole process yet, and I suspect there are still bugs.

The part I'm getting stuck on is in `director/origin_api.go`, line 115 in my changed files where we do `ar.Get(ctx, issuer_url)`. It looks like under the hood, `ar.Get` instantiates an `http.Client` and tries to get the `issuer.jwks` from the `issuer_url`. Of course, testing locally I don't have a valid TLS setup, and I can't figure out how to bypass that issue through the jwk library because the x509 certificate is obviously not signed by a known authority.

Otherwise, a few bugs I think I've nailed out:

1. The origin now sends a token to the director for verification (before no token was sent)
2. There were issues with the ttlcache mutex in `director/origin_api.go`, both in terms of the types of lock being set and an improperly scoped `defer` statement.
3. The origin now sends a slice of namespace ads to the director, with its own namespace ad as the only item. This should be reviewed because I'm hard coding things in the ad like "RequireToken", "MaxScopeDepth", "Strategy", and "BasePath". Currently I'm unsure of where these variables should be coming from or whether there are config variables that can be obtained through viper

### DIFF
--- a/origin_ui/advertise.go
+++ b/origin_ui/advertise.go
@@ -78,13 +78,12 @@ func AdvertiseOrigin() error {
 		Path:          prefix,
 		Issuer:        *namespaceUrl,
 		MaxScopeDepth: 3,
-		Strategy:     "OAuth2",
+		Strategy:      "OAuth2",
 		BasePath:      "/",
 	}
 	ad := director.OriginAdvertise{
 		Name:       name,
 		URL:        originUrl,
-		// Namespaces: make([]director.NamespaceAd, 0),
 		Namespaces: []director.NamespaceAd{nsAd},
 	}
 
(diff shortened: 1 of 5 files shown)


jhiemstrawisc:
Okay, this should be ready for a review @bbockelm. Two things to point out that can either be fixed as a part of this PR or another:

1. In `origin_ui/advertise.go`, there's a TODO that talks about generating an origin url and that code being in a branch I presume you were waiting to merge. I assume we don't to hardcode the origin url as localhost
2. I still have `RequireToken` set to false for the origin's namespace ad. I'm not sure how to determine what that setting should be based on the configuration variables I know of.

origin_ui/advertise.go
```diff
@@ -60,10 +60,31 @@ func AdvertiseOrigin() error {
...
+	if namespaceUrl.String() == "" {
+		return errors.New("No NamespaceUrl is set")
+	}
+
+	prefix := viper.GetString("NamespacePrefix")
+
+	// TODO: Need to figure out where to get some of these values
+	// 		 so that they aren't hardcoded...
+	nsAd := director.NamespaceAd{
+		RequireToken:  false,
```
bbockelm, line 77:
```suggestion
		RequireToken:  true,
```

even though these are mostly placeholders before we improve it next version, I want to default to authorization...

bbockelm (approved):
Approving to get things going forward...

One stylistic comment: I don't like big blocks of commented out code.  Either actually provide commentary of why it needs to be there or remove it (we have git history, we can bring it back when/if it's actually needed!).

PR #124, opened 2023-09-11
Remove hardcoded originUrl

### LINKED ISSUE #123: Send token and namespace ad to director fixes
Leftover TODOs for this PR (#121) that went into 7.0.0

1. In origin_ui/advertise.go, there's a TODO that talks about generating an origin url and that code being in a branch I presume you were waiting to merge. I assume we don't to hardcode the origin url as localhost
2. Still have RequireToken set to false for the origin's namespace ad. I'm not sure how to determine what that setting should be based on the configuration variables I know of.

jhiemstrawisc:
This commit removes the hardcoded origin url and instead makes the value configurable via the pelican config file. Specifying an origin url is required to serve an origin, and the command `pelican origin serve` will fail unless the value is set. It will also fail if the configured origin url specifies a port that doesn't match the port passed via the `-p` flag.

Closes #123

### DIFF
--- a/cmd/origin_serve.go
+++ b/cmd/origin_serve.go
@@ -27,6 +27,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


joereuss12 (approved):
LGTM

Issue #81, opened 2023-09-01
Test director registration


bbockelm:
Add a unit test for the director registration API.

Test should:
- Create keypair and corresponding and token allowing registration access.
- Inject the keypair into the internal key cache.  This may require some refactoring of the key cache to be an interface so we can replace the lookup into the namespace registry with the pubkey newly-created for the test.
- Invoke the registration endpoint, check for resulting HTTP status code.
- Do the above step again with an invalid token.
- Query the director for an object within the new origin's namespace.  Ensure the response includes the newly registered origin.

turetske:
@bbockelm I'm working on this now, but I realized this isn't quite clear (at least to a newbie) on what exact API functions is this supposed to be testing? Which module is the director registration API? I just want to make sure I don't bypass what I'm actually supposed to be writing the test for in the interest in making the test as small and focused as possible.

PR #126, opened 2023-09-12
Added redirect_test.go

turetske:
This is response to the issue [here](https://github.com/PelicanPlatform/pelican/issues/81) - note that after discussion with Brian, we determined the last bullet point (Query the director for an object within the new origin's namespace. Ensure the response includes the newly registered origin) should be a separate issue and is not covered in this pull request.

* Created redirect_test.go to test code in direct/redirect.go
* Tests the director registration with both a valid and invalid token
* Also added an interface called NameSpaceCache in order to mock the cache in tests in origina_api.go

### DIFF
--- a/director/origin_api.go
+++ b/director/origin_api.go
@@ -47,8 +47,15 @@ type (
 	}
 )
 
+// Create interface
+// Add it to namespacekeys in place of jwk.cache
+type NamespaceCache interface {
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (approved):
LGTM. I like the way you set this up!

PR #128, opened 2023-09-13
Makefile updates

CannonLock:
- Specify website input so only rebuilds on update
- Remove alternative docker build commands
  - The assumption is you are running in the appropriate env

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,27 +14,29 @@
 # limitations under the License.
 #
 
-CONTAINER_TOOL := docker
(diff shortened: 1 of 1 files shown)


jhiemstrawisc:
I'm certainly not a makefile guru. Outside of testing that I can still run `make` with the various arguments, are there other things I should look at testing or reviewing?

jhiemstrawisc (approved):
I tested each of the targets, and they all appeared to run and complete as expected.

PR #134, opened 2023-09-14
Added code coverage to github actions workflow

turetske:
- will comment on a pull request with current test coverage
	- only runs on the ubuntu-latest test to avoid duplicates
	- uses a third-party tool to add a message to a pull request called sticky-pull-request-comment whose source is located at https://github.com/marocchino/sticky-pull-request-comment

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,7 @@
 on: [push, pull_request]
 name: Test
+permissions:
+  pull-requests: write
 jobs:
   test:
     strategy:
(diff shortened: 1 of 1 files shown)


CannonLock:
@turetske LGTM.

A food for thought for @bbockelm is that this statistic might find a better home on the homepage with a [badge](https://github.com/marketplace/actions/coverage-badge). Since the number isn't really related to the PR at hand I am not sure how impactful it will be on a decision to merge.

bbockelm:
@CannonLock - I'm OK keeping it on GHA for now.  Let's see where we are at the end of the month on the metric.

@turetske - is it possible to show the before / after for the coverage?

turetske:
@bbockelm 

We could, there isn't a clean way to do before or after as the go test tool doesn't care about what's new or not, but we could do it if we keep artifacts around with the old values and update them. Potentially multiple artifacts as I imagine we would want it for each branch we want to merge into. (Though that would most often be main)

turetske:
@bbockelm It's now implementing a comparison between current branch and the base and is probably the best I can do without spending more time on this than I think it merits.

@CannonLock Can you look this over again?

CannonLock:
@turetske LGTM!

turetske:
@CannonLock Can you click the approve button so it can merge?

bbockelm (approved):
Hitting approve on behalf of @CannonLock

PR #137, opened 2023-09-14
Switched over remaining usage of http.Client to use common transport

joereuss12:
There were a few spots in newer code looking for the TLSSkipVerify on its own. Changed them over to use the common GetTransport function. I also had to rename the function to be capital letters since that's how Go reads functions from other packages.


jhiemstrawisc:
Bypassing merge restrictions because all I did was fix a small linter error.

Issue #136, opened 2023-09-14
Investigate timeouts in linter action


jhiemstrawisc:
Occasionally, the linter action fails on a merge due to a timeout. For example, this happened when I merged PR #128. It's not a big issue, but it's annoying and it spams my inbox. It looks like the solution might be as simple as bumping the timeout up a bit, but I haven't investigated beyond that.

bbockelm:
The default golang-ci timeout is 1 minute; we up'd it to 2 minutes here:

https://github.com/PelicanPlatform/pelican/blob/main/.golangci.yaml#L5

It looks like the action you mention timed out at 2 minutes.
- Good news: it appears that our custom configuration is working.
- Bad news: it seems 2 minutes is not enough

How about 3 minutes?

jhiemstrawisc:
Closed by PR #138

PR #138, opened 2023-09-14
Bump linter timeout to 3m

jhiemstrawisc:
After timeouts at 2m, we're going to try bumping the linter timeout to 3m

Fixes issue #136


turetske (commented):
LGTM, I'll set it to merge when the test pass.

PR #139, opened 2023-09-14
Run pre-commit against entire repo

jhiemstrawisc:
Now that we have pre-commit, we should be able to cut down on noise generated by some of the github actions. This commit runs pre-commit against the entire repo to whip everything into shape.


jhiemstrawisc:
@CannonLock I requested a review from you because I noticed that pre-commit modified a few svg's. in `origin_ui/src/public`. I'm guessing it was just dealing with some whitespace/new lines, but can you double check that it didn't do anything wonky?

bbockelm (approved):
LGTM.  I reviewed the SVG file changes and it appears that the only change is a newline.  Appears to be valid still.

Issue #140, opened 2023-09-14
Get namespace ad by longest match, not first


jhiemstrawisc:
@matyasselmeci experienced a bug with the director when requesting information for the path `/chtc/PUBLIC` where the information returned by the director corresponded to namespace `/chtc`. It looks like the director will return the first matching prefix without checking if there's a more complete one. This also appears to be a bug already solved in another spot by the old OSDF client.

PR #142, opened 2023-09-14
Match ns prefix by longest match, not first match

jhiemstrawisc:
Previously, when two namespaces like /chtc and /chtc/PUBLIC existed in the cache, the matchPrefix function wound up matching whichever happened to be read first. Now it matches against the longest prefix, so a request for path /chtc/PUBLIC/foo/bar will always match /chtc/PUBLIC.

Closes issue #140


bbockelm (changes requested):
Implementation looks correct.

However!  An important push for v7.1.0 is unit tests -- can you please add a regression test covering this function as part of the PR?

PR #145, opened 2023-09-19
Remove commenting code that causes errors when using a fork

turetske:
Removing the commenting code, at least until we decide on a fix so it's no longer causing a block.


jhiemstrawisc (approved):
LGTM. This block just removes the comment, so we can still see coverage metrics by looking into the logs. Let's follow up on how to get this to work later.

PR #149, opened 2023-09-25
First step toward self-testing: configure authorization

bbockelm:
The embedded XRootD server needs to be able to configure authorization for the built-in issuer.

This PR does exactly that -- reading in the (optional) admin-provided `scitokens.cfg` file, augmenting it with the builtin issuer, and then writing out the newly-generated configuration into the runtime directory.

Comes with fairly comprehensive unit tests -- it just doesn't have any user-facing changes.


jhiemstrawisc:
@bbockelm I think I mostly understand what's going on here, but I'm not sure how I can inspect the final output of the generated `scitokens-generated.cfg`. From reading the code, I'd expect serving a rootly origin to emit such a file in `/run/pelican/xrootd`.

It doesn't look like any of the new functions in `xrootd/authorization` are called outside of the tests. Perhaps all this code is reached through some mechanism I'm unaware of, or am I mistaken?

bbockelm:
Right now, the code indeed isn't reachable beyond the tests.  My thinking is to get the infrastructure merged now and then start working on the periodic self-test code.

jhiemstrawisc (changes requested):
Just a few small things, otherwise LGTM

jhiemstrawisc (approved):
LGTM

PR #152, opened 2023-09-27
Fixed Generate Private Key Problems in Windows

joereuss12:
Found an issue when trying to get a private key when running my unit test locally on my windows machine. From investigating found that privs.go only worked with uid's and gid's which is only functional with mac and linux. I modified a few things to work with windows SID's as well by checking what os is being ran. I ran my tests after modifying on both my linux container and locally on windows and it seems to work now. I also had to change around a few lines in GeneratePrivateKey since it calls a "chown" which is not a windows command.

### DIFF
--- a/config/init_server_creds.go
+++ b/config/init_server_creds.go
@@ -28,7 +28,9 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


turetske (approved):
LGTM! I would normally say we should add something to test this, but I think both our unit tests will take care of that.

PR #153, opened 2023-09-27
Added token checking for prometheus metrics

turetske:
-- Added a checkPromToken function in prometheus.go to check the tokens
    -- Added unit tests for checking the metrics
    -- Fixed missing "Bearer " in redirect_test

### DIFF
--- a/director/redirect_test.go
+++ b/director/redirect_test.go
@@ -181,7 +181,7 @@ func TestDirectorRegistration(t *testing.T) {
 	rInv.POST("/", RegisterOrigin)
 	cInv.Request, _ = http.NewRequest(http.MethodPost, "/", bytes.NewBuffer([]byte(`{"Namespaces": [{"Path": "/foo/bar", "URL": "https://get-your-tokens.org"}]}`)))
 
-	cInv.Request.Header.Set("Authorization", string(signedInv))
+	cInv.Request.Header.Set("Authorization", "Bearer "+string(signedInv))
 	cInv.Request.Header.Set("Content-Type", "application/json")
 
 	rInv.ServeHTTP(wInv, cInv.Request)
--- a/web_ui/prometheus.go
+++ b/web_ui/prometheus.go
@@ -17,6 +17,7 @@ package web_ui
 
 import (
 	"context"
+	"crypto/ecdsa"
 	"errors"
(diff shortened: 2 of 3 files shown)


bbockelm:
@jhiemstrawisc - I'd like you to pick up the review on this one.  One thing to check is that the graphs in the web UI that @CannonLock is working on continues to work.

@turetske - on its face, the approach looks correct (I'll let Justin dig into the details).  One thing this does *not* do is enable tokens that would be coming from the federation (but it's OK to do that part as a separate PR).  For the latter, you can get the JWKS information from the auto-discovery URL:

```
$ curl https://osg-htc.org/.well-known/pelican-configuration
{
  "director_endpoint": "https://director-caches.osgdev.chtc.io",
  "namespace_registration_endpoint": "https://registry.osgdev.chtc.io",
  "jwks_uri": "https://osg-htc.org/osdf/public_signing_key.jwks",
  "collector_endpoint": "condor://osdf-collector.osg-htc.org"
}
```

The `jwks_uri` is going to hold the public key for the Prometheus scraper from the federation.  The issuer in the token from the federation (in this example) is `https://osg-htc.org/`.

bbockelm:
Oh - one other thing I just spotted.  @turetske - can you make sure to work in your own fork and not create branches in the main repo for features?

turetske:
> Oh - one other thing I just spotted. @turetske - can you make sure to work in your own fork and not create branches in the main repo for features?

Yep. Starting to do that.

In regards to the using federation tokens, (I'm still fuzzy on tokens in general, so forgive me if this is a silly question) would it just be a case of changing where it's pulling the jwks it's verifying against from? Getting it from the auto-discovery url rather than `config.GetOriginJWKS`? Or would there be more to do?

bbockelm:
> In regards to the using federation tokens, would it just be a case of changing where it's pulling the jwks it's verifying against from? Getting it from the auto-discovery url rather than config.GetOriginJWKS? Or would there be more to do?

Yes.  You'd look at the `iss` ("issuer") attribute and decide where to pull the public key from.  So, somewhere in the existing code there will be a conditional to see if the issuer was the federation itself...

turetske:
Closing PR in order to redo as a fork

PR #154, opened 2023-09-27
Test Origin Login APIs

joereuss12:
This is a unit test for the origin login api's. The tests have been ran on both windows and in a linux container and shows everything running successfully. This effectively tests:
1. Code-based login when no password is set
2. Password reset API
3. Password-based login API
4. "whoami" API

### DIFF
--- a/origin_ui/origin_test.go
+++ b/origin_ui/origin_test.go
@@ -20,7 +20,6 @@ package origin_ui
 
 import (
 	"fmt"
-	"io/ioutil"
 	"math/rand"
 	"net/http"
 	"net/http/httptest"
@@ -29,8 +28,8 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/pelicanplatform/pelican/config"
 	"github.com/gin-gonic/gin"
+	"github.com/pelicanplatform/pelican/config"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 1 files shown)


bbockelm:
Changing the reviewer over to Emma.

origin_ui/origin_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+		recorder := httptest.NewRecorder()
+		router.ServeHTTP(recorder, req)
+		//Check http reponse code 401
+		assert.Equal(t, 401, recorder.Code)
+		assert.JSONEq(t, `{"error":"Login failed"}`, recorder.Body.String())
+	})
+
+	//Invoke with invalid user, should fail
+	t.Run("Without user", func(t *testing.T) {
```
turetske, line 274: You have tests for "Without user" and "Without password" but what about "Incorrect user" (where the user doesn't exist) or "Wrong password"

turetske (changes requested):
LG except for possibly some changes in how temp files are being created and the addition of two more test cases as discussed in the specific comments.

origin_ui/origin_test.go
```diff
@@ -0,0 +1,354 @@
...
+var (
+	tempPasswdFile *os.File
+	router         *gin.Engine
+)
+
+func TestMain(m *testing.M) {
+	gin.SetMode(gin.TestMode)
+
+	//set a temporary password file:
+	tempPasswdFile, err := os.CreateTemp("", "origin-ui-passwd")
```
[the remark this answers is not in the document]
joereuss12: From investigating, it looks like I can't use the packages TempDir() function (I did not know this existed but that's helpful to know now) in the TestMain() function and to me it doesn't make too much sense to create a test file for each function if it is just needed for general setup and use throughout the other functions. I did however, change `defer tempPasswdFile.Close()` to `defer os.Remove(tempPasswdFile.Name())` and that looks like the file is being cleaned up and removed after all of the tests have run. Working to add the other tests you mentioned now!

origin_ui/origin_test.go
joereuss12: Added with new commit

turetske (approved):
LGTM!

Issue #97, opened 2023-09-03
Create new "pelican origin token" CLI


bbockelm:
Since the signing key is on the origin, we can create a small suite of tools for generating tokens (take inspiration from the scitokens CLI).

This will be useful for origin admins and anyone wanting to test the origin.

jhiemstrawisc:
Token creation is accomplished by PR #157

PR #157, opened 2023-09-27
Expose CLI for creating tokens

jhiemstrawisc:
This PR exposes the ability for a user at an origin to create tokens for testing. It can be accessed via `pelican origin token create`. There are several claims that all default tokens will posses, namely iat, exp, nbf, iss, and jti. These can be overridden by passing them specifically to the tool as claims.

By passing a profile with the --profile flag, the tool will provide information about additional claims that are needed by the profile. NOTE: it only provides information about required values that ARE NOT auto-generated!

The lifetime and private-key flags can be used to set the token lifetime (in seconds) and the private signing key, respectively.

Next up is to create a CLI for verifying a provided token against a specific profile. While the basic infrastructure was set up for that via `pelican origin token verify`, the function remains unimplemented.

This is a partial implementation toward closing issue #97

### DIFF
--- a/cmd/origin.go
+++ b/cmd/origin.go
@@ -44,6 +44,34 @@ var (
 		RunE:         serveOrigin,
 		SilenceUsage: true,
 	}
+
+	// Expose the token manipulation CLI
+	originTokenCmd = &cobra.Command{
+		Use:   "token",
+		Short: "Manage Pelican origin tokens",
+	}
+
+	originTokenCreateCmd = &cobra.Command{
+		Use:   "create",
+		Short: "Create a Pelican origin token",
+		Long: `Create a JSON web token (JWT) using the origin's signing keys:
+
+Pelican origins use JWTs as bearer tokens for authorizing specific requests,
+such as reading from or writing to the origin's underlying storage, advertising
+to a director, etc. For more information about the makeup of a JWT, see
+https://jwt.io/introduction.
+
+Additional profiles that expand on JWT are supported. They include scitokens2 and
+wlcg1. For more information about these profiles, see https://scitokens.org/technical_docs/Claims
+and https://github.com/WLCG-AuthZ-WG/common-jwt-profile/blob/master/profile.md, respectively`,
+		RunE: cliTokenCreate,
+	}
+
+	originTokenVerifyCmd = &cobra.Command{
+		Use:   "verify",
+		Short: "Verify a Pelican origin token",
+		RunE:  verifyToken,
+	}
 )
 
 func configOrigin( /*cmd*/ *cobra.Command /*args*/, []string) {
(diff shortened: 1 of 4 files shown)


cmd/origin.go
```diff
@@ -59,4 +89,14 @@ func init() {
 		panic(err)
 	}
 	originServeCmd.Flags().AddFlag(portFlag)
+
+	originCmd.AddCommand(originTokenCmd)
+	originTokenCmd.AddCommand(originTokenCreateCmd)
+	originTokenCmd.PersistentFlags().String("profile", "", "Passing a profile ensures the created token adheres to the profile's requirements. Accepted values are scitokens2 and wlcg1")
```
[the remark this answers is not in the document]
bbockelm: Also, can you set a default value here?  I'd prefer `wlcg`

cmd/origin.go
```diff
@@ -44,6 +44,36 @@ var (
...
+		Use:   "token",
+		Short: "Manage Pelican origin tokens",
+	}
+
+	originTokenCreateCmd = &cobra.Command{
+		Use:   "create",
+		Short: "Create a Pelican origin token",
+		Long: `Create a JSON web token (JWT) using the origin's signing keys:
+Usage: pelican origin token create [FLAGS] claims
+E.g. pelican origin token create --profile scitokens2 aud=my-audience scope="read:/storage" scope="write:/storage"
```
bbockelm, line 59: I don't care for the arbitrary syntax here.  For well-known claims, provide flags:
```
--audience=foo --scope=read:/storage --scope=write:/storage
```

Then, for arbitrary overrides, add a separate claim:

```
--claim=foo=bar
```

bbockelm (changes requested):
A few changes requested, inline.

Most revolve around the command line flag approach and some nitpicks.

cmd/origin.go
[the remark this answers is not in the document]
jhiemstrawisc: My reasoning for not setting a default was that after inspecting token creation throughout our code base, not a single token conforms to either `wlcg` or `scitokens2`. It felt strange enforcing a particular profile when we ourselves weren't using one. 

I think the solution is to standardize our internal token creation with the `createToken` function this exposes. I'll set the default in this PR to `wlcg` and open a ticket for the standardization.

cmd/origin_token_test.go
```diff
@@ -0,0 +1,196 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ***************************************************************/
+
+package main
+
+import (
+	// "net/url"
```
bbockelm, line 22: Remove the commented-out imports.  We can always bring them back later.

cmd/origin_token.go
```diff
@@ -0,0 +1,310 @@
...
+			}
+		} else {
+			errMsg := "The provided profile '" + profile +
+				"' is not recognized. Valid options are 'scitokens2' or 'wlcg'"
+			return "", errors.New(errMsg)
+		}
+	}
+
+	lifetimeDuration := time.Duration(lifetime)
+	// Create a jti using uuid4. This will be added to all tokens.
```
bbockelm, line 149: It seems that UUIDs aren't built-in in golang; instead of adding a new dependency, let's do the following instead:

https://github.com/PelicanPlatform/pelican/pull/167/files#diff-9ecc873c07e2491d63b175c34ad6f290375f7932c2849c36cd7d106312953044R54-R58

bbockelm (changes requested):
Two very small changes then we're ready to go.

bbockelm (approved):
Revised changes LGTM.

Issue #159, opened 2023-09-28
Standardize token creation/verification throughout Pelican


jhiemstrawisc:
After starting work on the CLI for creating origin tokens, I realized we're pretty lackadaisical about how we create and verify tokens throughout the code base. We should pick an internal profile and stick to it. Currently, none of the tokens we create conform to either scitokens2 or wlcg.

A part of this should also be a standard profile-aware token verification tool.

haoming29:
The work from #320 should be able to finish the "token create" part of this ticket, and [#405](https://github.com/PelicanPlatform/pelican/issues/405) will finish the generic token verification part of the ticket. However, the "profile-specific" token verification is not part of the scope for these two PRs and thus need to be addressed separately. I'm working on #320 and best case scenario can get this out tomorrow but I won't expect this to show up in 7.3. Given that, do we want to slip over this ticket to 7.4? (I know we already did it once, life is hard... : (  @jhiemstrawisc )

jhiemstrawisc:
Truer words may never have been spoken -- life is indeed hard. Bumping to 7.4

Issue #119, opened 2023-09-07
CreateAdvertiseToken unit test


turetske:
Create a unit test for the createAdvertiseFunction in origin_api.go

PR #160, opened 2023-09-28
Added tests for the three functions in origin_api.go

turetske:
Tests tokens with various scopes for the `VerifyAdvertiseToken` function
	Tests the string created from `GetIssuerURL`
	Tests token creation using the `CreateAdvertiseToken` function
	Also adjusts the mocked cache in redirect_test.go to be used by both test files

### DIFF
--- a/director/origin_api_test.go
+++ b/director/origin_api_test.go
@@ -1,0 +1,162 @@
+package director
+
+import (
+	"crypto/elliptic"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/jellydator/ttlcache/v3"
+	"github.com/lestrrat-go/jwx/v2/jwa"
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/lestrrat-go/jwx/v2/jwt"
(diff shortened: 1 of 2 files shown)


jhiemstrawisc:
@turetske Just a heads up, we need to meddle with some of these tokens, and ultimately the functions that verify them (and their tests) soon. See issue #159. 

I think all this is okay for now, but we might have to come back to revisit at some point soon.

turetske:
> @turetske Just a heads up, we need to meddle with some of these tokens, and ultimately the functions that verify them (and their tests) soon. See issue #159.
> 
> I think all this is okay for now, but we might have to come back to revisit at some point soon.

@jhiemstrawisc Okay. I don't know enough about tokens to understand issue #159, but I will absolutely take your word for it.

bbockelm:
@turetske - is there a corresponding issue for this PR?

turetske:
@bbockelm Yeah, it's #119. I couldn't get started on my last v1.0 issue without clarification so I decided to tackle one of the unit tests issues I created. And then decided might as well test the whole file rather than one function.

PR #161, opened 2023-09-28
Fix federation auto-discovery bug

jhiemstrawisc:
That I'm aware of, nobody has encountered this bug yet. I just happened to notice it when reading through the file. I'm 99% sure the previous code was incorrect and would result in autodiscovery not being run if only the "DirectorUrl" was set, resulting in no valid NamespaceUrl.


turetske (approved):
LGTM

PR #162, opened 2023-09-28
Remove redundant viper call

jhiemstrawisc:
Another thing I happened to notice while scanning through some files.


bbockelm (approved):
LGTM

PR #163, opened 2023-09-29
Add development Dockerfile

### LINKED ISSUE #88: Create Dockerfile for development environment
Pelican's development environment is growing more complex -- whereas the "osdf-client" era only needed a go compiler, the modern dev + testing environment also needs nodejs/npm, xrootd, and (as soon as we get going with the issuer integration) Java.

To ease the process of onboarding new developers, let's create a common development environment container.  GitHub Action to build this can be triggered on an as-needed basis.

CannonLock:
99% of the work is courtesy @jhiemstrawisc and his Dockerfile.

Draft for now well I figure out why I can't run Goreleaser, everything else works as expected. 

Closes #88.

### DIFF
--- a/.github/workflows/publish-dev-container.yml
+++ b/.github/workflows/publish-dev-container.yml
@@ -1,0 +1,51 @@
+
+name: Release, Build, and Push Dev Image
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+-**
+    branches:
+      - main
+
+jobs:
+  build-dev-image:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: hub.opensciencegrid.org/pelican_platform/pelican-dev
+          tags: |
+            type=semver,pattern={{version}}
+            type=raw,value=latest-itb
+            type=ref,enable=true,prefix=itb-,suffix=-{{date 'YYYYMMDDHHmmss'}},event=tag
+            type=raw,value=sha-{{sha}}
+
+      - name: Log in to OSG Harbor
+        uses: docker/login-action@v2
+        with:
+          registry: hub.opensciencegrid.org
+          username: ${{ secrets.PELICAN_HARBOR_ROBOT_USER }}
+          password: ${{ secrets.PELICAN_HARBOR_ROBOT_PASSWORD }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-dev-container.yml
```diff
@@ -0,0 +1,48 @@
...
+        name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: hub.opensciencegrid.org/macrostrat/tileserver
```
jhiemstrawisc, line 25: Do we mean to be grabbing things from the macrostrate container, or is this a copy-paste error?

images/dev.Dockerfile
```diff
@@ -0,0 +1,39 @@
+#FROM mcr.microsoft.com/devcontainers/go:0-1-bullseye
```
jhiemstrawisc, line 1: Let's remove this comment

images/dev.Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+FROM almalinux:8
```
jhiemstrawisc, line 2: We should pin the platform as arm here, otherwise this Dockerfile (which is tailored for arm) will be built using the GHA runner's host OS -- most definitely amd64.

images/dev.Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+gpgcheck=0' > /etc/yum.repos.d/goreleaser.repo
+    
+# Install goreleaser and various other packages we need
+RUN yum install -y goreleaser npm xrootd xrootd-server xrootd-client nano xrootd-scitokens \
+    xrootd-voms jq procps docker make
+RUN yum clean all
+
+# Installing the right version of go
+SHELL ["/bin/sh", "-c"]
+RUN curl https://dl.google.com/go/go1.20.8.linux-arm64.tar.gz -o go1.20.8.linux-arm64.tar.gz
```
jhiemstrawisc, line 25: As it is, this only builds a container for arm64 -- great for us Mac users, but not useful for anyone else. Can we make two dev containers, one for amd64 and one for arm64? The amd64 container should use the OSG base containers. I couldn't use those for arm because the OSG repositories don't package xrootd for arm.

In the amd64 container, this line will need to be modified to grab the correct arch

jhiemstrawisc (changes requested):
The big things are that we should make sure a container is built for archs other than arm (as much as I'd love to flip the script on all those folks who don't build for arm64!)

CannonLock:
@jhiemstrawisc This addresses your comments. I am going to refrain from creating an amd64 image from the osg-software base until a need is identified. I think it makes sense to have us all use the same dev image across both platforms.

jhiemstrawisc (approved):
LGTM

PR #164, opened 2023-09-29
Turetske/parameter generation

turetske:
Implements parameter generation into structs with go generator based on the parameters.yaml file. Also wraps all calls to viper.Get* that are using generated params.


README.md
```diff
@@ -59,6 +59,8 @@ and it shares the same defaults and behavior as stashcp.
 Building
 --------
 
+If the `parameters.yaml` file was altered, regenerate `parameters.go` by running `go generate` in the `./config` directory
```
bbockelm, line 62: Is this comment needed?  I believe `goreleaser` (and just hitting `make`) automatically runs `go generate`, no?

bbockelm (changes requested):
I love the general approach here... exactly what I was thinking of.

However, two medium-sized changes:

1.  I don't like having this in the config module: I'm worried about conflicts between the parameter names and various configuration functions.  How about putting this into a standalone `param` module?
2. We shouldn't be putting the autogenerated file into the repo.  That's going to introduce unnecessary churn.  Instead, let's add the output file to `.gitignore` and remove it from the patch series.

README.md
turetske: Removed from the code

turetske:
@bbockelm 

I've updated with the changes you requested.

bbockelm (approved):
LGTM!  All the items from the prior review have been addressed.

Issue #25, opened 2023-07-28
Have webdav directory listing utilize token-based authorization


bbockelm:
We think, when stashcp does the directory existence test, it uses a library that doesn't leverage the bearer token.  Investigate this issue and, as necessary, add token support.

Alternately, if token support isn't available, rip out this code.

bbockelm:
@matyasselmeci points out this probably also breaks recursive transfers (which increases the priority to fix).

turetske:
Client side code. Interface for listing remote directories. The token isn't being passed to the web-dev library or it's being passed. "list-directory" function. Possibly customize the http headers.

turetske:
@bbockelm Looking into the code, gowebdav doesn't support this. However, there is an open issue for a fix on their repo, how do we feel about me attempting that rather than ripping out the code?

bbockelm:
What's the GH issue you found?  We should at least get that linked to this one.

turetske:
> What's the GH issue you found? We should at least get that linked to this one.

It's linked here: https://github.com/studio-b12/gowebdav/issues/61

 I did look into adding the support to the base web directory, but honestly my token/http knowledge still feels to weak to do so, so instead I'm removing the isDir code and we can fix it later.

bbockelm:
I'm OK with fixing the directory listing / upload on a future ticket.  Will merge the PR momentarily.

PR #165, opened 2023-09-29
Removed `isDir` from handle_http.go

turetske:
- The webdav library used does not do bearer token authorization so removed the directory existence check

   This is dealing with issue #25


bbockelm (approved):
LGTM.

PR #176, opened 2023-10-02
Put prometheus metrics behind federation/origin token

turetske:
This addresses issue Issue #6 and puts the accessing of prometheus metrics behind the federation or origin token.

### DIFF
--- a/web_ui/prometheus.go
+++ b/web_ui/prometheus.go
@@ -20,6 +20,7 @@ import (
 	"crypto/ecdsa"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"net/http"
 	"net/url"
@@ -37,6 +38,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/grafana/regexp"
 	"github.com/lestrrat-go/jwx/v2/jwa"
+	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/lestrrat-go/jwx/v2/jwt"
 	"github.com/mwitkow/go-conntrack"
 	"github.com/oklog/run"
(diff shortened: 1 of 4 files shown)


web_ui/prometheus.go
```diff
@@ -138,6 +143,83 @@ func runtimeInfo() (api_v1.RuntimeInfo, error) {
...
+			if err != nil {
+				c.JSON(400, gin.H{"error": "Failed to get federation key file"})
+			}
+			defer response.Body.Close()
+			contents, err := io.ReadAll(response.Body)
+			if err != nil {
+				c.JSON(400, gin.H{"error": "Failed to read federation key file"})
+				return
+			}
+			key, err := jwk.ParseKey(contents, jwk.WithPEM(true))
```
jhiemstrawisc, line 192: A few notes about this:

- The `jwk.ParseKey` function will fail if contents has more than one key. While the federation only makes available a single public key at the moment, I don't know if that will always be the case. Can you look at doing this with `jwk.Parse()` so we make sure to always look at the entire keyset?

- I'm not sure this should be passed with the `jwk.WithPEM` option, as the Federation's jwks isn't stored in PEM format. Does it work anyway?

web_ui/prometheus.go
```diff
@@ -138,6 +143,83 @@ func runtimeInfo() (api_v1.RuntimeInfo, error) {
...
+			}
+		}
+
+		var raw ecdsa.PrivateKey
+		if err = (*bKey).Raw(&raw); err != nil {
+			c.JSON(400, gin.H{"error": "Failed to extract signing key"})
+			return
+		}
+
+		_, err = jwt.Parse([]byte(strToken), jwt.WithKey(jwa.ES512, raw.PublicKey), jwt.WithValidate(true))
```
jhiemstrawisc, line 212: Another PR in the pipeline (#167) has to change token encryption to `ES256` for compatibility with the SciTokens library. Can you also change that here and on the token-creation side of things to prevent future merge conflicts? It'll also require that we create a new federation keypair to match, which could go in a GH issue.

web_ui/prometheus_test.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a token
+	issuerURL := url.URL{}
+	issuerURL.Scheme = "https"
+	issuerURL.Host = "test-http"
+	now := time.Now()
+	tok, err := jwt.NewBuilder().
```
jhiemstrawisc, line 84: Moving forward, we should probably start making sure our internal tokens adhere to a well-known JWT profile. Brian indicated he favors the WLCG1 profile, so let's stick with that for now. You can find more information about the additional claims to include [here](https://github.com/WLCG-AuthZ-WG/common-jwt-profile/blob/master/profile.md#capability-based-authorization-scope)

web_ui/prometheus.go
```diff
@@ -138,6 +143,83 @@ func runtimeInfo() (api_v1.RuntimeInfo, error) {
...
+
+		if err != nil {
+			c.JSON(403, gin.H{"error": "Permission Denied: Invalid token"})
+			return
+		}
+
+		av1.ServeHTTP(c.Writer, c.Request)
```
jhiemstrawisc, line 219: We've verified the token at this point, but we haven't checked that it actually grants the permissions we need. Can you add checks that the scopes, audience, etc all match what they're expected to? For some inspiration, see `namespace-registry/registry.go` around line 640.

jhiemstrawisc (changes requested):

web_ui/prometheus.go
```diff
@@ -138,6 +143,83 @@ func runtimeInfo() (api_v1.RuntimeInfo, error) {
...
+		if authzQuery := req.URL.Query()["authz"]; len(authzQuery) > 0 {
+			strToken = authzQuery[0]
+		} else if authzHeader := req.Header["Authorization"]; len(authzHeader) > 0 {
+			strToken = strings.TrimPrefix(authzHeader[0], "Bearer ")
+		} else {
+			c.JSON(403, gin.H{"error": "Permission Denied: Missing token"})
+			return
+		}
+
+		token, err = jwt.Parse([]byte(strToken), jwt.WithVerify(false))
```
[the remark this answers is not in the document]
turetske: Yes. This is just so we can get the issuer information from the token so we can check if the issuer is the federation. I'll add a comment to the code.

web_ui/prometheus.go
turetske: I think I got the withPEM from some other parsing we do, probably based on the origin, but I can change this

web_ui/prometheus.go
turetske: What scope do we want the token to have for the prometheus permissions?

jhiemstrawisc: I think something like `prometheus.read` would be be sensible.

turetske:
@jhiemstrawisc Addressed the code review comments, if you could review it again.

jhiemstrawisc (approved):
LGTM

PR #177, opened 2023-10-03
Refactor the client to be in a sub-package

### LINKED ISSUE #73: Migrate client go files into subpackage
Right now, the implementation of `pelican object` is primarily in the root of the package.  Migrate this out to a sub-package to make the modularity a bit clearer.

bbockelm:
It was confusing (esp. with a client file named `main.go`) to have the client be in the top-level module and have everything else in pelican be in a submodule.  Migrate the client to a submodule.

Fixes #73


turetske (approved):
LGTM!

Issue #21, opened 2023-07-28
Have director define a minimum supported version of stashcp


bbockelm:
(This came up during the Collaborations Day meeting for OSG)

We should add a response to the `pelican` client requests in the director to refuse to serve old clients.  This would provide a quite clear error message if there's unexpected version incompatibility.  Currently, the error messages can be extremely confusing when running an old client against the current OSDF.

jhiemstrawisc:
@bbockelm I'm hoping for a bit of clarification. Do we want future clients to send the director their version and have the director decide whether it can appropriately serve that version of the client, with an appropriate error message in the event that it cannot? Or do we want the director to broadcast the minimum client version it can support and have the client decide if it's capable of parsing the director's response?

bbockelm:
I think we want the client to advertise an appropriate `User-Agent` header

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

and then have the director decide if the client is compatible with the director.

In general, it's better to have the intelligence at the central level and keep the client as simple as possible.

PR #178, opened 2023-10-03
Add director min version detection capability

jhiemstrawisc:
According to issue #21, users have encountered issues with stashcp getting confusing error messages from the Director. While this PR doesn't necessarily handle old stashcp clients that don't set user-agent headers when querying the director, it makes things more robust moving forward.

Each service that queries the director now sets a User-Agent header of the form "pelican-<service>/<version>", eg "pelican-client/7.0.0". The director checks if the UA string matches the form, and if it doesn't it continues (this is to allow unknonwn UAs to continue working, for example if someone wants to curl the director). If the UA matches the expected form, the director will parse both the service and the version and make sure that it believes it can serve a service with that version. In the event that it cannot, it returns an error to the service and elaborates on both the version that the director detected, as well as what minimum version the service should update to for compatibility.

For now, I set those minimum versions to 7.0.0 for both the client and the origin.


director/redirect_test.go
```diff
@@ -183,9 +186,62 @@ func TestDirectorRegistration(t *testing.T) {
...
+		_, err := wInv.Write([]byte(":)"))
+		assert.NoError(t, err)
+	}))
+	defer tsInv.Close()
+	cInv.Request = &http.Request{
+		URL: &url.URL{},
+	}
+
+	// Create a private key to use for the test
+	privateKeyInv, err = ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
```
turetske, line 212: You should be able to reuse the keys that have been generated before. The only reason the earlier tests regenerate a private key is to sign a token with a different key than exists already. Actually, you should just be able to reuse the token from the first passing test.

director_test.go
```diff
@@ -188,7 +188,7 @@ func TestQueryDirector(t *testing.T) {
 	expectedLocation := "http://redirect.com"
 	handler := func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Location", expectedLocation)
-		w.WriteHeader(http.StatusFound)
+		w.WriteHeader(http.StatusTemporaryRedirect)
```
turetske, line 191: I'm trying to understand why this is changed from StatusFound to StatusTemporaryRedirect? This is probably my lack of http understanding, but I'd appreciate a clarification.

director.go
```diff
@@ -129,16 +136,38 @@ func QueryDirector(source string, directorUrl string) (resp *http.Response, err
...
 		return
 	}
-
 	defer resp.Body.Close()
+	log.Debugln("Director's response:", resp)
+
+	// Check HTTP response -- should be 307 (redirect), else something went wrong
+	body, _ := io.ReadAll(resp.Body)
+	if resp.StatusCode != 307 {
+		fmt.Printf("\n\n\nresp.StatusCode %d\n\n\n", resp.StatusCode)
```
turetske, line 163: Why so many newlines? Just for formatting?

director/redirect.go
```diff
@@ -85,7 +93,64 @@ func getFinalRedirectURL(rurl url.URL, authzEscaped string) string {
...
+	}
+
+	var minCompatVer *version.Version
+	switch service {
+	case "client":
+		minCompatVer = minClientVersion
+	case "origin":
+		minCompatVer = minOriginVersion
+	}
+	if err != nil {
```
turetske, line 135: This looks like an unneeded repeated error check and should be removed

turetske (changes requested):
Looks good except for a few minor changes. 

At some point I might want to refactor redirect_test.go to be tabularized now that there are three tests happening in there, but that change be for later.

director_test.go
jhiemstrawisc: The director has always returned a `307: Temporary Redirect`. The reason I had to change this was that I noticed the old code never actually checked the status code returned by the director. Now the client's director-handling code returns an error if the status code is not the expected 307. Since this test was mocking the director's response, it needed to be updated with the correct status code.

director.go
jhiemstrawisc: An errant debug comment! Thanks for catching this!

director/redirect_test.go
```diff
@@ -185,9 +188,43 @@ func TestDirectorRegistration(t *testing.T) {
...
+
+	// Sign token with the good key
+	signedInv, err = jwt.Sign(tok, jwt.WithKey(jwa.ES512, pKey))
+	assert.NoError(t, err, "Error signing token")
+
+	// Create the request and set the headers
+	rInv.POST("/", RegisterOrigin)
+	cInv.Request, _ = http.NewRequest(http.MethodPost, "/", bytes.NewBuffer([]byte(`{"Namespaces": [{"Path": "/foo/bar", "URL": "https://get-your-tokens.org"}]}`)))
+
+	cInv.Request.Header.Set("Authorization", string(signedInv))
```
turetske, line 221: I believe this should be "Bearer "+string(signedInv)

A different PR I have up changes it earlier in the test as well.

turetske (changes requested):
LGTM with some minor changes

bbockelm (changes requested):
Please rebase / clean up the commits.

There are some commits in the branch unrelated to the PR.  Rebase the branch to `origin/main` and remove the duplicates.

Issue #56, opened 2023-08-29
Don't request new token from file transfer plugins


joereuss12:
The plugin when it is run should assume it is fully automated and not request any feedback.

turetske:
@joereuss12 or @bbockelm Can I get some clarification as to what this means? Which plugin? Where is it located? Should this just remove token request/authentication from the plugin?

bbockelm:
The "staging plugin" (https://github.com/PelicanPlatform/pelican/blob/main/cmd/plugin_stage.go) is meant to only be run by HTCondor with all the authorization pre-configured.

However, if you run the plugin, it'll run the "token acquisition" workflow which tries to read the credential configuration file (potentially asking the user to provide the decryption password).  If a decryption password is needed -- or the token isn't readily available, return an error that a credential is required but is currently missing.

PR #180, opened 2023-10-03
Put the token acquisition workflow behind a plugin check

turetske:
This is addressing https://github.com/PelicanPlatform/pelican/issues/56

Basically, plugin runs should be provided with the authorization that they need. So there is a new pelican variable created and set to true whenever a plugin is run. Then, during getToken, it will check for this guard before running AcquireToken if a token hasn't been found and return an error instead of running the token acquisition workflow.


joereuss12 (approved):
LGTM

PR #181, opened 2023-10-03
Ship OA4MP in the Pelican docker image

bbockelm:
OA4MP is the selected token issuer for Pelican.  It's a tough one to wrangle as it's from the "Java universe" and the team has less experience in Java.

For now, just stage the application into the container allowing developers to start kicking the tires.


bbockelm:
Moving the review assignment from Cannon to Justin as Cannon's got a pretty full plate this week.

Issue #130, opened 2023-09-13
Consolidate configuration option hierarchies


jhiemstrawisc:
We should make the options configurable via `pelican.yaml` more consistent and hierarchical. For example, there's an option `Director.DefaultResponse` that is clearly an option to configure behavior for a director instance. Elsewhere, we have options that configure the director like "MaxMindKeyFile" and "GeoIPLocation", but it's not immediately obvious from the name what service(s) these affect. I think we should make these types of configurations hierarchical so that we wind up with a config more like:
```
Director:
    DefaultResponse: cache
    MaxMindKeyFile: /path/to/kf
    GeoIPLocation: /path/to/location
```

bbockelm:
I've been thinking about the same thing!

Can you create a spreadsheet with the following columns:
- Current parameter name
- Component name (or blank for top-level)
- New parameter name

and take a first stab at the renaming?  Implicit in this is that I'd like to avoid deeply nested hierarchies of parameter names.

The earlier we do this, the less backward compatibility around renaming we need to do.

PR #182, opened 2023-10-04
Overhaul config names

jhiemstrawisc:
In issue #130 we discussed making config variables hierarchical, such that each variable describes which service it affects. One goal was to keep the hierarchies flat (<= 2 layers). This PR breaks out the config variables and assigns them to various parent services. It also adds many new viper config variables to `docs/parameters.yaml` so that users can figure out which components of Pelican they have control over.

### DIFF
--- a/cmd/resources/xrootd.cfg
+++ b/cmd/resources/xrootd.cfg
@@ -15,46 +15,46 @@
 #
 
 {{if .Origin.UseCmsd}}
-all.manager {{.ManagerHost}}+ {{.ManagerPort}}
-{{end}}
+all.manager {{.Xrootd.ManagerHost}}+ {{.Xrootd.ManagerPort}}{{end}}
 all.role server
 if exec xrootd
-  xrd.port {{.Port}}
-  xrd.protocol http:{{.Port}} libXrdHttp.so
+  xrd.port {{.Xrootd.Port}}
+  xrd.protocol http:{{.Xrootd.Port}} libXrdHttp.so
 fi
-xrd.tls {{.TLSCertificate}} {{.TLSKey}}
-{{if .TLSCertDir}}
-xrd.tlsca certdir {{.TLSCertDir}}
+xrd.tls {{.Server.TLSCertificate}} {{.Server.TLSKey}}
+{{if .Xrootd.TLSCertDir}}
+xrd.tlsca certdir {{.Xrootd.TLSCertDir}}
 {{else}}
-xrd.tlsca certfile {{.TLSCACertFile}}
+xrd.tlsca certfile {{.Xrootd.TLSCertFile}}
 {{end}}
 {{if .Origin.UseMacaroons}}
 http.exthandler xrdmacaroons libXrdMacaroons.so
-macaroons.secretkey {{.MacaroonsKeyFile}}
+macaroons.secretkey {{.Xrootd.MacaroonsKeyFile}}
 ofs.authlib ++ libXrdMacaroons.so
 {{end}}
 http.header2cgi Authorization authz
 {{if .Origin.UseVoms}}
 http.secxtractor /usr/lib64/libXrdVoms.so
+http.staticpreload http://static/robots.txt {{.Xrootd.RobotsTxtFile}}
+{{if .Xrootd.Sitename}}
+all.sitename {{.Xrootd.Sitename}}
 {{end}}
-http.staticpreload http://static/robots.txt {{.RobotsTxtFile}}
-{{if .Sitename}}
-all.sitename {{.Sitename}}
+{{if .Xrootd.SummaryMonitoringHost}}
+xrd.report {{.Xrootd.SummaryMonitoringHost}}:{{.Xrootd.SummaryMonitoringPort}},127.0.0.1:{{.Xrootd.LocalMonitoringPort}} every 30s
 {{end}}
-{{if .SummaryMonitoringHost}}
-xrd.report {{.SummaryMonitoringHost}}:{{.SummaryMonitoringPort}},127.0.0.1:{{.LocalMonitoringPort}} every 30s
+{{if .Xrootd.DetailedMonitoringHost}}
+xrootd.monitor all auth flush 30s window 5s fstat 60 lfn ops xfr 5 {{if .Xrootd.DetailedMonitoringHost -}} dest redir fstat info files user pfc tcpmon ccm {{.Xrootd.DetailedMonitoringHost}}:{{.Xrootd.DetailedMonitoringPort}} {{-end}} dest redir fstat info files user pfc tcpmon ccm 127.0.0.1:{{.Xrootd.LocalMonitoringPort}}
 {{end}}
-xrootd.monitor all auth flush 30s window 5s fstat 60 lfn ops xfr 5 {{if .DetailedMonitoringHost -}} dest redir fstat info files user pfc tcpmon ccm {{.DetailedMonitoringHost}}:{{.DetailedMonitoringPort}} {{-end}} dest redir fstat info files user pfc tcpmon ccm 127.0.0.1:{{.LocalMonitoringPort}}
-all.adminpath {{.XrootdRun}}
-all.pidpath {{.XrootdRun}}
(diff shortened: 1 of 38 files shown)


jhiemstrawisc:
@bbockelm I think this is ready for review, even though tests aren't passing yet. It looks like the failed tests are fetching and testing against the head of the repo's main branch. Any idea what's going on there?

cmd/resources/xrootd.cfg
```diff
@@ -14,43 +14,43 @@
...
-  xrd.protocol http:{{.Port}} libXrdHttp.so
+  xrd.port {{.Xrootd.Port}}
+  xrd.protocol http:{{.Xrootd.Port}} libXrdHttp.so
 fi
-xrd.tls {{.TLSCertificate}} {{.TLSKey}}
-{{if .TLSCertDir}}
-xrd.tlsca certdir {{.TLSCertDir}}
+xrd.tls {{.Server.TLSCertificate}} {{.Server.TLSKey}}
+{{if .Xrootd.TLSCertDir}}
+xrd.tlsca certdir {{.Xrootd.TLSCertDir}}
```
bbockelm, line 25: Can you rename this one `Server.TLSCACertificateDirectory`?  It's not XRootD-specific, really -- just only used by XRootD (currently!).

cmd/resources/xrootd.cfg
```diff
@@ -14,43 +14,43 @@
...
 {{else}}
-xrd.tlsca certfile {{.TLSCertFile}}
+xrd.tlsca certfile {{.Xrootd.TLSCertFile}}
```
bbockelm, line 27: Similarly, rename this `Server.TLSCACertificateFile`.  It's not symmetric with the host certificate (`TLSCertificate`) but I find it useful to keep the symmetry of the file and directory.

config/config.go
```diff
@@ -294,47 +294,47 @@ func InitServer() error {
...
+	viper.SetDefault("Xrootd.RobotsTxtFile", filepath.Join(configDir, "robots.txt"))
+	viper.SetDefault("Xrootd.ScitokensConfig", filepath.Join(configDir, "xrootd", "scitokens.cfg"))
+	viper.SetDefault("Xrootd.Authfile", filepath.Join(configDir, "xrootd", "authfile"))
+	viper.SetDefault("Xrootd.MacaroonsKeyFile", filepath.Join(configDir, "macaroons-secret"))
 	viper.SetDefault("IssuerKey", filepath.Join(configDir, "issuer.jwk"))
-	viper.SetDefault("OriginUI.PasswordFile", filepath.Join(configDir, "origin-ui-passwd"))
-	viper.SetDefault("OIDC.ClientIDFile", filepath.Join(configDir, "oidc-client-id"))
-	viper.SetDefault("OIDC.ClientSecretFile", filepath.Join(configDir, "oidc-client-secret"))
+	viper.SetDefault("Origin.UIPasswordFile", filepath.Join(configDir, "origin-ui-passwd"))
+	viper.SetDefault("Registry.OIDCClientIDFile", filepath.Join(configDir, "oidc-client-id"))
```
bbockelm, line 305: Keep this at the `Server` level, not the `Registry` level.

Currently, only the registry uses OIDC ... but I see this for the origin and the director in the near-future as well.

config/config.go
```diff
@@ -294,47 +294,47 @@ func InitServer() error {
...
+	viper.SetDefault("Registry.OIDCClientSecretFile", filepath.Join(configDir, "oidc-client-secret"))
```
bbockelm, line 306: Similarly, put this under `Server`.

config/config.go
```diff
@@ -294,47 +294,47 @@ func InitServer() error {
...
-		viper.SetDefault("XrootdRun", "/run/pelican/xrootd")
-		viper.SetDefault("XrootdMultiuser", true)
-		viper.SetDefault("GeoIPLocation", "/var/cache/pelican/maxmind/GeoLite2-City.mmdb")
-		viper.SetDefault("NSRegistryLocation", "/var/lib/pelican/registry.sqlite")
-		viper.SetDefault("MonitoringData", "/var/lib/pelican/monitoring/data")
+		viper.SetDefault("Xrootd.RunLocation", "/run/pelican/xrootd")
+		viper.SetDefault("Origin.Multiuser", true)
+		viper.SetDefault("Director.GeoIPLocation", "/var/cache/pelican/maxmind/GeoLite2-City.mmdb")
+		viper.SetDefault("Registry.DbLocation", "/var/lib/pelican/registry.sqlite")
+		viper.SetDefault("Prometheus.MonitoringData", "/var/lib/pelican/monitoring/data")
```
bbockelm, line 312: How about `Monitoring.DataLocation`?  "Prometheus" shouldn't be exposed at this level.

config/config.go
```diff
@@ -464,7 +464,7 @@ func InitClient() error {
 		}
 		break
 	}
-	viper.Set("MinimumDownloadSpeed", downloadLimit)
+	viper.Set("Client.MinimumDownloadSpeed", downloadLimit)
```
bbockelm, line 467: I think some people actually tweak this.  Can you insert backward compatibility code so the value of `MinimumDownloadSpeed`, if set, is used as the default for `Client.MinimumDownloadSpeed` (that is, an explicit `Client.MinimumDownloadSpeed` supersedes everything else)?

config/resources/defaults.yaml
```diff
@@ -14,21 +14,24 @@
...
-SummaryMonitoringPort: 9931
-DetailedMonitoringPort: 9930
-MonitoringPortLower: 9930
-MonitoringPortHigher: 9999
-Mount: ""
-NamespacePrefix: ""
 Debug: false
+WebAddress: "0.0.0.0"
+Server:
+  WebPort: 8444
```
bbockelm, line 20: Let's change:
- `WebAddress` -> `Server.Address`
- `WebPort` -> `Server.Port`
The `Web` prefix was added to distinguish this from the Xrootd server's port number; this is currently achieved by putting it in the `Server` section.

docs/parameters.yaml
```diff
@@ -51,40 +44,140 @@ type: bool
...
+components: ["client"]
+---
+name: Client.SlowTransferWindow
+description: >-
+  A duration indicating the sliding window over which to consider transfer speeds for slow transfers.
+type: int
+default: 30
+components: ["client"]
+---
+name: Client.DisableHttpProxy
```
bbockelm, line 123: `DisableHttpProxy` and `DisableProxyFallback` are other backward compatibility knobs we should honor.

docs/parameters.yaml
```diff
@@ -172,33 +236,203 @@ root_default: /etc/pelican/oidc-client-secret
...
+name: Registry.OIDCDeviceAuthEndpoint
 description: >-
   A URL describing an OIDC Device Auth Endpoint. This is used by the namespace registry to establish OIDC information
   for authenticated registration.
 type: url
 default: none
 components: ["nsregistry"]
 ---
-name: OIDC.TokenEndpoint
+name: Registry.OIDCTokenEndpoint
```
bbockelm, line 247: As commented elsewhere, let's move the OIDC items outside the registry app... they can actually be top-level instead of `Server` like I suggested earlier if that makes sense.

docs/parameters.yaml
```diff
@@ -172,33 +236,203 @@ root_default: /etc/pelican/oidc-client-secret
...
+############################
+name: Prometheus.MonitoringData
+description: >-
+  A filepath where Prometheus should host its monitoring data.
+type: string
+root_default: /var/lib/pelican/monitoring/data
+default: $ConfigBase/monitoring/data
+components: ["origin"]
+---
+name: Prometheus.MonitoringPortLower
```
bbockelm, line 426: Change to a `Monitoring` section -- this has nothing to do with Prometheus.

bbockelm (changes requested):
Overall approach is good; some requests for renames embedded.

bbockelm (approved):
Updates post code-review look good.

This is a sufficiently large-scale change I'm going to go ahead and merge: any smaller issues can be done with follow-ups.

PR #183, opened 2023-10-04
Pelican transport fixes and generated after configs

joereuss12:
Fixed an issue with common transports for some reason not getting correct value from defaults.yaml. Also addressed issue moving the common transports to the config sub-module and allowed initClient and initServer to recreate the common transport object. In addition, moved the TLSSkipVerify to only run when setupTransfer is called again (either first time `GetTransfer()` is called or `initClient()` or `initServer()` is called).

### DIFF
--- a/client/director.go
+++ b/client/director.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


jhiemstrawisc (approved):
LGTM

Issue #147, opened 2023-09-20
Create and test querying the director for available namespaces


turetske:
This is a followup from issue #81 and it's last line, which wasn't included in the unit test created for that issue:

Query the director for an object within the new origin's namespace. Ensure the response includes the newly registered origin.

Techming:
Follow up on this issue and some breakdown of the tasks required:
* Add a new function to return a list of origins the director contains
* Use the newly created function to make a new test

Techming:
Closed with #185 merged.

PR #185, opened 2023-10-05
Query director for available namespaces

Techming:
According to issue #147 We want to make sure that when an origin was registered, the director cache actually contains the newly registered namespaces from the origin. Thus, to make the testing is possible, I added the ListNamespacesFromOrigins function, which as the name suggests, will return all the namespaces from all origin servers registered at the director.

With this function, I added the extra test clause in redirect_test.go to achieve the set goal.


turetske (approved):
LGTM!

PR #186, opened 2023-10-05
fixed typos within plugin_stage.go

joereuss12:
@jhiemstrawisc these are the typos we found way back when I was playing around with plugin staging. Realizing we never made these typos a permanent fix.


bbockelm (approved):
Reviewed the spelling used elsewhere in `plugin_stage.go`.  The changes match that.

@joereuss12 - before you merge, please create a new ticket for adding the `StagePlugin.*` parameters to the param table.

PR #191, opened 2023-10-06
Fix template rendering failure

bbockelm:
Additionally, add a simple regression test to ensure that the template at least compiles.


turetske (changes requested):
Fix the test if possible and I'll do another review. It's weird that it passes linux but fails on others.

turetske (approved):
LGTM

PR #193, opened 2023-10-09
Added check of cookie login for prometheus

turetske:
-- Refactored origin and federation check into their own module to be used by other functions
    -- Added a README explaining which tokens is it okay to sign with the Federation and which aren't
    -- Added a check for the authorization via the login cookie

### DIFF
--- a/web_ui/authorization.go
+++ b/web_ui/authorization.go
@@ -1,0 +1,136 @@
+/***************************************************************
+ *
+ * Copyright (C) 2023, Pelican Project, Morgridge Institute for Research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License.  You may
+ * obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ***************************************************************/
+package web_ui
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"errors"
+	"io"
+	"net/http"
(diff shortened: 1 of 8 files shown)


CannonLock:
LGTM

CannonLock (approved):
LGTM, in that it provides the promised function.

web_ui/authorization.go
```diff
@@ -0,0 +1,136 @@
...
+		if err = jwt.Validate(parsed, jwt.WithValidator(scopeValidator)); err != nil {
+			return
+		}
+
+		c.Set("User", "Federation")
+	}
+}
+
+// Checks that the given token was signed by the origin jwk and also checks that the token has the expected scope
+func OriginCheck(c *gin.Context, strToken string, expectedScope string) {
```
jhiemstrawisc, line 113: Can you adjust the two "check" functions introduced here so that they return errors in the event that one is encountered?

web_ui/prometheus_test.go
```diff
@@ -327,5 +346,54 @@ func TestPrometheusProtectionOriginHeaderScope(t *testing.T) {
...
+
+	// Set the request to go through the checkPromToken function
+	r.GET("/api/v1.0/prometheus/*any", checkPromToken(av1))
+
+	http.SetCookie(w, &http.Cookie{Name: "login", Value: string(signed)})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	//&http.Request{Header: http.Header{"Cookie": recorder.HeaderMap["Set-Cookie"]}}
```
jhiemstrawisc, line 391: If this isn't something we want to turn on, can you remove it?

jhiemstrawisc (changes requested):

web_ui/authorization.go
jhiemstrawisc: @turetske and I discussed the fact that we expect these functions to return errors in some cases, and that returning an error isn't necessarily indicative that something went wrong. While I think we should figure out a good way to handle actual errors in the future, doing so at this moment seems non-trivial, and we'd like to see this merged. Let's kick the can down the road.

jhiemstrawisc (approved):

PR #195, opened 2023-10-09
Testing common transport timeouts

joereuss12:
Created a unit test that tests functionality of common transport timeouts now that they can be configured. Did not include all timeouts since the test demonstrates functionality enough and it became too difficult to focus too much on.


jhiemstrawisc (approved):
While the regression tests don't test every option that's configurable via the `setupTransport` function, the nature of that function is such that if one option breaks, they probably all break. For that reason, I think it's fine that we test a few of the options until we can come back and cover more of them.

PR #196, opened 2023-10-09
Create Login Page for Post Initialization Login

CannonLock:


Techming (approved):
LGTM. Redirected to "Login with password" page when a user was authenticated using one-time code and setup the admin password, open a new browser session and revisit the `/view` endpoint. Entered the set password and got redirected back to the dashboard page.

PR #198, opened 2023-10-09
Strip "Bearer " from authz query and header if they're present

jhiemstrawisc:
I noticed while testing XRootD's ability to use a Pelican director as its redirector (defined via the `pss.origin` directive), that XRootD will take a token passed to it via an Authorization header and convert it directly into an authz URL query that gets passed to the director. For the director to return a valid redirect URL, the "Bearer " prefix must be stripped.

### DIFF
--- a/director/redirect.go
+++ b/director/redirect.go
@@ -81,8 +81,8 @@ func getAuthzEscaped(req *http.Request) (authzEscaped string) {
 		// even though it's coming via a URL
 		authzEscaped = strings.TrimPrefix(authzEscaped, "Bearer ")
 	} else if authzHeader := req.Header["Authorization"]; len(authzHeader) > 0 {
-		authzEscaped = url.QueryEscape(authzHeader[0])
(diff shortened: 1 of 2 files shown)


bbockelm:
@jhiemstrawisc - can you add a simple regression / unit test here?  Even if it's only capturing `getAuthzEscaped` and not the end-to-end, it'd capture a future regression.

bbockelm (changes requested):
Forgot to give a formal review, only typed the comment -- please add a simple regression test here.

bbockelm (approved):
I'm a little suspicious about automatically converting the
```
Authorization: token1234
```

(i.e., when it's missing the `Bearer` prefix)  Did we actually hit such a case?

Regardless, I think it's fine to go forward with it.

jhiemstrawisc:
Is the issue that we want to some kind of failure in the case that a header-embedded token is missing the `Bearer ` prefix? I have indeed managed to trigger XRootD to send an `Authorization: Basic blah` header, although I'm not sure what I did to achieve that.

Issue #150, opened 2023-09-26
Expose pelican version as a runnable flag


jhiemstrawisc:
Related to issue #21, it would be useful if the following command was supported:
`pelican --version`

It doesn't currently appear that there's a way to get the pelican version from the CLI without renaming the binary to `stashcp` which allows for `stashcp --version`

PR #199, opened 2023-10-09
Add --version flag to pelican-client

Techming:
According to issue #150 , we want to expose `--version` flag for pelican-client and display the version/build info to the user in a similar behavior as `stashcp` did. This PR addressed this issue but in a bit hacky way under various constraints from cobra.
* In order for all commands and subcommands to respond to a flag, we need to add logic to each individual command/subcommand in cobra. This can be done in a batch using `PersistentPreRun` [function](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#prerun-and-postrun-hooks), to add one handler for the root command and its child subcommands. However, cobra will not execute `PersistentPreRun` function for the command if the `Run` function of the command is not defined. See issue [2039](https://github.com/spf13/cobra/issues/2039). For us, many subcommands are only a level of hierarchy and won't handle any command logic themselves, such as `origin`. Adding `Run` functions will break it.
* One workaround is to handle the flag logic inside the `main` function of the command, i.e. in `root.go`. [However, the flag value won't be evaluated until `Execute()` function is called](https://github.com/spf13/cobra/issues/1066#issuecomment-604220701), which means we can't do anything using cobra in `root.go`
* This leads to the final solution this PR picked, which is to manually handle `--version` logic in `main.go` by reading the last command line argument. This means that for any command ending with `--version`, the version info will be popped up no matter what command/argument/flag before it was. This should be the expected behavior for the `--version` flag.
* I registered the flag in cobra so that the `--help` command will it as a global flag to users, but cobra didn't handle any of the actual logic.
* Also, `-v` shorthand was disabled because we had one reserved for `origin serve` where `-v` means the volume to attach to the origin server.

Another note for those who want to improve the test coverage for the `cmd` package. You might want to clear the `os.Args` before your test logic and restore it afterwards if you are not testing cobra but just the manually-added CLI logic. I had this issue with the test case in CI for Windows OS where the test will pass some weird arguments to my `main.go` (which wasn't even part of my testing code) and cause the test case fail by exiting the program with `command not found` error.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,20 +19,42 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 )
 
 func main() {
-	exec_name := filepath.Base(os.Args[0])
+	err := handleCLI(os.Args)
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func handleCLI(args []string) error {
+	exec_name := filepath.Base(args[0])
 	if exec_name == "stash_plugin" || exec_name == "osdf_plugin" || exec_name == "pelican_xfer_plugin" {
-		stashPluginMain(os.Args[1:])
+		stashPluginMain(args[1:])
 	} else if exec_name == "stashcp" {
 		err := copyCmd.Execute()
 		if err != nil {
-			os.Exit(1)
+			return err
 		}
 	} else {
+		// * We assume that os.Args should have minimum length of 1, so skipped empty check
+		// * Version flag is captured manually to ensure it's available to all the commands and subcommands
+		// 		This is becuase there's no gracefuly way to do it through Cobra
+		// * Note that append "--version" to CLI as the last argument will give the
+		// version info regardless of the commands and whether they are defined
+		// * Remove the -v shorthand since in "origin serve" flagset it's already used for "volume" flag
+		if args[len(args)-1] == "--version" {
+			fmt.Println("Version:", version)
+			fmt.Println("Build Date:", date)
+			fmt.Println("Build Commit:", commit)
+			fmt.Println("Built By:", builtBy)
+			return nil
+		}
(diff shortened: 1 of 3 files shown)


bbockelm:
On the Cobra side -- would it be better to take the approach here:

https://gianarb.it/blog/golang-mockmania-cli-command-with-cobra

where we invoke the cobra command object directly and provide it with a byte buffer?  E.g.,

```
	cmd := NewRootCmd()
	b := bytes.NewBufferString("")
	cmd.SetOut(b)
	cmd.SetArgs([]string{"--in", "testisawesome"})
	cmd.Execute()
	out, err := ioutil.ReadAll(b)
```

This is a little less thorough than invoking the top-level function (meaning it'd potentially miss bugs).  Would it simplify the unit test though?

Techming:
> On the Cobra side -- would it be better to take the approach here:
> 
> https://gianarb.it/blog/golang-mockmania-cli-command-with-cobra
> 
> where we invoke the cobra command object directly and provide it with a byte buffer? E.g.,
> 
> ```
> 	cmd := NewRootCmd()
> 	b := bytes.NewBufferString("")
> 	cmd.SetOut(b)
> 	cmd.SetArgs([]string{"--in", "testisawesome"})
> 	cmd.Execute()
> 	out, err := ioutil.ReadAll(b)
> ```
> 
> This is a little less thorough than invoking the top-level function (meaning it'd potentially miss bugs). Would it simplify the unit test though?

Yeah I think for Cobra this would be nice way to unit test instead of invoking the main function.

jhiemstrawisc (approved):
As is, it looks good to me. I'll approve for now and let you decide if you want to simplify the unit test. If you decide not to change anything, go ahead and merge when you get the chance. If you make changes, I'll re-review.

PR #207, opened 2023-10-10
Integrate component health status to Prometheus

Techming:
* Added a new Prometheus Gauge metric `last_update_component_health_status` with labels `component`, `status`, `message`, similar to existing in-memory data structure, and the value is Unix time in micro seconds
* This function is tested visually/using integrate test by calling `/query` API and verify the metric value can be distinguished between very small updates. Unit test for this function is a bit tricky and redundant given that 1. we can't mock the time value effectively. 2.Prometheus testutil only supports compare the result from a collector against an io stream but the times never match

### DIFF
--- a/metrics/health.go
+++ b/metrics/health.go
@@ -47,11 +47,15 @@ type (
 )
 
 var (
-	healthStatus               = sync.Map{}
-	PromHealthStatusLastUpdate = promauto.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "last_update_component_health_status",
-		Help: "Last update of components health status",
-	}, []string{"component", "status", "message"})
+	healthStatus     = sync.Map{}
+	PromHealthStatus = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "component_health_status",
+		Help: "The health status of various components",
+	}, []string{"component"})
+	PromHealthLastUpdate = promauto.NewGaugeVec(prometheus.GaugeOpts{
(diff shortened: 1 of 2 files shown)


metrics/health.go
```diff
@@ -44,7 +47,15 @@ type (
 )
 
 var (
-	healthStatus = sync.Map{}
+	healthStatus     = sync.Map{}
+	PromHealthStatus = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "component_health_status",
+		Help: "The health status of various components",
+	}, []string{"component"})
+	PromHealthLastUpdate = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
bbockelm, line 55: Should be a counter, not a gauge, as it will never decrease.

metrics/health.go
```diff
@@ -44,7 +47,15 @@ type (
 )
 
 var (
-	healthStatus = sync.Map{}
+	healthStatus     = sync.Map{}
+	PromHealthStatus = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "component_health_status",
```
bbockelm, line 52: Let's add `pelican_` as a prefix; otherwise, this is a pretty generic metric name.

bbockelm (changes requested):
Two small changes requested.

bbockelm (approved):
Tested revised code by hand - it's ready to go!

PR #208, opened 2023-10-10
Remove content length from upload to enable chunked

djw8605:
Removing the content length from a put changes the upload style from full file upload to a chunked encoded upload. A full file upload causes go to read in the entire file in memory while uploading, chunked does not.

**NOTE** chunked uploads only work after xrootd has merged and put into production xrootd/xrootd#2102

Corresponds to https://github.com/htcondor/osdf-client/pull/88


bbockelm:
@djw8605 - is there a way to check for server compatibility?  What happens when you try a chunk encoded PUT against an older server?

djw8605:
I suppose you could do a separate request to the xrootd server to determine the version.

If you do a chunked encoded PUT against an older (well, current as of now) server, it may or may not work.  It all depends on timing, buffers, bandwidth...  The larger the file, the more of a chance of failure.  Though, even small files frequently failed in my tests.

bbockelm:
@djw8605 - I'm tempted to merge this one.  The relevant server version has been released for about a month and I haven't seen any major complaints.  If we merge now, it'd make it into the v7.3.0 (early December) release, giving a few weeks for any straggling origins to upgrade.

Thoughts?

djw8605:
Sounds good to me, setting for ready for review.

turetske:
I suggest doing a rebase on main and running all the tests again. It's been over two months and many things have changed since then and I'd like the extra sanity check before merging in (I doubt it would fail, but just in case).

Issue #203, opened 2023-10-10
Refactor XrootD startup code


turetske:
Currently xrootd startup code is origin-specific - we want to move it to make it more generic

Separate out directory creation, config creation and running process mgmt

PR #216, opened 2023-10-11
Refactored xrootd specific code out of pelican/cmd

turetske:
This should close issue #203 as it takes out the xrootd setup/configuration from pelican/cmd.


bbockelm (commented):
Looks like a reasonable refactor of the code.

One thing that's not obvious to me: why did you create the function ` TestWriteOriginScitokensConfigAndXrootdConfig` that simply combines the two tests?  Wouldn't it be better to just have two distinct tests (to make it clearer what's being done)?

bbockelm:
Gotcha.

The failure due to a missing directory is the `xrootdRun` directory, right?

Instead of combining the tests, how about we create a temporary directory for the test and use `viper.Set()` to move `xrootdRun` to a new location?  I suspect the Scitokens test currently leaks its runtime directory... side-effects in tests are bad.

turetske:
@bbockelm Probably a lack of viper.reset() between tests

turetske:
Okay. This should fix the issue. Like I suspected, it really was just a lack of viper.Reset(), which I added to a few other tests in xrootd/authorization_tests.go. I'll try to better about making sure that's at the start of tests.

xrootd/authorization_test.go
```diff
@@ -157,4 +157,8 @@ func TestWriteOriginScitokensConfig(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.Equal(t, string(monitoringOutput), string(genCfg))
+
+	configPath, err := ConfigXrootd(true)
```
bbockelm, line 161: Now that we've revived the separate test for `ConfigXrootd`, can we remove these changes?

xrootd/xrootd_config_test.go
```diff
@@ -18,17 +18,19 @@
...
 
+	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestCreateXrootdConfig(t *testing.T) {
-	configPath, err := configXrootd()
+func TestXrootDConfig(t *testing.T) {
+	viper.Reset()
```
bbockelm, line 32: I don't understand how this works.  I would have expected something like this to be needed:

```
	dirname := t.TempDir()
	os.Setenv("PELICAN_XROOTD_RUNLOCATION", dirname)  # Or equivalent viper.Set call
	defer os.Unsetenv("PELICAN_XROOTD_RUNLOCATION")
	config_dirname := t.TempDir()
	viper.Reset()
	require.NoErr(t, err)
```

(and then the rest of the test)

bbockelm (commented):
Two more questions on the revised commits.

xrootd/xrootd_config_test.go
turetske: The issue is that before, the xrootdrunlocation was being set to a temporary directory by the scitokensconfig test that then gets destroyed, so when it tries to create the file, it can't because the directory doesn't exist. Currently the directory is unset, so it's "", which does exist. Which I realize definitely needs to be fixed in the way you are suggesting, but that's why it doesn't fail, because it can find the directory and create the file.

xrootd/authorization_test.go
turetske: Yeah. This was changed locally but didn't seem to have gotten pushed.

turetske:
@bbockelm Okay, now I believe it's truly fixed

bbockelm (approved):
LGTM!

Thanks @turetske - feels great to get all this helper code out of the command line module.

Issue #173, opened 2023-10-02
New "component" for origin health metrics -- federation integration


bbockelm:
Currently, if advertising to the director fails, then the "health API" still returns all green.

Let's add a new component to the health API, "federation", that indicates whether or not the origin is successfully advertised to the director.  If the periodic advertising fails, then we label this component as critical.

haoming29:
Closed with #217 merged

PR #217, opened 2023-10-11
Federation component health status

haoming29:
In #173, we want to monitor the advertising to federation status for origin. I added the "federation" component to the health status monitoring and mark it as "critical" if there's an error advertising to the federation, or "ok" otherwise.

I can confirm from visually testing the added metric with the origin UI running and that the federation status is returned by the `/health` api.

This PR did not come with a unit test as it requires us to mock "PeriodicAdvertiseOrigin" function but I can't do it without refactoring code and structure of `advertise.go` in origin_ui package, by adding interface and struct in that file, which will also cause refactors of code who imports those functions. Enable mocking functions in `advertise.go` can be a separate issue if we do want to improve test coverage for `origin_serve.go`.

### DIFF
--- a/origin_ui/advertise.go
+++ b/origin_ui/advertise.go
@@ -29,6 +29,7 @@ import (
 	"github.com/pelicanplatform/pelican/client"
 	"github.com/pelicanplatform/pelican/config"
 	"github.com/pelicanplatform/pelican/director"
+	"github.com/pelicanplatform/pelican/metrics"
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


bbockelm (changes requested):
NACK - doesn't work.

The `PeriodicAdvertiseOrigin` implementation launches a go routine; the return is whether launching the go routine failed (basically, it can never fail).

Inside the separate go routine is the real advertise and we're not capturing the result of that.

origin_ui/advertise.go
```diff
@@ -44,12 +45,27 @@ func PeriodicAdvertiseOrigin() error {
 		err := AdvertiseOrigin()
 		if err != nil {
 			log.Warningln("Origin advertise failed:", err)
+			if err = metrics.SetComponentHealthStatus("federation", "critical", "Error advertising origin to federation"); err != nil {
+				log.Warningln("Failed to update internal component health status:", err)
+			}
+		} else {
```
bbockelm, line 51: Here (and below), combine this into a single "else if".

bbockelm (changes requested):
One more small change I spotted.

Also, since it's a pretty small change with a few commits, let's squash and rebase before final approval.

bbockelm (approved):
Revisions LGTM.

Issue #226, opened 2023-10-13
Allow initialization on minimal environment


bbockelm:
v7.1.0 cannot be run as a `stash_plugin` because the config system errors out at `os.UserHomeDir()` when `$HOME` isn't set:

```
$ HOME= ~/stash_plugin -d -infile $PWD/stash_plugin.in -outfile /tmp/stash_plugin.out
Error: $HOME is not defined
```

Since this occurs in the `init()` function, the binary exits before the output file can be created (and has no place to put an error).

There's another error that occurs if `HOME` is defined but not a writable directory (a reasonably common situation). See:

```
$ HOME=/dev/null ~/stash_plugin -d -infile $PWD/stash_plugin.in -outfile /tmp/stash_plugin.out
Error: Failure when creating a directory for the monitoring data: mkdir /dev/null: not a directory
```

This is because `InitServer()` is called in the `init()` function in `origin_serve.go` (meaning that `InitServer` is _always_ called, regardless of what's done by the binary).  Rather, InitServer should probably be in a `PersistentPreRunE` function for the server-related objects initialized by cobra.

jhiemstrawisc:
@joereuss12 Can you confirm that the fixes we merged cover these failure cases in HTCondor?

joereuss12:
Yes, I tested it over on submit1 and it now works as `stash_plugin` without any issues and functioned how it should

jhiemstrawisc:
Closing as fixed by PR #230

PR #230, opened 2023-10-13
Update Pelican to not require a set $HOME in some configurations

jhiemstrawisc:
In some remote environments like HTCondor, we can't always assume that the user has a configured $HOME directory. Previously, this was enforced in both the root command and in the `main` package's chain of `init()` functions, preventing stash_plugin from running in HTCondor even though no $HOME directory is needed in that case.

This makes an unset $HOME a non-fatal warning at the root command level, and moves `InitServer()` into a pre-run for each cobra command that is actually responsible for starting a server. In the cases of servers, we do still require $HOME.

Closes issue #226

### DIFF
--- a/cmd/director.go
+++ b/cmd/director.go
@@ -19,6 +19,7 @@
 package main
 
 import (
+	"github.com/pelicanplatform/pelican/config"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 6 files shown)


joereuss12 (approved):
LGTM, tested on Submit1 and works with HTCondor as it should!

PR #233, opened 2023-10-13
Remove the collector from federation discovery

jhiemstrawisc:
We haven't used an HTCondor collector for some time now. Let's finish removing it.


bbockelm (approved):
LGTM.

PR #237, opened 2023-10-13
Fix executable alias not working on Windows

Techming:
This PR closes #124 where it supports `.exe` extension from windows and different uppercase/lowercase for the executable name and is able to distinguish between different executable alias.


matyasselmeci (approved):
LGTM

PR #242, opened 2023-10-17
Documentation for user installation of the Pelican Client

joereuss12:
First draft of user installation guide. Accepting any and all feedback since I am not too experienced in documentation.


docs/install.md
```diff
@@ -0,0 +1,31 @@
+Installing the Pelican Platform
+===============================
+
+This document explains how a user can download and install the Pelican client. 
+
+Installation Steps:
+--------------------
+
+### Install the Pelican Platform Binary
+Navigate to the [Pelican release page](https://github.com/PelicanPlatform/pelican/releases/). Download the proper binary for the system you are running on and select which version you would like to install.
```
jhiemstrawisc, line 10: I think we should provide a little more information here so that people who are less tech-savvy can figure out which platform they should download on their own. For example, it might not be immediately clear to people whether they should be downloading an RPM, a DEB or an APK file. For the .deb files, we call the x86_64 builds "amd64" which might also be unclear to people. For Mac folks, I wouldn't expect the average person to know that they should grab the Darwin builds. Can you perhaps make a table that people can use to pick the right download?

docs/install.md
```diff
@@ -0,0 +1,31 @@
...
+Navigate to the [Pelican release page](https://github.com/PelicanPlatform/pelican/releases/). Download the proper binary for the system you are running on and select which version you would like to install. 
+
+Our versions are built like so:
+For example: 7.1.2
+- 7 represents the latest major version release
+- 1 represents feature releases
+- 2 represents a bug fix/patch release
+
+### Extract the Binary
+Once the package has finished downloading, place it in your workspace and extract the binary
```
jhiemstrawisc, line 19: This comment is really only applicable to the `.tar.gz` files we release, and this step might be unclear to people. Can you offer some terminal commands that the average person with basic CLI experience could follow for extracting the binary?

Once done with that, we should also provide information for installing the `apk`, `rpm` and `deb` files on their respective systems. I think we're okay not getting into too much detail to explain how to build from source.

docs/install.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Test Functionality of the Pelican Platform
+Once extracted, make sure you are in the same directory as the **Pelican** executable. To test if everything works, we can do a simple **object copy** command:
```
jhiemstrawisc, line 22: When pelican is installed with the rpm, deb or apk, it should automatically be added to the path, so we don't actually need to be in the same directory in those cases. When the binary is added to the path, we can just do `pelican <sub commands>` wherever we are, just like with other terminal commands.

jhiemstrawisc (changes requested):

docs/install.md
turetske: One possibility is "check to ensure that the pelican executable is on your path with `which pelican` "

docs/install.md
```diff
@@ -0,0 +1,31 @@
...
+
+```console
+./pelican -f osg-htc.org object copy /osgconnect/public/osg/testfile.txt .
+testfile.txt 36.00 b / 36.00 b [=============================================================================================] Done!
+```
+
```
turetske, line 28: I think there should be an explanation as to what the flags mean. "-f" for example

turetske (changes requested):

docs/install.md
jhiemstrawisc: I'll cover more about the flags in the actual "usage" write up.

CannonLock (changes requested):
Can you merge with main and move this to the pages directory?

Also I think we will all be happier if we use # instead of '----' notation for headers. :)

docs/pages/install.md
```diff
@@ -0,0 +1,110 @@
...
+##### APK:
+To install APK packages on Alpine Linux, you can use the apk package manager:
+```console
+apk add pelican_amd64.apk
+```
+
+##### Tarballs:
+To install `.tar.gz` packages, you can extract with tar:
+```console
+tar -xzf pelican_Darwin_x86_64.tar.gz
```
jhiemstrawisc, line 81: We should probably make a note here that extracting the tarballs and zip files doesn't actually _install_ pelican -- it just extracts the binary. I know you cover it later, but I'm personally a fan of being explicit about things like this.

docs/pages/install.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+##### Zip:
+To install `.zip` packages, you can use [7zip](https://www.7-zip.org/) or other programs to unzip your file. Simply right click on the `.zip` file and extract all contents to a directory of your choice
```
jhiemstrawisc, line 85: Same as previous comment for the claim about installation.

docs/pages/install.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+##### For tarballs and zip, using the binary itself:
+Once extracted, make sure you are in the same directory as the **Pelican** executable. To test if everything works, we can do a simple **object copy** command:
+
+```console
+./pelican -f osg-htc.org object copy /osgconnect/public/osg/testfile.txt .
+testfile.txt 36.00 b / 36.00 b [=============================================================================================] Done!
+```
+**Note:** If you are running on windows, you will need to add '.exe' to './pelican' to be './pelican.exe ...'
```
jhiemstrawisc, line 106: Is this true? When I downloaded and extracted the tarball for Windows, the executable is already called `pelican.exe`.

jhiemstrawisc (changes requested):

docs/pages/install.md
joereuss12: yup, you're right. Thanks for catching that

turetske:
Should probably be ".mdx" rather than ".md" so that all are the same.

jhiemstrawisc (changes requested):
Other than this small change, I think we're good for now.

jhiemstrawisc (approved):
LGTM.

jhiemstrawisc:
The requested changes from other reviewers were already taken care of, but I can't indicate that to Github without dismissing the reviews. For the sake of getting this merged, I'm going to bypass branch protections so Cannon and Emma don't have to come back to this.

PR #243, opened 2023-10-18
Cleanup of pelican config files

joereuss12:
Fixes issues with the 3 config files used with Pelican: defaults.yaml, osdf.yaml, and pelican.yaml. Before this fix there was a bit of a mess of where the configs are set up with viper. This fix moves InitConfig to config.go and sets up all the config files in the priority they need to be with pelican.yaml taking higher precedence than defaults.yaml. Also fixed issue in object_copy.go with some weird issues with the -d flag. Only need to check for the -d flag with stashcp since it does not go through root, any other modes are checked within root/initconfig and trigger debug mode.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -295,6 +295,7 @@ func GetTransport() *http.Transport {
 }
 
 func InitConfig() {
+	viper.SetConfigType("yaml")
 	// 1) Set up defaults.yaml
 	err := viper.MergeConfig(strings.NewReader(defaultsYaml))
 	if err != nil {
@@ -308,8 +309,8 @@ func InitConfig() {
 			cobra.CheckErr(err)
 		}
 	}
-	if viper.GetString("config") != "" {
-		viper.SetConfigFile(viper.GetString("config"))
+	if configFile := viper.GetString("config"); configFile != "" {
+		viper.SetConfigFile(configFile)
 	} else {
 		home, err := os.UserHomeDir()
 		if err != nil {
@@ -323,7 +324,7 @@ func InitConfig() {
 		viper.SetConfigName("pelican")
 	}
 
-	viper.SetEnvPrefix(GetPreferredPrefix())
+	viper.SetEnvPrefix(prefix)
 	viper.AutomaticEnv()
 	// This line allows viper to use an env var like ORIGIN_VALUE to override the viper string "Origin.Value"
 	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
--- a/config/config.go
+++ b/config/config.go
@@ -311,10 +311,10 @@ func InitConfig() {
 	if viper.GetString("config") != "" {
 		viper.SetConfigFile(viper.GetString("config"))
 	} else {
-	home, err := os.UserHomeDir()
-	if err != nil {
-		log.Warningln("No home directory found for user -- will check for configuration yaml in /etc/pelican/")
-	}
+		home, err := os.UserHomeDir()
+		if err != nil {
+			log.Warningln("No home directory found for user -- will check for configuration yaml in /etc/pelican/")
+		}
 
 		// 3) Set up pelican.yaml (has higher precedence)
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -293,6 +294,51 @@ func GetTransport() *http.Transport {
...
+	}
+	// 2) Set up osdf.yaml (if needed)
+	prefix := GetPreferredPrefix()
+	if prefix == "OSDF" {
+		err := viper.MergeConfig(strings.NewReader(osdfDefaultsYaml))
+		if err != nil {
+			cobra.CheckErr(err)
+		}
+	}
+	if viper.GetString("config") != "" {
```
jhiemstrawisc, line 311: We can avoid an additional viper call here:
```
if configFile := viper.GetString("Config"); configFile != "" {
    viper.SetConfigFile(configFile))
}
```

config/config.go
```diff
@@ -293,6 +294,51 @@ func GetTransport() *http.Transport {
...
+		}
+
+		// 3) Set up pelican.yaml (has higher precedence)
+		viper.AddConfigPath(filepath.Join(home, ".config", "pelican"))
+		viper.AddConfigPath(filepath.Join("/etc", "pelican"))
+		viper.SetConfigType("yaml")
+		viper.SetConfigName("pelican")
+	}
+
+	viper.SetEnvPrefix(GetPreferredPrefix())
```
jhiemstrawisc, line 326: We can avoid an extra call to `GetPreferredPrefix` here --> we already have the value stored in a variable called `prefix`

config/config.go
```diff
@@ -293,6 +294,51 @@ func GetTransport() *http.Transport {
 	return transport
 }
 
+func InitConfig() {
+	// 1) Set up defaults.yaml
+	err := viper.MergeConfig(strings.NewReader(defaultsYaml))
```
jhiemstrawisc, line 299: Neither of these first two `viper.MergeConfig()` calls appear to actually load anything for me. I began by trying to serve a director using a binary named `osdf`. Doing this prints an error that the `Federation.TopologyNamespaceURL` isn't set -- even though it should be by osdf.yaml.

I followed up by adding a print statement right after the first merge statement:
`fmt.Println("Value of Xrootd.ManagerPort: ", viper.GetInt("Xrootd.ManagerPort"))`
This should print `1213` as it appears in defaults.yaml, but it actually prints `0` -- the default int when viper doesn't have another value to print.

As a final sanity check, I tried this print statement right after the first merge:
`fmt.Println("Entire Viper: ", viper.AllSettings())`
this printed:
```
 map[config: debug:true director:map[defaultresponse:] federation:map[discoveryurl:] issuerkey: issuerurl: namespaceurl: origin:map[exportvolume:] server:map[port:8445] stageplugin:map[hook:false mountprefix: originprefix: shadoworiginprefix: token:]]
```
To me this looks like we only have default values defined from the main module's stack of `init` functions.

config/config.go
```diff
@@ -293,6 +294,51 @@ func GetTransport() *http.Transport {
...
+	if err != nil {
+		cobra.CheckErr(err)
+	}
+	// 2) Set up osdf.yaml (if needed)
+	prefix := GetPreferredPrefix()
```
jhiemstrawisc, line 304: See previous comment -- these aren't actually getting any of the configs.

jhiemstrawisc (changes requested):
An additional comment - you have a test for `TestDefaultsYaml`, but you should probably also make sure `osdf.yaml` is tested as appropriate as well.

jhiemstrawisc:
Ahh, @joereuss12 I think I figured out what the source of your headache is with the calls to `MergeConfig()` on the embedded yamls -- viper has to be told how it should interpret the io reader because it can parse many types of configuration files. Adding `viper.SetConfigType("yaml")` as the very first line of `InitConfig` solved the issue for me.

jhiemstrawisc (approved):
LGTM. Thanks for cleaning up all of the config, @joereuss12. It's been a long time coming!

PR #245, opened 2023-10-18
Origin Installation Documentation

turetske:
First draft of origin installation documentation. Please feel free to comment.


CannonLock (changes requested):
Couple items to align with standard practice: 

- Merge with main and move this to /docs/pages/.
- Make the extension .mdx ( It isn't necessary but consistency is nice )
- Remove the top yaml
  - The docs framework will grab the file update time by default 
  - The title is presumed from file name, but you can also adjust with _meta.json file as specified in the linked docs of the README
  
This is a personal preference so no need to change but # are far superior for creating headers then the line ['---', '==='] notation. 

https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#headers

docs/origin_install.md
```diff
@@ -0,0 +1,154 @@
...
+   [University of Florida site](https://github.com/opensciencegrid/topology/blob/b14218d6e9d9df013a42e4d8538b2eeea615514c/topology/University%20of%20Florida/UF%20HPC/UFlorida-HPC.yaml#L250)
+   should set the following configuration:
+
+        set resourcename = UFlorida-XRD
+
+   For example, the XRootD service registered at the
+   [University of Florida site](https://github.com/opensciencegrid/topology/blob/b14218d6e9d9df013a42e4d8538b2eeea615514c/topology/University%20of%20Florida/UF%20HPC/UFlorida-HPC.yaml#L250)
+   should set the following configuration:
+
+        set resourcename = UFlorida-XRD
```
joereuss12, line 88: You have this example twice here.

docs/origin_install.md
```diff
@@ -0,0 +1,154 @@
...
+### Optional configuration
+
+The following configuration steps are optional and will likely not be required for setting up a small site.
+If you do not need any of the following special configurations, skip to
+[the section on using XRootD](#using-xrootd).
+
+
+#### Enabling multi-user support
+
+!!! warning "Requirements for XRootD-Multiuser with VOMS FQANs"
```
joereuss12, line 105: I don't know much about markdown so I could be wrong, but maybe reformat these warnings to look more visually appealing. Could just be github not recognizing '!!!' or something too.

joereuss12 (changes requested):
Just have a few suggestions but otherwise I think it looks pretty good. I haven't looked into installing an origin too much but could follow this along quite well. I feel everything should make sense especially for someone who would be an admin.

docs/origin_install.md
```diff
@@ -0,0 +1,154 @@
...
+    Support is available in `xrootd-voms 5.4.2-1.1`, in the OSG 3.6 repos, though it is expected in XRootD 5.5.0.
+    If you want to use multiuser, ensure you are getting `xrootd-voms` from the OSG repos.
+
+To install an XRootD Standalone server, run the following command:
+
+```console
+root@xrootd-standalone # yum install osg-xrootd-standalone
+```
+
+Configuring XRootD
```
jhiemstrawisc, line 41: Is there actually any requirement that an admin configure XRootD? Looking through the xrootd-related steps below, I think most of this stuff is specifically stuff we auto-configure via the origin.

docs/origin_install.md
```diff
@@ -0,0 +1,154 @@
...
+``` console
+root@xrootd-standalone # yum install xrootd-multiuser
+```
+
+Installing Pelican
+------------------
+
+Grab the appropriate binary for your system from the [pelican repository](https://github.com/PelicanPlatform/pelican/releases)
+
+Launching an Origin
```
jhiemstrawisc, line 125: We should probably be consistent with what we call this step in our CLI and say "Serving an Origin"

docs/origin_install.md
```diff
@@ -0,0 +1,154 @@
...
+-------------------
+
+To launch a pelican origin, run:
+
+```./pelican origin -f <federation> -v <volume>```
```
jhiemstrawisc, line 130: This is missing the `serve` sub-command, and the fact that the argument to the volume flag takes two pieces. We need ```pelican origin serve -f <federation url> -v <local directory>:<namespace prefix>```

docs/origin_install.md
```diff
@@ -0,0 +1,154 @@
...
+
+Where `<federation>` is the address of the federation the origin will be a part of and `<volume>` is the location where the data hosted on the origin will be mounted.
```
jhiemstrawisc, line 132: `<local directory>` -- the directory containing objects to be exported to the federation
`<namespace prefix>` -- the namespace at which files from `<local directory>` will be made available in the federation. For example, if I want to export `/test/foo.txt` to the federation, passing `-v /test:/my/namespace` will allow others to get `foo.txt` by doing an object copy against the path `/my/namespace/foo.txt`.

docs/origin_install.md
```diff
@@ -0,0 +1,154 @@
...
+
+[code entry page](code_entry.png)
+
+[password entry page](password_entry.png)
+
+You should now see a webpage that looks like so:
+
+[metrics page](metrics_view.png)
+
+This will refresh every 10 minutes with the xrootd health metrics so that, as an admin, you can check the status of your origin.
```
jhiemstrawisc, line 154: I'm not sure what we should say about registering the origin with the namespace registry. Eventually we want this to happen automatically. For now, the reality is that after serving the origin (which generates the asymmetric key pair needed by the registry) we also need to `pelican namespace register --prefix <namespace prefix> --namespace-url <registry url>`. This gives the registry the public key for the origin, and makes that public key available to all of the other services in the federation for verifying tokens.

jhiemstrawisc (changes requested):

docs/origin_install.md
turetske: Hmm. I can't see a good way to do this. At least in the markdown nextra uses

turetske:
Addressed the changes, unfortunately, these instructions will all probably change with Brian's latest PR, but it's correct as of now.

docs/pages/origin_install.mdx
```diff
@@ -0,0 +1,74 @@
...
+To launch a pelican origin, run:
+
+```./pelican origin serve -f <federation> -v <local_directory>:<namespace_prefix>```
+
+Where `<federation>` is the address of the federation the origin will be a part of, `<local directory>` is the directory containing objects to be exported to the federation, and `<namespace prefix>` is the namespace at which files from `<local directory>` will be made available in the federation.
+
+The first time the origin is started, you will see something that looks like the following:
+
+
+<ExportedImage width={1000} height={1000} src={"/pelican/origin_start.png"} alt={"Image of an origin startup terminal prompt with the initializaiton code"} />
```
jhiemstrawisc, line 56: The image and other console examples all show Pelican being run as root - is this something we require, or is it just an artifact from our developing in a rootly container?

docs/pages/origin_install.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+<ExportedImage width={1000} height={1000} src={"/pelican/invalid_certificate.png"} alt={"Image of a safari browser invalid certificate warning"} />
+
+Proceed despite the warning to get to the code entry page. Enter the code specified in the terminal and create a root metrics password. You should now see a webpage that looks like so:
+
+<ExportedImage width={1000} height={1000} src={"/pelican/metrics_view.png"} alt={"Image of prometheus metrics graphs for a Pelican origin"} />
+
+This will refresh every 10 minutes with the xrootd health metrics so that, as an admin, you can check the status of your origin.
+
+### Register the origin with the namespace-registry
```
jhiemstrawisc, line 70: This step should technically occur before serving the origin so that the health tests querying the director work. Can you move it up and explain what this step means?

jhiemstrawisc (changes requested):

jhiemstrawisc:
One other quick comment -- it looks like each tab in the rendered index just reads the file name. Can you name the document something like `serving_an_origin.mdx`?

jhiemstrawisc (approved):
LGTM

jhiemstrawisc:
@joereuss12 @CannonLock, when you get a second can you confirm that Emma made your requested changes and approve the PR? I can't merge until we've all signed off.

joereuss12 (approved):
LGTM!

CannonLock (approved):
LGTM

PR #250, opened 2023-10-19
Moved transport config options to parameters.yaml

joereuss12:
Moved the timeout options for transports into param documentation


jhiemstrawisc (changes requested):
Failing tests notwithstanding (I think this is an artifact of something else), two notes about this:
- First, we should keep the configuration levels at a maximum of two layers. That will mean changing `Transport.Dialer.Timeout` to `Transport.DialerTimeout` (we only want one period per name).
- Second, we need to go through the code and make sure that anywhere we do a `viper.Get()` on one of these, we instead use the param module.

jhiemstrawisc (approved):
Looks good to me, and it's great to be one step closer to finishing up documentation for all the parameters!

PR #252, opened 2023-10-19
Improve test of stash_plugin to be more like HTCSS environment

joereuss12:
This adds two test cases, when `HOME` is not set and when `HOME` is set to a non-writable dir and ensures we exit(0) from `stash_plugin -classad` for both those cases. Meant to mimic an environment like HTCondor.


jhiemstrawisc (approved):
I tested locally with the instructions Joe provided. Everything passed as expected, and the test looks like it does test what we care about. LGTM.

PR #253, opened 2023-10-19
Web UI improvements

### LINKED ISSUE #247: Improve bootstrapping of web UI
The Pelican web UI is really annoying if you're not a developer launching from a terminal.  It would be pretty painful if pelican was started as a systemd service where you have to go rooting around for the stdout.

1. Startup of the origin shouldn't be blocked by the web UI.  The origin should work reasonably well without it, no need to block.
2. The current passcode should be written to a file, mode 0600, in the config directory.  That simplifies the install instructions - you only need to look in a single location if you don't start the origin at the terminal.
3. The "reset admin password" functionality should be a sub-command of the origin.  This would allow a script to enable the web UI, for example if we're automating the origin install inside a container.
4. Create a config option for disabling the UI completely.

bbockelm:
Fixes #247


joereuss12:
It looks like when EnableUI is set to true or false, it always prompts (if origin password file does not exist):
```
Pelican admin interface is not initialized
To initialize, login at https://e66d18373dad:8444/view/initialization/code/ with the following code:
191229
```
The only difference that this config setting seems to make is adding this line:
`INFO[2023-10-24T20:30:23Z] Origin web UI is not enabled; skipping setup`

Otherwise the behavior of the EnableUI config seems very similar in both cases and with `-d` flag set the output is basically the same and the webpage being generated is the same. Is this the intended behavior?

joereuss12 (approved):
LGTM!

PR #254, opened 2023-10-19
Expose director's public key via Open ID style URLs

haoming29:
* Generate private/public key pair for director at the director's start
* Expose director's discovery url at `<directorURL>/.well-known/pelican-configuration` which contains the `jwks_uri` attribute for accessing the public keyset
* Expose director's public keyset uri at `<directorURL>/.well-known/public_signing_key.jwks`
* Added an API in `director_api.go` `LoadDirectorPublicKey` to access director's public key from anywhere in the application

### DIFF
--- a/cmd/director_serve.go
+++ b/cmd/director_serve.go
@@ -29,22 +29,10 @@ import (
 	"github.com/pelicanplatform/pelican/director"
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/web_ui"
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
-
-func generateTLSCertIfNeeded() error {
-
-	// As necessary, generate a private key and corresponding cert
-	if err := config.GeneratePrivateKey(param.Server_TLSKey.GetString(), elliptic.P256()); err != nil {
-		return err
-	}
-	if err := config.GenerateCert(); err != nil {
-		return err
-	}
-
-	return nil
-}
 
 func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 	log.Info("Initializing Director GeoIP database...")
@@ -61,8 +49,15 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 	}
 	go director.PeriodicCacheReload()
 
-	err := generateTLSCertIfNeeded()
+	// The director needs its own private key. If one doesn't exist, this will generate it
+	issuerKeyFile := param.IssuerKey.GetString()
+	err := config.GeneratePrivateKey(issuerKeyFile, elliptic.P256())
 	if err != nil {
+		return errors.Wrap(err, "Failed to generate director private key")
+	}
+
+	// Generate a TLS certificate if needed
+	if err := config.GenerateCert(); err != nil {
 		return err
 	}
 
@@ -80,7 +75,9 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 	}
 	log.Debugf("The director will redirect to %ss by default", defaultResponse)
 	engine.Use(director.ShortcutMiddleware(defaultResponse))
-	director.RegisterDirector(engine.Group("/"))
+	rootGroup := engine.Group("/")
+	director.RegisterDirector(rootGroup)
+	director.RegisterDirectorAuth(rootGroup)
 
 	log.Info("Starting web engine...")
 	go web_ui.RunEngine(engine)
--- a/director/authentication.go
+++ b/director/authentication.go
@@ -1,0 +1,61 @@
+package director
+
+import (
+	"encoding/json"
+
+	"github.com/gin-gonic/gin"
+	"github.com/pelicanplatform/pelican/config"
+	"github.com/pelicanplatform/pelican/param"
+	log "github.com/sirupsen/logrus"
+)
+
+type DiscoveryResponse struct {
+	JwksUri string `json:"jwks_uri"`
+}
+
+// Returns links related to director's authentication, including the link
+// to get the public key from the director
+func discoveryHandler(ctx *gin.Context) {
+	directorUrl := param.Federation_DirectorUrl.GetString()
+	if len(directorUrl) == 0 {
+		ctx.JSON(500, gin.H{"error": "Bad server configuration: Director URL is not set"})
+		return
+	}
+	rs := DiscoveryResponse{
+		JwksUri: directorUrl + "/.well-known/public-signing-key.jwks",
+	}
+	jsonData, err := json.MarshalIndent(rs, "", "  ")
+	if err != nil {
+		ctx.JSON(500, gin.H{"error": "Failed to marshal director's discovery response"})
+		return
+	}
+	// Append a new line to the JSON data
+	jsonData = append(jsonData, '\n')
+	ctx.Header("Content-Disposition", "attachment; filename=pelican-director-configuration.json")
+	ctx.Data(200, "application/json", jsonData)
+}
+
+// Returns director's public key
+func jwksHandler(ctx *gin.Context) {
+	issuerKeyfile := param.IssuerKey.GetString()
+	key, err := config.LoadPublicKey("", issuerKeyfile)
+	if err != nil {
+		log.Errorf("Failed to load director's public key: %v", err)
+		ctx.JSON(500, gin.H{"error": "Failed to load director's public key"})
+	} else {
+		jsonData, err := json.MarshalIndent(key, "", "  ")
+		if err != nil {
+			ctx.JSON(500, gin.H{"error": "Failed to marshal director's public key"})
+			return
+		}
+		// Append a new line to the JSON data
+		jsonData = append(jsonData, '\n')
+		ctx.Header("Content-Disposition", "attachment; filename=public-signing-key.jwks")
+		ctx.Data(200, "application/json", jsonData)
+	}
+}
+
+func RegisterDirectorAuth(router *gin.RouterGroup) {
+	router.GET("/.well-known/pelican-configuration", discoveryHandler)
+	router.GET("/.well-known/public-signing-key.jwks", jwksHandler)
+}
--- a/director/director_api.go
+++ b/director/director_api.go
@@ -18,6 +18,20 @@
 
 package director
 
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/pelicanplatform/pelican/config"
+	"github.com/pelicanplatform/pelican/param"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
 // List all namespaces from origins registered at the director
 func ListNamespacesFromOrigins() []NamespaceAd {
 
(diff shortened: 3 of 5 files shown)


cmd/director_serve.go
```diff
@@ -66,6 +67,13 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 		return err
 	}
 
+	// The director needs its own private key. If one doesn't exist, this will generate it
+	issuerKeyFile := param.IssuerKey.GetString()
+	err = config.GeneratePrivateKey(issuerKeyFile, elliptic.P521())
```
jhiemstrawisc, line 72: Due to limitations in other ares of Pelican, we can't support `elliptic.P521`. Can you bump this down to `elliptic.P256`?

A second note is that we also generate a private key in the func `generateTLSCertIfNeeded`. Can you remove it from there.

A third note is that ALL servers need a private key to work correctly. I vote we just remove this code from the Director (and from the origin and namespace registry, if it appears there) and instead move the key generation into `config.InitServer`. Then all servers will be guaranteed to generate keys in the same way.

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+		return nil, errors.Errorf("Director URL is unset; Can't load director's public key")
+	}
+	log.Debugln("Director URL:", directorUrlStr)
+	directorUrl, err := url.Parse(directorUrlStr)
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintln("Invalid director URL:", directorUrlStr))
+	}
+	directorUrl.Scheme = "https"
+
+	discoveryUrl, _ := url.Parse(directorUrl.String())
```
jhiemstrawisc, line 65: Elsewhere in Pelican we use `discoveryUrl` to indicate the Federation's discovery url. Can you modify the variable name to make it clear this is meant to be the Director's discovery url?

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+	discoveryUrl.Path = path.Join(".well-known/pelican-configuration", discoveryUrl.Path)
```
jhiemstrawisc, line 66: In this case, we're probably better off hardcoding the separator because we're dealing with a URL path, which is universal -- if you were to run this on Windows, `path.Join()` would join the paths with `\` instead of `/`

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+
+	tr := config.GetTransport()
+	client := &http.Client{Transport: tr, Timeout: time.Second * 5}
```
jhiemstrawisc, line 69: Any particular reason we set the timeout here instead of using the default? It's fine if there is a reason, but if that's the case can you explain with a quick comment?

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+
+	req, err := http.NewRequest(http.MethodGet, discoveryUrl.String(), nil)
+	if err != nil {
+
```
jhiemstrawisc, line 73: Can you remove the extra whitespace here?

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+		return nil, errors.Wrap(err, fmt.Sprintln("Failure when doing director metadata request creation for: ", discoveryUrl))
```
jhiemstrawisc, line 74: I don't think we actually need the sprintln:
`return nil, errors.Wrapf(err, "Failure when doing director metadata request creation for: %s", discoveryUrl.String()`

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+	}
+
+	result, err := client.Do(req)
+	if err != nil {
+
```
jhiemstrawisc, line 79: Extra whitespace

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+		return nil, errors.Wrap(err, fmt.Sprintln("Failure when doing director metadata lookup to: ", discoveryUrl))
+	}
+
+	if result.Body != nil {
+		defer result.Body.Close()
+	}
+
+	body, err := io.ReadAll(result.Body)
+	if err != nil {
+
```
jhiemstrawisc, line 89: Extra whitespace

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+		return nil, errors.Wrap(err, fmt.Sprintln("Failure when doing director metadata read to: ", discoveryUrl))
+	}
+
+	metadata := DiscoveryResponse{}
+
+	err = json.Unmarshal(body, &metadata)
+	if err != nil {
+
```
jhiemstrawisc, line 97: whitespace

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+		return nil, errors.Wrap(err, fmt.Sprintln("Failure when parsing director metadata at: ", discoveryUrl))
+	}
+
+	jwksUri := metadata.JwksUri
+
+	response, err := client.Get(jwksUri)
+	if err != nil {
+
```
jhiemstrawisc, line 105: whitespace

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+		return nil, errors.Wrap(err, fmt.Sprintln("Failure when requesting director Jwks URI: ", jwksUri))
+	}
+	defer response.Body.Close()
+	contents, err := io.ReadAll(response.Body)
+	if err != nil {
+
```
jhiemstrawisc, line 111: whitespace

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+		return nil, errors.Wrap(err, fmt.Sprintln("Failure when requesting director Jwks URI: ", jwksUri))
+	}
+	keys, err := jwk.Parse(contents)
+	if err != nil {
+
```
jhiemstrawisc, line 116: whitespace

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+		return nil, errors.Wrap(err, fmt.Sprintln("Failure when parsing director's jwks: ", jwksUri))
+	}
+	key, ok := keys.Key(0)
+	if !ok {
+
```
jhiemstrawisc, line 121: whitespace

director/director_api.go
```diff
@@ -33,3 +49,78 @@ func ListNamespacesFromOrigins() []NamespaceAd {
 	}
 	return namespaces
 }
+
+func LoadDirectorPublicKey() (*jwk.Key, error) {
```
jhiemstrawisc, line 53: I don't see that this function is called anywhere. Should something be hooked into it? Maybe in the jwks handler logic from `director/authentication.go`?

director/authentication.go
```diff
@@ -0,0 +1,40 @@
...
+type DiscoveryResponse struct {
+	JwksUri string `json:"jwks_uri"`
+}
+
+// Returns links related to director's authentication, including the link
+// to get the public key from the director
+func discoveryHandler(ctx *gin.Context) {
+	director := param.Federation_DirectorUrl.GetString()
+	rs := DiscoveryResponse{
+		JwksUri: director + "/.well-known/public_signing_key.jwks",
```
jhiemstrawisc, line 19: Three comments about this line:

I tested without setting the director URL, and we get here without a warning. When that happens, the jwksuri is just the path. We should add logic to make sure that doesn't happen. I think Brian B added a function somewhere in the last few weeks to generate a server's hostname -- in the event that no director URL is set, we should probably make sure to at least use that.

Second comment is easy -- can you look into whether these types of things should have a newline at the end? When I try curling the director for the key, I get a file that doesn't end in a newline, which goes against the POSIX standard (although I'm not confident that's something we need to follow here).

Third comment is also easy -- unless there's another reason I don't know about, we should probably stick to using `-` separators here instead of suddenly switching to `_`.

director/authentication.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Returns director's public key
+func jwksHandler(ctx *gin.Context) {
+	issuerKeyfile := param.IssuerKey.GetString()
+	key, err := config.LoadPublicKey("", issuerKeyfile)
+	if err != nil {
+		log.Errorf("Failed to load director's public key: %v", err)
+		ctx.JSON(500, gin.H{"error": "Failed to load director's public key"})
+	} else {
+		ctx.JSON(200, key)
```
jhiemstrawisc, line 33: This is another spot where the JSON we return doesn't end with a newline character. Not sure if it's needed, but it's probably not a bad thing to always have.

jhiemstrawisc (changes requested):

jhiemstrawisc:
One other quick note -- can you add regression tests for the functionality you introduce in this PR?

director/director_api.go
Techming: I referred to the similar behavior of [DiscoverFederation() function](https://github.com/PelicanPlatform/pelican/blob/e7fc4aa86870acd444e3f43b7388bba6a99748db/config/config.go#L169) in config.go. I agree we should just use default setting here.

cmd/director_serve.go
Techming: For this part, I will remove the code from generateTLSCertIfNeeded and change the encryption method. For the third note, I will open a separate issue for it.

director/director_api.go
Techming: This is a helper function for origin/other service to call to obtain the public key from the director. It won't be called anywhere in this PR but will be used for the following director-based health test PR.

director/authentication.go
Techming: For the first note, I throw 500 error instead of falling back to hostname just to be consistent with behaviors from other places.

For the third note, I was following the current osg-htc.org issuer's format where they have https://osg-htc.org/.well-known/pelican-configuration for discovery and https://osg-htc.org/osdf/public_signing_key.jwks for jwks. I can change to use `-` if we don't need to take consideration of this.

Techming:
> One other quick note -- can you add regression tests for the functionality you introduce in this PR?

Sure, although I'm not entirely sure how regression test work in our setting? I would be very helpful if we could schedule a time to talk about this!

From what I understand, it's to make sure this new feature won't break any existing functionality but I can't find a place where this feature might affect it. I can run a file transfer if that helps?

cmd/director_serve.go
Techming: Looking at `namespace_registry_serve.go` which uses `generateTLSCertIfNeeded`, there's also a duplicate private key generation: https://github.com/PelicanPlatform/pelican/blob/e7fc4aa86870acd444e3f43b7388bba6a99748db/cmd/namespace_registry_serve.go#L47 and https://github.com/PelicanPlatform/pelican/blob/e7fc4aa86870acd444e3f43b7388bba6a99748db/cmd/namespace_registry_serve.go#L54 And the encryption there is also `P521`. 

With that, I will restructure the code in `director_serve.go` so that logic from `generateTLSCertIfNeeded` is pulled to `serveDirector` and I will move `generateTLSCertIfNeeded` to `namespace_registry_serve.go` just so that it won't break anything. I will open another issue for consolidating/fixing encryption in`namespace_registry_serve.go`

jhiemstrawisc: Can you open another issue for switching any remaining `elliptic.P521` curves to `elliptic.P256`? I thought we had caught them all.

director/director_api.go
jhiemstrawisc: Ahh, gotcha.

director/director_api.go
```diff
@@ -33,3 +47,68 @@ func ListNamespacesFromOrigins() []NamespaceAd {
...
+	discoveryUrlStr := param.Federation_DirectorUrl.GetString()
```
jhiemstrawisc, line 52: Just to make greping the repo a little easier, can you modify `discoveryUrl` stuff to include some reference to `director`? We use `discoveryUrl` elsewhere to indicate the federation's discovery url

jhiemstrawisc (changes requested):

haoming29:
Tried to add a unit test for `LoadDirectorPublicKey` but it can't work. To mock the director's endpoint I used a mock server package [httpmock](https://github.com/jarcoal/httpmock), which only intercept the `http.DefaultClient` and we used a custom client, so it won't work: https://github.com/jarcoal/httpmock/issues/107#issuecomment-919959154 

Similarly I has the same issue with another famous HTTP mocking library [gock](https://github.com/h2non/gock). It supports custom client but you have to expose the client to the test function which I feel it's not necessary.

Therefore, we should hold on testing this function using unit test but should wait for our integration tests infra and test this over there.

jhiemstrawisc (approved):
LGTM

PR #257, opened 2023-10-22
Implement creation of a sharing URL

### LINKED ISSUE #24: [client] Pelican should generate a "sharing URL" for users

bbockelm:
This allows a user to create a "sharing URL", a URL that includes the authorization token but has minimal privileges.

The intended usage is something like this:
```
$ pelican object share osdf://chtc/staging/bbockelm/my_test_file
osdf:///chtc/staging/bbockelm/my_test_file?authz=$LONG_TOKEN_STRING
```

Includes test cases covering token generation and the director URL discovery but, honestly, it's a bit difficult to test end-to-end as the `director-cache.osg-htc.org` is not working with authenticated CHTC URLs right now.

Fixes #24


client/sharing_url.go
```diff
@@ -0,0 +1,106 @@
...
+			viper.Set("Federation.DirectorUrl", "")
+			viper.Set("Federation.DiscoveryUrl", discoveryUrl.String())
+			if err := config.DiscoverFederation(); err != nil {
+				return "", errors.Wrapf(err, "Failed to discover location of the director for the federation %s", objectUrl.Host)
+			}
+			if directorUrl = param.Federation_DirectorUrl.GetString(); directorUrl == "" {
+				return "", errors.Errorf("Director for the federation %s not discovered", objectUrl.Host)
+			}
+		}
+	} else if objectUrl.Scheme == "osdf" && configDirectorUrl == "" {
```
jhiemstrawisc, line 55: Small bug that I happened to hit here because I forgot to clear my config. If we have the federation defined in the config and we are using `osdf://` as our scheme, we skip this block and hit line 74, telling us the scheme `osdf://` is invalid -- even if the federation information from the config is identical to the OSDF configuration. The mere fact it's defined in the configuration causes this to fail.

client/sharing_url.go
```diff
@@ -0,0 +1,106 @@
...
+			return "", errors.Errorf("Director for the OSDF not discovered")
+		}
+	} else if objectUrl.Scheme == "" {
+		if configDirectorUrl == "" {
+			return "", errors.Errorf("Must provide a federation name for path %s (e.g., pelican://osg-htc.org/%s)", objectUrl.Path, objectUrl.Path)
+		} else {
+			directorUrl = configDirectorUrl
+		}
+	} else if objectUrl.Scheme != "" {
+		return "", errors.Errorf("Unsupported scheme for pelican: %s://", objectUrl.Scheme)
```
jhiemstrawisc, line 74: This isn't a change request, per say, but I wonder how we could support having schemeless sharing. If I do `osdf object share /chtc/staging/jhiemstra/chtc-auth-test.txt`, might it be reasonable to generate a sharing URL that points to a cache or an origin? For example, if I `curl -v https://origin-auth2000.chtc.wisc.edu:1095/chtc/staging/jhiemstra/chtc-auth-test.txt?authz=<TOKEN FROM RUNNING OBJECT SHARE>`, I also get the file. We could generate this particular sharing URL when no scheme is provided.

jhiemstrawisc (changes requested):
Some of the changes to configuration are conflicting with a recent PR Joe put together. Can you rebase?

client/sharing_url.go
bbockelm: Resolved in the latest push.

client/sharing_url.go
bbockelm: Not sure I'm following -- the prior conditional handles the case for schemaless sharing, no?

jhiemstrawisc: Ah, I think I tested in a way that pointed at the wrong director, which caused a failure I misinterpreted. After pointing back to the production director, this appears to work as intended.

jhiemstrawisc (approved):
LGTM

PR #267, opened 2023-10-23
Add some documentation on developing the website

CannonLock:


bbockelm (changes requested):
I don't particularly like having developer home directories in the source code -- is there some other way to express the `run.sh` script to make it relocatable?

bbockelm (approved):
Revised PR looks good to me.

PR #269, opened 2023-10-23
Replaced some remaining viper.Get* calls with the appropriate param.GetString() call

turetske:
Just some replacements to viper.Get* that didn't occur before. Should be a very quick review.


jhiemstrawisc (approved):
LGTM

PR #270, opened 2023-10-23
Added the StagePlugin.* params to the param table

joereuss12:
Kept them all named StagePlugin but changed the token one to be just Plugin.Token since future plugins might have a token as well.


turetske (approved):
LGTM!

Issue #232, opened 2023-10-13
Origin UI crashes when /health API returns with a missing component


haoming29:
Related to #231 , currently, If the `/api/v1.0/health` returns data that has missing fields in the component object, saying "cmsd", the web UI will crash with a blank page and error message saying "Client error: check the console for info". 

We might want to more gracefully handle such error and don't crash the whole application when there is some component health status missing.

PR #272, opened 2023-10-23
Change /health API components data from map to array

### LINKED ISSUE #231: Improve /health api returned data structure
If the `/api/v1.0/health` returns data that has missing fields in the `component` object, saying "cmsd", the web UI will crash with blank page and error message to check the console for info.

To make it more robust, we want to switch the data structure from an object to an array of objects for "component", so in stead of 

```
{
    "status": "critical",
    "components": {
        "cmsd": {
            "status": "ok",
            "last_update": 1697212808
        },
        "director": {
            "status": "critical",
            "message": "No director test report received within the time limit",
            "last_update": 1697212838
        },
    }
}
```

We might want to return:

```
{
  "status": "critical",
  "components": [
    {
      "name": "cmsd",
      "status": "ok",
      "last_update": 1697212808
    },
    {
      "name": "xrootd",
      "status": "critical",
      "message": "No director test report received within the time limit",
      "last_update": 1697212838
    }
  ]
}
```

So that we can handle optional component health status.

haoming29:
This PR should close #231 which changes the return value of `/health` API to make it more robust. The `components` field is changed from a map of components health status `"componentName": {}`, to an array of component health status: 

```json
"components": [
      {
          "name": "cmsd",
          "status": "ok",
          "last_update": 1698087191
      },
      {
          "name": "federation",
          "status": "critical",
          "message": "Error advertising origin to federation",
          "last_update": 1698087191
      }
]
```

In hope by doing this, we can eliminate errors from missing component health status at the front end.

**This PR should be merged jointly with the PR for #232.**

### DIFF
--- a/metrics/health.go
+++ b/metrics/health.go
@@ -29,6 +29,7 @@ import (
 
 type (
 	ComponentStatus struct {
+		Name       string `json:"name"`
 		Status     string `json:"status"`
 		Message    string `json:"message,omitempty"`
 		LastUpdate int64  `json:"last_update"`
@@ -41,8 +42,8 @@ type (
 	}
 
 	HealthStatus struct {
-		OverallStatus   string                     `json:"status"`
-		ComponentStatus map[string]ComponentStatus `json:"components"`
+		OverallStatus   string            `json:"status"`
+		ComponentStatus []ComponentStatus `json:"components"`
 	}
 )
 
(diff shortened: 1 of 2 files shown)


bbockelm:
@haoming29 I don't understand this approach.

Doesn't @CannonLock simply need to check if the key exists?  It's not clear to me why an array is inherently better than a map...

haoming29:
I thought with more and more components added to the metric, if we want to display all them on dashboard, it's more flexible to use array instead of map so that Cannon can just iterate over available components and display them. With map, Cannon has to manually add the name of the new components each time and I thought that's not very efficient. I could be wrong though.
________________________________

PR #268, opened 2023-10-23
Auto-generate config struct from parameters.yaml and provide REST API

haoming29:
Per #225 , This PR introduce a new origin RESTful API `/api/v1.0/config` that returns the configurations of the server. It reads the `docs/parameter.yaml` and parse it as a Go struct in `param/parameter_struct.go`. Later at the server/client init stage (InitServer/InitClient), it unmarshal Viper config into such struct and made available to UI via RESTful API.

A note to the reviewer, right now I placed the API at `origin_ui/origin.go` but we might want to put this route to `web_ui/ui.go` in the future given that this is a generic API for all servers. The reason I did this is that there seems to be no authentication middleware in `web_ui` and I don't want to expose server config to unauthorized user. In the long term, we should figure out auth mechanism for `web_ui`

Issue #72, opened 2023-08-31
Use go generator to minimally auto-create NPM output dir


bbockelm:
Right now, `src/out/*` has to match some file in order for `goreleaser` to successfully build.  In the GHA, we pre-create a file before running the linters.

We should migrate this to a `go generator` step so it doesn't break the build in other contexts.  Long term (or, possibly, as a side-effect of this work), we should run the `npm` command as a generator.

bbockelm:
Reassigning to @haoming29 in order to free up Cannon's time toward other UI issues.

PR #273, opened 2023-10-23
Move npm path placeholder creation from GHA to Go

haoming29:
According to #72, we want to move the "placeholder file" creation for GHA linter from the GHA to `go:generate`. I created a `generator/next_generator.go` to programmatically generate path and placeholder file to `origin_ui/src/out/placeholder`, removed the same logic in GHA, and updated the command that calls `go:generate`

However, this change, as well as #268 makes `generate` and `param` package grow out of its original scope. I suggest that we move the `main` function from `param_generator.go` to a separate file like `main.go` and potentially move the call to `go:generate` from `param/param.go` to `generate/main.go` so that the logic is more clear as we are not just generating files for params.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -23,8 +23,6 @@ jobs:
       - name: Generate placeholder files
         id: generate-placeholder
         run: |
-          mkdir -p origin_ui/src/out
-          touch origin_ui/src/out/placeholder
           go generate ./...
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
--- a/generate/next_generator.go
+++ b/generate/next_generator.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"log"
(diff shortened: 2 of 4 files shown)


bbockelm:
Why not put this in the origin_ui subdirectory? Seems good to keep the generator in the associated subdirectory, no?

haoming29:
I tried to move the `next_generate.go` to `origin_ui` but it broke all the generator pipes which won't generate `parameters.go` neither.

turetske:
> Why not put this in the origin_ui subdirectory? Seems good to keep the generator in the associated subdirectory, no?

@bbockelm Because a go generator must be in package main, so if it's moved into origin_ui, there will be conflicting package names.

haoming29:
Reverted to previous commit where generator files are at the same place.

bbockelm (approved):
I understand Haoming and Emma discussed this approach and all are happy.  Merging!

Issue #258, opened 2023-10-23
Investigate Failures in /chtc namespace for director-caches.osg-htc.org


bbockelm:
The current director instance doesn't return the expected credential generation headers for the `/chtc` namespace, breaking authenticated downloads with `pelican object copy`.

Investigate the issue and fix as necessary.

jhiemstrawisc:
Closed by PR #274

PR #274, opened 2023-10-23
Fix broken Director auth headers and add regression test

jhiemstrawisc:
Since the bug isn't super easy to catch in the diff, it was coming from:
```
cacheNS := NamespaceAd{}
cacheNS.Path = ns.Path
cacheNS.RequireToken = ns.UseTokenOnRead
cacheAdMap[cacheAd] = append(cacheAdMap[cacheAd], cacheNS)
```

I'm not sure why I ever did this because the appended value (`cacheNS`) should have just been `nsAd` like it is for the similar origin logic. I'll probably figure it out around the same time I figure out how many licks it takes to get to the center of a tootsie pop.

Once we're comfortable merging this, I'll patch it into 7.1 as this constitutes a bugfix.

Fixes issue #258

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -65,6 +65,38 @@ type (
 	}
 )
 
+func parseServerAd(server Server, serverType ServerType) ServerAd {
+	serverAd := ServerAd{}
+	serverAd.Type = serverType
+	serverAd.Name = server.Resource
+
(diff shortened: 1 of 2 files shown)


turetske (approved):
LGTM

PR #279, opened 2023-10-24
Fix some issues with Pelican RPMs

joereuss12:
Fixes only some of the issues with Pelican RPMs. The issues fixed are that the LICENSE and the README were world writable, so I made them read only. There was a typo with the homepage, and the license was changed to an accepted format for Apache 2.0.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -58,10 +58,10 @@ nfpms:
     file_name_template: '{{ .PackageName }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
     id: pelican
     vendor: OSG Consortium
-    homepage: https://github.com/PelicanProject/pelican
+    homepage: https://github.com/PelicanPlatform/pelican
     maintainer: Brian Bockelman <bbockelman@morgridge.org>
     description: Command-line copy tool for the Open Science Data Federation
-    license: Apache 2.0
+    license: ASL 2.0
     formats:
       - apk
       - deb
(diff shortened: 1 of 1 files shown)


matyasselmeci:
You can ignore the `E: no-binary` error. Checking for a binary is based on the FILECLASS RPM tag (you can view that for files in a package with something like `rpm -ql wget --fileclass`. nfpm (the GoReleaser plugin that makes RPMs) does not populate that tag so it's not something you can fix.

.goreleaser.yml
```diff
@@ -74,15 +74,27 @@ nfpms:
       rpm:
         contents:
           - src: LICENSE
-            dst: "/usr/share/doc/{{ .PackageName }}-{{ .Version }}/LICENSE.txt"
+            dst: "usr/share/doc/{{ .PackageName }}-{{ .Version }}/LICENSE.txt"
```
matyasselmeci, line 77: This change does not appear to do anything; the files in the RPM are still created with leading slashes, which you can check by running
```
rpm2cpio dist/pelican-7.1.1-next-1.arm64.rpm | cpio -t
```
for the pelican RPM the files start with `/`. You can compare that with an rpm you download via `yumdownloader` (so `yumdownloader wget` for example) and see that those have files that start with `./`.

I've also tried using `./` as the prefix (so `./usr/share/doc/...`) but that has the same results too.  So I think it's just another GoReleaser/NFPM limitation.

joereuss12: Okay, I can change it back then. Tim T's rpmlint was yelling at him because we were using absolute paths instead of relative but that could have just been something weird on his end, I'm not super sure

matyasselmeci: Maybe it was complaining about absolute paths as symlink targets? Those aren't allowed; fortunately that is something you can change - see https://github.com/htcondor/osdf-client/pull/86/files. (Pelican doesn't have symlinks yet but this is something to keep in mind for later since nFPM's example (https://goreleaser.com/customization/nfpm/) does it wrong.)

joereuss12: So should I not worry about it and keep the leading /'s since we don't have symlinks yet?

matyasselmeci: yep

joereuss12:
Alright so if we don't have to worry about the `no-binary` and the `no-group-tag` error, the only error left is `no-changelogname-tag` which I don't feel like is a fatal error especially since our changelog gets generated just fine within github. Therefore, I think this should be good to go

matyasselmeci:
Yep, agreed.

matyasselmeci (approved):

PR #282, opened 2023-10-24
Update Dockerfile

CannonLock:
Update Dockerfile
- Download the node_modules at root
- Set the NODE_PATH to look there


matyasselmeci (approved):
Tried it out and it works just fine. Thanks!

Issue #251, opened 2023-10-19
Browser text sometimes unreadable in browsers with dark mode enabled


jhiemstrawisc:
I'd consider this pretty low priority, but when I'm using the browser interface in Firefox with dark mode enabled, some of the text in form fields are barely visible (harder to see on my screen than it is in the image). It would be nice if this text could be color-coded dynamically like other text.
![Screenshot 2023-10-19 at 3 19 36 PM](https://github.com/PelicanPlatform/pelican/assets/75916364/bc7e5993-2291-4138-8541-e00dd08d11e7)

CannonLock:
This was bothering me as I worked on it so I fixed it here: #284

PR #284, opened 2023-10-25
Api view

CannonLock:
I will wait to merge until #268 is merged and then drop my final commit on that.


bbockelm:
@CannonLock - I think this is ready for a rebase now that #268 is in.

bbockelm:
Also -- before merging, can you writeup a corresponding issue?  I don't see any obvious one that this is associated with.

CannonLock:
Close #251

haoming29:
For some reason when I navigate to `/view/config` the page is blank, nothing is under `<body>` tag, no console error, can I can see `/view/config` returns 200 at backend. @CannonLock Do you think I missed anything in setting up?

origin_ui/src/app/(dashboard)/config/page.tsx
```diff
@@ -0,0 +1,379 @@
...
+import TextField from "@mui/material/TextField";
+import {ArrowDropDown, ArrowDropUp} from '@mui/icons-material';
+import {fontSize} from "@mui/system";
+
+interface Config {
+    ConfigDir: string;
+    Debug: boolean;
+    TLSSkipVerify: boolean;
+    IssuerKey: string;
+    Transport: {
```
haoming29, line 40: I think all the types under `Transport` are number not string

origin_ui/src/app/(dashboard)/config/page.tsx
```diff
@@ -0,0 +1,379 @@
...
+            disabled
+            size="small"
+            id={`${id}-text-input`}
+            label={name}
+            variant={"outlined"}
+            value={value}
+        />
+    }
+
+    if(typeof value === 'number' || value instanceof Number){
```
haoming29, line 163: Is it possible that we add unit for the values under `Transport`? Right now having 10 digit number as values for various "duration" fields can confuse people, or we might want to convert the durations into second when displaying them (I can do the conversion at the backend if that works better)

Screenshot 2023-11-02 at 9 07 57 AM

origin_ui/src/app/globals.css
haoming29, line 1: We might also want to change the tick color of the graph to be lighter. It's hard to see it in dark mode right now.

Ref:
Screenshot 2023-11-02 at 9 20 39 AM

origin_ui/src/app/(dashboard)/Sidebar.tsx
```diff
@@ -17,25 +17,44 @@
...
+                    </Box>
+                    <Box pt={1}>
+                        <Link href={"/config/"}>
+                            <IconButton>
+                                <BuildIcon/>
+                            </IconButton>
+                        </Link>
+                    </Box>
                 </div>
                 <div>
```
haoming29, line 59: It seems that GitHub icon is not aligned with others in the column. Not sure if it's an existing bug or not.

Ref:
Screenshot 2023-11-02 at 9 28 17 AM

haoming29 (changes requested):
Thanks for taking care of the dark mode! It looks fantastic and very cool. A couple of small changes requested plus the `/config/` page not showing up when navigated from index through the sidebar. 

I'm not entirely sure why we would want to check if there's a trailing slash before feeding the next.js `index.html`? I thought it's more common not to append the slash to the end of the page url.

origin_ui/src/app/(dashboard)/Sidebar.tsx
```diff
@@ -17,25 +17,44 @@
...
-                <div style={{display:"flex"}}>
-                    <Image
-                        src={PelicanLogo}
-                        alt={"Pelican Logo"}
-                        width={32}
-                        height={32}
-                    />
+            <div className={styles.header} style={{display: "flex", flexDirection: "column", justifyContent:"space-between", padding:"1rem", top:0, position:"fixed", zIndex:"1", overflow: "hidden", height: "100vh"}}>
+                <div style={{display:"flex", flexDirection: "column"}}>
+                    <Link href={"/index.html"}>
```
haoming29, line 36: This doesn't seem to be the correct way to patch the issue. Why do we need to append the `index.html` to call the URLs? Isn't this supposed to be handled by the web server (either Go or Nginx)? For running the origin without Nginx, Go should be able to direct all page traffic to`index.html`

CannonLock: I think we agreed the although it isn't great that we have to be explicit in our request this is a fix.

origin_ui/src/app/globals.css
CannonLock: I double checked and I have removed all the dark mode code. Do a hard reload to clear out the existing stylesheets.

CannonLock:
@haoming29 Lets move the enhanced view of duration type objects to the next milestone. #322

haoming29 (approved):
LGTM! But let's resolve the merge conflict against main first. I don't want to mess up with the discrepancy in `package.json` and @CannonLock you should have the best knowledge in resolving the conflicts.

PR #286, opened 2023-10-25
Fix dev container build failure by specify base image platform

haoming29:
Hot fix for the latest build-dev-container GHA failure on main branch as `scitokens-oauth2-server` base image isn't built for arm platform. Fix by specifying `--platform` on base image as `amd64`.


jhiemstrawisc (approved):
LGTM

PR #288, opened 2023-10-25
Add Web URL in origin's advertisement message

### LINKED ISSUE #285: Add API endpoint URL in origin's advertisement message
In implementing #175 , I figured that in order for director to send a HTTP request to origin to report the health test result, it must know the hostname and port number of origin's web engine, not the xrootd service. Currently, in the origin advertisement message send to director, we only include `URL` for file transfer which is the port where xrootd resides. We want to add an extra field, `WebUrl` for origin's web engine endpoint.

1. Add `WebUrl` field to OriginAdvertise here: https://github.com/PelicanPlatform/pelican/blob/497cd4dc14c05ad1a8af23bc196d494f4dda0b7c/director/origin_api.go#L43
2. Document the difference between the two fields
3. Add unit tests to ensure we didn't break anything

haoming29:
This should close #285 where we need origin's web engine URL as part of the origin advertisement so that the director can talk to origin's web api.

For reviewer, I did check that `AdvertiseOrigin()` now carries the correct port number for `WebURl`, which is `Server.Port`. I didn't add test to it because I can't get away from making errors in generating ad token in the test case and I think until we can mock the token generation function, we can't really test `AdvertiseOrigin()`.

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -94,6 +94,10 @@ func parseServerAd(server Server, serverType ServerType) ServerAd {
 	}
 	serverAd.URL = *serverUrl
 
+	// We assume that the WebURL we get from topology is the same as the origin XRootD URL for now
+	// Would like to have a separate field for WebURL in the future
+	serverAd.WebURL = *serverUrl
+
 	return serverAd
 }
 
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -94,9 +94,7 @@ func parseServerAd(server Server, serverType ServerType) ServerAd {
 	}
 	serverAd.URL = *serverUrl
 
-	// We assume that the WebURL we get from topology is the same as the origin XRootD URL for now
-	// Would like to have a separate field for WebURL in the future
-	serverAd.WebURL = *serverUrl
+	// We will leave serverAd.WebURL as empty when fetched from topology
 
 	return serverAd
 }
--- a/director/origin_api.go
+++ b/director/origin_api.go
@@ -42,7 +42,8 @@ import (
 type (
 	OriginAdvertise struct {
 		Name       string        `json:"name"`
-		URL        string        `json:"url"`
+		URL        string        `json:"url"`     // This is the url for origin's XRootD service and file transfer
+		WebURL     string        `json:"web_url"` // This is the url for origin's web engine and APIs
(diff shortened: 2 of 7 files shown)


director/advertise.go
```diff
@@ -94,6 +94,10 @@ func parseServerAd(server Server, serverType ServerType) ServerAd {
 	}
 	serverAd.URL = *serverUrl
 
+	// We assume that the WebURL we get from topology is the same as the origin XRootD URL for now
+	// Would like to have a separate field for WebURL in the future
+	serverAd.WebURL = *serverUrl
```
haoming29, line 99: On other caveat is that right now I assume that when we fetch from OSDF Topology to populate namespace cache, we can get the WebUrl from `Server.Endpoint`, the same as XRootD URL. It worth checking if this is true or not.

haoming29:
Reassigned review to @bbockelm as Justin is on sick leave

director/origin_api.go
```diff
@@ -42,7 +42,8 @@ import (
 type (
 	OriginAdvertise struct {
 		Name       string        `json:"name"`
-		URL        string        `json:"url"`
+		URL        string        `json:"url"`     // This is the url for origin's XRootD service and file transfer
+		WebURL     string        `json:"web_url"` // This is the url for origin's web engine and APIs
```
bbockelm, line 46: What happens when the `WebURL` isn't set?  I think we should have a test for when this isn't set (as this will be the case when importing from topology) and ensure that the `web_url` isn't set.

director/advertise.go
bbockelm: I don't think this is true.  I would simply just leave it unset (and we should ensure that, in a unit test, it's OK for this to be omitted).

1.  If the origin is not running Pelican, then there is no `WebURL`.
2. If the origin is running Pelican, then it should also advertise its own `WebURL`.

director/origin_api.go
haoming29: I thought `OriginAdvertise` struct is only used for pelican origin to register at director and the topology is not involved? I don't quite understand when `WebURL` will be empty in this struct. I do believe that `ServerAd` struct can have empty `WebURL` as that's where we set topology response to.

turetske (approved):

Issue #96, opened 2023-09-03
Auto-configure access for origin issuer


bbockelm:
We currently require the origin admin to somehow generate the scitokens.cfg used by xrootd.

However, for the simple cases (namespace registry is the issuer; origin has its own signing key), a reasonable scitokens.cfg can be auto-generated: simply allow the named issuer the ability to manage authorization for all exported directories.

bbockelm:
This got much easier starting with the work toward closing out #4.  There's now code for generating/manipulating and writing out the scitokens configuration:

https://github.com/PelicanPlatform/pelican/blob/main/xrootd/authorization.go#L223

That function could take a list of exported directories and setup the local issuer for those base directories.  The following would be pseudocode for the target entry in the config file:

```
[Issuer Origin]
issuer = {{.OriginUrl}}
base_path = {{ StringsJoin .ExportPaths ", " }}
```

Then, there's some question about how to map the remote requests to Unix users when the multi-user plugin is enabled.  I'd suggest simply enumerating the different options in the scitokens config as Pelican options for the origin.  See [this link](https://github.com/xrootd/xrootd/tree/master/src/XrdSciTokens) for information about the available scitokens configuration.

PR #299, opened 2023-10-30
Auto-configure origin issuer access

jhiemstrawisc:
A note -- I tried but failed to figure out how to use `mapstructure` to get a clean unmarshal of the data from our config. To get this moving, I just grabbed the values manually.

Closes issue #96

### DIFF
--- a/generate/param_generator.go
+++ b/generate/param_generator.go
@@ -63,6 +63,7 @@ func GenParamEnum() {
 	}
 
 	stringParamMap := make(map[string]string)
+	stringSliceParamMap := make(map[string]string)
 	intParamMap := make(map[string]string)
 	boolParamMap := make(map[string]string)
 	durationParamMap := make(map[string]string)
@@ -102,6 +103,8 @@ func GenParamEnum() {
 			fallthrough
 		case "string":
 			stringParamMap[name] = rawName
+		case "stringSlice":
+			stringSliceParamMap[name] = rawName
 		case "int":
 			intParamMap[name] = rawName
 		case "bool":
@@ -123,11 +126,12 @@ func GenParamEnum() {
 
 	// Generate the code based on the template
 	err = packageTemplate.Execute(f, struct {
-		StringMap   map[string]string
-		IntMap      map[string]string
-		BoolMap     map[string]string
-		DurationMap map[string]string
-	}{StringMap: stringParamMap, IntMap: intParamMap, BoolMap: boolParamMap, DurationMap: durationParamMap})
+		StringMap      map[string]string
+		StringSliceMap map[string]string
+		IntMap         map[string]string
+		BoolMap        map[string]string
+		DurationMap    map[string]string
+	}{StringMap: stringParamMap, StringSliceMap: stringSliceParamMap, IntMap: intParamMap, BoolMap: boolParamMap, DurationMap: durationParamMap})
 
 	if err != nil {
 		panic(err)
@@ -286,6 +290,10 @@ type StringParam struct {
 	name string
 }
 
+type StringSliceParam struct {
+	name string
+}
+
 type BoolParam struct {
 	name string
 }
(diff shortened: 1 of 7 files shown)


jhiemstrawisc:
@bbockelm, I wasn't totally confident about the scope of the assigned issue, so let me know if I'm off the mark on anything.

generate/param_generator.go
```diff
@@ -102,14 +103,16 @@ func GenParamEnum() {
...
+			stringSliceParamMap[name] = rawName
 		case "int":
 			intParamMap[name] = rawName
 		case "bool":
 			boolParamMap[name] = rawName
 		case "duration":
 			durationParamMap[name] = rawName
 		default:
-			errMsg := "UnknownType, add a new struct and return method to the generator or add it to one of the already handles types"
+			errMsg := fmt.Sprintf("UnknownType '%s', add a new struct and return method to the generator or add it to one of the already handles types", pType)
```
turetske, line 115: Noticed a grammar error, probably mine originally.  Change `handles` to `handled`

generate/param_generator.go
```diff
@@ -218,14 +222,16 @@ func GenParamStruct() {
...
+			goType = "[]string"
 		case "int":
 			goType = "int"
 		case "bool":
 			goType = "bool"
 		case "duration":
 			goType = "time.Duration"
 		default:
-			errMsg := "UnknownType, add a new struct and return method to the generator or add it to one of the already handles types"
+			errMsg := fmt.Sprintf("UnknownType '%s', add a new struct and return method to the generator or add it to one of the already handles types", pType)
```
turetske, line 234: Same grammar error as above

xrootd/authorization_test.go
```diff
@@ -153,7 +153,7 @@ func TestWriteOriginScitokensConfig(t *testing.T) {
 	err = os.WriteFile(scitokensCfg, []byte(toMergeOutput), 0640)
 	require.NoError(t, err)
 
-	err = WriteOriginScitokensConfig()
+	err = WriteOriginScitokensConfig([]string{"/foo/bar"})
```
turetske, line 156: Can you add a test that checks that the ScitokensConfig file looks like you except it to with this added? Or create a test that will change what the file looks like and check that it is what is expected?

xrootd/xrootd_config.go
```diff
@@ -263,7 +263,9 @@ to export the directory /mnt/foo to the path /bar in the data federation`)
 		return err
 	}
 
-	if err := WriteOriginScitokensConfig(); err != nil {
+	// Set up exportedPaths, which we later use to grant access to the origin's issuer.
+	exportedPaths := viper.GetStringSlice("Origin.NamespacePrefix")
```
turetske, line 267: Is this specific to origins, will a cache setup care about this?

turetske (changes requested):
I have a few questions and some minor changes and tests I would like to see.

xrootd/xrootd_config.go
jhiemstrawisc: I'm not sure. How does a cache handle setting up the scitokens config? Does this concept map to caches?

bbockelm: The caches will get their authorization config information from the namespace entries in the director -- it's going to look completely different from how the origins do this.

jhiemstrawisc: @turetske is there a change with this you'd like to see?

turetske: @jhiemstrawisc Nope. I was just confused.

xrootd/authorization_test.go
```diff
@@ -153,7 +169,7 @@ func TestWriteOriginScitokensConfig(t *testing.T) {
 	err = os.WriteFile(scitokensCfg, []byte(toMergeOutput), 0640)
 	require.NoError(t, err)
 
-	err = WriteOriginScitokensConfig()
+	err = WriteOriginScitokensConfig([]string{})
```
turetske, line 172: Can you add a test which does a write with a non-empty array and test that the file generated contains what is expected?

Or am I missing seeing that being tested somewhere?

jhiemstrawisc: Okay, with latest push both the generation of the `Issuer` struct and the file writing is tested.

turetske (approved):
LGTM!

Issue #201, opened 2023-10-10
Progress bar does not have correct totals


bbockelm:
The progress bar appears to be broken -- the "total" amount given is almost always smaller than the amount transferred.  See screenshot with v7.1.0:

<img width="1004" alt="image" src="https://github.com/PelicanPlatform/pelican/assets/1093447/b2c5d175-5740-4e4e-84c2-abbf997f6a11">

bbockelm:
Shouldn't be that -- when I download the same URL with `curl`, `curl` gets the correct length and its progress bar works!

joereuss12:
So from further investigation, I feel like that download you are performing is not returning `Content-Length` within its header (at least from trying to print it within Go, I could not figure out the curl command you used and check the header that way). I also found that for curl, if it cannot find the `Content-Length` value it uses other ways to calculate the total size (at least I am pretty sure). I'm working on some sort of fix right now so the progress bar looks better and does not appear to 'lag behind' for those types of downloads. I feel like no matter if I am right or wrong about that download not having `Content-Length` set, we should probably have something in place to work when it is not set.

I've come up with a few solutions:
1. I can dynamically update the total by doing some calculations with the currently downloaded bytes and the speed at which it is doing it. This way the total will always be higher than the current download time but won't really be correct until close to the end of the download
2. I can just make the total say `unknown` or a `-` or something
3. I can move updating the total until after the progress bar updates, this way the total will just equal the current amount downloaded whenever it updates

If I am wrong and the `Content-Length` header is actually set, then there might be some sort of issue within the function itself, I just keep getting either `-1` or an empty string when trying to print out the `Content-Length`

bbockelm:
@joereuss12 - I think the "problem" here is the fact we're setting the `X-Transfer-Status` and asking for the `trailers`.  That should enable the chunk encoding in the response; if chunk encoding is enabled, then the `Content-Length` header is omitted.

What about doing a `HEAD` request to get the size followed by the actual download?

PR #302, opened 2023-10-30
Fixed issues with Progress Bars

joereuss12:
Fixed issue when Content-Length header was ommitted when chunk encoding is enabled. Now if the original response size is not found it will do a HEAD request to get the proper `Content-Length`. The progress bars should always have the proper total now. 

Fixes issue #201


turetske (approved):
LGTM!

PR #295, opened 2023-10-28
Automatically register origin with the namespace daemon

bbockelm:
Fixes #171

PR #305, opened 2023-10-30
Director-based origin health tests

### LINKED ISSUE #175: Director-based health tests
In #167, we add some simple origin-based self-tests.  However, this only really touches the local host -- there's no indication whether the origin is open to the outside world.

The director has a list of all active origins and there's a federation key pair we can use to generate transfer tokens.  Let's create a new set of tests that have the director periodically upload/download from all the connected origins.

The results can be polled from the origin itself (allowing them to be incorporated into the origin's local monitoring dashboard) and exposed via Prometheus.

haoming29:
Closes #175 

## Test Instructions

Make sure in your `pelican.yaml`, the host of your `Federation.DirectorUrl` is the same as your `hostname`. If you run things in container, it should be a string instead of `localhost`

Have director & registry & origin running at debug mode (`-d`) and watch for the log message at director's end: 

```console
DEBUG[2023-11-07T14:48:11Z] Starting Director test for origin fae8c2865de4 at https://fae8c2865de4:8443 
DEBUG[2023-11-07T14:48:41Z] Starting a new Director test cycle for origin: fae8c2865de4 at https://fae8c2865de4:8443 
......
DEBUG[2023-11-07T14:48:41Z] Director test cycle succeeded at Tue Nov  7 14:48:41 UTC 2023
```

Also, at origin's end, the console message should reflect that the director uploaded/downloaded/deleted a file to/from origin and also you **shouldn't** see any errors from Xrootd suggesting a file upload is failed:

```console
INFO[2023-11-07T14:48:41Z] Served Request                                client=172.17.0.2 daemon=gin fields.time=1.792959ms method=POST resource=/api/v1.0/origin-api/directorTest status=200
```


haoming29:
When implementing this, I found that for director's scitoken to be verified by XRootD, its public key needs to be part of the jwks generated at the start of the xrootd (ref: https://github.com/PelicanPlatform/pelican/blob/42b92154b39a5304d84b34fc5f7ab40df1e7b039/xrootd/xrootd_config.go#L185) but it's impossible right now for us to manipulate this jwks as it's a pointer to interface. I will need #295 to be merged first which addressed this issue.

turetske (changes requested):
I have a few changes to request. I definitely think someone else should look over this as well, as it's a large PR and I'm sure I might have missed some things.

cmd/director_serve.go
```diff
@@ -61,6 +61,10 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 		return err
 	}
 
+	if err != nil {
```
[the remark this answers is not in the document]
haoming29: This should be a tragedy after a very large a merge conflict. Thanks for catching it!

origin_ui/origin_api.go
```diff
@@ -0,0 +1,188 @@
...
+	directorTimeoutTimer *time.Timer
+	// Duration to wait before timeout
+	// TODO: Do we want to make this a configurable value?
+	directorTimeoutDuration = 30 * time.Second
+	exitSignals             = make(chan os.Signal, 1)
+	exitLoop                = make(chan struct{})
+)
+
+func apiAuthHandler(ctx *gin.Context) {
+	authHeader := ctx.Request.Header.Get("Authorization")
```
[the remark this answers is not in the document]
haoming29: Yeah... This handler for right now only serves as a check for director's health test report, so I don't think we need to check authz. The caller check is at here around line 75 where `valid, err := director.VerifyDirectorTestReportToken(token)` checks if the token `issuer` is the director. With that being said, I think I will rename this handler to `directorApiAuthHandler` or something like that and ensure it's only used for checking requests from the director.

director/origin_api.go
```diff
@@ -185,6 +185,83 @@ func VerifyAdvertiseToken(token, namespace string) (bool, error) {
...
 
+// Create a token for director to report the health status to the
+// origin
+func CreateDirectorTestReportToken(originWebUrl string) (string, error) {
+	directorURL := param.Federation_DirectorUrl.GetString()
+	if directorURL == "" {
+		return "", errors.New("Director URL is not known; cannot create director test report token")
+	}
+
+	tok, err := jwt.NewBuilder().
```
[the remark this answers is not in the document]
haoming29: That's good to know. I will add WLCG components to the token. If this is the future, then token generated in #289 should also adhere to the WLCG profile.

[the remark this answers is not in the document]
haoming29: Looking at the link you pointed to me, I thought this is the token mainly for file transfer and we had a bunch of scope claims that are only for internal API calls, i.e. advertise/monitoring/reporting like `Claim("scope", "pelican.advertise")`. How can we adhere to WLCG profile while having our unique scopes?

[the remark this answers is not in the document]
turetske: Good point. Let's leave it for now.

haoming29:
I still can't make origin happy when accepting director's test files, with Brian's patch merged. I added director as an issuer to scitoken config. I tried to add director's public JWK but it turned out to be the same as the origin's JWK (which makes sense as they are generated based off of the same private key). Adding director's public JWK or not generates the same error as following:

**Director:**
```console
DEBUG[2023-11-06T17:26:34Z] Starting Director test for origin fae8c2865de4 at https://fae8c2865de4:8443 
DEBUG[2023-11-06T17:26:34Z] Starting a new Director test cycle for origin: fae8c2865de4 at https://fae8c2865de4:8443 
WARNING[2023-11-06T17:26:34Z] Director test cycle failed during test upload: Error response 403 from director test upload: 403 Forbidden 
```

**Director's generated scitoken:**
```
eyJhbGciOiJFUzI1NiIsImtpZCI6IjV3NE1ZYTRPUnFndjk4eUZleFpDZXFPZklacTE0QlRBMUJqX2w3Z0hDRVEiLCJ0eXAiOiJKV1QifQ.eyJhdWQiOlsiaHR0cHM6Ly9mYWU4YzI4NjVkZTQ6ODQ0MyJdLCJleHAiOjE2OTkyOTE2NTQsImlhdCI6MTY5OTI5MTU5NCwiaXNzIjoiaHR0cHM6Ly9sb2NhbGhvc3Q6ODg4OCIsImp0aSI6IlBuX0l2STB6RVZaNXNvVVFwaEVibWciLCJzY29wZSI6InN0b3JhZ2UucmVhZDovIHN0b3JhZ2UubW9kaWZ5Oi8iLCJzdWIiOiJkaXJlY3RvciIsIndsY2cudmVyIjoiMS4wIn0.NwaPezTY6NCi2oAuS8DJ_IKG_7_vXN8XuLO3KpveyUKgyIAVJP_0GwONv9Sa0so2hp8hkYsg85ABWIbW34zAEw
```

`/run/pelican/xrootd/export/.well-known/issuer.jwks`
```json
{
 "keys": [
  {
   "alg": "ES256",
   "crv": "P-256",
   "kid": "5w4MYa4ORqgv98yFexZCeqOfIZq14BTA1Bj_l7gHCEQ",
   "kty": "EC",
   "x": "LW_iiCcUz0kBnqQnVWQDlq4wsZu4A0u6ClH9kSfxOwU",
   "y": "KqXC0Lfx3BMsuPJWznQD00Q8xmafv56XeBrqF7DjyOk"
  }
 ]
}
```

And I confirm that the signature of the token can be verified by the public key

**Origin**
```console
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 scitokens_GenerateAcls: Failed to deserialize SciToken: SSL peer certificate or SSH remote key was not OK  daemon=xrootd
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 scitokens_Access: Failed to generate ACLs for token  daemon=xrootd
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 acc_Audit: unknown.2:36@fae8c2865de4 deny https *@[::ffff:172.17.0.2] create /pelican/monitoring/director-test-2023-11-06T17:26:34Z.txt  daemon=xrootd
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 scitokens_GenerateAcls: Failed to deserialize SciToken: SSL peer certificate or SSH remote key was not OK  daemon=xrootd
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 scitokens_Access: Failed to generate ACLs for token  daemon=xrootd
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 acc_Audit: unknown.2:36@fae8c2865de4 deny https *@[::ffff:172.17.0.2] excl_create /pelican/monitoring/director-test-2023-11-06T17:26:34Z.txt  daemon=xrootd
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 ofs_open: unknown.2:36@fae8c2865de4 Unable to create /pelican/monitoring/director-test-2023-11-06T17:26:34Z.txt; permission denied  daemon=xrootd
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 unknown.2:36@fae8c2865de4 Xrootd_Response: sending err 3010: Unable to create /pelican/monitoring/director-test-2023-11-06T17:26:34Z.txt; permission denied  daemon=xrootd
INFO[2023-11-06T17:26:34Z] 231106 17:26:34 3218 XrootdXeq: unknown.2:36@fae8c2865de4 disc 0:00:00 (send failure)  daemon=xrootd
```

Regarding the errors from the origin, I looked up the reason `SSL peer certificate or SSH remote key was not OK` which seems to be related to the public key but I'm not sure how to resolve it. 

@bbockelm Any thoughts on how I should navigate?

haoming29:
OK. After some investigation, I think this is an issue with self-signed CA and I shouldn't add director's public key to origin's `issuer.jwks` anyway because the intent should be let xrootd to query token's `iss` endpoint to get the issuer's JWK.

I believe in querying director's `iss` endpoint when scitoken tries to verify the token, there' s an SSL/SSH error and scitoken can't establish the connection to director's discovery endpoint. When I change the `iss` claim in director's token from `directorURL` to `originURL`, the error goes away and I get success for director's file upload to origin.

bbockelm:
@haoming29 - I think you need to rebase onto `main`.  Looks like you're picking up commits from an older branch.

bbockelm:
> OK. After some investigation, I think this is an issue with self-signed CA and I shouldn't add director's public key to origin's issuer.jwks anyway because the intent should be let xrootd to query token's iss endpoint to get the issuer's JWK.

I don't think it's an issue with the self-signed CA.  In that case, you usually get a different error message (more along the lines of "this CA isn't trusted").

> When I change the iss claim in director's token from directorURL to originURL, the error goes away and I get success for director's file upload to origin.

Ok, that's a hint.  What is the director URL versus the origin URL?  Is it possible the director URL is something like `https://localhost:8444` (not matching the certificate) whereas the origin URL is `https://fae8c2865de4:8443` (matching the certificate)?

director/origin_api.go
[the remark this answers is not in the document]
bbockelm: My $0.02:
- There's no problem with using custom scopes; this is commonly done.  The WLCG profile just outlines the known ones, it doesn't ban unknown ones.
- This is an internal API so we have flexibility in the future.  However, if we change to use WLCG tokens, we will need to think about version compatibility -- will an old director still work with a new origin?

So, provided that we're OK with potentially needing compatibility code, I'm OK with leaving this as-is.

haoming29:
> > OK. After some investigation, I think this is an issue with self-signed CA and I shouldn't add director's public key to origin's issuer.jwks anyway because the intent should be let xrootd to query token's iss endpoint to get the issuer's JWK.
> 
> I don't think it's an issue with the self-signed CA. In that case, you usually get a different error message (more along the lines of "this CA isn't trusted").
> 
> > When I change the iss claim in director's token from directorURL to originURL, the error goes away and I get success for director's file upload to origin.
> 
> Ok, that's a hint. What is the director URL versus the origin URL? Is it possible the director URL is something like `https://localhost:8444` (not matching the certificate) whereas the origin URL is `https://fae8c2865de4:8443` (matching the certificate)?

Yes! Hostname is the catch.

turetske (approved):
So this LGTM, but I want another pair of eyes to look at it as well.

origin_ui/origin_api.go
```diff
@@ -0,0 +1,189 @@
...
+		return errors.New("Origin configuration passed a nil pointer")
+	}
+	if err := metrics.SetComponentHealthStatus("director", "warning", "Initializing origin, unknown status for director"); err != nil {
+		return err
+	}
+	// start the timer for the director test report timeout
+	resetDirectorTimeoutTimer()
+
+	// When program exits
+	signal.Notify(exitSignals, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
```
bbockelm, line 169: As follow-up work, it's probably better to provide `ConfigureOriginAPI` a context object that can be cancelled on shutdown.

Having signal handlers scattered throughout the code is a bit scary -- too much indirect action -- and are fairly difficult to unit test compared to a context.

origin_ui/origin_api.go
```diff
@@ -0,0 +1,189 @@
...
+	dt := DirectorTest{}
+	if err := ctx.ShouldBind(&dt); err != nil {
+		log.Errorf("Invalid director test response")
+		ctx.JSON(400, gin.H{"error": "Invalid director test response"})
+		return
+	}
+	// We will let the timer go timeout if director didn't send a valid json request
+	resetDirectorTimeoutTimer()
+	if dt.Status == "ok" {
+		if err := metrics.SetComponentHealthStatus("director", "ok", fmt.Sprintf("Director timestamp: %s", dt.Timestamp)); err != nil {
```
bbockelm, line 141: For future work -- I hate all the strings in this interface.  We should cast it to an enum.

IIRC, the reason why `SetComponentHealthStatus` returns an error is because it internally casts to an enum.  If we forced type safety on the caller, then we could eliminate the error return and simplify the code base.

origin_ui/origin_api.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/pelicanplatform/pelican/metrics"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+type (
+	DirectorTest struct {
+		Status    string `json:"status"`
+		Message   string `json:"message"`
+		Timestamp string `json:"timestamp"`
```
bbockelm, line 41: Future work: this might be better-served as a Unix timestamp (and hence an unsigned 64-bit integer).

director/authentication.go
```diff
@@ -10,9 +10,15 @@ import (
 )
 
 type DiscoveryResponse struct {
+	Issuer  string `json:"issuer"`
 	JwksUri string `json:"jwks_uri"`
 }
 
+var (
```
bbockelm, line 17:
```suggestion
const (
```

xrootd/authorization.go
```diff
@@ -305,11 +318,14 @@ func WriteOriginScitokensConfig(exportedPaths []string) error {
 		cfg.Issuers = append(cfg.Issuers, issuer)
 		cfg.Global.Audience = append(cfg.Global.Audience, issuer.Issuer)
 	}
-
```
bbockelm, line 308: In the future, please avoid spurious whitespace changes.

director/origin_monitor.go
```diff
@@ -0,0 +1,320 @@
...
+		return "", err
+	}
+
+	return string(signed), nil
+}
+
+func UploadTestfile(originUrl string) (string, error) {
+	tkn, err := generateMonitoringScitoken(originUrl)
+	if err != nil {
+		return "", errors.Wrap(err, "Failed to create a token for the diretor test upload")
```
bbockelm, line 106:
```suggestion
		return "", errors.Wrap(err, "Failed to create a token for the director test upload")
```

director/origin_monitor.go
```diff
@@ -0,0 +1,320 @@
...
+		serverAdMutex.RLock()
+		if !serverAds.Has(originAd) {
+			log.Debug(fmt.Sprintf("End of director test cycle for origin: %s at %s", originName, originUrl))
+			serverAdMutex.RUnlock()
+			return
+		}
+		serverAdMutex.RUnlock()
+
+		log.Debug(fmt.Sprintf("Starting a new Director test cycle for origin: %s at %s", originName, originUrl))
+		url, err := UploadTestfile(originUrl)
```
bbockelm, line 289: Future work: I don't like how much duplication there is between here and the origin self-test.  Feels like it's something that could have been factored out.

director/origin_monitor.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+		// End the periodic test if the ttlcache doesn't have the origin ad
+		// meaning the origin is offline and the cache is expired
+		serverAdMutex.RLock()
+		if !serverAds.Has(originAd) {
+			log.Debug(fmt.Sprintf("End of director test cycle for origin: %s at %s", originName, originUrl))
+			serverAdMutex.RUnlock()
+			return
+		}
+		serverAdMutex.RUnlock()
```
bbockelm, line 286: Better to remove this and change to `defer serverAdMutex.RUnlock()`.  If anything in the control block above panics, you would end up with the lock held forever and potentially deadlock in the future.

director/origin_monitor.go
```diff
@@ -0,0 +1,320 @@
...
+			time.Sleep(5 * time.Second)
+			log.Debug(fmt.Sprintf("Starting Director test for origin %s at %s", originName, originUrl))
+			firstRound = false
+		} else {
+			time.Sleep(15 * time.Second)
+		}
+		// End the periodic test if the ttlcache doesn't have the origin ad
+		// meaning the origin is offline and the cache is expired
+		serverAdMutex.RLock()
+		if !serverAds.Has(originAd) {
```
bbockelm, line 281: This means that, if an origin is decommissioned, we'd still get a log line about it forever, right?  Seems better to timeout...

director/redirect.go
```diff
@@ -389,8 +389,17 @@ func registerServeAd(ctx *gin.Context, sType ServerType) {
...
+	hasOriginAdInCache := serverAds.Has(sAd)
 	RecordAd(sAd, &ad.Namespaces)
 
+	// Start director periodic test of origin's health status if origin AD
+	// has WebURL field AND it's not already been registered
+	serverAdMutex.RLock()
+	if ad.WebURL != "" && !hasOriginAdInCache {
+		go PeriodicDirectorTest(sAd)
+	}
+	serverAdMutex.RUnlock()
```
bbockelm, line 401: Defer this instead.

director/redirect.go
```diff
@@ -389,8 +389,17 @@ func registerServeAd(ctx *gin.Context, sType ServerType) {
 		Type:    sType,
 	}
 
+	hasOriginAdInCache := serverAds.Has(sAd)
 	RecordAd(sAd, &ad.Namespaces)
 
+	// Start director periodic test of origin's health status if origin AD
+	// has WebURL field AND it's not already been registered
+	serverAdMutex.RLock()
+	if ad.WebURL != "" && !hasOriginAdInCache {
```
bbockelm, line 398: What happens if the origin disappears from the cache, the cache gets garbage collected, then it comes up again?  As far as this logic goes, we now have a "new" origin and thus a second goroutine testing it.

Why not pass the go routine a `context.WithCancel(context.Background()), keep a cancel function in the cache, and then call the `cancel()` on invalidation?  That would remove the running goroutine when caches disappear.

bbockelm (changes requested):
Almost there.

Review notes a number of "future work" items.  You're welcome to either fix them or create a follow-up issue with them noted.

The critical item is thinking about how to better stop testing origins when they disappear from the director... I think there's some bugs around the current code.

director/redirect.go
haoming29: I thought the logic in `origin_monitor.go` will return and end the goroutine when origin disappears from the cache.

director/origin_monitor.go
haoming29: I didn't get why there would be a log line if the origin is evicted from the cache. But I agree it's safer and more robust to end the goroutine in `onEviction` handler of the cache. Will make the changes.

haoming29:
With code review feedback, I will integrate most of the suggestions in this PR but leave the followings as future works and will create separate PRs for them:
* Refactor file test logic into a central place
* Provide `ConfigureOriginAPI` a context object that can be cancelled on shutdown
* Refactor `SetComponentHealthStatus` to use enum as status code instead of string literals

director/cache_ads_test.go
```diff
@@ -176,3 +180,62 @@ func TestGetAdsForPath(t *testing.T) {
...
+		}()
+
+		serverAds.Delete(mockPelicanOriginServerAd) // This should call onEviction handler and close the context
+
+		require.False(t, serverAds.Has(mockPelicanOriginServerAd), "serverAds didn't delete originAd")
+
+		// OnEviction is handled on a different goroutine than the cache management
+		// So we want to wait for a bit so that OnEviction can have time to be
+		// executed
+		time.Sleep(2 * time.Second) // This is a bit flaky and can delay the test run, so any thoughts on improvement is very welcome
```
haoming29, line 237: As comment says, any thoughts on improving this is very welcomed.

director/redirect.go
haoming29: Introduced a new map to keep all cancel functions as the cache key needs to be comparable while the cancelFunc is not.

director/cache_ads_test.go
bbockelm: I would suggest something like this:

```
                 cancelChan := make(chan int)
		go func() {
			<-ctx.Done()
			if ctx.Err() == context.Canceled {
				cancelChan <- 1
			}
		}()

                 ...

                 select {
                 case <-cancelChan:
                 	require.True(t, true)
                 case <-time.After(3 * time.Second):
                 	require.True(t, false)
                 }
```

So, instead of polling (and in this case, once), you get a notification via the channel from the other go routine that the eviction is done.  It also includes a reasonable timeout so, on failure, the test doesn't hang forever.

bbockelm (approved):

Issue #205, opened 2023-10-10
Implement “pelican cache serve”


turetske:
Implement this for “public” cache only

Issue #204, opened 2023-10-10
Programmatically Generation simple config


turetske:
Get a cache config example
Get a cache running by hand
Read/understand config
Programmatically generation simple config

PR #306, opened 2023-10-31
Pelican cache noauth

turetske:
This implements the pelican cache serve command and starts up a pelican cache with no authentication. Some of the code is cludgy in an attempt to get functionality and proof of concept in before fixing/streamlining.

Each commit gives a pretty good step-by-step overview of the changes. This will resolve issue #205, #204, #206 

### Cache Config Generation

[Added xrootd cache config generation](https://github.com/PelicanPlatform/pelican/commit/f15ecaa32a192bfe255503084a76a75b8f4283cc) 
    -- Renamed xrootd/resources/xrootd.cfg to xrootd/resources/xrootd-origin.cfg
    -- Added xrootd/resources/xrootd-cache.cfg
    -- Adjusted xrootd/xrootd_config.go to use the correct cfg template depending on whether we are in pelican cache or pelican origin mode
    
 This addresses issue #204 (Note that the issue with the hardcoded directorUrl in xroot-cache.cfg is addressed in a later commit
    

### Pelican Cache Serve Command

[Added the "pelican cache serve" command to the cli](https://github.com/PelicanPlatform/pelican/commit/9b8f7eef53c3f9081454759554ba6134daa66011)

Adds the pelican cache serve cobra command

### Register the cache prefix with the namespace registry

[Added cache prefix registration for the pelican cache](https://github.com/PelicanPlatform/pelican/commit/ad66e4199c24fcf3a92dcbcfc016e55889a2f8eb)

This registers the cache prefix `"/caches/<XrootD.SiteName>"` with the namespace registry. This is done within `cacheServe` rather than as a separate command line call to `pelican namespace register`.

Note that the check to see if the prefix is already registered is cludgy and could probably be done better.

### Advertise to the director

[Added cache advertisement to the director](https://github.com/PelicanPlatform/pelican/commit/0eaeeb6f783fba7e346886f35667c2e98868340b) 
    -- Added a director endpoint to access the list of all origin spaces
    -- Adjusted the namespaceAds and serverAds to be marshalled into json
    -- Fixed redirect tests to be in line with adjusted structs
    -- Cache serve now grabs the origins that are advertised to the director and implements periodic cache advertisement
        -- Note, this is normally done via the web_ui, but this first implementation is without the web_ui
        
At some point, we will want to implement a way for the cache to choose which origin namespaces it can access, but it currently just gets all the ones that the director has advertised to it. 

### Generate Authfile

[pelican cache now gets auth from director](https://github.com/PelicanPlatform/pelican/commit/77963b001aebe95bd181792357b2571c626ffe2f) 
    -- Uses the namespaces advertised from the director to populate the auth file

Addressed issue #206


turetske:
I realize when looking through PR #295 does the namespace registration much better. When addressing other PR reviews, I'll refactor to mimic that.

cmd/cache_serve.go
```diff
@@ -0,0 +1,286 @@
...
+
+	return body, nil
+}
+
+func periodicAdvertiseCache(prefix string, nsAds []director.NamespaceAd) error {
+	ticker := time.NewTicker(1 * time.Minute)
+	go func() {
+		err := advertiseCache(prefix, nsAds)
+		if err != nil {
+			log.Warningln("Vsvhr advertise failed:", err)
```
bbockelm, line 85: What's `Vsvhr` mean here?

cmd/cache_serve.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	req, err := http.NewRequest("POST", directorUrl.String(), bytes.NewBuffer(body))
+	if err != nil {
+		return errors.Wrap(err, "Failed to create POST request for director registration")
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "Bearer "+token)
+	userAgent := "pelican-origin/" + client.ObjectClientOptions.Version
```
bbockelm, line 164: Should be `cache`, not `origin`, right?

cmd/cache_serve.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	// TODO: waiting on a different branch to merge origin URL generation
+	// The checkdefaults func that runs before the origin is served checks for and
+	// parses the originUrl, so it should be safe to just grab it as a string here.
+	originUrl := param.Origin_Url.GetString()
+
+	// TODO: Need to figure out where to get some of these values
+	//               so that they aren't hardcoded...
+	cAd := director.NamespaceAd{
```
bbockelm, line 123: The namespace ad here doesn't make much sense, does it?  Can we skip it?

cmd/cache_serve.go
```diff
@@ -0,0 +1,286 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Failed to create POST request for director registration")
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "Bearer "+token)
+	userAgent := "pelican-origin/" + client.ObjectClientOptions.Version
+	req.Header.Set("User-Agent", userAgent)
+
+	// We should switch this over to use the common transport, but for that to happen
```
bbockelm, line 167: Copy/paste comment?  This is using the common transport!

docs/parameters.yaml
```diff
@@ -336,6 +336,27 @@ default: none
 components: ["origin"]
 ---
 
+############################
+#   Cache-level configs    #
+############################
+name: Cache.DataFileLocation
```
bbockelm, line 342: It's not a "data file", is it?  It assumes that there's a directory -- how about `Cache.DataLocation`?

docs/parameters.yaml
```diff
@@ -336,6 +336,27 @@ default: none
...
+description: >-
+  The filepath of the oss.space files for a cache instance. This should *not* be in the same path as XRootD.Mount or
```
bbockelm, line 344: There's a lot of XRootD jargon here.  Can we demystify it?

bbockelm (changes requested):
Didn't make my way through the whole thing but figured you'd appreciate the first round of review comments!

cmd/cache_serve.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+type directorResponse struct {
+	Error string `json:"error"`
+}
+
+func makeRequest(url string, method string, data map[string]interface{}, headers map[string]string) ([]byte, error) {
```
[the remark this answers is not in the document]
turetske: Yes, this is from the namespace-registry and should be refactored out as a utility. Should that be part of these changes or should I make an issue for it? There are a bunch of functions that are in specific server code that can be used by all the servers that should be refactored out as utility functions, so I feel that's a whole separate issue.

cmd/cache_serve.go
turetske: What do you mean skip it? CacheAds are expected to advertise to the director with a list of namespaces they have access to. At least that's how it was explained to me.

turetske:
> I realize when looking through PR #295 does the namespace registration much better. When addressing other PR reviews, I'll refactor to mimic that.

Actually. I'll wait until the PR #295 is merged in and I think the namespace registration should be a fix with a refactoring out of common code.

director/redirect.go
```diff
@@ -343,17 +343,17 @@ func RegisterOrigin(ctx *gin.Context) {
...
 		if !ok {
-			log.Warningf("Origin %v advertised to namespace %v without valid registration\n",
-				ad.Name, namespace.Path)
-			ctx.JSON(400, gin.H{"error": "Origin not authorized to advertise to this namespace"})
+			log.Warningf("%s %v advertised to namespace %v without valid registration\n",
+				sType, ad.Name, namespace.Path)
+			ctx.JSON(400, gin.H{"error": sType + " not authorized to advertise to this namespace"})
 			return
 		}
 	}
```
bbockelm, line 352: When we have `sType` set to the cache, we should call something like this:
```
ok, err := VerifyAdvertiseToken(token, "/caches/$SITENAME")
```
When `sType` is a origin, we should do the for-loop as above

That way, we still make sure the cache is registered (separately, we need to do a validation workflow for registered caches!) but don't need to fake a namespace ad as done previously.

director/cache_ads.go
```diff
@@ -35,13 +35,13 @@ import (
...
-		Strategy      StrategyType
-		BasePath      string
-		VaultServer   string
+		RequireToken  bool         `json:"requireToken"`
+		Path          string       `json:"path"`
+		Issuer        url.URL      `json:"url"`
+		MaxScopeDepth uint         `json:"maxScopeDepth"`
+		Strategy      StrategyType `json:"strategy"`
+		BasePath      string       `json:"basePath"`
+		VaultServer   string       `json:"vaultServer"`
```
bbockelm, line 44: Does this have any forward/backward compatibility concerns?

bbockelm (changes requested):
Quite close on this one - just a few more items.

director/cache_ads.go
turetske: I don't know. It caused issues with the redirect tests. But it didn't break anything that was already using the NamespaceAds in the code otherwise. As long as it's marshaling properly into a struct, I think it should be okay?

turetske: I did this and it seems to advertise properly, looking at the debug logs.

bbockelm (approved):
Ok, let's go!

Issue #202, opened 2023-10-10
Resolve issues with split origin/cache endpoint for director


jhiemstrawisc:
As a temporary work around to interfacing with XRootD, the director was modified before the 7.0.0 release to serve either caches or origins at its default endpoint based on how the director was configured at launch. This was due to a limitation in XRootD that prevents a redirector service (indicated by the `pss.origin` directive) from taking any paths after its domain, making it impossible to point the redirector to `<my-director>/api/v1.0/director/origins`.

The problem is that we now have two director services, one for caches and one for origins, and synchronization between these two is a problem because Pelican only advertises to one.

Off the top of my head, there are two routes that could be taken to resolve this:
- Follow up on the [open issue](https://github.com/xrootd/xrootd/issues/2068) for xrootd to make redirector configuration with a server path doable. This would allow us to nix one of the directors.
- Adjust Pelican so that the federation discovery step can handle a list of director services. Origins should then advertise to each director, and the directors should probably periodically merge their advertisement caches as well.

bbockelm:
I'll add a third option:

Make the default route be based on the incoming Hostname header, meaning one service can serve either, depending on what the client 'thinks' the hostname is.

At some point, we need to think about high availability / redundancy at the director level.  I'd just like avoid doing so right now.

jhiemstrawisc:
@bbockelm Following up on this because of the recent NACK on a PR that was my interpretation of your comment. Can you provide more detail about what you were thinking? My reading of your suggestion was to have the director read the incoming `Host` header and direct based on whether it thought the original request was for an origin vs a cache.

bbockelm:
I think you got the approach right but I want the feature to be explicitly configured.  That is, make the administrator explicitly enable this mode of operation and then explicitly state what hostname header is associated with the operation.  I.e., something like this in the YAML config:

```
Director:
   DefaultResponse: cache
   CacheResponseHostnames: ["cache.example.com"]
   OriginResponseHostnames: ["origin.example.com"]
```

Your current PR looks at the incoming hostname and does some substring searching to guess intent.  It's always better to make intent explicit instead of implicit.

PR #313, opened 2023-10-31
Make director service host-aware

jhiemstrawisc:
In some cases, the director service should be aware of the originally-requested host. This is useful in the case where a host `https://director-origins.osg-htc.org` contains a CNAME record to `https://director-caches.osg-htc.org`, because this indicates to the cache director that it should actually redirect to an origin.

This works because the host header from the original request to the origin director is forwarded through the CNAME redirect instead of being swapped out for the canonical host from DNS. This essentially makes the director aware of who the request was originally intended for, thus allowing it to act appropriately on the original request.

For now, if the Host or X-Forwarded-Host headers contain the word "origin", a cache director will redirect to an origin, and vice-versa for an origin director. We may eventually need to make these checks more intelligent. Note that the X-Forwarded-Host header is checked because of the way we deploy services on Kubernetes through Traefic

### DIFF
--- a/cmd/director.go
+++ b/cmd/director.go
@@ -63,10 +63,4 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
-
-	directorServeCmd.Flags().BoolP("enable-hostname-redirects", "", false, "Enabling host-aware redirects allows the director to bypass its default response for hosts specified via the Pelican configuration file")
-	err = viper.BindPFlag("Director.HostAwareRedirects", directorServeCmd.Flags().Lookup("enable-hostname-redirects"))
-	if err != nil {
-		panic(err)
-	}
 }
--- a/cmd/director.go
+++ b/cmd/director.go
@@ -63,4 +63,10 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
+
+	directorServeCmd.Flags().BoolP("enable-hostname-redirects", "", false, "Enabling host-aware redirects allows the director to bypass its default response for hosts specified via the Pelican configuration file")
+	err = viper.BindPFlag("Director.HostAwareRedirects", directorServeCmd.Flags().Lookup("enable-hostname-redirects"))
+	if err != nil {
+		panic(err)
+	}
 }
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -30,6 +30,7 @@ import (
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
 )
 
 type (
@@ -172,6 +173,17 @@ func AdvertiseOSDF() error {
 
 	for cacheAd, namespacesSlice := range cacheAdMap {
 		RecordAd(cacheAd, &namespacesSlice)
+
+		// In addition to recording the ads, we can add these to the string slice the Director uses
+		// when checking if specific hosts should receive a cache or an origin redirect. For now, when
+		// a cache uses the director as its redirector, it only makes sense to redirect the cache to an
+		// origin, regardless of what the director is configured to serve by default.
+		if param.Director_HostAwareRedirects.GetBool() {
+			oRedirectHosts := param.Director_OriginResponseHostnames.GetStringSlice()
+			oRedirectHosts = append(oRedirectHosts, cacheAd.URL.Host)
+			oRedirectHosts = append(oRedirectHosts, cacheAd.AuthURL.Host)
+			viper.Set("Director.OriginResponseHostnames", oRedirectHosts)
+		}
 	}
 
 	return nil
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -30,7 +30,6 @@ import (
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
-	"github.com/spf13/viper"
 )
 
 type (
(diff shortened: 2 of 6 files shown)


turetske (changes requested):
Can you write tests for this? Something which sends a request checks that the redirect matches what you'd expect if the header/x-forward contains a different value than the default, etc.?

turetske:
@jhiemstrawisc For my own edification, I'm trying to wrap my head around when this would occur. What would cause this?

bbockelm:
NACK.  I don't like the magic of searching through the hostname. Let's force the admin to explicitly configure it if they use this mode.

I don't understand the comment about CNAMEs.  To first order, there shouldn't be any dependency on DNS or CNAMEs server-side.  Everything should be based on what the client puts in the header.

jhiemstrawisc:
RE your first comment -- I'm not sure how to test this unless I can run a director service as a part of a test. Any thoughts?

RE your second comment -- due to a few limitations in XRootD (see issue #202 for more detail), we've been trying to find a way to split a single director instance apart in a way that allows it to serve both caches and origins without using a path to specify which to serve. The idea is that we create a CNAME record for `https://director-origins.osg-htc.org` to point to `https://director-caches.osg-htc.org`. Right now we've been testing caches with the director by setting `pss.origin https://director-origins.osg-htc.org`, but that gets into a lot of the issues we've discussed. With this modification, we can create the CNAME and have `https://director-caches.osg-htc.org` be the service that actually handles the request -- until we get around to coming up with a better way to support multiple directors.

bbockelm:
> RE your first comment -- I'm not sure how to test this unless I can run a director service as a part of a test. Any thoughts?

Yes -- we should just "run" a director service as part of the test.  You can either launch a full-blown server:

https://github.com/PelicanPlatform/pelican/blob/main/director/advertise_test.go#L109

or just create a mockup object that invokes the handler directly:

https://github.com/PelicanPlatform/pelican/blob/bf553a28938555bd0b5c03ea8cb5be0199a2be22/origin_ui/origin_test.go#L168

I tend to prefer the latter approach as there's less chance for weird side-effects from multiple tests running in parallel.

director/advertise.go
```diff
@@ -172,6 +173,17 @@ func AdvertiseOSDF() error {
 
 	for cacheAd, namespacesSlice := range cacheAdMap {
 		RecordAd(cacheAd, &namespacesSlice)
+
+		// In addition to recording the ads, we can add these to the string slice the Director uses
+		// when checking if specific hosts should receive a cache or an origin redirect. For now, when
+		// a cache uses the director as its redirector, it only makes sense to redirect the cache to an
+		// origin, regardless of what the director is configured to serve by default.
+		if param.Director_HostAwareRedirects.GetBool() {
```
bbockelm, line 181: I'm not following this.

It appears it adds the hostname of each cache in the OSDF to the list of the hostnames the director identifies itself as.  Why is that?

It also adds it as a list, meaning this will grow every 10 minutes when this function is invoked.

cmd/director.go
```diff
@@ -63,4 +63,10 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
+
+	directorServeCmd.Flags().BoolP("enable-hostname-redirects", "", false, "Enabling host-aware redirects allows the director to bypass its default response for hosts specified via the Pelican configuration file")
```
bbockelm, line 67: Let's skip the command line flag if it's not possible to use this functionality completely from CLI.

director/redirect.go
```diff
@@ -278,21 +280,64 @@ func RedirectToOrigin(ginCtx *gin.Context) {
...
 		// If this is a request to the Prometheus API, don't modify the path
-		if strings.HasPrefix(c.Request.URL.Path, "/.well-known") ||
-			strings.HasPrefix(c.Request.URL.Path, "/api/v1.0/prometheus") {
+		if strings.HasPrefix(c.Request.URL.Path, "/.well-known/") ||
+			(strings.HasPrefix(c.Request.URL.Path, "/api/v1.0/") && !strings.HasPrefix(c.Request.URL.Path, "/api/v1.0/director/")) {
 			c.Next()
 			return
 		}
+
+		if param.Director_HostAwareRedirects.GetBool() {
```
bbockelm, line 323: Why is this an option?  What's the harm in unconditionally enabling this?

In other words, any time an admin wants to enable this, they have to set two parameters (flip a boolean then add a name to a list).  It seems that can be simplified to one action -- adding a name to the list implies it should be used, no?

docs/parameters.yaml
```diff
@@ -349,6 +349,29 @@ type: string
...
+  When enabled, the director can be configured to redirect specified host names to either a cache or an origin, bypassing
+  the default response. This allows to a client to get either a cache or an origin from a director at its default path
+  by setting the `Host` or `X-Forwarded-Host` header.
+type: bool
+default: false
+components: ["director"]
+---
+name: Director.CacheResponseHostnames
+description: >-
+  A list of hostnames to be pinned to cache redirects at the director's default path when `Director.HostAwareRedirects` is enabled.
```
bbockelm, line 363: I'm having trouble following this description.  How about this:

> A list of virtual hostnames for the director.  If a request is sent by the client to one of these hostnames,
>  the director assumes it should respond with a redirect to a cache.
> 
> If present, the hostname is taken from the `X-Forwarded-Host` header in the request.  Otherwise, `Host` is used.

(and a similar update for the origin)

bbockelm (changes requested):
I like the config-based approach better!  A few comments:
- I don't understand what's being done with the OSDF integration here -- why is the list of cache hostnames being added as potential virtual hostnames for the director?
- I think we could simplify the configuration of this feature by only requiring the list to be set.
- I found the description in the documentation confusing; suggestion is in inline comment.

director/advertise.go
jhiemstrawisc: I think I got a bit confused as I was working on this, essentially mixing up the "from" and the "to". This feature would be nice to explore if the `Host` header contained the sender's host, which would allow us to say "requests **from** `Host: foo` should always be redirected to an origin," but the host header tells us **to**. I'll remove this chunk.

bbockelm (approved):
Revisions LGTM.

PR #316, opened 2023-11-02
Upgrade status component

### LINKED ISSUE #232: Origin UI crashes when /health API returns with a missing component

CannonLock:
- Remove some padding
- More robust to errors parsing response
- Status components are no longer hardcoded
- Added message view
- Closes #232

### DIFF
--- a/origin_ui/src/components/StatusBox.tsx
+++ b/origin_ui/src/components/StatusBox.tsx
@@ -24,18 +24,61 @@ import { alpha, useTheme } from "@mui/material";
 interface StatusDisplayProps {
     component: string;
     status: string;
+    message?: string;
 }
 
-function StatusDisplay({component, status}: StatusDisplayProps) {
+function StatusDisplay({component, status, message}: StatusDisplayProps) {
 
     const theme = useTheme()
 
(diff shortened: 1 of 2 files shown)


origin_ui/src/components/StatusBox.tsx
```diff
@@ -24,18 +24,61 @@ import { alpha, useTheme } from "@mui/material";
...
+            break
+        case "web-ui":
+            component = "Web UI"
+            break
+        case "cmsd":
+            component = "CMSD"
+            break
+        case "federation":
+            component = "Federation"
+            break
```
haoming29, line 63: We will need to add another case when #175 is merged.

haoming29 (approved):
LGTM! Manually and visually confirmed that when `/heath` api returns error, status box will catch it and render an error box while the rest of the page can render as is. Also confirmed that a new component key won't break the rendering.

PR #318, opened 2023-11-02
Graph Robustness #276

### LINKED ISSUE #276: Graphs do not auto-update
I loaded up the origin UI and then shut down the Pelican process.  However, it appears that (a) the graphs aren't updating since the original timestamp and (b) no error message is being recorded (the database is not returning empty, it's unreachable!).

image

CannonLock:
- Catches and displays errors for graph queries
- The uncaught errors prevented refresh to that is fixed as well
- Close #276

### DIFF
--- a/origin_ui/src/components/graphs/Graph.tsx
+++ b/origin_ui/src/components/graphs/Graph.tsx
@@ -82,7 +82,7 @@ export default function Graph({getData, options, boxProps, drawer}: GraphProps) 
             }
         } catch (e: any) {
             let date = new Date(Date.now()).toLocaleString()
-            setError(date + " : " + e.message)
+            setError(date + " : " + e.message + "; plot will auto-refresh")
         }
     }
 
--- a/origin_ui/src/components/graphs/Graph.tsx
+++ b/origin_ui/src/components/graphs/Graph.tsx
@@ -48,7 +48,7 @@ const defaultOptions: Partial<ChartOptions<"line">> = {
     scales: {
         x: {
             type: 'time',
-                time: {
+            time: {
                 round: 'second',
             }
         }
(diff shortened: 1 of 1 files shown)


origin_ui/src/components/graphs/Graph.tsx
```diff
@@ -70,14 +70,19 @@ export default function Graph({getData, options, boxProps, drawer}: GraphProps)
...
+            setLoading(false)
+            if(response.datasets[0].data.length == 0){
+                let date = new Date(Date.now()).toLocaleTimeString()
+                setError(`No data returned by database as of ${date}; plot will auto-refresh`)
+            } else {
+                setError("")
+            }
+        } catch (e: any) {
+            let date = new Date(Date.now()).toLocaleString()
+            setError(date + " : " + e.message)
```
haoming29, line 85: Do we want to `setLoading(false)` here? It seems that if `getData()` returns error, we will never set loading to false.

haoming29 (changes requested):
One small catch as mentioned in the comment. Otherwise it looks good to me. I can confirm that 1. Graph displays error message when `/api/v1.0/prometheus` returns non 200 response. 2. Graph displays error message indicating it failed to fetch (instead of no data) when origin is down.

origin_ui/src/components/graphs/Graph.tsx
haoming29: That makes sense to me! I think one easy improvement can be adding something like "Retry in 1 minute" at the end of the error message so that the user knows it's still updating not just died

haoming29 (approved):
LGTM! Will wait for tests to finish before merging.

PR #321, opened 2023-11-03
Fix bypassed authentication for /config API

haoming29:
This fixes the issue with `/api/v1.0/config` not checking `User` field in ctx and allows unauthed user to access config.


CannonLock (approved):
Looks good, works good!

PR #324, opened 2023-11-03
Add draft of SECURITY.md

jhiemstrawisc:
Since we're dealing with data, we should have some information about how community/federation members can report any security concerns they might have.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -11,6 +11,11 @@ We currently offer support for the following versions of Pelican:
 
 ## Reporting a Vulnerability or Security Issue
 
+We recognize that trust is hard to earn and easy to lose. That's why we take security incidents
+very seriously and strive to answer all concerns in as timely a manner as we are able. When you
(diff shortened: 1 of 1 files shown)


haoming29 (approved):
Not sure if @bbockelm wants to take a look at it as well but LGTM

jhiemstrawisc:
Let's wait for a review from Brian. I drafted this without certainty about how we delineate "OSDF as a service, maintained under the OSG" and "Pelican proper". It might be that we want to remove references to OSDF here and instead have this security policy focus only on Pelican code vulnerabilities.

jhiemstrawisc:
For the 7.3 release, I think this can go in if you're okay with it @haoming29. I removed references to OSDF and OSG, which are technically not really a part of the pelican software suite (well, if we pretend topology doesn't exist). This way, if there's an issue specific to the way OSDF is being run that isn't an actual vulnerability with pelican, people don't think they should be emailing us. If there's a pelican issue, they shouldn't be emailing OSG.

Issue #325, opened 2023-11-03
Director bug


jhiemstrawisc:
The director function `GetAdsForPath()` de-references a nil pointer when queried for a namespace prefix that does not exist in the ttlcache.

jhiemstrawisc:
Fixed by PR #326

Issue #300, opened 2023-10-30
Accessing APIs from browser causes director to panic


haoming29:
When accessing any of the director's API endpoints through the browser i.e. `/api/v1.0/prometheus/query`, director server will panic as we are requesting favicon from the server, which is recognized as a redirect traffic but there's no namespace attached to the URL, thus a null pointer exception occurred in handling the redirecting.

```console
2023/10/30 13:57:08 [Recovery] 2023/10/30 - 13:57:08 panic recovered:
runtime error: invalid memory address or nil pointer dereference
/usr/local/go/src/runtime/panic.go:260 (0x45c77)
        panicmem: panic(memoryError)
/usr/local/go/src/runtime/signal_unix.go:841 (0x5ece3)
        sigpanic: panicmem()
/workspaces/pelican-dev/pelican/director/cache_ads.go:192 (0x663ae4)
        GetAdsForPath: originNamespace = *best
/workspaces/pelican-dev/pelican/director/redirect.go:170 (0x666f03)
        RedirectToCache: namespaceAd, _, cacheAds := GetAdsForPath(reqPath)
/workspaces/pelican-dev/pelican/director/redirect.go:285 (0x667f97)
        ShortcutMiddleware.func1: RedirectToCache(c)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0xc5fd3f)
        (*Context).Next: c.handlers[c.index](c)
/root/go/pkg/mod/github.com/zsais/go-gin-prometheus@v0.1.0/middleware.go:364 (0xc5fd18)
        (*Prometheus).HandlerFunc.func1: c.Next()
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0xc67fb3)
        (*Context).Next: c.handlers[c.index](c)
/workspaces/pelican-dev/pelican/web_ui/ui.go:69 (0xc67f8c)
        GetEngine.func1: ctx.Next()
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x62699b)
        (*Context).Next: c.handlers[c.index](c)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/recovery.go:102 (0x62697c)
        CustomRecoveryWithWriter.func1: c.Next()
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x62599f)
[...]

PR #326, opened 2023-11-03
Fix null dereference bug in Director

jhiemstrawisc:
The Director had a null dereference bug that occurred when it tried to find ads for non-existent namespace prefixes. This commit fixes the bug, adds a regression test to catch similar bugs in the future, and re-orders the error reporting to the client in the event that it tries to get something from a non-existent namespace.

### DIFF
--- a/director/cache_ads.go
+++ b/director/cache_ads.go
@@ -190,6 +190,8 @@ func GetAdsForPath(reqPath string) (originNamespace NamespaceAd, originAds []Ser
 		}
 	}
 
-	originNamespace = *best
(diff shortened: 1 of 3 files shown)


jhiemstrawisc:
Fixes issue #325

haoming29:
Related: #300 Which can also trigger director to panic, but the fix for #300 is still required so that we don't trigger redirect for director's web ui requests.

bbockelm (approved):
LGTM.  Thanks for including the regression test.

Issue #246, opened 2023-10-19
Improve RPM for Pelican


bbockelm:
The Pelican RPM only ships the `pelican` binary.  More is needed:

1.  Create a `pelican-osdf-compat` sub-RPM that includes symlinks for `stash_plugin`, `osdf`, and `stashcp`.
2. Have the `pelican-osdf-compat` RPM contain an `Obsoletes` statement, indicating it obsoletes all the prior OSDF/stashcp RPMs we've shipped.
3.  Add a `pelican-server` RPM that contains the systemd unit files needed to startup / shutdown Pelican.

I *think* this is all doable via goreleaser (but it may be that we need to break out a handwritten RPM).

bbockelm:
Also from @timtheisen -- we also need to include the condor config file to enable the `stash_plugin`.

matyasselmeci:
Can we split `pelican-server` into a separate issue? I can get `pelican-osdf-compat` done in time for 7.2.0.

matyasselmeci:
For the compat RPM, Goreleaser does not appear to expand templates within `replaces` and `provides` lists - that is, this
```
provides:
- "osdf-client (= {{ .Version }})"
```
does not result in a deb with a Provides of `osdf-client (= 7.1.1)` but with `osdf-client (= {{ .Version }})` which results in a syntax error when installing.  The RPMs are similarly broken; in addition, I can't use the RPM macros either (`osdf-client = %{version}`) is inserted literally.

Either we have to add a step where we generate `.goreleaser.yml` with the correct version coded in, or just say that we provide `osdf-client=7` and call it a day.

matyasselmeci:
I forked off pelican-server to #331 so I'm marking this resolved.

PR #327, opened 2023-11-03
 Add pelican-osdf-compat package

matyasselmeci:
This adds compat symlinks for `/usr/bin/stashcp`, `/usr/bin/osdf`, and `condor-stash-plugin` to use the `pelican` binary isntead, and a condor configuration for enabling the condor-stash-plugin.  This allows pelican to replace the old "stashcp" and "condor-stash-plugin" RPMs.

The appropriate `obsoletes` (`replaces` for Deb) and `provides` clauses are included, with the caveat that pelican-osdf-compat provides version "7" of the packages it's replacing, as opposed to whatever specific version the pelican-osdf-compat package itself is.  This is because GoReleaser doesn't expand templates inside `replaces` and `provides` lists so the value has to be hardcoded.

Deb, RPM, and APK use different syntax for specifying versions clauses:
- APK does not use `replaces` (`provides` is sufficient for replacing a package).
- Deb puts parens around the version restriction (so `stashcp (= 7)`   whereas RPM and APK use `stashcp = 7`).
- Deb uses `<<` instead of `<` as the strictly-less-than operator.

The APK doesn't have a symlink and config file for condor-stash-plugin because condor is not available on Alpine.

See #246

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -98,3 +98,97 @@ nfpms:
             file_info:
               mode: 0644
             type: doc
+  # end package pelican
+
+  - package_name: pelican-osdf-compat
+    builds: []
+    # file_name_template: '{{ .PackageName }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
+    file_name_template: "{{ .ConventionalFileName }}"
+    id: pelican-osdf-compat
+    vendor: OSG Consortium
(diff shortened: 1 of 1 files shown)


bbockelm (approved):
LGTM.

Separately, we should migrate the homepage to pelicanplatform.org; no need to hold up the bus for that though.

Issue #170, opened 2023-10-02
End-to-end integration test including XRootD


bbockelm:
Add a go-based integration test that:
1.  Launches the origin.  May need some modest refactoring to move necessary code from `cmd/origin_serve.go` into the `xrootd` or `origin_ui` module.
2. Checks the xrootd server is alive and ready.  Perhaps expose the running PID for xrootd and allow the test to get the resource usage (`getrusage` - should be exposed in the syscalls package).
3. Checks the health API is set appropriately.
4. Invokes the self-test cycle: upload a file, download a file, delete the file.  Verify contents.
5. Shuts down the origin.

PR #328, opened 2023-11-04
End-to-end test for XRootD

bbockelm:
@turetske - not sure if this is helpful but I was tinkering with testing xrootd while on the airplane and cobbled together this PR.

Now, the tests shouldn't pass for the exact same thing you're struggling with -- waiting on the new release of `scitokens-cpp`.  But perhaps the approach is interesting for you?

One item we'll need to think through: How do we want to approach Mac OS X?  The homebrew-packaged version of XRootD comes without SciTokens, meaning that maintaining a test there would be fairly difficult.

### DIFF
--- a/xrootd/origin_test.go
+++ b/xrootd/origin_test.go
@@ -58,7 +58,7 @@ func TestOrigin(t *testing.T) {
 	err = config.GenerateCert()
 	require.NoError(t, err)
 
-	err = CheckXrootdEnv(true)
+	err = CheckXrootdEnv(true, nil)
 	require.NoError(t, err)
 
 	err = SetUpMonitoring()
@@ -67,7 +67,7 @@ func TestOrigin(t *testing.T) {
 	configPath, err := ConfigXrootd(true)
 	require.NoError(t, err)
 
-	launchers, err := ConfigureLaunchers(false, configPath)
+	launchers, err := ConfigureLaunchers(false, configPath, false)
 	require.NoError(t, err)
 
 	ctx, cancel := context.WithCancel(context.Background())
@@ -78,7 +78,7 @@ func TestOrigin(t *testing.T) {
 
 	testExpiry := time.Now().Add(10 * time.Second)
 	testSuccess := false
-	for !testSuccess || time.Now().After(testExpiry) {
+	for !(testSuccess || time.Now().After(testExpiry)) {
 		time.Sleep(50 * time.Millisecond)
 		req, err := http.NewRequest("GET", param.Origin_Url.GetString(), nil)
 		require.NoError(t, err)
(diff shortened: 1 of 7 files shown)


xrootd/xrootd_config.go
```diff
@@ -99,6 +99,20 @@ func CheckXrootdEnv(origin bool) error {
 			" to desired daemon user %v", runtimeDir, username)
 	}
 
+	// The scitokens library will write the token cache into the user's home
+	// directory.  By setting $XDG_CACHE_HOME, we will move the cache into our runtime dir.
```
turetske, line 103: Took me a few seconds to realize which "cache" this was referring to, due to overloaded terms. I was thinking the pelican cache for a second rather than the token cache. Maybe rename things to be a bit more specific? "tokenCacheDir" or something like that? Mostly for someone taking a look at this code a year from now.

turetske (commented):
Obviously waiting on the SciTokens config, but definitely good to see how the end-to-end setup would work. Very much like how I imagined after you described it for me and definitely a template that can be used for later integration tests (say for the cache, registry, and director).

turetske:
Looking at this further, I was under the impression you didn't want this to be a unit test and instead a test that ran during integration testing. At least that what issue #170 seems to imply. Yet the TestOrigin code will run as part of the `go test ./...` command workflow. Is that how you want it implemented?

turetske (changes requested):
@bbockelm I ran this in a local dev container with the correct versions of sciTokens-cpp and the correct xrootd running with some minor changes (see my other review comments.)

The major problem I see is that if the xrootd server is not up, it will constantly get into a loop of the following error: ```INFO[2023-11-08T20:08:51Z] Failed to send request to XRootD; likely, server is not up (will retry in 50ms): Get "https://ccd8174190bf:8443": dial tcp 172.17.0.3:8443: connect: connection refused```

That is only visible if running `go test` in the xrootd folder. Otherwise the tests will just hang indefinitely if running `go test ./...` from the base directory with any good error output. Another issue is that you can't cancel out of the test easily if this is happening `ctrl-c` doesn't work. I had to open another terminal and kill multiple processes by pid.

This could be a major issue in a dev environment where we just want to make sure the tests still pass after pushing. I think my earlier comment about not making this a unit test still stands.

xrootd/origin_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	configPath, err := ConfigXrootd(true)
+	require.NoError(t, err)
+
+	launchers, err := ConfigureLaunchers(false, configPath)
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		_ = daemon.LaunchDaemons(ctx, launchers)
```
turetske, line 75: There is a failure when launching the daemons, it fails due to the lack of permission to use the key file: `XrdTLS: TLS_Context: Unable to use key file /tmp/TestOrigin1137965101/002/certificates/tls.key; permission denied`, which seems odd to me because that's a tmp directory and I'm pretty sure we add xrootd to the users/permissions when we do so.

turetske (changes requested):
@bbockelm Found the infinite loop issue and got a closer look at why the xrootd launcher was failing. I'm not sure of the fix yet, I'll try to look into the xrdtls fix later.

turetske:
@bbockelm It passes! I can't request a review from you as it's your own PR originally, but if you could take a look at my changes, that would be appreciated.

bbockelm:
Merging PR as both @turetske and I are happy with it.

PR #337, opened 2023-11-06
Unit test for namespace-registry

joereuss12:
Testing the main functionalities of namespace-registry: Adding a namespace (without indentity), listing/getting contents from a namespace, and deleting a namespace.

### DIFF
--- a/cmd/namespace_registry_test.go
+++ b/cmd/namespace_registry_test.go
@@ -43,11 +43,18 @@ func TestServeNamespaceRegistry(t *testing.T) {
 	viper.Set("Registry.DbLocation", filepath.Join(issuerTempDir, "test.sql"))
 	err := config.InitServer()
 	require.NoError(t, err)
+
 	err = nsregistry.InitializeDB()
 	require.NoError(t, err)
+	defer nsregistry.ShutdownDB()
 
 	gin.SetMode(gin.TestMode)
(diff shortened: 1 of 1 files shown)


bbockelm:
@joereuss12 - I think the issue with the unit test on Windows is you'll need to call an explicit `ShutdownDB()` before tearing down the unit test.  This is because Windows won't let you delete a file that is currently open.

cmd/namespace_registry_test.go
```diff
@@ -0,0 +1,131 @@
...
+		err = nsregistry.NamespaceGet(svr.URL + "/api/v1.0/registry")
+		require.NoError(t, err)
+		w.Close()
+		os.Stdout = oldStdout
+
+		capturedOutput := make([]byte, 1024)
+		n, _ := r.Read(capturedOutput)
+		stdoutCapture = string(capturedOutput[:n])
+		assert.Equal(t, "[]\n", stdoutCapture)
+	})
```
jhiemstrawisc, line 130: Can you bookend your code with a `viper.Reset()`? If the next test to run happens not to call `viper.Reset()` at the start, it might leak some of your viper config.

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM

PR #339, opened 2023-11-07
Renamed getOriginJWK to getServerJWK

### LINKED ISSUE #314: Change name of getOriginJWK
Origin is an overloaded term in Pelican and can lead to some confusion. Potentially rename this to `getHostJWK` or `getServerJWK`?

turetske:
Close #314 

`getOriginJWK` is a confusing name because of the overloaded `origin` term. This changes that function to `getServerJWK`. If that's not an acceptable name, vscode makes refactoring very simple, so it can also be something like `getIssuerJWK` or `getHostJWK`.

### DIFF
--- a/web_ui/authorization.go
+++ b/web_ui/authorization.go
@@ -109,9 +109,9 @@ func FederationCheck(c *gin.Context, strToken string, expectedScope string) {
 	}
 }
 
-// Checks that the given token was signed by the origin jwk and also checks that the token has the expected scope
-func OriginCheck(c *gin.Context, strToken string, expectedScope string) {
-	bKey, err := pelican_config.GetOriginJWK()
+// Checks that the given token was signed by the issuer jwk and also checks that the token has the expected scope
+func IssuerCheck(c *gin.Context, strToken string, expectedScope string) {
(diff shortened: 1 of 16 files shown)


haoming29 (changes requested):
Just one small change hidden in the comment. LGTM otherwise.

turetske:
> When talking to @jhiemstrawisc on #359, I was thinking we should be more clear on different "private keys" used in the system, so I would suggest we use something like `GetIssuerPrivateJWK` to be extra clear on what this function is doing. @turetske Let me know your thoughts.

That sounds good to me and is a very quick fix.

haoming29:
@turetske I think somehow you were on an older branch, the serveNamespaceRegister has been merged to main.

turetske:
@haoming29 Yeah. Something went weird with my rebase, thanks for catching it.

turetske:
@haoming29 Okay, those commit issues should be fixed now.

web_ui/authorization.go
haoming29, line 1: Thanks for catching this!

haoming29 (approved):
LGTM. Confirmed that no obvious legacy `getOriginJWK` left. Thanks for managing the conflicts.

PR #341, opened 2023-11-07
Code cleanup in serveNamespaceRegistry

### LINKED ISSUE #255: Incompatible key encryption method and duplicate key gen in namespace_registry_serve.go
Quote from https://github.com/PelicanPlatform/pelican/pull/254#discussion_r1367473513:

Looking at `namespace_registry_serve.go` which uses `generateTLSCertIfNeeded`, there's also a duplicate private key generation: https://github.com/PelicanPlatform/pelican/blob/e7fc4aa86870acd444e3f43b7388bba6a99748db/cmd/namespace_registry_serve.go#L47 and https://github.com/PelicanPlatform/pelican/blob/e7fc4aa86870acd444e3f43b7388bba6a99748db/cmd/namespace_registry_serve.go#L54 And the encryption there is also `P521`. 

We want to 1. remove `generateTLSCertIfNeeded` and pull the logic from there to `serveNamespaceRegistry` and we only generate private key once 2. Ensure that key encryption curve is `elliptic.P256()`

turetske:
close #255

### DIFF
--- a/cmd/namespace_registry_serve.go
+++ b/cmd/namespace_registry_serve.go
@@ -34,17 +34,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func generateTLSCertIfNeeded() error {
-	// As necessary, generate a private key and corresponding cert
-	if err := config.GeneratePrivateKey(param.Server_TLSKey.GetString(), elliptic.P256()); err != nil {
-		return err
-	}
-	if err := config.GenerateCert(); err != nil {
-		return err
-	}
-	return nil
-}
-
 func serveNamespaceRegistry( /*cmd*/ *cobra.Command /*args*/, []string) error {
 	log.Info("Initializing the namespace registry's database...")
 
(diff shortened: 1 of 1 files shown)


cmd/namespace_registry_serve.go
```diff
@@ -54,17 +43,20 @@ func serveNamespaceRegistry( /*cmd*/ *cobra.Command /*args*/, []string) error {
...
-	err = config.GeneratePrivateKey(issuerKeyFile, elliptic.P521())
+	err = config.GeneratePrivateKey(issuerKeyFile, elliptic.P256())
 	if err != nil {
 		return errors.Wrap(err, "Failed to generate registry private key")
 	}
+
+	if err := config.GenerateCert(); err != nil {
+		return err
+	}
+	if err != nil {
```
haoming29, line 56: This code is not reachable as the above code checks error and returns.

haoming29 (changes requested):

haoming29 (approved):
LGTM!

PR #346, opened 2023-11-07
Made Pelican logging able to output to file

joereuss12:
Added an option for Pelican logs to output to a specified file rather than stdout. This is used with either the `-l` flag or by specifying Logging.File within the Pelican config.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -353,9 +353,7 @@ func InitConfig() {
 	} else {
 		logLevel := param.Logging_Level.GetString()
 		level, err := log.ParseLevel(logLevel)
-		if err != nil {
-			cobra.CheckErr(err)
-		}
+		cobra.CheckErr(err)
 		SetLogging(level)
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -357,12 +357,20 @@ func InitConfig() {
 		SetLogging(level)
 	}
 
-	logLocation := param.Logging_File.GetString()
+	logLocation := param.Logging_LogLocation.GetString()
 	if logLocation != "" {
+		dir := filepath.Dir(logLocation)
+		if dir != "" {
+			if err := os.MkdirAll(dir, 0644); err != nil {
+				log.Errorf("Failed to access/create specified directory. Error: %v", err)
+				os.Exit(1)
+			}
+		}
 		// Note: do not need to close the file, logrus does it for us
 		f, err := os.OpenFile(logLocation, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


docs/parameters.yaml
```diff
@@ -113,6 +113,13 @@ type: string
 default: Error
 components: ["*"]
 ---
+name: Logging.File
```
jhiemstrawisc, line 116: How about we go with `Logging.LogLocation` here to match other configuration options

config/config.go
```diff
@@ -358,6 +358,16 @@ func InitConfig() {
 		}
 		SetLogging(level)
 	}
+
+	logLocation := param.Logging_File.GetString()
+	if logLocation != "" {
+		// Note: do not need to close the file, logrus does it for us
+		f, err := os.OpenFile(logLocation, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644)
+		if err != nil {
```
turetske, line 366: I don't think you need this `err != nil` line as cobra.CheckErr does that check already. It's redundant.

turetske (changes requested):

config/config.go
joereuss12: Ah, good catch

config/config.go
```diff
@@ -358,6 +358,16 @@ func InitConfig() {
 		}
 		SetLogging(level)
 	}
+
+	logLocation := param.Logging_File.GetString()
+	if logLocation != "" {
+		// Note: do not need to close the file, logrus does it for us
+		f, err := os.OpenFile(logLocation, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644)
```
jhiemstrawisc, line 363: Two things to look into:
- How does this react if I feed it a location that the user running a Pelican command doesn't have permissions to write to?
- Assume I have an empty directory `/test`. If I run Pelican with the flag `-l /test/logs/my_log`, this will fail because we don't handle creating the directory. Check out `os.MkdirAll` elsewhere in `config.go`.

jhiemstrawisc (changes requested):

docs/parameters.yaml
```diff
@@ -113,6 +113,13 @@ type: string
 default: Error
 components: ["*"]
 ---
+name: Logging.LogLocation
+description: >-
+  A string defining a file to write log outputs to, if the user desires.
+type: string
```
turetske, line 119: Technically, this should be the filename type.

turetske (changes requested):
Minor change requested, then I'll approve

turetske (approved):
LGTM!

jhiemstrawisc (approved):

PR #347, opened 2023-11-07
Add unit tests to XRootD monitoring handler

### LINKED ISSUE #129: Unit tests for the XRootD UDP monitoring
XRootD produces a very specific binary format and sequence of packets for transfers that occur, documented extensively here:

https://xrootd.slac.stanford.edu/doc/dev56/xrd_monitoring.htm

For a unit test:

- We could dump these packets as binary files (or as a JSON-formatted list of raw bytes) inside the git repository.
- Use the `go:embed` mechanism to have the packets inside the unit test file
- In the unit test, invoke the processing handlers directly for `xrootd_metrics.go`, and
- Query the global Prometheus gauge to see if the desired values.

Particularly, I'm suspicious the code that does the accounting for different file prefixes is somewhat buggy... the unit test could focus on seeing if it's aggregating the file transfers correctly.

haoming29:
Closes #129 

A couple of lingering questions regarding the handler code, which will affect how I write the test code, see `TODO` and other comments in the code:
* Why we want to use SID as user id for u-stream and record it in our sessions cache? `SID` is an `int64` value per xrootd documentation while the `UserID.Id` is a `uint32` value, which can't hold `SID`.
* What `ComputePrefix` function is doing? It looks buggy because it will only return "/" no matter what

Besides, I fixed some bugs in the handler code as I write the tests. See code and comments below for details.

Finally, this PR introduced a considerable amount of code to serialize xrootd monitoring struct as well as adding such structs to our project. For the serializers, they look a bit redundant but I don't think reflection can help as it can't preserve the order of the fields in the structs.

### DIFF
--- a/metrics/xrootd_metrics.go
+++ b/metrics/xrootd_metrics.go
@@ -46,8 +46,8 @@ type (
 	XrdUserId struct {
 		Prot string
 		User string
-		Pid  string
-		Sid  string
+		Pid  int
+		Sid  int
 		Host string
 	}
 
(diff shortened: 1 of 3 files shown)


bbockelm:
> Why we want to use SID as user id for u-stream and record it in our sessions cache? SID is an int64 value per xrootd documentation while the UserID.Id is a uint32 value, which can't hold SID.

Took a deep look at the code here -- it appears the that SID is not a unique (docs say "Server’s fingerprint for the connection to user:pid at host") but rather a static value computed at server startup.  I filed a bug with upstream: https://github.com/xrootd/xrootd/issues/2133.  The full `userid` (what you call `XrdUserId`) _is_ unique for HTTP.

We can use dictid in the monitoring packet as a unique identifier for the connection.  Unfortunately, it's not repeated in the `d` record type (meaning that we'd have to use `XrdUserId` to join the information together for HTTP).  The most reliable approach might be to just ignore the `d` record type as file open information is also sent via the `f` stream.

bbockelm (approved):
Did some improvements to the PR - liked the basic approach.

1.  Switch to using the dictid as the unique key for user sessions.  The prior approach, based on the SID, wasn't workable as the SID is the "server ID" (a unique identifier of the server instance), not the "session ID".
2. Add support for the token authentication packet type.  This is needed as there's no support for establishing a user session at HTTPS connection time in Pelican.

PR #348, opened 2023-11-07
Get rid of superfluous symlink

timtheisen:


matyasselmeci (approved):
LGTM. Glad we don't need this anymore!

PR #356, opened 2023-11-08
Use Server.Hostname than os.Hostname() in generating CA

### LINKED ISSUE #303: Use Server.Hostname instead of os.Hostname() in generating CA
We generate the CA using `os.Hostname()`; However, when we set `Server.Hostname` different from os.Hostname, there will be TLS error saying the CA was not issued for the host. Thus, we want to read hostname from config.

https://github.com/PelicanPlatform/pelican/blob/bdd7616ae3de882bda4975c26c33965082bef1be/config/init_server_creds.go#L159

haoming29:
Closes #303 

With this PR merged, we should be consistently using Server.Hostname from parameters at all places. This PR also changes the param type of hostname from url to string, and noted the default value will be `os.Hostname()`

Note to the reviewer. I've tested that by changing the `Server.Hostname` from its default to `localhost`, remove existing CA file, spinning up an origin to let it generate new CA, and copy the CA file to `/etc/pelican/certificates/tlsca.crt`, and re-run the origin, the self-test works as expected.

### DIFF
--- a/config/init_server_creds.go
+++ b/config/init_server_creds.go
@@ -161,10 +161,7 @@ func GenerateCACert() error {
 	if err != nil {
 		return err
 	}
-	hostname, err := os.Hostname()
-	if err != nil {
-		return err
-	}
+	hostname := param.Server_Hostname.GetString()
 	notBefore := time.Now()
 	template := x509.Certificate{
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (approved):
LGTM

Issue #256, opened 2023-10-20
Consolidate generating server private key and CA certificate into InitServer()


haoming29:
Currently, each service (origin/director/namespace) needs a private key and CA certificate and the same logic repeats. There are also inconsistency issues when a new service is added and 
generating a key. https://github.com/PelicanPlatform/pelican/blob/e7fc4aa86870acd444e3f43b7388bba6a99748db/cmd/origin_serve.go#L72

According to discussion here: https://github.com/PelicanPlatform/pelican/pull/254#discussion_r1367399193 and as an addendum to #255 , we want to consolidate server private key generation as well as CA certificate generation into `config.InitServer` so that we ensure each server will have a private key and a CA certificate.

1. Remove private key and CA cert generation from `cmd.*_serve.go`
2. Add private key and CA cert generation in `config.InitServer`
3. Run regression test

PR #362, opened 2023-11-08
Standardize server init and cleanup a few other small items

jhiemstrawisc:
The big part of this PR is standardizing some of how servers are initialized. Importantly, this moves federation discovery and some key creation from the various `*_serve.go` files into `InitServer()`.

In the process of putting this together, I noticed a few other cleanup items. One was changing a parameter.yaml default value and type to match the description. Another was modifying config_test to use `assert.Equal` instead of `assert.True` so that if the test fails, it prints out what the expected and actual values are.

### DIFF
--- a/cmd/director.go
+++ b/cmd/director.go
@@ -32,7 +32,7 @@ var (
 	directorCmd = &cobra.Command{
 		Use: "director",
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			err := config.InitServer()
+			err := initDirector()
 			return err
 		},
 		Short: "Launch a Pelican Director",
@@ -68,6 +68,13 @@ func getDirectorEndpoint() (string, error) {
 
 	// Return the string, as opposed to a pointer to the URL object
 	return directorEndpointURL.String(), nil
+}
+
+func initDirector() error {
+	err := config.InitServer()
+	cobra.CheckErr(err)
+
+	return err
 }
 
 func init() {
--- a/cmd/director_serve.go
+++ b/cmd/director_serve.go
@@ -19,7 +19,6 @@
 package main
 
 import (
-	"crypto/elliptic"
 	"fmt"
 	"os"
 	"os/signal"
@@ -29,17 +28,11 @@ import (
 	"github.com/pelicanplatform/pelican/director"
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/web_ui"
-	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
 func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
-	err := config.DiscoverFederation()
-	if err != nil {
-		log.Warningln("Failed to do service auto-discovery:", err)
-	}
-
 	log.Info("Initializing Director GeoIP database...")
 	director.InitializeDB()
 
@@ -53,25 +46,6 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 		}
 	}
 	go director.PeriodicCacheReload()
-
-	// The director needs its own private key. If one doesn't exist, this will generate it
-	issuerKeyFile := param.IssuerKey.GetString()
-	err = config.GeneratePrivateKey(issuerKeyFile, elliptic.P256())
-	if err != nil {
-		return errors.Wrap(err, "Failed to generate director private key")
-	}
-
-	// Since director will provide a public JWK for token authentication
-	// We need to initialize public JWK and private JWK at the start
-	_, err = config.GenerateIssuerJWKS()
-	if err != nil {
-		return errors.Wrap(err, "Failed to load director's public and private jwk")
-	}
-
-	// Generate a TLS certificate if needed
-	if err := config.GenerateCert(); err != nil {
-		return err
-	}
 
 	engine, err := web_ui.GetEngine()
 	if err != nil {
--- a/config/config_linux.go
+++ b/config/config_linux.go
@@ -21,10 +21,49 @@
 package config
 
 import (
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 )
 
+// Different distros have different default CACertificate locations. Instead of trying to figure out
+// the version of linux we use, check all the well-known places until we find one that exists.
+func findLinuxCACert() (string, error) {
+	// These values pulled from the same place x509 package looks:
+	// https://go.dev/src/crypto/x509/root_linux.go
+	certFileLocations := []string{
+		"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
+		"/etc/pki/tls/cert.pem",                             // One RHEL-based possibility
+		"/etc/pki/tls/certs/ca-bundle.crt",                  // Another RHEL-based possibility
+		"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
+		"/etc/pki/tls/cacert.pem",                           // OpenELEC
+		"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // Often the think previous RHEL options symlink to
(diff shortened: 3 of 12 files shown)


jhiemstrawisc:
@haoming29, I know we happened to stumble on some of the same stuff. If you think I got any of this wrong, or if you found anything we need to add that I didn't catch, feel free to use this PR as a base and open a new PR from your own branch. If that's the case, I'll close this one and we can switch to yours.

haoming29:
This should also close #256

jhiemstrawisc:
@bbockelm @haoming29 What do we think the solution to the failing test should look like? It looks the problem is caused by trying to create the `TLSCACertificateFile` in `/etc/pki`, which the test runner doesn't have perms to do. If this is the default for Linux, we probably don't want to change the ultimate target location, but this seems to imply we can't create the key in the intended spot unless run with sudo or as root...

cmd/director.go
```diff
@@ -70,6 +70,13 @@ func getDirectorEndpoint() (string, error) {
 	return directorEndpointURL.String(), nil
 }
 
+func initDirector() error {
+	err := config.InitServer()
```
haoming29, line 74: Why do we want to pull this into a separate function if we don't have anything else to add?

jhiemstrawisc: It matches what we do with the origin [see here](https://github.com/PelicanPlatform/pelican/blob/469d87bf64a5f410f0dc751fcfcefec3609ba450/cmd/origin.go#L36) and it may be useful if we add any other components to the director like a web ui or monitoring.

jhiemstrawisc: (in short, partially because I'm fond of the symmetry. If you think it adds unneeded cruft at this point, I'm not opposed to removing)

haoming29: That makes sense to me, although I think we put other components to `serveDirector` now. Let's keep this function and hopefully we put new components here in the future.

turetske:
> @bbockelm @haoming29 What do we think the solution to the failing test should look like? It looks the problem is caused by trying to create the `TLSCACertificateFile` in `/etc/pki`, which the test runner doesn't have perms to do. If this is the default for Linux, we probably don't want to change the ultimate target location, but this seems to imply we can't create the key in the intended spot unless run with sudo or as root...
> 
> What are the consequences of trying to write to `/etc/pki` and if unable, using a default of `/etc/pelican/certificates`?

Why not overwrite the default value for `TLSCACertificateFile` location for the test to use the `t.tempDir`? Just for testing purposes?

cmd/director_serve.go
```diff
@@ -54,25 +47,6 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
...
-	// The director needs its own private key. If one doesn't exist, this will generate it
-	issuerKeyFile := param.IssuerKey.GetString()
-	err = config.GeneratePrivateKey(issuerKeyFile, elliptic.P256())
-	if err != nil {
-		return errors.Wrap(err, "Failed to generate director private key")
-	}
-
-	// Since director will provide a public JWK for token authentication
-	// We need to initialize public JWK and private JWK at the start
-	_, err = config.GenerateIssuerJWKS()
```
haoming29, line 66: Can we add back this line to where it was? I got this error at the start of the director with the changes:

```console
WARNING[2023-11-08T22:09:35Z] Failed to verify director service discovery token: key provider 0 failed: failed to find matching key: no key ID ("kid") specified in token 
```

jhiemstrawisc:
> @jhiemstrawisc Why not overwrite the default value for `TLSCACertificateFile` location for the test to use the `t.tempDir`? Just for testing purposes?

👆

haoming29 (changes requested):
Also, let's remove this as well. Thanks! https://github.com/PelicanPlatform/pelican/blob/ffbe0d54d63cee7c2a3e3894d3934ca3c3a9d080/cmd/namespace_registry_serve.go#L48

turetske:
> @jhiemstrawisc Why not overwrite the default value for `TLSCACertificateFile` location for the test to use the `t.tempDir`? Just for testing purposes?

More specifically, this is a parameter that can be set via the config file. If we can't handle that location being different in any way, then why is it something that is alterable in a pelican.yaml file? Also, if the user isn't root, then the location is being set via the InitServer() to be some other location already, I believe. The issue here is that the tests are being run with the assumption that the user is root, but the test driver doesn't have root permissions, hence the need for the temp director.

haoming29:
> @bbockelm @haoming29 What do we think the solution to the failing test should look like? It looks the problem is caused by trying to create the `TLSCACertificateFile` in `/etc/pki`, which the test runner doesn't have perms to do. If this is the default for Linux, we probably don't want to change the ultimate target location, but this seems to imply we can't create the key in the intended spot unless run with sudo or as root...
> 
> What are the consequences of trying to write to `/etc/pki` and if unable, using a default of `/etc/pelican/certificates`?

Do we actually set `Server.TLSCACertificateDirectory` to its default value in the code? Not sure if I missed anything but I can't find any viper.SetDefault("Server.TLSCACertificateDirectory") anywhere

turetske:
> Member



> > @bbockelm @haoming29 What do we think the solution to the failing test should look like? It looks the problem is caused by trying to create the `TLSCACertificateFile` in `/etc/pki`, which the test runner doesn't have perms to do. If this is the default for Linux, we probably don't want to change the ultimate target location, but this seems to imply we can't create the key in the intended spot unless run with sudo or as root...
> > What are the consequences of trying to write to `/etc/pki` and if unable, using a default of `/etc/pelican/certificates`?
> 
> Do we actually set `Server.TLSCACertificateDirectory` to its default value in the code? Not sure if I missed anything but I can't find any `viper.SetDefault("Server.TLSCACertificateDirectory")` anywhere

@haoming29 https://github.com/PelicanPlatform/pelican/blob/ffbe0d54d63cee7c2a3e3894d3934ca3c3a9d080/config/config.go#L402C1-L403C1

I think you were looking for the wrong name.

haoming29:
@turetske Sorry I shouldn't quote reply that comment, but I meant `Server.TLSCACertificateDirectory`. It's default value is being changed in this PR.

turetske:
@haoming29 Ah. I see, I was getting confused. I understand now. Then we don't set it. It's possibly something we want to if that's possible.

@jhiemstrawisc This changes my above comment. If we can't set it, then that's a problem for the reasons you described above, I didn't realize the directory was different than the file.

jhiemstrawisc:
@haoming29 @turetske, I think this is mostly sorted out. It boiled down to the fact that apparently there are a variety of default CA Certificate paths that should be checked based on which version of linux you're running. Rather than assuming we're on a RHEL derivative (which we did before by hardcoding `/etc/pki/tls/cert.pem`), the additions here now check all the same spots the x509 package checks for a certificate when the host OS is linux. I figure the folks who wrote the official Golang x509 package probably know a bit more than I do about all of this 😜  If none can be found, we create one in `$configDir/certificates`, which Pelican guarantees is writable.

On that note, we might eventually take inspiration from the x509 package in setting this up for Darwin and Windows: https://go.dev/src/crypto/x509/

haoming29 (approved):
LGTM! Tested locally and can confirm all servers can start at a fresh container. @jhiemstrawisc Feel free to add that default to TLS directory or push back. Thanks for taking care of these cleanups and refactoring!

config/config_linux.go
```diff
@@ -21,10 +21,49 @@
...
+	configDir := viper.GetString("ConfigDir")
+	return filepath.Join(configDir, "certificates", "cert.pem"), nil
+}
+
 func InitServerOSDefaults() error {
-	viper.SetDefault("Server.TLSCACertificateFile", "/etc/pki/tls/cert.pem")
-	return nil
+	if certLoc, err := findLinuxCACert(); err == nil {
+		viper.SetDefault("Server.TLSCACertificateFile", certLoc)
+		return nil
```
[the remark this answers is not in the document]
jhiemstrawisc: It looks like we primarily use this in auto generating the XRootD config, and if we have a cert dir, we use that instead of the cert file. So that we don't alter the behavior of this logic, I think we should leave it with no default... for now.

I'll update `parameters.yaml` to reflect this.

PR #363, opened 2023-11-08
Fix problems with config_test.go

joereuss12:
The tests are no longer affected by changes in the user's Pelican.yaml


jhiemstrawisc (approved):
Tested, and all looks good!

PR #365, opened 2023-11-09
Namespace registry serve cleanup

turetske:
These are the same commits that cleaned up the namespace registry code previously, just cherry-picked onto a new branch after the revert.

Hopefully it can be a quick approval and merge process.


haoming29 (approved):
LGTM!

PR #369, opened 2023-11-10
Fix issue with progress bar panic on download

joereuss12:
Fixes issue where if the cache sends a `permission denied` error, the next download tries to utilize the completed bar and panic. Now on failure it calls abort and removes the bar instance.


joereuss12:
@jhiemstrawisc I think this fixes the issue you were encountering however, I am having trouble recreating the issue so I kept this as a draft for now until tested. Do you want to give this fix a try and see if it looks right?

jhiemstrawisc (approved):
@joereuss12, of course, since the errors I was getting were transient in the first place, I can't seem to recreate them either with or without your additions here. Your changes look sensible and don't seem to cause any new errors, though, so let's fly with them.

Issue #308, opened 2023-10-31
Refactor general server_ui code out of origin_ui


turetske:
Due to implementing origin_ui first, there is a lot of origin specific code that actually code be more general. This should be refactored into server_ui instead of origin_ui so it can be used by other servers such as the cache, director, or namespace. Some of the code in cmd/origin_serve is also more general than just origin code.

haoming29:
In #291 I refactored much of the authorization check code in `web_ui` which can be helpful in implementing this issue

haoming29:
This can be very helpful for #264 as it will greatly reduce repetitive logic we would create.

haoming29:
Looks like #355 will be heavily dependent on this PR in order not to generate duplicate code and #355 was marked as part of 7.3.0. @bbockelm Do you think we should prioritize this issue to get #355 unblocked?

PR #373, opened 2023-11-13
Add authorization to Prometheus /metrics endpoint

### LINKED ISSUE #291: Add authorization to Prometheus scraping endpoint
Right now, if you go to `<server-url>/metrics` for any server, the Prometheus metrics will return even if you don't have any authentication token. We want to protect this endpoint by only allowing authenticated user to access it, although it might be a bit tricky to do so.

1. Issue a token signed by the server and attach it to the header of the Prometheus static scraper
2. Check token authentication at `/metrics` endpoint
3. Might need to come up a way to refresh the token periodically
4. Add regression tests

Note that this will break director's scraping mechanism. Will need a separate PR to allow director to carry a token when scraping origins' metrics.

haoming29:
Closes #291 

* Added Bearer authorization to Prometheus `/metrics` endpoint and issue token for the scraper of both director and server itself, with periodic reloading.
* Refactored authorization logic in `web_ui` to make it more flexible, clear, and reusable.
* Changed LoadDirectorPublicKey to return key, not pointer to the key

## Testing Instructions
* Spin up `director` and you should expect `/metrics` requests are all successful, and their frequency should be around 15s.
* Change `Monitoring.TokenExpiresIn` and `Monitoring.TokenRefreshInterval` to smaller value to ensure token refreshment is successful
* Run `director`, `registry` and `origin`. At `origin` end, confirm that its `/metrics` requests are also all successful.
* Go to origin's web APIs and check auth works as expected. i.e. `/api/v1.0/prometheus` as well as `/metrics` (You should be able to see it through web only when you logged in)

### DIFF
--- a/origin_ui/origin.go
+++ b/origin_ui/origin.go
@@ -187,6 +187,8 @@ func setLoginCookie(ctx *gin.Context, user string) {
 	issuerURL.Host = ctx.Request.URL.Host
 	now := time.Now()
 	tok, err := jwt.NewBuilder().
+		// TODO: We might want to come up with some names broader than this for a
+		// generic token for Web APIs, like web_ui.access
 		Claim("scope", "prometheus.read").
 		Issuer(issuerURL.String()).
 		IssuedAt(now).
--- a/origin_ui/origin.go
+++ b/origin_ui/origin.go
@@ -184,7 +184,7 @@ func setLoginCookie(ctx *gin.Context, user string) {
 
 	issuerURL := url.URL{}
 	issuerURL.Scheme = "https"
-	issuerURL.Host = ctx.Request.URL.Host
+	issuerURL.Host = config.ComputeExternalAddress()
 	now := time.Now()
 	tok, err := jwt.NewBuilder().
 		// TODO: We might want to come up with some names broader than this for a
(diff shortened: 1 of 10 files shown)


haoming29:
@turetske Sorry I just found a new bug in this PR. Trying to address it so please hold on reviewing it. Thanks!

haoming29:
> @turetske Sorry I just found a bug in this PR. Trying to address it so please hold on reviewing it. Thanks!

Fixed! Now it should be good to go.

origin_ui/origin.go
```diff
@@ -184,9 +184,11 @@ func setLoginCookie(ctx *gin.Context, user string) {
 
 	issuerURL := url.URL{}
 	issuerURL.Scheme = "https"
-	issuerURL.Host = ctx.Request.URL.Host
+	issuerURL.Host = config.ComputeExternalAddress()
 	now := time.Now()
 	tok, err := jwt.NewBuilder().
+		// TODO: We might want to come up with some names broader than this for a
+		// generic token for Web APIs, like web_ui.access
 		Claim("scope", "prometheus.read").
```
haoming29, line 192: This can be addressed in #308

turetske (approved):
LGTM!

PR #375, opened 2023-11-13
Always generate a CA if no host certificate present

### LINKED ISSUE #301: Always generate a CA certificate at server's start
Looking at the code I'm a bit confusing about the task of this issue so @bbockelm might need to give the details about this issue. Just serve as a placeholder now so that we don't forget.

haoming29:
Closes #301 

Since `SciToken` doesn't like self-signed host certificate, we want to always generate our own CA at pelican config directory when there is no host certificate present, so that we can sign our generated host certificate by the certificate and private key of our generated CA.

This PR also introduced more robust checking of required files to set up TLS. It now checks that:
1. A valid TLS host certificate is present (via Server.TLSCertificate)
2. The private key of TLS cert if present (via Server.TLSKey)
3. A valid CA certificate is present (via Server.TLSCACertificateFile)

Or it will generate a new host certificate, and in this process, it will need a CA cert/private key to sign the host certificate (to avoid self-signed host cert). It will check if:

1. A valid TLS CA certificate is present (via Server.TLSCACertificateFile)
2. The private key of TLS CA cert is present (via Server.TLSCAKey) (This is very rare situation as no CA wants to provide its private key to anyone else, but we leave the option here to be flexible)

And if the check passes, it will grab the CA cert and private key as provided to proceed, or it will generate new CA cert and its private key instead.

## Testing Instructions

* Make sure you have patched your XRootD and Scitoken plugin following #167, with XRootD `version >= 5.6.3` and SciToken `version >= 1.1.0`
* In your `pelican.yaml` file, remove previously set `Server.TLSCACertificate` value
* Run `cd /etc/pelican/` and `rm -rf ./certificates` to remove already generated certificates
* Run `director` `registry` and `origin`
* Note that origin self-test works as expected, no 401 or 403 reported for any file trasnfers.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -19,7 +19,6 @@
 package config
 
 import (
-	"crypto/elliptic"
 	"crypto/tls"
 	"crypto/x509"
 	_ "embed"
@@ -481,10 +480,8 @@ func InitServer() error {
 	if err != nil {
 		return err
 	}
-	err = GeneratePrivateKey(param.Server_TLSKey.GetString(), elliptic.P256())
-	if err != nil {
-		return err
-	}
+
+	// Check if we have required files in place to set up TLS, or we will generate them
 	err = GenerateCert()
 	if err != nil {
 		return err
--- a/config/config_default.go
+++ b/config/config_default.go
@@ -29,9 +29,9 @@ import (
 
 func InitServerOSDefaults() error {
 	// Windows / Mac don't have a default set of CAs installed at
-	// a well-known location as is expected by XRootD.  Hook up a
-	// fake one instead.
-
+	// a well-known location as is expected by XRootD. We want to always generate our own CA
+	// if Server_TLSCertificate (host certificate) is not explicitly set so that
+	// we can sign our host cert by our CA instead of self-signing
 	tlscaFile := filepath.Join(viper.GetString("ConfigDir"), "certificates", "tlsca.pem")
 	viper.SetDefault("Server.TLSCACertificateFile", tlscaFile)
 
--- a/config/config_linux.go
+++ b/config/config_linux.go
@@ -21,49 +21,16 @@
 package config
 
 import (
-	"os"
 	"path/filepath"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 )
 
-// Different distros have different default CACertificate locations. Instead of trying to figure out
-// the version of linux we use, check all the well-known places until we find one that exists.
-func findLinuxCACert() (string, error) {
-	// These values pulled from the same place x509 package looks:
-	// https://go.dev/src/crypto/x509/root_linux.go
-	certFileLocations := []string{
-		"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
-		"/etc/pki/tls/cert.pem",                             // One RHEL-based possibility
-		"/etc/pki/tls/certs/ca-bundle.crt",                  // Another RHEL-based possibility
-		"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
-		"/etc/pki/tls/cacert.pem",                           // OpenELEC
-		"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // Often the think previous RHEL options symlink to
-		"/etc/ssl/cert.pem",                                 // Alpine Linux
-	}
-
-	for _, certLoc := range certFileLocations {
-		if file, err := os.Open(certLoc); err == nil {
-			file.Close()
-			return certLoc, nil
-		} else if !errors.Is(err, os.ErrNotExist) {
-			return "", err
-		}
-	}
-
-	// If we didn't find a cert in a well-known location, we'll wind up creating one later.
-	// In that case, we need to make sure it's someplace we have permission to write to, so we'll
-	// put it where we put other Pelican certs
-	configDir := viper.GetString("ConfigDir")
-	return filepath.Join(configDir, "certificates", "cert.pem"), nil
-}
-
 func InitServerOSDefaults() error {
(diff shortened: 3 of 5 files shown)


haoming29:
@jhiemstrawisc Sorry

jhiemstrawisc:
@haoming29, I'm trying to test this, but I still get a 403 for the self-monitoring cycle. Here's what I'm attempting:

`yum list xrootd xrootd-scitokens` reports that both packages are at version `1:5.6.3-1.el8`
`yum list scitokens-cpp` reports version `1.1.0-1.el8`

I then `rm -rf /etc/pelican/certificates/*` and in the following order across different terminal instances do:
`pelican registry serve -p 8446 -d`
`pelican director serve -p 8445 -d`
`pelican origin serve -p 8444 -v /test:/jhiemstra/test -d`

My Pelican configuration for the test:
```
Federation:
  NamespaceUrl: https://823ed179c67b:8446
  DirectorUrl: https://823ed179c67b:8445

Registry:
  DbLocation: /workspaces/pelican/pelican.sqlite
```

The reason I think something isn't working properly for me is that at the origin I get:
```
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 XrootdBridge: unknown.9:35@823ed179c67b login as nobody  daemon=xrootd
DEBUG[2023-11-15T16:14:33Z] MonPacket: Received a user login packet      
DEBUG[2023-11-15T16:14:33Z] GetSIDRest inputs: https/unknown.9:215444624327513@823ed179c67b
&p=https&n=&h=[::ffff:172.17.0.2]&o=&r=&g=&m=&I=4 
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 scitokens_GenerateAcls: Failed to deserialize SciToken: Metadata resource https://823ed179c67b:8443/.well-known/oauth-authorization-server is missing 'jwks_uri' string value  daemon=xrootd
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 scitokens_Access: Failed to generate ACLs for token  daemon=xrootd
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 acc_Audit: unknown.9:35@823ed179c67b deny https *@[::ffff:172.17.0.2] create /pelican/monitoring/self-test-2023-11-15T16:14:33Z.txt  daemon=xrootd
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 scitokens_GenerateAcls: Failed to deserialize SciToken: Metadata resource https://823ed179c67b:8443/.well-known/oauth-authorization-server is missing 'jwks_uri' string value  daemon=xrootd
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 scitokens_Access: Failed to generate ACLs for token  daemon=xrootd
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 acc_Audit: unknown.9:35@823ed179c67b deny https *@[::ffff:172.17.0.2] excl_create /pelican/monitoring/self-test-2023-11-15T16:14:33Z.txt  daemon=xrootd
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 ofs_open: unknown.9:35@823ed179c67b Unable to create /pelican/monitoring/self-test-2023-11-15T16:14:33Z.txt; permission denied  daemon=xrootd
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 unknown.9:35@823ed179c67b Xrootd_Response: sending err 3010: Unable to create /pelican/monitoring/self-test-2023-11-15T16:14:33Z.txt; permission denied  daemon=xrootd
INFO[2023-11-15T16:14:33Z] 231115 16:14:33 92558 XrootdXeq: unknown.9:35@823ed179c67b disc 0:00:00 (send failure)  daemon=xrootd
WARNING[2023-11-15T16:14:33Z] Self-test monitoring cycle failed during test upload: Error response 403 from monitoring upload: 403 Forbidden
```

The errors seem to point to a missing `jwks_uri` at the origin. A curl of that endpoint returns the JSON:
```
{
  "token_endpoint":"https:\/\/localhost:8443\/.oauth2\/token"
}
```
Any ideas what I'm doing wrong?

haoming29:
> Member

Ah. @jhiemstrawisc  Can you try adding the following to your `pelican.yaml` config file? Not sure if this is going to fix it but I remember we had issue without this when trying to get self-test working. Sorry for missing this info in the description!

```yaml
Origin:
  EnableMacaroons: false
```

jhiemstrawisc:
Yep that was the fix. Do you know if we need to be worried about disabling macaroons? If you're confident we don't, I'll go ahead and approve/merge. I just don't fully understand the implications of turning off macaroons when it's something we enable by default. If we enable both macaroons and self testing by default, we should probably either figure out how to make those play well together, or turn one of them off. That sounds like something for a different pull request, though.

haoming29:
I think the initial thoughts when Brian figured out this is that Macaroons matches all the paths as xrootd file directory which includes [what we set as the xrootd issuer location](https://github.com/PelicanPlatform/pelican/blob/c87f4286c9f3a5850118795e287e4c80f45689a2/xrootd/authorization.go#L347) i.e. `/.well-known/openid-configuration`. As you can tell from your previous debug output, when SciToken is trying to reach out to the issuer to grab the public key, it instead was directed to `/.well-known/oauth-authorization-server` which we never configured. This is causing xrootd or scitoken unable to find the correct public key for token verification.

I remember Brian says we are not using Macaroons for anything right now so it's safe to turn it off. But we'd better confirm with @bbockelm 

Either way, I think this issue is unrelated to this PR as the self-test won't work without turning off `EnableMacaroons`, but I agree we should create a new issue documenting this and try to find a fix (hopefully not by patching xrootd again).

jhiemstrawisc:
Sounds good. Since the scope of this PR is to handle creating the CA certificate and corresponding key, I'll approve and merge, but let's come back to disabling macaroons at another point.

PR #376, opened 2023-11-15
Introduce enum for health status and component names for SetComponentHealthStatus

### LINKED ISSUE #351: Create enum type healthStatus and Refactor SetComponentHealthStatus
Ref: https://github.com/PelicanPlatform/pelican/pull/305#discussion_r1385266375

We want to create an enum type `healthStatus` and read through code and refactor instances of  `SetComponentHealthStatus` to use enum as status code instead of string literals.

haoming29:
Closes #351 

We want to be more robust in setting component health status. This PR introduces enum for both the health status and the component name to record. Related code are updated to adhere to this new convention.

I manually tested that `SetComponentHealthStatus` work without causing errors at `origin` side where the web ui dashboard can correctly display health status of all existing components, and when there's an error from a component, the health status updates as expected.

For reviewer, I recommend spinning up your `origin` and go to `/view` and check that all status on top right side is expected. You can then turn off your `director` and wait for the next origin advertisement to fail to see if that's correctly displayed.

### DIFF
--- a/cmd/cache.go
+++ b/cmd/cache.go
@@ -45,11 +45,8 @@ var (
 func initCache() error {
 	err := config.InitServer()
 	cobra.CheckErr(err)
-	err = metrics.SetComponentHealthStatus("xrootd", "critical", "xrootd has not been started")
-	cobra.CheckErr(err)
-	err = metrics.SetComponentHealthStatus("cmsd", "critical", "cmsd has not been started")
-	cobra.CheckErr(err)
-
+	metrics.SetComponentHealthStatus(metrics.OriginCache_XRootD, metrics.StatusCritical, "xrootd has not been started")
+	metrics.SetComponentHealthStatus(metrics.OriginCache_CMSD, metrics.StatusCritical, "cmsd has not been started")
 	return err
 }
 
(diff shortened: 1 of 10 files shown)


cmd/cache.go
```diff
@@ -45,11 +45,8 @@ var (
 func initCache() error {
 	err := config.InitServer()
 	cobra.CheckErr(err)
-	err = metrics.SetComponentHealthStatus("xrootd", "critical", "xrootd has not been started")
-	cobra.CheckErr(err)
-	err = metrics.SetComponentHealthStatus("cmsd", "critical", "cmsd has not been started")
-	cobra.CheckErr(err)
-
+	metrics.SetComponentHealthStatus(metrics.OriginCache_XRootD, metrics.StatusCritical, "xrootd has not been started")
```
jhiemstrawisc, line 48: Do we lose anything here by not checking the error from this? The previous `cobra.CheckErr` performs an `os.Exit(1)` if an error exists, so if we don't redefine `err` at the call to `metrics.SetComponentHealthStatus()`, this function will always return nil.

jhiemstrawisc (approved):
Looks clean to me and , see my one question in the comment. I'll approve, and if you don't think it's anything to worry about, go ahead and merge.

cmd/cache.go
haoming29: I think that's the expected behavior. Now with the enum type ensured, we don't return anything from `SetComponentHealthStatus`. The error returned before is simply checking if our status string literals matches what we expect.

PR #377, opened 2023-11-15
Reimplimented recursive downloads

joereuss12:
The functionality of recursive downloads has been restored. When specifying with the -r flag and targetting a dir, all the files will be downloaded within that directory. This functionality currently works however the progress bars are a little funky so that still needs fixing.


joereuss12:
From playing around with things, it seems like the progress bars mostly look okay when the debug log is not activated. When debug is activated, it keeps wanting to print things during the download causing multiple bars for 1 file to be printed. I will create a new issue for this but I think we should implement what we have since it does work correctly. It also seems like it cannot get the correct total for the files when doing this recursively either.

client/bearer_auth.go
```diff
@@ -0,0 +1,61 @@
+package client
```
bbockelm, line 1: Missing the copyright statement.

client/bearer_auth.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	//"fmt"
```
bbockelm, line 4: Don't include commented-out lines.

client/bearer_auth.go
```diff
@@ -0,0 +1,61 @@
...
+	"io"
+	"net/http"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/studio-b12/gowebdav"
+)
+
+// BasicAuth structure holds our credentials, this is the authorizer
+type BearerAuth struct {
```
bbockelm, line 13: Type should not be exported unless required.  Can you change to `bearerAuth`?

client/bearer_auth.go
```diff
@@ -0,0 +1,61 @@
...
+	token string
+}
+
+// BearerAuthenticator is an Authenticator for BearerAuth
+type BearerAuthenticator struct {
```
bbockelm, line 18: Similar to above type, do not export.

client/bearer_auth.go
```diff
@@ -0,0 +1,61 @@
...
+// Authorize the current request
+func (b *BearerAuthenticator) Authorize(c *http.Client, rq *http.Request, path string) error {
+	rq.Header.Add("Authorization", "Bearer "+b.token) //set the header with the token
+	return nil
+}
+
+// Verify verifies the authentication
+func (b *BearerAuthenticator) Verify(c *http.Client, rs *http.Response, path string) (redo bool, err error) {
+	if rs.StatusCode == 401 {
+		//err = NewPathError("Authorize", path, rs.StatusCode)
```
bbockelm, line 41: Why is this commented out?  There should be an error generated, no?

client/bearer_auth.go
```diff
@@ -0,0 +1,61 @@
...
+		log.Errorf("Authorize: %s, %v", path, rs.StatusCode)
```
bbockelm, line 42: Seems a bit loud as the higher-up user should catch this as well.  Consider `Debug`-level logging.

client/bearer_auth.go
```diff
@@ -0,0 +1,61 @@
...
+	return nil
+}
+
+// Clone creates a Copy of itself
+func (b *BearerAuthenticator) Clone() gowebdav.Authenticator {
+	// no copy due to read only access
+	return b
+}
+
+// // String toString
```
bbockelm, line 58: Remove commented-out code.

director/redirect.go
```diff
@@ -223,8 +223,8 @@ func RedirectToCache(ginCtx *gin.Context) {
 			ginCtx.Writer.Header()["X-Pelican-Token-Generation"] = []string{tokenGen}
 		}
 	}
-	ginCtx.Writer.Header()["X-Pelican-Namespace"] = []string{fmt.Sprintf("namespace=%s, require-token=%v",
-		namespaceAd.Path, namespaceAd.RequireToken)}
+	ginCtx.Writer.Header()["X-Pelican-Namespace"] = []string{fmt.Sprintf("namespace=%s, require-token=%v, dirlisthost=%s",
```
bbockelm, line 226: I know that for topology, we've committed to calling this `dirlisthost`.  However, that seems like a really bad name for Pelican (as we don't make any assumptions about a directory!).

How about `collections-url`?

bbockelm (changes requested):
A small number of minor items.

Please also add a unit test covering the bearer token authentication before we do the final merge.

joereuss12:
Changed to be a draft because now I can make unit tests for this functionality from the merging of #390

joereuss12:
Marking this as ready for review since the end-to-end tests are currently stalled. See #391

bbockelm (approved):
LGTM after the prior round of changes.

PR #381, opened 2023-11-15
Fix Config Bug

### LINKED ISSUE #380: Config page breaks on receiving a null value
Currently config page content breaks when there is a null value in the config response.

CannonLock:
- Add strong typing to avoid future bugs
- Add null, string[] and int[] support

Fixes #380

### DIFF
--- a/origin_ui/src/app/(dashboard)/config/page.tsx
+++ b/origin_ui/src/app/(dashboard)/config/page.tsx
@@ -35,122 +35,21 @@ import {isLoggedIn} from "@/helpers/login";
 
 type duration = number | `${number}${"ns" | "us" | "µs" | "ms" |"s" | "m" | "h"}`;
 
-interface Config {
-    ConfigDir: string;
-    Debug: boolean;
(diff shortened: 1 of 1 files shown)


origin_ui/src/app/(dashboard)/config/page.tsx
```diff
@@ -35,100 +35,21 @@ import {isLoggedIn} from "@/helpers/login";
...
-        PortLower: number;
-        PortHigher: number;
-    };
+type Config = {
+    [key: string]: ConfigValue
 }
 
-function sortConfig (a: any, b: any) {
-    if(typeof a[1] == 'object' && typeof b[1] != 'object'){
+type ConfigValue = Config | string | number | boolean | null | string[] | number[] | duration
```
haoming29, line 42: Thanks for making this change! It will be greatly improve the robustness of the way we displaying config values

haoming29 (approved):
LGTM! Manually added a new config locally and the `/config` page still works as expected. Thanks for making this so flexible!

PR #382, opened 2023-11-16
Refactor init_server_cred to be more clear on their usage

### LINKED ISSUE #360: Clarify "private key" / "public key" and its usage and functions in init_server_cred
We really overloaded the name "private key" in our codebase and there are multiple references to private keys in adjacent places which really confuse people and cause bug. As a follow-up to #359, #339 and #256, we want to be more clear on what do we mean by private keys, where we generate them, and what's the usage of them.

* Add function header comment to all the functions inside of `init_server_cred.go` to explain their purposes. Pay close attention to `GeneratePrivateKey` which is used for generating ECDSA private key for both TLS certificate and issuer key (for signing JWT)
* Rename top-level variable `privateKey` to `issuerPrivateJWK` and explain that this is the private JWK for signing JWT
* Rename function `GenerateIssuerJWKS` to `GetIssuerJWKS`, to emphasize that people should use this function to get the public JWK (encapsulated in a JWKS)
* Rename function `LoadPublicKey` to something like `loadIssuerJWKS` to reflect that this generates the private JWK for signing JWT and returns the public JWK for verifying a JWT, which happened to be wrapped in a JWKS. Also, this function should be a "private" helper function and comment should be added to let people use `GetIssuerJWKS` for most of the cases, unless they know what they are doing
* Rename argument `tlsKey` in `LoadPrivateKey` to `keyLocation`, as this function is very generic and can load any ecdsa private key, not just the one for TLS certficate.

haoming29:
Closes #360

### DIFF
--- a/cmd/namespace_client.go
+++ b/cmd/namespace_client.go
@@ -37,7 +37,6 @@ import (
 var withIdentity bool
 var prefix string
 var pubkeyPath string
-var privkeyPath string
 
 func getNamespaceEndpoint() (string, error) {
 	namespaceEndpoint := param.Federation_NamespaceUrl.GetString()
@@ -61,15 +60,6 @@ func registerANamespace(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	privkey := privkeyPath
-	if privkey == "" {
-		privkey = param.IssuerKey.GetString()
-	}
-	if privkey == "" {
-		log.Error("Private key file is not set; specify its location with the --privkey option or by setting the IssuerKey configuration variable")
-		os.Exit(1)
-	}
-
 	namespaceEndpoint, err := getNamespaceEndpoint()
 	if err != nil {
 		log.Errorln("Failed to get NamespaceURL from config: ", err)
@@ -87,7 +77,7 @@ func registerANamespace(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	publicKey, err := config.LoadPublicKey("", privkey)
+	publicKey, err := config.GetIssuerPublicJWKS()
 	if err != nil {
 		log.Error("Error: Failed to retrieve public key: ", err)
 		os.Exit(1)
@@ -105,7 +95,7 @@ func registerANamespace(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	privateKeyRaw, err := config.LoadPrivateKey(privkey)
+	privateKeyRaw, err := config.LoadPrivateKey(param.IssuerKey.GetString())
 	if err != nil {
 		log.Error("Failed to load private key", err)
 		os.Exit(1)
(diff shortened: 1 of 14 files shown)


cmd/namespace_client.go
```diff
@@ -247,13 +237,21 @@ func init() {
 	deleteCmd.Flags().StringVar(&prefix, "prefix", "", "prefix for delete namespace")
 
 	namespaceCmd.PersistentFlags().String("namespace-url", "", "Endpoint for the namespace registry")
-	err := viper.BindPFlag("NamespaceURL", namespaceCmd.PersistentFlags().Lookup("namespace-url"))
-	if err != nil {
+	// TODO: I don't think this flag is bound to any viper config name we have.
+	// Should instead be Federation.NamespaceURL
+	if err := viper.BindPFlag("NamespaceURL", namespaceCmd.PersistentFlags().Lookup("namespace-url")); err != nil {
```
jhiemstrawisc, line 242: Absolutely, this one slipped past us when we refactored all of the viper variables. In fact, I'd go so far as to say we should change all of them to `Federation.RegistryUrl`, because I think that's a lot more clear, but we might be locked in because of our concerns about backwards compatibility... For now, can you make the change to `Federation.NamespaceUrl`?

config/config.go
```diff
@@ -476,7 +476,7 @@ func InitServer() error {
 
 	// As necessary, generate a private keys, JWKS and corresponding certs
 	// Note: GenerateIssuerJWKS will also generate a private key in the location stored by the viper var "IssuerKey"
-	_, err = GenerateIssuerJWKS()
+	_, err = GetIssuerPublicJWKS()
```
jhiemstrawisc, line 479: Can you also change the previous comment to reflect the new function name?

docs/parameters.yaml
```diff
@@ -46,7 +46,10 @@ components: ["origin", "nsregistry", "director"]
 ---
 name: IssuerKey
 description: >-
-  The name of the file containing a service's private `issuer.jwk` key.
+  A filepath to the file containing a ecdsa-encrypted private key which later will be parsed
```
jhiemstrawisc, line 49: I think it's more accurate to say "... containing a PEM-encoded ecdsa private key" -- is the key itself actually encrypted with ecdsa?

jhiemstrawisc (changes requested):
We've been really hurting for a refactor like this. Thanks for taking care of it! Just a few small changes, and then I think we're good to go

docs/parameters.yaml
haoming29: I think yours is more accurate.

cmd/namespace_client.go
haoming29: I will post renaming to `Federation.RegistryUrl` as a separate issue and @bbockelm might have a better idea of whether this will have compatibility issue for clients that's currently running as `stashcp` or `osdf` as that's the only place where I think we have people actually using them and will potentially require an update in the future.

haoming29: Ref #388

jhiemstrawisc (approved):
Changes LGTM. Thanks!

PR #385, opened 2023-11-16
Git test action updates

turetske:
Updated the GitHub test workflow to use our linux dev image (so the xrootd e2e test will work). Also updates the image to use a more recent version of nodejs.


joereuss12 (approved):
LGTM

PR #386, opened 2023-11-16
Adjusted dev.Dockerfile to run make web-build properly

turetske:
Updated dev image to have a later version of nodejs in order to run make web-build properly


haoming29:
Do we also want to explicitly say which node version we are using as the engine? Something like the following in the `package.json` @CannonLock 

```json
{
    "engines": {
        "node" : 18
    }
}
```

CannonLock (approved):
LGTM

PR #392, opened 2023-11-17
Fixed scitokens generation bug

### LINKED ISSUE #390: Remove multiple duplicate Issuer entries for generated SciTokens files
When we have multiple ads with the same Issuer, we are creating multiple entries which look like:

```
[Issuer 1]
issuer: https://my-issuer
basepath: p1

[Issuer 2]
issuer: https://my-issuer
basepath: p2
```

This is a bug as SciTokens doesn't allow for duplicate urls for the same issuer and will deny requires trying to reach p2

It should be:

```
[Issuer 1]
issuer: https://my-issuer
basepath: p1, p2
```

turetske:
Close #390 

Scitokens generation will now create a map of issuer urls and add a new base path to an already existing issuer url if it exists rather than adding a new Issuer struct

### DIFF
--- a/xrootd/authorization.go
+++ b/xrootd/authorization.go
@@ -63,8 +63,8 @@ type (
 
 	// Top-level configuration object for the template
 	ScitokensCfg struct {
-		Global  GlobalCfg
-		Issuers []Issuer
+		Global    GlobalCfg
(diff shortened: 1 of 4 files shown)


jhiemstrawisc (approved):
Tested and it looks good to me.

Issue #262, opened 2023-10-23
Issue OA4MP tokens using CILogon


bbockelm:
Extend the OA4MP integration to make the device code authorization mode function via CILogon authorization.

This should only go as far as "template based" scopes and authorization -- e.g., specify that a user can have scope `storage.read:/some/prefix/$USERNAME`.  Assume that the username claim is configurable as well as the group name(s) claim.

PR #393, opened 2023-11-20
Add authorization policies for OA4MP

bbockelm:
This adds a simple authorization policy templates to the OA4MP issuer, enabling CILogon-based authentication to be used.


bbockelm:
Fixes #262

docs/parameters.yaml
```diff
@@ -565,33 +670,51 @@ description: >-
...
+name: OIDC.Issuer
+description: >-
+  The URL of the OIDC issuer.  If set, OIDC auto-discovery may be used to find other endpoints (token, user info,
+  device auth).
+type: url
+default: none
+components: ["nsregistry", "origin"]
 ---
 
+############################
```
jhiemstrawisc, line 715: I think this section is already covered by line 530

jhiemstrawisc (changes requested):
One small change and one general question.  My question is whether we want to start moving toward having Pelican handle things like creating the `xrootd` and `tomcat` user along with all the other setup those require. We currently assume that origins, issuers, etc are all being run from one of the containers we export because those containers do a lot of the setup needed to handle our dependencies. Do we ever plan to support simple setup of an issuer-enabled origin outside of this context?

bbockelm:
> Do we want to start moving toward having Pelican handle things like creating the xrootd and tomcat user along with all the other setup those require.

Nah - that's probably where I draw the line.  All the other setup we do is self-contained within Pelican: adding a user mucks with the system configuration.

Now, it's traditional that an RPM package may create these users.  I could see the `pelican-server` RPM doing this.

jhiemstrawisc (approved):

Issue #399, opened 2023-11-21
Federation URL is not parsed correctly if passed via CLI with a path


jhiemstrawisc:
When I tried to do:
`pelican namespace list -f https://osg-htc.org/osdf/itb`
I got the following error:
`ERROR[2023-11-21T19:50:56Z] Failed to initialize the client:  Failure when parsing federation metadata at https://osg-htc.org/.well-known/pelican-configuration/osdf/itb: invalid character '<' looking for beginning of value `

Pelican is still trying to use `https://osg-htc.org` as the federation when it should be using `https://osg-htc.org/osdf/itb` as the federation base.

PR #400, opened 2023-11-21
Fix small bugs in DiscoverFederation()

jhiemstrawisc:
Previously, running pelican with `-f foo.com/a/path` would result in DiscoverFederation() trying to get metadata from `foo.com/.well-known/pelican-configuration/a/path` instead of the correct `foo.com/a/path/.well-known/pelican-configuration`. This fixes that.

It also uses the `url.JoinPath()` method for parsing URL paths instead of `path.Join()`. We never really do much with Windows, but if we did, that would prevent federation discovery from working, because it would try to use \ as a path separator in a URL. Whoops.

Closes issue #399

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -178,7 +178,10 @@ func DiscoverFederation() error {
 	}
 
 	discoveryUrl, _ := url.Parse(federationUrl.String())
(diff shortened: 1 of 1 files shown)


turetske (approved):
LGTM!

PR #401, opened 2023-11-21
Add swagger spec and include auth APIs and common APIs

### LINKED ISSUE #352: Swagger schema for Pelican
Right now, it's difficult to get an outsider's view of the APIs of Pelican -- and, as things grow, it'll only get worse.

The Swagger ecosystem has a variety of tools for checking schema compliance which would be useful in unit tests and a nice UI which allows external developers to hand-invoke / test the APIs via browser.

We should write a Swagger / OpenAPI schema describing our existing APIs and have an agreed-upon rule that new APIs should come with the Swagger schema.  A quick perusal of the golang ecosystem, it seems that OpenAPI v2 is the most widely supported - let's target that.

haoming29:
Closes #352 

This is a hand-coded Swagger specification of APIs in Pelican servers. For now, I included APIs for authentication and common APIs (/config, /health, etc.)

As a next step, we might want to programmatically generate this spec using annotations on handler functions, and there's existing Go package to handle spec generation for us: [swaggo/swag](https://github.com/swaggo/swag)

If you are using VSCode as your IDE, you can render the Swagger UI from this `.yaml` file by installing the [OpenAPI (Swagger) Editor](https://marketplace.visualstudio.com/items?itemName=42Crunch.vscode-openapi)

### DIFF
--- a/swagger/pelican-swagger.yaml
+++ b/swagger/pelican-swagger.yaml
@@ -1,0 +1,282 @@
+swagger: '2.0'
+info:
+  title: Pelican Server APIs
+  description: "[Pelican](https://pelicanplatform.org/) provides an open-source software platform for federating 
+      dataset repositories together and delivering the objects to computing capacity such as the [OSPool](https://osg-htc.org/services/open_science_pool.html)
+
+
+      This is the API documentation for various APIs in Pelican servers (director, registry, origin, etc) 
+      to communicate with each other and in-between users accessing the servers. 
+      
+
+      For how to set up Pelican servers, please refer to the documentation at [docs.pelicanplatform.org](https://docs.pelicanplatform.org/)"
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0
+  contact:
+    name: API Support via Pelican GitHub Issue
+    url: https://github.com/PelicanPlatform/pelican/issues
+  version: '1.0'
+basePath: /api/v1.0/
+consumes:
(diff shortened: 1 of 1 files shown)


CannonLock:
@haoming29 Is there a reason you picked yaml over json? 

I am supposed to consume this in the frontend so it will be more convenient for me to have it as json.

haoming29:
> @haoming29 Is there a reason you picked yaml over json?
> 
> I am supposed to consume this in the frontend so it will be more convenient for me to have it as json.

@bbockelm suggests to use `yaml` to be more human-readable and I thought the swagger-ui npm package support `.yaml` by default? https://github.com/swagger-api/swagger-ui/blob/HEAD/docs/usage/configuration.md#configuration

CannonLock:
Okay, yaml works for now. When this is autogenerated we should do JSON as well as yaml. 

The npm package only consumes json, its not a hard translation but I don't see any reason to make it if we can auto generate the file. -> https://www.npmjs.com/package/swagger-ui-react

haoming29:
> Okay, yaml works for now. When this is autogenerated we should do JSON as well as yaml.
> 
> The npm package only consumes json, its not a hard translation but I don't see any reason to make it if we can auto generate the file. -> https://www.npmjs.com/package/swagger-ui-react

That makes sense to me

CannonLock (approved):
Approved.

Make sure to add another commit to your auth restructure so that these paths are maintained.

PR #410, opened 2023-11-27
Data Exports Overview Table #220

### LINKED ISSUE #220: Data Exports Overview Table
Create table view of Data exports.

Each location exported to the federation has at least three pieces of data: the type (POSIX, S3), the local path (/foo), and the federation routing key (/bar).

CannonLock:
- Implements data export table with existing config

Closes #220


CannonLock:
Screenshot 2023-11-27 at 11 55 08 AM (2)
Screenshot 2023-11-27 at 11 53 55 AM (2)

CannonLock:
Handles extra long, null, empty values and a broken config url gracefully.

@haoming29 I did notice that I was getting `"NamespacePrefix": ""` and `"Mount": ""`. If those values are not set I think they should be null.

CannonLock:
I don't like the wasted space below the status box, but I don't want to spend time on perfection until things are more settled.

haoming29:
> Handles extra long, null, empty values and a broken config url gracefully.
> 
> @haoming29 I did notice that I was getting `"NamespacePrefix": ""` and `"Mount": ""`. If those values are not set I think they should be null unless "" is a valid value.
> 
> Above are examples of a extra long value, and a value that isn't in the config, defaulting to NULL.

NamespacePrefix value should not be null. This is a bug and I fixed it here: #395  That value should be present once 395 is merged.

origin_ui/src/components/DataExportTable.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    const getData = async () => {
+        let response = await fetch("/api/v1.0/config")
+        if (response.ok) {
+            const responseData = await response.json()
+
+            setData([{
+                "Type": "POSIX",
+                "Local Path": ["", undefined].includes(responseData?.Xrootd?.Mount) ? "NULL" : responseData?.Xrootd?.Mount,
+                "Routing Key": ["", undefined].includes(responseData?.Origin?.NamespacePrefix) ? "NULL" : responseData?.Origin?.NamespacePrefix
```
haoming29, line 82: Routing Key sounds a bit strange to me (I don't understand what this means at the first glance, especially when the values were missing). Do we have a strong reason to name it this way rather than "Namespace Prefix"?

haoming29 (changes requested):
I got this weird double line for the table row on my browser. Is this intended?

Screenshot 2023-11-27 at 3 52 11 PM

origin_ui/src/components/DataExportTable.tsx
CannonLock: Nah, I had 'Routing Key' in my notes from a meeting with Brian, but "Namespace Prefix" is definetly better.

CannonLock:
@haoming29 I can't reproduce the extra space in the table. Any insight on your end what might be causing that? Can you inspect element and see where that gap is coming from?

haoming29:
Hey @CannonLock. Could you rebase the branch against main and fix the merge conflict first? The last time I switch between the old branch and the new web_ui frontend branch (main), my git crashed locally and I have to delete my project and re-pull from remote.

CannonLock:
@haoming29 Are merge commits not allowed? Any reason I should specifically rebase?

haoming29:
> @haoming29 Are merge commits not allowed? Any reason I should specifically rebase?

Merge is totally fine! Sorry I just meant to resolve the conflict.

CannonLock:
@haoming29 What browser are you using? Safari?

haoming29:
> @haoming29 What browser are you using? Safari?

@CannonLock Happened to me on both Chrome and Safari.

CannonLock:
@haoming29 I can't reproduce in any browser and I can't remove the border as that is what discriminate cell borders. 

Are you alright with merging and I will fix when I can reproduce in the future?

haoming29:
> @haoming29 I can't reproduce in any browser and I can't remove the border as that is what discriminate cell borders.
> 
> Are you alright with merging and I will fix when I can reproduce in the future?

Yeah that's not a big deal. I'm pretty sure `border-bottom: none` also solve the issue but it's OK if the cell can't be discriminated with that.

haoming29 (approved):
LGTM. The extra line in the table is a minor visual issue that's not critical to fix.

CannonLock:
@haoming29 Yeah my concern is that this isn't always going to be a single row so that border will be used eventually.

PR #411, opened 2023-11-27
Federation Overview #221

### LINKED ISSUE #221: Federation Overview
Create a info box on the Federation overview.

What's the name of the federation the origin is in?  Where's the namespace registry?  Director?  I'm envisioning links to each even if the director and namespace don't have a UI (yet!).

CannonLock:
- Implements Federation overview which provides a series of links to all Federation components
- Quiet Failure

Closes #221


haoming29 (approved):
LGTM. The URLs can't work if I ran origin in a container (the hostnames are invalid) but I assume in production it should work as expected.

CannonLock:
@haoming29 What do you mean they do not work? I have then set to ingest the config values, did you run into a case where they did not do that correctly?

PR #412, opened 2023-11-27
Adjusted advertiseOSDF namespaceAd generation

turetske:
-- Now uses basepaths and issuers if a token is required

Also, in the advertise_test unit testing code, I switched the `actual` vs `expected` values since they were reversed and it was annoying me.

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -48,6 +48,12 @@ type (
 		VaultServer   string `json:"vault_server"`
 	}
 
+	Scitokens struct {
+		BasePath   []string `json:"base_path"`
+		Issuer     string   `json:"issuer"`
+		Restricted []string `json:"restricted_path"`
+	}
+
 	Namespace struct {
 		Caches               []Server             `json:"caches"`
 		Origins              []Server             `json:"origins"`
@@ -55,6 +61,7 @@ type (
 		DirlistHost          string               `json:"dirlisthost"`
 		Path                 string               `json:"path"`
 		ReadHTTPS            bool                 `json:"readhttps"`
+		Scitokens            []Scitokens          `json:"scitokens"`
 		UseTokenOnRead       bool                 `json:"usetokenonread"`
 		WritebackHost        string               `json:"writebackhost"`
 	}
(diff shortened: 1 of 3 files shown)


jhiemstrawisc (changes requested):
@turetske, what's the expected header response for a namespace from topology that has multiple base paths? For example, the namespace `/igwn/cit` reports multiple basepaths across two issuers in topology. If I fire up a director built on this code and I curl for `/igwn/cit`, this is the output from the `X-Pelican-Token-Generation` header:
`< x-pelican-token-generation: issuer=https://cilogon.org/igwn, max-scope-depth=0, base-path=/igwn/cit`
After some testing, there appears to be no way to get a response with any of the other base paths.

When the director is picking a namespace advertisement to associate with a requested resource, it matches by the first namespace path that has the longest prefix of the resource -- If I request `/igwn/cit/foobar`, I get the first nsAd that has the path `/igwn/cit`, but none of the others. As such, of the 6 namespace ads/basepath combos for `/igwn/cit` that this generates, only the first one in the cache will ever make it back to the client.

Since this nsAd struct is used internally only by the director, that implies to me that all of this information somehow needs to make it into the director's header responses.

turetske:
@jhiemstrawisc That's a good point. I suppose it should really attempt a match on the basepath first and then the path. I think at least. This will be a larger fix, so I may have to punt this off to 7.4

turetske:
> My $0.02:
> 
> * We should match on most specific path first then base path.  This is because there are some oddball cases where base path (for issuing / verification) don't line up with the path itself.
> * For a given prefix, there should only be one `CredentialGeneration` block (see https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/GLOW.yaml#L109 versus https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/LIGO.yaml).  There may be multiple permitted issuers that can sign tokens but only one that we can redirect people to create new tokens.
>   
>   * In terms of headers, we should have the `x-pelican-authorization` be multi-valued.  How about `x-pelican-authorization: issuer=https://issuer1.example.com, issuer=https://issuer2.example.com`?  Would that be backward-compatible with existing clients?
> * We shouldn't be using the name `scitokens` in the wire format.  SciTokens is a token profile and, in many cases, these issuers use other token formats like the WLCG profile.  How about something more generic like `token_authorization`?

@bbockelm  - I'm somewhat confused by your specific comments about `CredentialGeneration` and `scitokens`. OSDF I think only has one `CredentialGeneration` per origin already and the `scitokens` that is currently being passed is because `scitokens` in the json field from OSDFTopology. I can adjust the name of the struct that is mirroring the topology so that it has a different name, but that `json:scitokens` bit has to stay.

turetske:
The below is our NamespaceAd struct - which is not serving our purposes.

```	
NamespaceAd struct {
		RequireToken   bool         `json:"requireToken"`
		Path           string       `json:"path"`
		Issuer         url.URL      `json:"url"`
		MaxScopeDepth  uint         `json:"maxScopeDepth"`
		Strategy       StrategyType `json:"strategy"`
		BasePath       string       `json:"basePath"`
		VaultServer    string       `json:"vaultServer"`
		DirlistHost    string       `json:"dirlisthost"`
		RestrictedPath []string     `json:"restrictedPath"`
}
```

The issue now is what we want to determine a single NamespaceAd. It should be one ad per namespace. We've determined that topology has multiple issuers and base paths per namespace. So we will need to adjust this NamespaceAd significantly to account for the multiple Issuers and BasePath combinations and will almost certainly have backwards compatibility concerns.

bbockelm:
Ok - I got around to reviewing the [Google document](https://docs.google.com/document/d/1Sp0886JJsv1SGJcjz9U0rdw4C7WhV6kkssllfhbzOPg/edit) that @turetske had created based on the discussion.  I think a lot of the "bones" of this PR can stay in place but I did some minor reworking of field names and other parts of the structure.

We'll still need to update the clients and the response headers from the director.  Those aren't covered in length in the document, just as an aside in the implementation notes.

PR #414, opened 2023-11-27
Add Oauth2 client with CILogon

### LINKED ISSUE #355: Authentication for the registry UI
As with the origin, the registry web UI will need some authentication / accounts setup so we can do things like have admins approve registrations.  In fact, it'll probably have three modes:
- Public (view registrations but can't mutate anything).
- OAuth2 (for remote users)
- User/password (for the admin)

The user/password setup should share the same routines as the origin UI (in fact, that ought to be refactored out -- it'll be needed for the cache as well!).

For OAuth2, I want to do this in-process instead of using any sort of oauth2 proxy to keep things standalone.  Here's a good example of how to use the go oauth2 library to do a code flow:

https://github.com/douglasmakey/oauth2-example

The only problem with that example is I don't like the use of a cookie the client can tamper with.  We should switch the cookie management to use https://github.com/gin-contrib/sessions with the "cookies" backend.  For the cookie backend, we'll need a shared secret for decryption; let's concatenate the byte array `pelican` with the DER form of the service's private key, take a hash, and use the hash's bytes as the secret.

This ticket is preparatory work for #264.

haoming29:
Close #355 

* Added a generic OAuth2 client to do third-party login for servers, with default provider being CILogon
* Enabled OAuth2 for `registry` server

## Testing Instructions

**PLEASE NOTE THE TEST INSTRUCTIONS ARE UPDATED AS OF 2023/12/08**

* Make sure you have `OIDC.ClientID` configured to your CILogon client ID
* Also ensure that you created a file under `/etc/pelican/oidc-client-secret` and copy-paste your  CILogon client secret there
* Also ensure that you either ask Cannon to register your redirect URL at CILogon as  `https://<registry-host>/api/v1.0/auth/cilogon/callback` or set your namespace registry web port to be `9999`
* Spin up `registry`: `./pelican registry serve -p 9999`
* Go to `https://<registry-host>/api/v1.0/auth/cilogon/login` and you should be redirected to CILogon for authentication
* After you authenticated, the page should be redirected back to `https://<registry-host>/view/`, which should be a server selection page and if you pick any server (say origin or director), you should be able to see their dashboard page without need to "login" 
* This auth flow supports `next_url` which can take you to where-ever you were before redirected to login. To test this function, go to `https://<registry-host>/api/v1.0/auth/cilogon/login?next_url=%2Fview%2Fconfig` and you should be eventually taken back to `/view/config/`


jhiemstrawisc:
@Haoming29, before I review, is there a difference between the parameters you added and the `OIDC` block of parameters that are already in `parameters.yaml`? I know we already have an implementation of CILogon/OAuth2 for origin registration when we do `pelican namespace register --with-identity <prefix>`, and I'm wondering if these need to be consolidated.

haoming29:
Hey @jhiemstrawisc I _think_ I've consolidated OAuth client logic to the OIDC stuff we already had so this PR should be good to go for review. No rush though, it's for 7.4 anyway

jhiemstrawisc (approved):

PR #415, opened 2023-11-28
Implement recursive uploads

joereuss12:
First draft of recursive uploads. This will go through a directory and upload every file to the destination. The only problem is that it will not maintain the same directory structure since I could not find a way to have PUT work with directories.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -22,12 +22,14 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/fs"
 	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


joereuss12:
@bbockelm Would you like me to look more into somehow keeping the directory structure on the server side? As of now, it only spits all of the files into the specified directory on the origin because I could not figure out how to do a client side PUT with directories.

joereuss12:
Marking this ready for review. Cannot create a unit test because it requires a director to be spinning so it can find the `namespace.DirListHost`. Therefore, I will just wait to create an end-to-end test for it when Emma completes her "federation in a box".

bbockelm:
@joereuss12 - there are some conflicts to fix up here.

turetske (approved):
Approved when merge conflicts are resolved

PR #416, opened 2023-11-28
Ensure TTL caches evict expired item

### LINKED ISSUE #330: Add unit tests to ttlcache to make sure cache eviction works
I'm not sure if we want the ttlcache to automatically evict expired cache entry but if we do, according to the documentation [here](https://github.com/jellydator/ttlcache#usage), we seem not to call `go cache.Start()` anywhere in our code and we don't seem to call `cache.DeleteExpired()` periodically which are the only two ways to make the eviction work. So I propose we add some unit tests to at least one instance of ttlcache we used to ensure the eviction work.

haoming29:
Closes #330 

The added unit tests to director's `serverAds` cache and `namespaceKeys` cache suggest that expired cache item won't be evicted with existing code, so this PR added `cache.Start()` to all TTL cache instances to ensure they work as expected.

The part I'm not sure about is whether we should call `cache.Stop()` at the shutdown of the program. I called `Stop()` for now but I don't know if this is redundant or not.

### DIFF
--- a/cmd/cache_serve.go
+++ b/cmd/cache_serve.go
@@ -149,8 +149,11 @@ func serveCache( /*cmd*/ *cobra.Command /*args*/, []string) error {
 	shutdownCtx, shutdownCancel := context.WithCancel(context.Background())
 	var wg sync.WaitGroup
 
-	defer config.CleanupTempResources()
-	defer shutdownCancel()
+	defer func() {
+		shutdownCancel()
+		wg.Wait()
+		config.CleanupTempResources()
+	}()
 
 	err := config.DiscoverFederation()
 	if err != nil {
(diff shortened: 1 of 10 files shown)


cmd/cache_serve.go
```diff
@@ -145,14 +146,22 @@ func advertiseCache(prefix string, nsAds []director.NamespaceAd) error {
 }
 
 func serveCache( /*cmd*/ *cobra.Command /*args*/, []string) error {
-	defer config.CleanupTempResources()
+	shutdownCtx, shutdownCancel := context.WithCancel(context.Background())
+	var wg sync.WaitGroup
```
jhiemstrawisc, line 150: Can you add a brief comment in one of the `*_serve.go` files that explains how/why you're using the wait group? I think it makes sense after looking up the `sync` package, but it would be nice to have at least one comment in the code that explains the logic.

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM!

PR #419, opened 2023-11-29
Issue 169

jhiemstrawisc:


xrootd/xrootd_config.go
```diff
@@ -136,72 +141,82 @@ func CheckXrootdEnv(origin bool, nsAds []director.NamespaceAd) error {
 
 	// If we use "volume mount" style options, configure the export directories.
 	if origin {
```
turetske, line 143: Just as a note, this is going to cause conflicts once my code is moved in. I recommend doing a rebase off of main once that happens earlier than later.

turetske:
This seems like the correct approach to me, but I'm assuming @bbockelm will have more comments.

jhiemstrawisc:
Tests are currently failing because the test CI doesn't pull in the updates I've made to the dev container we use to run tests. Those updates are reflected in PR #450. I will re-run once that's merged, built, and released.

cmd/origin.go
```diff
@@ -109,12 +109,63 @@ func initOrigin() error {
...
+	if err := viper.BindPFlag("Origin.S3AccessKeyfile", originServeCmd.Flags().Lookup("bucket-access-keyfile")); err != nil {
+		panic(err)
+	}
+	if err := viper.BindPFlag("Origin.S3SecretKeyfile", originServeCmd.Flags().Lookup("bucket-secret-keyfile")); err != nil {
+		panic(err)
+	}
+
+	// Would be nice to make these mutually exclusive to mode=posix instead of to --volume, but cobra
+	// doesn't seem to have something that can make the value of a flag exclusive to other flags
+	// Anyway, we never want to run the S3 flags with the -v flag.
```
turetske, line 153: Is there a check somewhere to ensure volume is set if mode is posix? I understand it can't be a cobra check, but it's something that should exist. Maybe it does and I just can't find it in the code.

turetske (changes requested):

xrootd/origin_test.go
```diff
@@ -134,4 +146,117 @@ func TestOrigin(t *testing.T) {
...
+	minIOServerCmd := exec.Command("minio", "server", "--quiet", tmpDir)
+	minIOServerCmd.Env = []string{fmt.Sprintf("HOME=%s", tmpDir)}
+
+	// Create a few buffers to grab outputs
+	var outb, errb bytes.Buffer
+	minIOServerCmd.Stdout = &outb
+	minIOServerCmd.Stderr = &errb
+
+	err := minIOServerCmd.Start()
+	// Wait for the server to initialize. Hopefully this always happens in under 2 seconds!
```
turetske, line 166: Can this do a small loop to test for server initialization like we do for the origin and xrootd?

utils/server_utils.go
```diff
@@ -58,3 +60,26 @@ func MakeRequest(url string, method string, data map[string]interface{}, headers
 
 	return body, nil
 }
+
```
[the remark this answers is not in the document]
turetske: Why resolved? It's not a general server utility function, it's an xrootd server function.

turetske (changes requested):
Small test changes and other questions.

PR #423, opened 2023-11-29
Pelican cache with auth

turetske:
Adds authentication to the pelican cache

-- Adds scitokens authenticated to the pelican cache
-- Fixes some errors that were in the namespace registration
-- Does a refactor to move common code out of origin specific folders/functions to be used by both the cache and origin

For ease of understanding, the first 6 commits (up through https://github.com/PelicanPlatform/pelican/pull/423/commits/b0606ce22f6f9ddd5648b94ea2886c965d12ad6d) are the actual cache with auth functionality and all the commits afterwards are the refactor.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -76,11 +76,20 @@ type (
 		JwksUri                       string `json:"jwks_uri"`
 	}
 
+	ServerType string
+
 	TokenOperation int
 
 	TokenGenerationOpts struct {
 		Operation TokenOperation
 	}
+)
+
+const (
+	CacheType    ServerType = "Cache"
+	OriginType   ServerType = "Origin"
+	DirectorType ServerType = "Director"
+	RegistryType ServerType = "Registry"
 )
 
 const (
@@ -388,9 +397,15 @@ func initConfigDir() error {
 	return nil
 }
 
-func InitServer() error {
+func InitServer(sType ServerType) error {
 	if err := initConfigDir(); err != nil {
 		return errors.Wrap(err, "Failed to initialize the server configuration")
+	}
+	xrootdPrefix := ""
+	if sType == OriginType {
+		xrootdPrefix = "origin"
+	} else if sType == CacheType {
+		xrootdPrefix = "cache"
 	}
 	configDir := viper.GetString("ConfigDir")
 	viper.SetConfigType("yaml")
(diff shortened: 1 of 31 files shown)


server_utils/server_struct.go
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ *
+ ***************************************************************/
+
+package server_utils
+
+import "github.com/pelicanplatform/pelican/director"
+
+type (
+	XRootDServer struct {
```
bbockelm, line 24: Convert this to an interface and have the implementations with composition, instead of a struct with a function pointer.

That is,

```
type XRootDServer interface {
  GetNamespaceAds func() []director.NamespaceAd
  CreateAdvertisement func(string, string, string) (director.OriginAdvertise, error)
}
```

config/config.go
```diff
@@ -76,13 +76,22 @@ type (
...
+
 	TokenOperation int
 
 	TokenGenerationOpts struct {
 		Operation TokenOperation
 	}
 )
 
+const (
+	CacheType    ServerType = "Cache"
```
bbockelm, line 89: Switch these to a bitmask.  We'll eventually want these to be a multi-selection (e.g., call `InitServer` for both origin and director).

xrootd/authorization.go
```diff
@@ -134,23 +134,30 @@ func EmitAuthfile(nsAds []director.NamespaceAd) error {
...
 	foundPublicLine := false
-	for sc.Scan() {
-		lineContents := sc.Text()
-		words := strings.Fields(lineContents)
-		if len(words) >= 2 && words[0] == "u" && words[1] == "*" {
-			output.Write([]byte("u * /.well-known lr " + strings.Join(words[2:], " ") + "\n"))
-			foundPublicLine = true
-		} else {
-			output.Write([]byte(lineContents + "\n"))
+	if nsAds == nil {
```
bbockelm, line 137: This could probably use some explaining.  It looks like the input auth file is ignored `nsAds` isn't null.  Why?

To support some legacy use cases, we'll need to merge in an authfile for the cache.

bbockelm (changes requested):
A few requested changes; see inline comments.

The only substantial request is switching from a struct with a function pointer to an interface.

PR #426, opened 2023-11-29
Update integration test file - update stashcp source

haoming29:
Just figured out the `./stashcp` source wasn't updated and the integration test keeps failing. I thought for stashcp it by default add `stash:///` to the url you provided but turned out you can directly feed `osdf:///` to use the osdf URL


turetske (approved):
LGTM

PR #428, opened 2023-11-30
Add supervisor script to make origins servable

jhiemstrawisc:
Now our production container can be started with "origin_serve" and the origin supervisord daemon will launch. To make the origin functional, add any origin arguments to the environment variable `OSDF_ORIGIN_ARGS`. For example, to serve an origin on port 8444 that mounts /foo to /namespace/bar in the ITB federation, you'd create the environment variable
```
OSDF_ORIGIN_ARGS="-f https://osg-htc.org/osdf/itb -v /foo:/namespace/bar -p 8444"
```

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -47,9 +47,10 @@ RUN useradd -o -u 10940 -g 10940 -s /bin/sh xrootd
 
 # Install dependencies
 RUN yum -y update \
-    && yum -y install xrootd xrootd-client xrdcl-http xrootd-server \
+    && yum -y install xrootd xrootd-client xrdcl-http xrootd-server xrootd-scitokens xrootd-voms \
+    xrootd-devel xrootd-client-devel xrootd-server-devel \
+    curl-devel git cmake3 gcc-c++ openssl-devel java-17-openjdk-headless \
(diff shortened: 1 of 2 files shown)


matyasselmeci:
I am not a big fan of having a single environment variable for all the arguments. Remember the Kustomize workflow: having a base that you then patch over. If you have individual environment variables for your options then the patch can append, replace, or remove options. If you just have the arguments then the patch has to replace all of them.

jhiemstrawisc:
I don't see that there's anything about this that prevents leaving the environment variable blank and setting up the entire configuration via individual environment variables or a regular pelican configuration file. For example, you could serve the exact same origin by leaving `OSDF_ORIGIN_ARGUMENTS` blank and adding the configuration file:
```
Federation:
  DiscoveryUrl: https://osg-htc.org/osdf/itb
  
Server:
  WebPort: 8443
  
Origin:
  ExportVolume: /foo
  NamespacePrefix: /namespace/bar
```

Or even by setting these each individually with an environment variable (eg `OSDF_ORIGIN_EXPORTVOLUME="/foo")

turetske (approved):
Tested.

LGTM!

Issue #417, opened 2023-11-28
Add director's public key to federation's discovery URL


haoming29:
Currently director has it's own discovery url for other servers to fetch the public JWK for token validation. We want to merge this discovery with federation's discovery URL so that federation can be a token issuer.

Based on the talk with @bbockelm, we need to wait until we can serve a federation through pelican, which in long term should be that the director can be a federation itself.

PR #429, opened 2023-11-30
Refactor `checkAPIToken` to be more generic and less verbose

### LINKED ISSUE #405: Revisit `checkAPIToken` and similar functions
(I think this one is important to get done before 7.3.0 is out!)

There's a few issues in `checkAPIToken` and friends I spotted when working on #404:
1.  We should give it a more precise name or make the issuer sources configurable.  There's very few cases where we accept any of the 4 different sources of authorization.  A precise name would be like `checkAnyAuthorization`; making it configurable would be providing a bitmask for the different sources of authorization as an input variable.
2.  It's ambiguous what scopes are allowed from what issuer.  There are three different scopes listed (`pelican.directorScrape`, `pelican.promMetric`, and `prometheus.read`) within `promMetricAuthHandler`; I'm pretty sure we shouldn't be accepting `pelican.directorScrape` from the origin's issuer.
3. Speaking of scopes, it's not immediately clear which scope does what.  For example, I don't think there's a distinction between `pelican.directorScrape` and `prometheus.read`: it's about the permission being granted, not who is doing the action.
   - If there's a reason to keep multiple scopes around (e.g., backward compatibility), then we should add a comment as to why those scopes are present.
4.  I've always assumed that the federation issuer is the director issuer: that is, there's no reason to have both checked for and the director would set its `iss` claim to be the federation URL.  Is there a strong reason why we'd need to separate them?
5. When we do check the tokens, we always look up the JWKS from the remote service (this would become a DoS in cases where someone pings the `/metrics` endpoint repeatedly with a fake token).  Instead, we should follow what's done [in the director](https://github.com/PelicanPlatform/pelican/blob/main/director/origin_api.go#L110-L186) and use the `jwk` library's `Cache` object to maintain the key.
   - As part of the consolidation of the key handling we need to do, perhaps we should centralize this logic to make it more reusable?

haoming29:
Closes #405 

* Refactored out `checkAPIToken` as `CheckAnyAuth` into `utils` package and migrate dependency functions to `utils` as well
* Rename Prometheus token scopes
* Use cache for JWKS

### DIFF
--- a/utils/server_auth.go
+++ b/utils/server_auth.go
@@ -20,17 +20,19 @@ package utils
 import (
 	"context"
 	"crypto/ecdsa"
+	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
 	"github.com/gin-gonic/gin"
+	"github.com/lestrrat-go/httprc"
 	"github.com/lestrrat-go/jwx/v2/jwa"
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/lestrrat-go/jwx/v2/jwt"
 	"github.com/pelicanplatform/pelican/config"
-	"github.com/pelicanplatform/pelican/director"
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -49,7 +51,11 @@ type (
 		DirectorCheck(ctx *gin.Context, token string, scopes []string) error
 		IssuerCheck(ctx *gin.Context, token string, scopes []string) error
 	}
-	AuthCheckImpl struct{}
+	AuthCheckImpl     struct{}
+	DiscoveryResponse struct { // This is a duplicate from director/authentication to ensure we don't have cyclic import
+		Issuer  string `json:"issuer"`
+		JwksUri string `json:"jwks_uri"`
+	}
 )
 
 const (
@@ -65,8 +71,10 @@ const (
 )
 
 var (
-	federationJWK *jwk.Cache
-	authChecker   AuthChecker
+	federationJWK    *jwk.Cache
+	directorJWK      *jwk.Cache
+	directorMetadata *httprc.Cache
+	authChecker      AuthChecker
 )
 
 func init() {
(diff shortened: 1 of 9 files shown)


haoming29:
@bbockelm Requested your review in case you want to peek at it and check if it fits your suggestions

utils/server_auth.go
```diff
@@ -0,0 +1,342 @@
...
+			return jwt.NewValidationError(errors.New("No scope is present; required for authorization"))
+		}
+		scope, ok := scope_any.(string)
+		if !ok {
+			return jwt.NewValidationError(errors.New("scope claim in token is not string-valued"))
+		}
+
+		for _, tokenScope := range strings.Split(scope, " ") {
+			// As long as there's one scope in the token that matches the pool of desriedScopes
+			// we say it's valid
```
jhiemstrawisc, line 107: I wonder about the semantics of this. Let's say I want to know if a token gives me permission to read and write some resource, and let's assume that the token does actually contain both scopes needed for those actions. To actually verify that using this validator, wouldn't I need to validate it twice, once for each scope?

Is there a drawback to requiring that every scope passed in the `anyScopes` is matched in the token?

utils/server_auth.go
```diff
@@ -0,0 +1,342 @@
...
+			if scopeContains(tokenScope, anyScopes) {
+				return nil
+			}
+		}
+		return jwt.NewValidationError(errors.New(fmt.Sprint("Token does not contain any of the scopes: ", anyScopes)))
+	})
+}
+
+// Checks that the given token was signed by the federation jwk and also checks that the token has the expected scope
```
jhiemstrawisc, line 116: Not a request for change, but for clarification -- what is the `federation jwk`, and where does it live? If it lives at the metadata discovery endpoint returned by `param.Federation_DiscoveryUrl`, which is currently a static server, how would one actually get a token signed by the private key? Would that have to be hand-created?

utils/server_auth.go
```diff
@@ -0,0 +1,342 @@
...
+		return errors.New(fmt.Sprint("Issuer is not a federation: ", token.Issuer()))
+	}
+
+	fedURIFile := param.Federation_JwkUrl.GetString()
+	ctx := context.Background()
+	if federationJWK == nil {
+		client := &http.Client{Transport: config.GetTransport()}
+		federationJWK = jwk.NewCache(ctx)
+		if err := federationJWK.Register(fedURIFile, jwk.WithRefreshInterval(15*time.Minute), jwk.WithHTTPClient(client)); err != nil {
+			return errors.Wrap(err, "Failed to register cache for federation's public JWKS")
```
jhiemstrawisc, line 137: Same deal as before, let's go with something like "Failed to create the federation's internal JWKs cache".

utils/server_auth.go
```diff
@@ -0,0 +1,342 @@
...
+
+	parsed, err := jwt.Parse([]byte(strToken), jwt.WithKey(jwa.ES256, raw.PublicKey))
+
+	if err != nil {
+		return errors.Wrap(err, "Failed to verify JWT by federation's key")
+	}
+
+	scopeValidator := createScopeValidator(anyOfTheScopes)
+	if err = jwt.Validate(parsed, jwt.WithValidator(scopeValidator)); err != nil {
+		return errors.Wrap(err, "Failed to verify the scope of the token")
```
jhiemstrawisc, line 163: For this spot and elsewhere that we print the same error message, is there any harm in including in the error message the specific scopes that failed?

utils/server_auth.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Invalid JWT")
+	}
+
+	serverURL := param.Server_ExternalWebUrl.GetString()
+	if serverURL != token.Issuer() {
+		if param.Origin_Url.GetString() == token.Issuer() {
+			return errors.New(fmt.Sprint("Wrong issuer; expect the issuer to be the server's web address but got Origin.URL, " + token.Issuer()))
+		} else {
+			return errors.New(fmt.Sprint("Issuer is not server itself: ", token.Issuer()))
```
jhiemstrawisc, line 183: This isn't something I know how to grapple with at the moment, but in general we don't require that the issuer is the same as the server. We do want to allow people to use their own issuers instead of the built-in issuer we provide. I don't know if there's a change for this, but just something to think about moving forward because eventually we'll have to come back to these sections.

utils/server_auth.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+	}
+
+	bKey, err := config.GetIssuerPrivateJWK()
```
jhiemstrawisc, line 187: Are we able to just load the public key and use it directly for validation instead of loading the private key into memory?

I see we have a `GetIssuerPublicJWKS`, which still loads the private key into memory, but returns the public JWKS. Maybe there's a way we can validate the tokens with the keyset?

utils/server_auth.go
```diff
@@ -0,0 +1,342 @@
...
+				return exists
+			}
+		case Director:
+			err := authChecker.DirectorCheck(ctx, token, authOption.Scopes)
+			if _, exists := ctx.Get("User"); err != nil || !exists {
+				errMsg += fmt.Sprintln("Director Check failed: ", err)
+				log.Debug("Director Check failed: ", err)
+				break
+			} else {
+				log.Debug("Director Check succeed")
```
jhiemstrawisc, line 320: So when I use this new code to serve a director, I see the log output `Director Check succeeded`. It's always good to see success, but I'm having trouble understanding what was being checked and what success implies. Can you elaborate a bit?

utils/server_auth_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	w := httptest.NewRecorder()
+	ctx, _ := gin.CreateTestContext(w)
+	ctx.Request = r
+
+	return ctx
+}
+
+func TestCheckAnyAuth(t *testing.T) {
```
jhiemstrawisc, line 56: Can you add a few more comments to the tests that explain what each test is checking? I'm not grokking what's going on here (most likely my fault, sorry!)

jhiemstrawisc (changes requested):

utils/server_auth.go
haoming29: I'd like to elaborate more but I'm not sure if the extra verbosity is what we want for the original ticket #405 where Brian suggests not to have too much verbose message, but I guess for the DEBUG level output it should be fine?

utils/server_auth.go
haoming29: Sure! Will do

utils/server_auth.go
haoming29: Will add comment as a hint for the future

utils/server_auth.go
haoming29: You are right, this is also the part that I don't have a lot of confidence when implementing it. The `anyToken` check was designed for the case where a single API can accept multiple valid scopes for the same resource as long as there is one exists in the token (so that we don't need to check each of the scopes). This is no longer the case with the scope refactoring we made as part of this PR so I guess it's less useful. What I will to is to add the additional parameter (like `checkAllScopes`) to the function and enable the feature you described, which was supposed to be done in the previous ticket but I just didn't get enough effort in making that happen (my bad)

utils/server_auth.go
haoming29: Yes! We don't have any token created by federation now and this function will soon be merged with DirectorCheck because director will soon become a federation itself in #417 

With this being said, It seems safe to remove `FederationCheck` from the code if it's never being used. However, I would ask for @turetske first to see if there's a reason to keep it as I vaguely remember Emma created this function originally as part of the token check process for Prometheus query engine endpoint.

utils/server_auth.go
haoming29: `createScopeValidator` will return the error message with scopes that failed and since this one wraps the `err` returned by `createScopeValidator`, I would assume it will carry the error from `createScopeValidator`

utils/server_auth.go
jhiemstrawisc: It doesn't even have to be at the level of logging output. I'm just curious what this does and how I should interpret it (even if we don't change any code for now).

utils/server_auth.go
turetske: I believe the idea was from Brian, basically if someone has a token signed by the federation key then they have access to the prometheus metrics. It would have to be hand created. I believe it was mostly there for the future.

jhiemstrawisc (approved):
LGTM

PR #430, opened 2023-11-30
Refactor director SD tests and add test for duplicated origins

### LINKED ISSUE #418: Ensure director service discovery endpoint does not return duplicated origins
The current test to  director service discovery endpoint didn't cover the case where multiple origin entries are present. We should improve this test coverage by adding some cases where multiple origin servers are present in serverAds cache with some intentional duplicate `Set()` calls.

haoming29:
Closes #418


turetske (approved):
LGTM!

PR #433, opened 2023-11-30
Fix bad error message from director

jhiemstrawisc:
When a client tries to `object copy` a file from a non-existant namespace, the client spits out an error of the form "Error from Director: Failure to unmarshal Director response: Unknown character N", or something like that. That's gobbeldygook, when a curl of the director clearly states "Namespace does not exist" or "No caches for that path" or something helpful. This makes the client check specifically for a 404, and in that case, it prints the body of the response, which _should_ have a helpful error message.


joereuss12 (approved):
LGTM

PR #442, opened 2023-12-01
Refactor `generate` package

### LINKED ISSUE #304: Reorganize generate package and go generator
As mentioned in #273 , we want to reorganize `generator` package to pull `main` function out of `param_generator.go` into a separate file, `main.go`. We can also move `go generate` command from `param` package back to `generator/main.go` so that things are more clear.

In the future, with any new generator functions, we should put them in different files under `generator` package and call them from `generator/main`.

haoming29:
Close #304 

One question for the reviewer is that do we _actually_  want to exclude the generator files for the release? I tried to use `//go:build ignore` in the `generate` package but that will make `go generate` malfunction with missing function import in between Go files.

### DIFF
--- a/generate/main.go
+++ b/generate/main.go
@@ -1,0 +1,13 @@
+package main
+
+//go:generate go run ../generate
+
+// This should not be included in any release of pelican
+
+// Include more generator functions here but keep them encapsulated
+// in their separate files under `generate` package
+func main() {
+	GenParamEnum()
+	GenParamStruct()
+	GenPlaceholderPathForNext()
+}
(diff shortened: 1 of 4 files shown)


turetske:
@haoming29 I don't think we want to exclude the generator files in the release, at least for now. I think the solution might be some fussing the with the GitHub ci that allows us to generate the generated files and then exclude the generator files solely in the release pipeline. I also think that's a separate issue than this review.

turetske (approved):
LGTM!

haoming29:
> @haoming29 I don't think we want to exclude the generator files in the release, at least for now. I think the solution might be some fussing the with the GitHub ci that allows us to generate the generated files and then exclude the generator files solely in the release pipeline. I also think that's a separate issue than this review.

Make sense to me. I will create another ticket for what we discussed here

PR #449, opened 2023-12-04
Web UI Move

CannonLock:
- Update website structure where the root contains leaf nodes of each pelican service
- Build new top level page to direct to the running services
- Move the origin ui to /origin sub dir


CannonLock:
I have poked around and triple checked everything worked as before. 

I plan to have it auto direct if only one service is being used once the active services api becomes available.

haoming29 (changes requested):
LGTM for most part. Some small changes that I think worth integrated into this PR:
* With changes introduced by this PR, the function `configureWebUI` is kind of confusing as it is actually configuring static assets of the web, so I proposed a rename to `configureWebResource`
* `ConfigureServerWebAPI` has the same issue and I would say we should rename it to `ConfigureServerWeb`

Some other observations that can be turned into additional tickets to fix in the future:
* Without the "active server" hint, currently the Web UI can be accessed by _any_ of server type (director/registry/origin etc). However, this introduced a problem (and will persist even when we have the active server hint) when the user switches between a server UI to another server UI in the same session (browser), where the login token issued by `server A` can't be validated by `server B`. This is because the token issuer claim is not what the server validating the token expects. We can fix this by removing port number of the issuer when issuing token for web login. I'll add a separate issue for this.
* Related to above but no need to solve in this PR. Currently when I access the web UI, it will show up the grid of servers UI I can select (although it's only "Origin" for now). This might not be what we want because this means a user can access _any_ server web UI from _any_ server URL. Say I have director runs at 8888 and registry at 9999 but I can see registry's UI in `8888/view` if I select "Registry" to go to. We might want to come up with a better way of directing traffic to correct server UI, potentially based on the config value of director/registry/origin URL

web_ui/ui.go
```diff
@@ -52,6 +61,26 @@ func getConfigValues(ctx *gin.Context) {
 	ctx.JSON(200, config)
 }
 
+func configureWebUI(engine *gin.Engine) error {
```
CannonLock, line 64:
```suggestion
func configureWebResource(engine *gin.Engine) error {
```

CannonLock:
@haoming29 I agree with renaming `configureWebUI` but I think we should keep the api function name. 

`ConfigureServerWeb` is to consuming and makes me think that it includes everything when really half the website is in those static resources.

haoming29 (approved):
LGTM. Thanks for the moving!

haoming29 (approved):

PR #454, opened 2023-12-05
Upgrade NextJS to ^14

CannonLock:
- Implementation of #323

### DIFF
--- a/web_ui/frontend/package-lock.json
+++ b/web_ui/frontend/package-lock.json
@@ -22,7 +22,7 @@
         "eslint": "8.45.0",
         "eslint-config-next": "13.4.12",
         "luxon": "^3.4.3",
-        "next": "^13.5.6",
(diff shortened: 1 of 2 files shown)


CannonLock:
@haoming29 I am a reviewer on another PR from the CI that I approve but since it is dependencies I have to update then rebuild to get it right. So this is the same PR as linked above and that I have approved just merged in.

CannonLock:
@haoming29 NPM handles all that. When I change next to v14 and run `npm i` all the v14 next dependencies are updated to the correct versions as well.

haoming29:
> @haoming29 NPM handles all that. When I change next to v14 and run `npm i` all the v14 next dependencies are updated to the correct versions as well. You can see these updates in the package-lock.json.

Great! Then this should be good to go.

haoming29 (approved):

Issue #265, opened 2023-10-23
Web UI for the director


bbockelm:
The director should have a web UI that contains:

- A view of all the attached origins and caches (structured list of caches, client-side filterable)
- The ability to specify an object in the federation and click "download" to download.
   - Based on the director response, this may generate an OAuth2 workflow.

haoming29:
Do we also want to show the health status of origins the director knows? At least it's API-wise ready by the time this is under the way.

CannonLock:
Going to cut this in two parts as it is not possible to download anything via web UI yet as the director does not have the appropriate CORS headers.

PR #458, opened 2023-12-05
Add `listServer` API to director Web UI

haoming29:
As part of the implementation to #265, this PR introduced a new API for director Web UI to get a list of advertised origins and caches: 

`/api/v1.0/director_ui/servers?server_type=origin` or `=cache` or without query parameter to get all servers advertised to the director.

Response:

```json
[
  {
    "name": "fae8c2865de4",
    "authUrl": "https://fae8c2865de4:8443",
    "url": "https://fae8c2865de4:8443",
    "webUrl": "https://fae8c2865de4:8444",
    "type": "Origin",
    "latitude": 0,
    "longitude": 0
  }
]
```


haoming29:
@CannonLock here's the API to for querying and get a list of origins and caches registered at the director (which is different from what's showing up in the namespace registry UI)

haoming29:
> @haoming29 Should this be added to the openapi spec?

I'm holding off adding new APIs to the spec as @bbockelm mentioned that we want to introduce the automated spec generation with annotations. I'm not sure what's the timeline for that though.

director/director_ui.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"net/http"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+)
+
+type (
+	listServerRequest struct {
+		ServerType string `form:"server_type,omitempty"` // "cache" or "origin"
```
turetske, line 12: What is this omit empty doing?

turetske (changes requested):

director/director_ui.go
haoming29: Thanks for pointing out. It should be removed it as this is a request struct not response

CannonLock:
@haoming29 Are the lat lon 0 because I don't have the Maxmind db?

haoming29:
> @haoming29 Are the lat lon 0 because I don't have the Maxmind db?

Unfortunately for Pelican Origins, we don't specify Lat/Long when we advertise to the director

CannonLock:
> > @haoming29 Are the lat lon 0 because I don't have the Maxmind db?
> 
> Unfortunately for Pelican Origins, we don't specify Lat/Long when we advertise to the director. I believe this field is only populated when we fetch origins from OSDF for now

Is it a new ticket to have unset values come in as null? null is a better proxy for a unset then 0.

turetske (approved):
LGTM!

PR #460, opened 2023-12-05
Enable key chaining to protect super/sub namespaces

jhiemstrawisc:
Another tool for admins, this PR allows a registry to be run such that it requires key chaining to register a sub/super namespace of an already-registered namespace. For example, if `/foo/bar` is registered, when key chaining is enabled, neither `/foo` nor `/foo/bar/baz` can be registered unless by a key paired with `/foo/bar`. Another example -- if `/foo` and `/foo/bar` are both registered, then we cannot register `/foo/bar/baz` unless the key is also registered to one of the superspace set (`/foo` and `/foo/bar`. 

I think eventually we should enable these types of registrations via tokens as well -- if I own `/foo` and I want to enable Emma to register `/foo/bar`, I should be able to generate a token with the key registered to `/foo` to enable any public key Emma wants to use. Based on feedback to this PR, I'll open that as an issue for future development.

### DIFF
--- a/namespace_registry/client_commands_test.go
+++ b/namespace_registry/client_commands_test.go
@@ -157,6 +157,10 @@ func TestRegistryKeyChaining(t *testing.T) {
 	err = NamespaceRegister(privKey, svr.URL+"/api/v1.0/registry", "", "/foo/bar")
 	require.NoError(t, err)
 
+	// Perform one test with a subspace and the same key -- should succeed
+	err = NamespaceRegister(privKey, svr.URL+"/api/v1.0/registry", "", "/foo/bar/test")
+	require.NoError(t, err)
+
 	// Now we create a new key and try to use it to register a super/sub space. These shouldn't succeed
 	viper.Set("IssuerKey", t.TempDir()+"/keychaining")
 	_, err = config.GetIssuerPublicJWKS()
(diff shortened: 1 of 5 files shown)


jhiemstrawisc:
Notes on testing:

1. Build and fire up a registry. Then launch an origin with a namespace like `/foo/bar` ensuring that the origin is pointed at your local registry. Double check that the namespace is saved in the sqlite database. Once confirmed, terminate the origin.
2. If running as root, delete `/etc/pelican/certificates/*` and `/etc/pelican/issuer.jwk`. Adjust accordingly for non-rootly execution.
3. Run a new origin, this time with a namespace that either prefixes or suffixes `/foo/bar`. For example, you could do `/foo` or `/foo/bar/baz`. Because you deleted the origin's registration key in the previous step, the origin will generate a new key and attempt to register it. Registration should fail at the origin with a periodic `ERROR` message that explains why registration fails. The registry should report a `WARNING` with similar information.
4. Now re-do this whole shebang, but add `Registry.RequireKeyChaining: false` to your pelican config. This should turn the feature off and allow registration under different keys.

namespace_registry/client_commands_test.go
```diff
@@ -129,3 +136,57 @@ func TestServeNamespaceRegistry(t *testing.T) {
 	})
 	viper.Reset()
 }
+
+func TestRegistryKeyChaining(t *testing.T) {
+	viper.Reset()
+	// On by default, but just to make things explicit
+	viper.Set("Registry.RequireKeyChaining", true)
```
turetske, line 143: If Registry.RequireKeyChaining is set to false in /etc/pelican/pelican.yaml, this test will use that rather than the viper.Set that occurs in on line. Somehow the pelican.yaml file is being read when it shouldn't be for this command.  Or being read to late so it's overwriting this line.

turetske (changes requested):

turetske (approved):
LGTM!

Issue #293, opened 2023-10-28
Allow re-registration


bbockelm:
If a given registration has already been performed, then it should not be an error when done again:

```
$ ./dist/pelican_darwin_arm64/pelican -f osg-htc.org namespace register --prefix /test-bbockelm
$ ./dist/pelican_darwin_arm64/pelican -f osg-htc.org namespace register --prefix /test-bbockelm
ERROR[2023-10-28T11:45:02-05:00] Failed to register prefix /test-bbockelm: Failed to make request: server encountered an error during key-sign challenge: The URL https://registry.osg-htc.org/api/v1.0/registry replied with status code 500 
```

If the user attempts to re-register with the same prefix and public key, we should respond with a known code that the client can handle and count as a success.

PR #461, opened 2023-12-06
Provide better message (with no error) when we register an already-registered namespace prefix

jhiemstrawisc:


jhiemstrawisc:
Closes issue #293

PR #462, opened 2023-12-06
Namespace registry admin

turetske:
This adds a requirement for admin approval before a cache can advertise to the director.

Currently there is no endpoint to approve caches, but the registry database can be altered via sqlite with the following commands after the cache has been created with `pelican cache serve`:

run sqlite

```
.open "/var/lib/pelican/registry.sqlite"
SELECT * FROM namespace;
```

There will be a list of origins and caches that looks like this:

```
id          prefix                pubkey             identity    admin_metadata         
----------  --------------------  -----------------  ----------  -----------------------
1           /caches/f1ec0afbfc09  <key_info>                     {"admin_approved":false}
```

Then run the following with `1` replaced by whatever id is in the id column of the cache you wish to approve
```
UPDATE namespace SET admin_metadata = '{"admin_approved":true}' WHERE id = 1;
```

### DIFF
--- a/director/origin_api.go
+++ b/director/origin_api.go
@@ -148,6 +148,9 @@ func VerifyAdvertiseToken(token, namespace string) (bool, error) {
 			return false, errors.Wrap(err, "failed to marshal the public keyset into JWKS JSON")
 		}
 		log.Debugln("Constructed JWKS from fetching jwks:", string(jsonbuf))
+		if jsonbuf == nil {
+			return false, errors.Wrap(err, "Cache does not have admin approval")
+		}
 	}
 
 	if err != nil {
--- a/director/origin_api.go
+++ b/director/origin_api.go
@@ -267,6 +267,6 @@ func GetRegistryIssuerURL(prefix string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	namespace_url.Path = path.Join(namespace_url.Path, "api", "v1.0", "registry", prefix, ".well-known", "issuer.jwks")
+	namespace_url.Path = path.Join(namespace_url.Path, "api", "v2.0", "registry", "metadata", prefix, ".well-known", "issuer.jwks")
 	return namespace_url.String(), nil
 }
--- a/director/origin_api.go
+++ b/director/origin_api.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/url"
-	"path"
 	"strings"
 	"sync"
 	"time"
@@ -270,6 +269,9 @@ func GetRegistryIssuerURL(prefix string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	namespace_url.Path = path.Join(namespace_url.Path, "api", "v2.0", "registry", "metadata", prefix, ".well-known", "issuer.jwks")
+	namespace_url.Path, err = url.JoinPath(namespace_url.Path, "api", "v2.0", "registry", "metadata", prefix, ".well-known", "issuer.jwks")
+	if err != nil {
+		return "", err
+	}
 	return namespace_url.String(), nil
 }
--- a/director/origin_api.go
+++ b/director/origin_api.go
@@ -58,6 +58,8 @@ type NamespaceCache interface {
 var (
 	namespaceKeys      = ttlcache.New[string, NamespaceCache](ttlcache.WithTTL[string, NamespaceCache](15 * time.Minute))
 	namespaceKeysMutex = sync.RWMutex{}
+
+	adminApprovalErr error
 )
 
 func CreateAdvertiseToken(namespace string) (string, error) {
@@ -149,7 +151,8 @@ func VerifyAdvertiseToken(token, namespace string) (bool, error) {
 		}
 		log.Debugln("Constructed JWKS from fetching jwks:", string(jsonbuf))
 		if jsonbuf == nil {
-			return false, errors.Wrap(err, "Cache does not have admin approval")
+			adminApprovalErr = errors.New(namespace + " has not been approved by an administrator.")
+			return false, adminApprovalErr
 		}
 	}
 
--- a/director/redirect.go
+++ b/director/redirect.go
@@ -402,8 +402,13 @@ func registerServeAd(ctx *gin.Context, sType ServerType) {
 		prefix := path.Join("caches", ad.Name)
 		ok, err := VerifyAdvertiseToken(token, prefix)
 		if err != nil {
-			log.Warningln("Failed to verify token:", err)
-			ctx.JSON(400, gin.H{"error": "Authorization token verification failed"})
+			if !ok {
+				log.Warningln("Failed to verify token:", err)
+				ctx.JSON(404, gin.H{"error": "Cache is not admin approved"})
+			} else {
+				log.Warningln("Failed to verify token:", err)
+				ctx.JSON(400, gin.H{"error": "Authorization token verification failed"})
+			}
 			return
 		}
 		if !ok {
--- a/director/redirect.go
+++ b/director/redirect.go
@@ -363,20 +363,20 @@ func ShortcutMiddleware(defaultResponse string) gin.HandlerFunc {
 func registerServeAd(ctx *gin.Context, sType ServerType) {
 	tokens, present := ctx.Request.Header["Authorization"]
 	if !present || len(tokens) == 0 {
-		ctx.JSON(401, gin.H{"error": "Bearer token not present in the 'Authorization' header"})
+		ctx.JSON(http.StatusUnauthorized, gin.H{"error": "Bearer token not present in the 'Authorization' header"})
 		return
 	}
 
 	err := versionCompatCheck(ctx)
 	if err != nil {
 		log.Debugf("A version incompatibility was encountered while registering %s and no response was served: %v", sType, err)
-		ctx.JSON(500, gin.H{"error": "Incompatible versions detected: " + fmt.Sprintf("%v", err)})
+		ctx.JSON(http.StatusInternalServerError, gin.H{"error": "Incompatible versions detected: " + fmt.Sprintf("%v", err)})
 		return
 	}
 
 	ad := OriginAdvertise{}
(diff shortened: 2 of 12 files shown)


namespace_registry/registry_db_test.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+func TestCacheAdminNoApproval(t *testing.T) {
+	viper.Reset()
+	registry_db_dir := t.TempDir()
+
+	viper.Set("Registry.DbLocation", registry_db_dir)
+
+	err := InitializeDB()
+	assert.NoError(t, err, "error initializing registry database")
```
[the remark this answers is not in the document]
turetske: So, I looked this up as it seems to be personal preference. There is a minor difference. Basically, `require` will cause a test to quit and exit immediately, which may be what we want more often, but with go routines, it can cause some issues.

To quote someone else:
"However, if the test has goroutines, we need to be careful that the test waits for goroutines' completion. Otherwise, the test errors get buried/missed; leaving you wondering why a test case failed!"

director/redirect.go
```diff
@@ -402,8 +402,13 @@ func registerServeAd(ctx *gin.Context, sType ServerType) {
 		prefix := path.Join("caches", ad.Name)
 		ok, err := VerifyAdvertiseToken(token, prefix)
 		if err != nil {
-			log.Warningln("Failed to verify token:", err)
-			ctx.JSON(400, gin.H{"error": "Authorization token verification failed"})
+			if !ok {
+				log.Warningln("Failed to verify token:", err)
+				ctx.JSON(404, gin.H{"error": "Cache is not admin approved"})
```
jhiemstrawisc, line 407: A few smallish notes:

There don't appear to be any paths in `VerifyAdvertiseToken` that return `true, err`, so all errors from the function will appear like an admin approval error. Instead of trying to parse combinations of `ok` and `err`, you can define a custom error type and check directly for an admin approval error via type assertion like you do elsewhere with `serverCredsErr`.

Also, I think we should be careful about the HTTP status code we reply with. Replying with a `404` here implies to the cache that the director resource it's attempting to `PUT` to couldn't be found. If we're denying its request to advertise because the token doesn't have the correct scopes or the cache isn't approved, we should reply with `403` -- permission denied.

While we're meddling with these sections of code, let's switch the `40X` to the appropriate `http.StatusXXX` methods. For `403`, this is `http.StatusForbidden`

director/redirect.go
```diff
@@ -402,8 +402,13 @@ func registerServeAd(ctx *gin.Context, sType ServerType) {
...
+			} else {
+				log.Warningln("Failed to verify token:", err)
+				ctx.JSON(400, gin.H{"error": "Authorization token verification failed"})
```
jhiemstrawisc, line 410: I don't think we need to change anything here, but we should make sure that the `400` (`http.StatusBadRequest`) response is only reachable in the event that the error is due to the advertise token itself being malformed. From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400), `400` is for bad client requests.

namespace_registry/registry.go
```diff
@@ -763,13 +802,22 @@ func getOpenIDConfiguration(c *gin.Context) {
 */
 
 func RegisterNamespaceRegistry(router *gin.RouterGroup) {
-	registry := router.Group("/api/v1.0/registry")
+	registry1 := router.Group("/api/v1.0/registry")
```
jhiemstrawisc, line 805: Can we make these a little more clear that we're versioning? Maybe something like `v1registry` and `v2registry`?

namespace_registry/registry_db.go
```diff
@@ -78,7 +80,7 @@ func createNamespaceTable() {
         prefix TEXT NOT NULL UNIQUE,
         pubkey TEXT NOT NULL,
         identity TEXT,
-        admin_metadata TEXT
+        admin_metadata TEXT CHECK (length("admin_metadata") <= 4000)
```
jhiemstrawisc, line 83: Is there some kind of logic behind restricting the length of what admins can set in the `admin_metadata` column to 4000? If so, could you add a comment?

namespace_registry/registry.go
```diff
@@ -722,8 +737,21 @@ func metadataHandler(ctx *gin.Context) {
 	if filepath.Base(path) == "issuer.jwks" {
 		// do something
 		prefix := strings.TrimSuffix(path, "/.well-known/issuer.jwks")
-		jwks, err := dbGetPrefixJwks(prefix)
+		var checkAdmin bool
+		if filepath.Base(prefix) == "getNamespaceKey" {
```
jhiemstrawisc, line 741: I think this still needs to be cleaned up to reflect the change in approach you and Brian settled on.

namespace_registry/registry_db.go
```diff
@@ -227,15 +229,35 @@ func getPrefixJwksById(id int) (jwk.Set, error) {
 	return set, nil
 }
 
-func dbGetPrefixJwks(prefix string) (*jwk.Set, error) {
-	jwksQuery := `SELECT pubkey FROM namespace WHERE prefix = ?`
+func dbGetPrefixJwks(prefix string, adminApproval bool) (*jwk.Set, error) {
```
jhiemstrawisc, line 232: Can you remove the adminApproval bool from the function signature since we always check if it's a cache anyway?

server_ui/advertise.go
```diff
@@ -132,6 +132,9 @@ func Advertise(server server_utils.XRootDServer) error {
 		if unmarshalErr := json.Unmarshal(body, &respErr); unmarshalErr != nil { // Error creating json
 			return errors.Wrapf(unmarshalErr, "Could not unmarshall the director's response, which responded %v from director registration: %v", resp.StatusCode, resp.Status)
 		}
+		if resp.StatusCode == 404 {
```
jhiemstrawisc, line 135: Should be changed to `http.StatusForbidden` to match the change server-side.

server_ui/register_namespace.go
```diff
@@ -58,12 +58,15 @@ func keyIsRegistered(privkey jwk.Key, url string) (keyStatus, error) {
...
 
-	req, err := http.NewRequest("GET", url+"/.well-known/issuer.jwks", nil)
+	req, err := http.NewRequest("GET", url, nil)
+
 	if err != nil {
 		return noKeyPresent, err
 	}
 
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("prefix", prefix)
```
jhiemstrawisc, line 68: I couldn't figure if/where we check this header. Why do we set it here? Technically I think the IETF has [deprecated](https://datatracker.ietf.org/doc/html/draft-saintandre-xdash-00) the recommendation of prefixing custom headers with `X-`, but we since every other custom Pelican header we use starts with `X-Pelican-`, we should be consistent.

jhiemstrawisc (changes requested):

namespace_registry/registry_db.go
turetske: Will do, but basically admin_metadata might not necessarily just mean data set by the admins. We might eventually add things such as "creator" or something else for example, which could be determined by the person who creates the namespace. This is to put a hard limit on what is written to disc to prevent someone maliciously potentially inserting something here of a huge length.

server_ui/register_namespace.go
turetske: This is checked in `namespace_registry/registry.go` in `dbGetNamespace`

director/origin_api.go
```diff
@@ -264,6 +270,6 @@ func GetRegistryIssuerURL(prefix string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	namespace_url.Path = path.Join(namespace_url.Path, "api", "v1.0", "registry", prefix, ".well-known", "issuer.jwks")
+	namespace_url.Path = path.Join(namespace_url.Path, "api", "v2.0", "registry", "metadata", prefix, ".well-known", "issuer.jwks")
```
jhiemstrawisc, line 273: I'm worried that on Windows this will use the wrong path separator, because `path.Join` is os-specific. Can you use the `url.JoinPath` method here?

director/redirect.go
```diff
@@ -363,20 +363,20 @@ func ShortcutMiddleware(defaultResponse string) gin.HandlerFunc {
 func registerServeAd(ctx *gin.Context, sType ServerType) {
 	tokens, present := ctx.Request.Header["Authorization"]
 	if !present || len(tokens) == 0 {
-		ctx.JSON(401, gin.H{"error": "Bearer token not present in the 'Authorization' header"})
+		ctx.JSON(http.StatusUnauthorized, gin.H{"error": "Bearer token not present in the 'Authorization' header"})
```
jhiemstrawisc, line 366: `401 -- unauthorized` is a bit of a misnomer in that it actually indicates unauthenticated (which is not something we check). The [HTTP specification RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-401-unauthorized) for `401` also states that if we want to return a `401`, we also need to start returning a `WWW-Authenticate` header with instructions for authentication. How about we just switch this to `http.StatusForbidden`.

director/redirect.go
```diff
@@ -402,29 +402,34 @@ func registerServeAd(ctx *gin.Context, sType ServerType) {
...
+				log.Warningln("Failed to verify token:", err)
+				ctx.JSON(http.StatusForbidden, gin.H{"error": "Authorization token verification failed"})
+			}
 			return
 		}
 		if !ok {
 			log.Warningf("%s %v advertised to namespace %v without valid registration\n",
 				sType, ad.Name, prefix)
-			ctx.JSON(400, gin.H{"error": sType + " not authorized to advertise to this namespace"})
+			ctx.JSON(http.StatusUnauthorized, gin.H{"error": sType + " not authorized to advertise to this namespace"})
```
jhiemstrawisc, line 417: `http.StatusForbidden`

jhiemstrawisc (changes requested):
Few small things and a rebase, and then I think we're good to go!

jhiemstrawisc (approved):

PR #466, opened 2023-12-06
Director UI

### LINKED ISSUE #265: Web UI for the director

CannonLock:
Closes #265

Screenshot 2023-12-06 at 12 44 47 PM (2)


CannonLock:
I will rebase on main once #458 is merged

haoming29:
@CannonLock Do we want to bring back the config page for the director? I'm pretty sure it's still going to be very useful to our director users to see their configs

web_ui/frontend/components/ServerTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+        },
+        "authUrl": {
+            name: "Auth URL",
+            cellNode: TableCellOverflowLink
+        },
+        "url": {
+            name: "URL",
+            cellNode: TableCellOverflowLink
+        }
+        // webUrl: "Web URL"  TODO: Uncomment when the Web URL is populated someday - Cannon Lock 2023/12/06
```
haoming29, line 105: It is populated if there is a Pelican origin advertised to the director and we should display this column. It is very confusing for now that the OSDF and Pelican origins each have different behaviors in advertisement.

haoming29 (changes requested):
LGTM for most part. One small change is to include WebURL in the column and the other is to include Config page in director's UI

CannonLock:
> @CannonLock Do we want to bring back the config page for the director? I'm pretty sure it's still going to be very useful to our director users to see their configs

Yeah good call, I will add it.

CannonLock:
@haoming29 I moved the config to the top level rather then create it twice.

haoming29:
@CannonLock  One last thing: it seems like we don't check login status for director's page, is this what we want? I happened to find this when I click the config page from the director and I was redirected to login

CannonLock:
@haoming29 I moved the auth redirects to server side. It is a bad user experience to have them on client side.

haoming29 (approved):
LGTM. We can fix the server router in a separate ticket.

haoming29:
> > LGTM. We can fix the server router in a separate ticket.
> 
> @haoming29 Fix? Is it doing something incorrectly?

Everything is good, I meant for the "matching and directing to index.html" thing that Go couldn't do a path rewrite. Not related to the auth change to server.

PR #467, opened 2023-12-06
Build containers but don't push on PR

CannonLock:


CannonLock:
@jhiemstrawisc If anything I think the test failure is proof that this does what it is supposed to ;)

haoming29:
> @jhiemstrawisc If anything I think the test failure is proof that this does what it is supposed to ;)

 Could you try to merge against the latest main? The error in release GH is fixed by #469 which has been merged in

haoming29 (approved):
LGTM. Thanks for making this change

PR #468, opened 2023-12-06
Adding dir listing capability to Pelican Origins

### LINKED ISSUE #465: Add directory listing capability to Pelican origins
Having issues doing recursive downloads because this is set to false and there is no way within Pelican to set this config. Add this option within Pelican within xrootd origin config

joereuss12:
fixes #465


turetske (changes requested):

turetske (approved):
LGTM assuming the tests pass

PR #469, opened 2023-12-06
Arm-compatible Pelican container and fix 404 minio RPM

### LINKED ISSUE #438: Pelican container not compatible with arm
Trying to build or work with the Pelican Docker container that comes pre-bundled with things like OA4MP and the scitokens-oauth2-server is very painful for people on ARM. In particular, trying to build the docker container on ARM immediately fails because of the line:
```
FROM hub.opensciencegrid.org/sciauth/scitokens-oauth2-server:release AS scitokens-oauth2-server
```

Do we have the ability to poke these upstream dependencies and have them built for arm as well?

haoming29:
Fixes #438 
Also fixed the current CI/CD release error with minio RPM missing by changing the package URL to archive with the pinned version.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -37,7 +37,7 @@ COPY --from=website-build /webapp/out ./web_ui/frontend/out
 
 RUN goreleaser --clean --snapshot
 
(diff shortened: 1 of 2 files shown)


turetske (approved):
LGTM

PR #470, opened 2023-12-06
Make the registry aware of topology

jhiemstrawisc:
I still need to sort out how this will play with the key chaining PR. Let's get that merged first.


turetske (changes requested):
Minor comment changes + the discussion we have about `parseTopologyServerAd`

turetske (approved):
LGTM assuming everything passes.

PR #471, opened 2023-12-06
Fix progress bar for recursive downloads

joereuss12:
Recursive download progress bar was panicing and exiting when we did not want it to. This patch is just a band-aid to keep functionality and to stop the panic, but they still do not look pretty when debug logging is enabled.


jhiemstrawisc (approved):
LGTM

PR #473, opened 2023-12-07
Fix progress bars w/ recursive uploads and downloads

joereuss12:
The progress bars now appear below the debug log as they are running and are not duplicated. Also, disabled progress bars with specified debug log output. Figured you don't need the bars to be printed to a specific log file.


jhiemstrawisc (approved):
Tested, and I've never seen such awesome progress bars for all my recursive downloads 😉

Even though I think we're safe to merge while the NPM build step is failing, I'm still going to wait for the tests to pass before I merge.

haoming29:
@joereuss12 and @jhiemstrawisc Just so you know, Cannon's PR to fix the npm build issue is merged in and we should be good to go

PR #475, opened 2023-12-08
Lock NextJS to 14.0.3

CannonLock:
14.0.4 breaks the website build so lock it to 14.0.3


haoming29 (approved):
LGTM! Thanks for the fix

Issue #335, opened 2023-11-06
Refactor commonly used util code into util package


turetske:
Refactor general utility functions out of specific directories into pelican/util. This is similar to the web_ui refactoring issue, but broader.

There is a bunch of code that shouldn't be specific to certain packages - for example `MakeRequest`, which has since been moved from the namespace-registry package into the utils package.

A read through of the code and a refactor of obvious functions should be done to move other general functions out of specific packages.

jhiemstrawisc:
Linking issue #320 and issue #159 to this.

turetske:
> Linking issue #320 and issue #159 to this.

@jhiemstrawisc I understand #320, but I don't quite understand why #159 gets linked in?

haoming29:
Link issue #447 and #405 here and PR #429 here that closes #405 
We are making progress!

jhiemstrawisc:
More or less this has been accomplished through lots of other PRs. Closing as done.

PR #483, opened 2023-12-11
Refactor and move token create logic to utils

haoming29:
Instead of passing a `claimMap` to create a token, now you can config your token by constructing `TokenConfig` and call `CreateToken` off this instance.
* Replaced `CreateEncodedToken` by `CreateToken`
* Ran regression tests and ensure existing code are not affected


haoming29:
Another progress towards #335 and #159

jhiemstrawisc (approved):
Thanks for this refactor, I think we'll be able to be a lot more consistent with how to create tokens internally thanks to it!

PR #486, opened 2023-12-11
Add location prefix to Prometheus data store

### LINKED ISSUE #476: Add prefix to the data location of Prometheus instance based on `ServerType`
Right now, if you run Pelican server instances in a single container, you will only be able to run one Prometheus instance at a time (depends on which server you spin up first, excluding cache which right now has a separate data location config parameter) and we want to be able to see all of them running.

To fix it, we want to add a prefix to Prometheus data location path based on which the `ServerType` config value which tells us which server instance is running for a specific process. We might also want to refactor cache's data location parameter to reuse the same logic.

haoming29:
Closes #476 

Also fixed an uncaught config value not updated error in `/api/config`


bbockelm:
I'm not entirely sure I want to go this direction:

1.  With the new "federation in a box" functionality, we should be able to have a single Pelican instance inside a test container for all the different server types, no?
      * IIRC, the cache component is not done yet but that's just a matter of "not done yet".
3. This breaks backward compatibility with existing data directories.

@haoming29 - this work was done prior to the new "federation in a box" mode.  Do you still think the PR is relevant?

haoming29:
> I'm not entirely sure I want to go this direction:
> 
> 1. With the new "federation in a box" functionality, we should be able to have a single Pelican instance inside a test container for all the different server types, no?
>    
>    * IIRC, the cache component is not done yet but that's just a matter of "not done yet".
> 2. This breaks backward compatibility with existing data directories.
> 
> @haoming29 - this work was done prior to the new "federation in a box" mode. Do you still think the PR is relevant?

For one. The "federation in a box" does not have any web component enabled as of now (including Prometheus) and I believe during local testing, Emma found bigger issue when enabling them, where multiple Prometheus instances are causing themselves panic. I'm not sure how Emma want to proceed with this.

For two. Yes, this will unfortunately break the backward compatibility. I think a better way to handle this is to check and migrate existing data folder to our new prefixed ones, but the issue is we don't really know which server was generating the data to the old folder. Any other ideas are welcomed.

jhiemstrawisc:
@turetske @haoming29, has there been any movement on this? I'm wondering if I should approve and merge, or if we decided to take a different approach. It would be nice to squeeze this into 7.4, since this causes a headache any time we need to run a registry and a director side by side.

haoming29:
> @turetske @haoming29, has there been any movement on this? I'm wondering if I should approve and merge, or if we decided to take a different approach. It would be nice to squeeze this into 7.4, since this causes a headache any time we need to run a registry and a director side by side.

@bbockelm and I had a conversation last week about backward compatibility issue (see above). I didn't find the best way to handle the data location migration for the existing system and I'm not sure if Brian wants to merge this in with such issue. The data files before this change won't get lost, just that we reroute new files to a different location and Prometheus won't read the old data. So it won't be the end of the day, but it also won't be a smooth/silent upgrade.

haoming29:
@jhiemstrawisc Also, this is tied to how "federation in a box" is implemented and my assumption when replying Brian above was make prior to new changes in #570. I'm not sure if I need to make major changes with the update in #570. I'll take a look at that PR and figure out the way to proceed.

jhiemstrawisc:
I'll hold off on a review for now then. Do you think we can get this in 7.4 or should we push back to 7.5?

haoming29:
> I'll hold off on a review for now then. Do you think we can get this in 7.4 or should we push back to 7.5?

7.4 is still reasonable. I will confirm with Brian.

haoming29:
After the conversation, I think we want to push this off to 7.5 and revisit how we want to implement this with #570 

As major changes are expected, I will close this PR for now and re-open new one later.

PR #491, opened 2023-12-12
Exclude node_modules from calculating Next.js hash in GHA

### LINKED ISSUE #488: GHA macos runner always return `next.js` cache error
In the last several GHA runs, macos runner almost always returned with an error message `Error: The template is not valid. .github/workflows/test.yml (Line: 25, Col: 14): hashFiles('**/*.js, **/*.jsx, **/*.ts, **/*.tsx') couldn't finish within 120 seconds.`. Usually a re-run will fix the issue. However, it's annoying to have code-unrelated test failure, so that's fix this issue

haoming29:
Fixes #488 

Basically fix the `hashFiles` timeout error after 120s when doing hashing for Next.js runner cache by excluding `node_modules` from calculating the hash.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
           ~/.npm
           ${{ github.workspace }}/.next/cache
         # Generate a new cache whenever packages or source files change.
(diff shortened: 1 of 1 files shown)


CannonLock (approved):
LGTM

PR #493, opened 2023-12-12
Include cache in director's service discovery endpoint

### LINKED ISSUE #481: Update director service discovery endpoint to include caches
Currently we only scrape `/metrics` endpoint from the director to origins, but with caches integrated to the system, we want also to include them in director's monitoring. So let's add caches in service discovery endpoint and give them different labels to distinguish.

haoming29:
Closes #481 

## Testing Instructions
* Spin up your `director` `registry` `origin` and `cache`
* Go to your director's Prometheus query engine endpoint for discovered servers, i.e. `/api/v1.0/prometheus/query?query=up`
* You should be able to see your cache and origin in the list
* Go to an arbitrary metric and both origin and cache shared, say `xrootd_monitoring_packets_received`, i.e. `/api/v1.0/prometheus/query?query=xrootd_monitoring_packets_received`, you should be able to see the metric from both your origin and cache, distinguished by their label.

### DIFF
--- a/director/redirect.go
+++ b/director/redirect.go
@@ -452,9 +452,9 @@ func registerServeAd(ctx *gin.Context, sType ServerType) {
 	ctx.JSON(200, gin.H{"msg": "Successful registration"})
 }
 
-// Return a list of available origins URL in Prometheus HTTP SD format
+// Return a list of registered origins and caches in Prometheus HTTP SD format
 // for director's Prometheus service discovery
-func DiscoverOrigins(ctx *gin.Context) {
+func DiscoverOriginCache(ctx *gin.Context) {
 	// Check token for authorization
 	tokens, present := ctx.Request.Header["Authorization"]
 	if !present || len(tokens) == 0 {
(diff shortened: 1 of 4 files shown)


director/redirect.go
```diff
@@ -479,24 +479,21 @@ func DiscoverOrigins(ctx *gin.Context) {
 	serverAds := serverAds.Keys()
 	promDiscoveryRes := make([]PromDiscoveryItem, 0)
 	for _, ad := range serverAds {
-		// We don't include caches in this discovery for right now
-		if ad.Type != OriginType {
-			continue
-		}
 		if ad.WebURL.String() == "" {
-			// Oririgns fetched from topology can't be scraped as they
+			// Oririgns and caches fetched from topology can't be scraped as they
```
turetske, line 483: Type - should be "Origins" instead of "Oririgns"

director/redirect.go
```diff
@@ -522,7 +519,7 @@ func RegisterDirector(router *gin.RouterGroup) {
 	router.GET("/api/v1.0/director/object/*any", RedirectToCache)
 	router.GET("/api/v1.0/director/origin/*any", RedirectToOrigin)
 	router.POST("/api/v1.0/director/registerOrigin", RegisterOrigin)
-	router.GET("/api/v1.0/director/discoverOrigins", DiscoverOrigins)
+	router.GET("/api/v1.0/director/discoverOrigins", DiscoverOriginCache)
```
turetske, line 522: Maybe leave a comments for if we ever make a v2.0 version of the api to change it to `/director/discoverOriginCache`

turetske (changes requested):
Very minor changes

turetske (approved):
LGTM!

PR #494, opened 2023-12-13
Add topology component to Prometheus health metric

### LINKED ISSUE #482: Add # of errors from topology to Prometheus metrics
For director and namespace registry, add a new Prometheus counter metric `pelican_error_topology` that counts the number of errors director/registry have when talking to topology. Add label "server": "director" or "registry" to distinguish (although this metrics is independent to the server)

haoming29:
Closes #482 

Instead of adding # of errors as a separate Prometheus metric, I took advantage of existing component health metric by adding topology as a new component. If we want to look at number of errors (Status.Critical) over time, we should use PromQL to get what we want. Something like `query=count_over_time((pelican_component_health_status{component="topology"}%20==%201)[1h:5m])` which will give you total number of critical status over the period of time.

### DIFF
--- a/cmd/director_serve.go
+++ b/cmd/director_serve.go
@@ -51,9 +51,8 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 	log.Info("Initializing Director GeoIP database...")
 	director.InitializeDB()
 
-	metrics.SetComponentHealthStatus(metrics.DirectorRegistry_Topology, metrics.StatusWarning, "Start requesting from topology, status unknown")
-
 	if config.GetPreferredPrefix() == "OSDF" {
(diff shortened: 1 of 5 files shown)


cmd/director_serve.go
```diff
@@ -50,6 +51,8 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 	log.Info("Initializing Director GeoIP database...")
 	director.InitializeDB()
 
+	metrics.SetComponentHealthStatus(metrics.DirectorRegistry_Topology, metrics.StatusWarning, "Start requesting from topology, status unknown")
```
turetske, line 54: I think this should be behind the config.GetPreferredPrefix()==OSDF tag as we only care about Topology when running in OSDF mode.

turetske (changes requested):

cmd/director_serve.go
haoming29: Make sense to me.

turetske (approved):
LGTM!

PR #496, opened 2023-12-13
Allow GeoIP overrides and fix server geoip randomization

jhiemstrawisc:
First and foremost, this allows a director to configure IP blocks that we want to directly map to a lat/long pair. Valid IPs include IPv4 and IPv6, both masked and unmasked. In the event that MaxMind either doesn't contain a coordinate for an IP or we just don't like the coordinate it returns, we can now set where we want a director to think the server is. Note that this affects both servers (origins/caches) as well as clients, because we update server coordinates at ad recording time, and we get client coordinates at redirect time.

One small change is that we now log at the Director when an IP address results in the null coordinate. This should allow us to filter such requests in the logs when we're collecting metrics.

This also fixes another small bug that's been annoying me -- we don't treat the null address as a "bad" coordinate, which means that if MaxMind can't resolve a lat/long for a particular incoming client request, we send them to null island. Instead, the change in this commit that checks for lat/long == (0,0) in SortServers will cause a client IP resolving to the null coordinate to be redirected to a randomly-ordered list of caches.

One thing to note is that I don't test some of the functions I modify if those functions require access to a MaxMind database. That's definitely something we want to figure out at some point, and while we could create an API key for caching the database in our GH runners, we'd also want to set up the tests to skip if no database/key are present. That seems like it's outside the scope of what I'm trying to do here.

### DIFF
--- a/director/sort_test.go
+++ b/director/sort_test.go
@@ -1,0 +1,149 @@
+package director
+
+import (
+	"bytes"
+	"net"
+	"strings"
+	"testing"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckOverrides(t *testing.T) {
+	viper.Reset()
+
+	// We'll also check that our logging feature responsibly reports
+	// what Pelican is telling the user.
+	logOutput := &(bytes.Buffer{})
+	log.SetOutput(logOutput)
+	log.SetLevel(log.DebugLevel)
+
+	// Mockup of the YAML we need to kick off the tests. Note that if you mess with it, you need to
+	// make sure the indentations are spaces.
+	yamlMockup := `
+GeoIPOverrides:
+  # Valid IPv4
+  - IP: "192.168.0.1"
+    Coordinate:
+      Lat: 123.4
+      Long: 987.6
+  # Valid IPv4 CIDR
+  - IP: "10.0.0.0/24"
+    Coordinate:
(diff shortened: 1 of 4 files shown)


director/sort_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckOverrides(t *testing.T) {
+	viper.Reset()
+
+	// We'll also check that our logging feature responsibly reports
+	// what Pelican is telling the user.
+	logOutput := &(bytes.Buffer{})
```
haoming29, line 19: Like this approach!

director/sort_test.go
```diff
@@ -0,0 +1,149 @@
...
+	log.SetOutput(logOutput)
+	log.SetLevel(log.DebugLevel)
+
+	// Mockup of the YAML we need to kick off the tests. Note that if you mess with it, you need to
+	// make sure the indentations are spaces.
+	yamlMockup := `
```
haoming29, line 25: It might be more tidy if we put this under `director/resources/*.yaml` and use `go:embed` to embed this file and read it directly. Up to you to decide.

haoming29 (approved):
LGTM! One note on the test file but it's non-blocking. Thanks for fixing this!

bbockelm:
@jhiemstrawisc - I think this is ready to go once you address the minor refactor / use of `go:embed` suggested by Haoming.

PR #497, opened 2023-12-13
Fix Origin UI RateGraph data loss/issues

### LINKED ISSUE #487: Prometheus metric `xrootd_server_bytes` not updated in Origin
In current main, if you spin up origin (without director and registry, so that your Origin's prometheus query endpoint can actually run) and go to web UI, the dashboard graph will show "No data returned by database as of 4:08:59 PM; Plot will auto-refresh". 

Go to the URL the graph requests: `/api/v1.0/prometheus/query?query=rate%28xrootd_server_bytes`, it will return empty array. Go to Prometheus metric discovery endpoint:`/api/v1.0/prometheus/label/__name__/values`, it does not include `xrootd_server_bytes`, which suggests this metrics does not exist.

Not sure what's the cause on top of my head, but we might break some xrootd config files in the last several major changes.

haoming29:
Fix #487 

It turned out that issue #487 is expected as we set the default data resolution to 3h, and at the server fresh start (no history metric data), Prometheus doesn't have any data at `now() - 3h`. However, this should be adjust so that the default resolution can always give _some_ result to the user at the server start.

This PR fixes anther issue where the "reset" button sets `time` differently than the default value, causing inconsistent result (although in the later commits it was consolidated to update time accordingly, see details in next point)

This PR also fixes a bug where the RateGraph visualizes data from `t= now()` to `t= now() + range (by default 3h)` which means it "generates" fake data from now to 3 hours in the future and we don't want that. Instead, now the data will be from `t=now() - range` to `t=now()`, which is the **history data from last `<range>` until now**.

This PR finally addressed the issue where we use `xrootd_server_bytes{direction="rx"}` metric for both `rx` and `tx` data. Now the two data series will give different values (as reported by Prometheus)


bbockelm:
@CannonLock - can you take a look at this?

web_ui/frontend/app/origin/page.tsx
```diff
@@ -39,10 +39,10 @@ export default function Home() {
                     <Box sx={{backgroundColor: "#F6F6F6", borderRadius: "1rem"}} p={2}>
                         <Box minHeight={"200px"}>
                             <RateGraph
-                                rate={TimeDuration.fromString("3h")}
-                                duration={TimeDuration.fromString("7d")}
-                                resolution={TimeDuration.fromString("3h")}
-                                metric={['xrootd_server_bytes{direction="rx"}', 'xrootd_server_bytes{direction="rx"}']}
```
CannonLock, line 45: Lets keep these defaults. I realize they aren't the best when you first enter but for a origin that is actually in use somewhere I think they are more helpful then just todays numbers. 

If you disagree, we should at least kick up the resolution a little bit, maybe to "20m"? 1440 data points on the graph is going to get busy.

CannonLock (changes requested):
Looking for your input on one thing before I approve.

web_ui/frontend/app/origin/page.tsx
haoming29: I agree we should think about users who run origin for days. How about I add a note to the graph to tell the user if they just started the origin and they can't see incoming data, they should modify the resolution?

haoming29: Actually, I can't run origin UI with latest main and I got the white screen with client side error that leads to `Objects are not valid as a React child (found: object with keys {Type, Value}). If you meant to render a collection of children, use an array instead.` I've rolled back the values to default but I won't be able to add the note without seeing the UI. We can talk about more in detail tomorrow.

haoming29:
Given the limited time we have for 7.4, I will back port this in 7.4.x

haoming29:
@CannonLock I added some instructions in the "no data" error message for how to adjust settings to get data out for fresh servers and I tested everything looks good.

haoming29:
> @haoming29 When I run the origin I am getting `{"error":"Correct authorization required to access Prometheus query engine APIs"}`

Did you login when you hit this? Where did this come from?

CannonLock (approved):
Bug I saw was related to the director being down? LGTM.

PR #500, opened 2023-12-14
Pelican authfile fix

turetske:
Two fixes, one which fixes the xrootd run file locations and one which moves the topology PeriodicReload to be behind the OSDF flag.

### DIFF
--- a/cmd/director_serve.go
+++ b/cmd/director_serve.go
@@ -58,8 +58,8 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 		if err := director.AdvertiseOSDF(); err != nil {
 			panic(err)
 		}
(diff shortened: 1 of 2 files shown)


cmd/director_serve.go
```diff
@@ -58,8 +58,8 @@ func serveDirector( /*cmd*/ *cobra.Command /*args*/, []string) error {
 		if err := director.AdvertiseOSDF(); err != nil {
 			panic(err)
 		}
+		go director.PeriodicCacheReload()
```
haoming29, line 61: lol Looking at this code yesterday I was wondering why we wanted to talk to topology if we are not running in `osdf` mode. This explains.

haoming29 (approved):
LGTM!

Issue #172, opened 2023-10-02
"pelican serve" mode


bbockelm:
Add a hidden "pelican serve" option that starts Pelican with a list of enabled modules (current possibilities: origin, registry, director).

The goal of "pelican serve" mode will be to enable better end-to-end and integration testing, providing a "federation in a box".

bbockelm:
Fixed as part of #570.

PR #503, opened 2023-12-14
Refactor `config.ServerType` to provide a globally accessible source for server(s) enabled

### LINKED ISSUE #501: Refactor `config.ServerType` to a bit map and set the map at the server start
In #280 We wanted to have ServerType as a bit map to indicate which server(s) is running. This variable should be set at server start (for only once) and it's globally accessible afterwards.

We indeed had a `config.ServerType` in our codebase right now but it's not quite what we wanted originally and web UI will need it to redirect user to the correct landing page of server UI.

Let's refactor the existing `config.ServerType`: https://github.com/PelicanPlatform/pelican/issues/280#issuecomment-1784181744

> Agreed that this is a bit of a mess. However, I'd like to preserve the future ability to have multiple daemon types all running inside a single process.
> 
> So, instead of a single value, how about a type like this:
> 
> type ServerComponent int
> 
> func (component ServerComponent) Set(ServerComponent newComp) {
>     component |= newComp
> }
> 
> func (component ServerComponent) IsEnabled(ServerComponent testComp) bool {
>     return (component & testComp) == testComp
> }
> 
> const (
>    ComponentOrigin ServerComponent = 1
>    ComponentDirector ServerComponent = 2
>    ComponentRegistry ServerComponent = 4
> )
> (typing this outside a golang compiler so I'm not checking correctness -- hopefully the pseudocode above makes sense)
> 
> With this, we could refactor the different components to have a single startup function. The implementation of pelican origin serve would simply be setting the ComponentOrigin flag inside the config then calling the common startup function. This way, one could enable other components through configuration to implement https://github.com/PelicanPlatform/pelican/issues/172.

haoming29:
Closes #501 

Now you can use `config.IsServerEnabled` anywhere in the server code to test if a server is enabled. A private function `setEnabledServer` is called in `initServer()` to indicate which server(s) are enabled.

This would be a necessary step for UI to redirect server web UI user to correct server UI landing page.

This can also be used in #172

When browsing through places where we used ServerType, it seems that we had a `XrootDServer` interface that relies on implementation to tell which server is running. I didn't touch that part of the code in this refactoring but if we want I can remove `GetServerType()` from `XrootDServer` interface and replace existing ones to `config.IsServerEnabled`. I'm fine either way. @turetske

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -76,14 +76,14 @@ type (
 		JwksUri                       string `json:"jwks_uri"`
 	}
 
-	ServerType int
-
 	TokenOperation int
 
 	TokenGenerationOpts struct {
 		Operation TokenOperation
 	}
 )
+
+type ServerType int // ServerType is a bit mask indicating which Pelican server(s) are running in the current process
 
 const (
 	CacheType ServerType = 1 << iota
@@ -116,10 +116,43 @@ var (
 	// Our global transports that only will get reconfigured if needed
 	transport     *http.Transport
 	onceTransport sync.Once
+
+	// A global variable to be set and access to obtain enabled Pelican servers in the current process
+	enabledServers ServerType
+	setServerOnce  sync.Once
 )
 
-func (sType ServerType) IsSet(otherVal ServerType) bool {
-	return sType&otherVal == otherVal
+// Set sets a list of newServers to ServerType instance
+//
+// DON'T call this function on EnabledServers but call setEnabledServer instead
+func (sType *ServerType) Set(newServers []ServerType) {
+	for _, server := range newServers {
+		*sType |= server
+	}
+}
+
+// IsEnabled checks if a testServer is in the ServerType instance
+func (sType ServerType) IsEnabled(testServer ServerType) bool {
+	return sType&testServer == testServer
+}
+
+// setEnabledServer sets the global variable config.EnabledServers to include newServers.
+// Since this function should only be called in config package, we mark it "private" to avoid
+// reset value in other pacakge
+//
+// This will only be called once in a single process
+func setEnabledServer(newServers []ServerType) {
+	setServerOnce.Do(func() {
+		// For each process, we only want to set enabled servers once
+		enabledServers.Set(newServers)
+	})
+}
+
+// IsServerEnabled checks if testServer is enabled in the current process.
+//
+// Use this function to check which server(s) are running in the current process.
+func IsServerEnabled(testServer ServerType) bool {
+	return enabledServers.IsEnabled(testServer)
 }
 
 func (sType ServerType) String() string {
@@ -323,7 +356,7 @@ func setupTransport() {
 	}
 }
 
-func parseServerIssuerURL(sType ServerType) error {
+func parseServerIssuerURL() error {
 	if param.Server_IssuerUrl.GetString() != "" {
 		_, err := url.Parse(param.Server_IssuerUrl.GetString())
 		if err != nil {
@@ -345,7 +378,7 @@ func parseServerIssuerURL(sType ServerType) error {
 		return errors.New("If Server.IssuerHostname is configured, you must provide a valid port")
 	}
 
-	if sType == OriginType {
+	if enabledServers.IsEnabled(OriginType) {
 		// If Origin.Mode is set to anything that isn't "posix" or "", assume we're running a plugin and
 		// that the origin's issuer URL actually uses the same port as OriginUI instead of XRootD. This is
 		// because under that condition, keys are being served by the Pelican process instead of by XRootD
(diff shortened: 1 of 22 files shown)


turetske:
> When browsing through places where we used ServerType, it seems that we had a `XrootDServer` interface that relies on implementation to tell which server is running. I didn't touch that part of the code in this refactoring but if we want I can remove `GetServerType()` from `XrootDServer` interface and replace existing ones to `config.IsServerEnabled`. I'm fine either way. @turetske

@haoming29 I'm hesitant to do this, because in this case it specifically needs to know if the server is the one of that "type", not whether the server is running. Because the "pelican serve" code can run multiple servers, we need to distinguish between "is <x> server type running" vs "is <x> the server type that is calling this function".

config/config.go
```diff
@@ -345,7 +376,7 @@ func parseServerIssuerURL(sType ServerType) error {
 		return errors.New("If Server.IssuerHostname is configured, you must provide a valid port")
 	}
 
-	if sType == OriginType {
+	if IsServerEnabled(OriginType) {
```
turetske, line 379: What occurs if multiple servers are running from the same process? If I start up an origin and then a cache -- will this still show true when I start the cache? That's a problem since we don't want the cache to execute this code block.

I suggest going back to having a "this is the server calling the function" variable and a separate "enabled server" variable, the latter of which can be used by the web uis.

config/config.go
```diff
@@ -469,14 +500,17 @@ func initConfigDir() error {
...
+func InitServer(servers []ServerType) error {
 	if err := initConfigDir(); err != nil {
 		return errors.Wrap(err, "Failed to initialize the server configuration")
 	}
+
+	setEnabledServer(servers)
+
 	xrootdPrefix := ""
-	if sType.IsSet(OriginType) {
+	if IsServerEnabled(OriginType) {
```
turetske, line 511: Same concerns as above re: multiple servers in the same process.

xrootd/authorization.go
```diff
@@ -148,7 +148,7 @@ func EmitAuthfile(server server_utils.XRootDServer) error {
 		words := strings.Fields(lineContents)
 		// There exists a public access already in the authfile
 		if len(words) >= 2 && words[0] == "u" && words[1] == "*" {
-			if server.GetServerType().IsSet(config.OriginType) {
+			if server.GetServerType().IsEnabled(config.OriginType) {
```
turetske, line 151: Again. I'm concerned about the "pelican serve" command and running multiple servers from the same process. Mostly marking it again so as to have a record of all the locations it could be an issue.

xrootd/xrootd_config.go
```diff
@@ -313,7 +313,7 @@ func CheckXrootdEnv(server server_utils.XRootDServer) error {
 		}
 	}
 
-	if server.GetServerType().IsSet(config.OriginType) {
+	if server.GetServerType().IsEnabled(config.OriginType) {
```
turetske, line 316: Here is one last place where we want to differentiate between "enabled" vs. "calling" server.

turetske (changes requested):
Comments made regarding the discussions we had re: multiple servers in a single process.

xrootd/authorization.go
haoming29: In _think_ this part shouldn't be any issue because `GetServerType()` only returns the implementation of the `server` instance, which will only return one server type ([ref](https://github.com/PelicanPlatform/pelican/blob/b27a6edfb97357b2db1a109bd78dd2a89f2fcb8a/origin_ui/advertise.go#L39)). And the logic in `IsEnabled` is robust to tell if a single server is enabled.

xrootd/xrootd_config.go
haoming29: Same as the one in `xroot/authorization`, the method override will ensure it behaves as expected for running multiple servers.

xrootd/authorization.go
turetske: Ah, okay. Then this is fine, but I think the issues in config/config.go still apply.

config/config.go
```diff
@@ -547,7 +584,7 @@ func InitServer(sType ServerType) error {
 	// they have overridden the defaults.
 	hostname = viper.GetString("Server.Hostname")
 
-	if sType.IsSet(CacheType) {
+	if IsServerEnabled(CacheType) {
```
turetske, line 587: This should be a current server check. If you are running both an origin and a cache from the same process and you start up the cache first, then this will overwrite the origin's xrootd port, which we don't want.

turetske (changes requested):
One issue in config.go that needs to be dealt with.

turetske (approved):
LGTM. The prometheus issue will need to be deal with in the future when the fed in a box implements web-ui, but that's for a different time.

PR #512, opened 2023-12-19
Origins now advertise themselves as writebackhosts

joereuss12:
This is a bandaid fix for now to allow origins to show themselves as writeback hosts to allow for puts. This helps seperate them from relying on topology to get this information. This is just a band-aid for now, as we should make the writebackhost configurable and able to be set or not set when an origin is spun up.


bbockelm (changes requested):
I'd like to completely rework the approach here.

- No client header changes should be necessary.  The target of the redirect of the PUT command should be all that's needed.  We had to do this as a separate header originally because we weren't using the HTTP redirect functionality.
   - That is, a standard `curl` should be able to do the upload.
- There should be a change in the advertisement from origin to the director.  However, instead of advertising the `WriteBackHost`, we should simply advertise whether writes are enabled (a boolean flag) as the endpoint for PUTs should be the same as the endpoint for any other operation.
- For the fix around the periodic advertisement, that looks like it's potentially an important bugfix.  Please separate that out and have @turetske review it as a standalone PR.

joereuss12:
Closed by #622

Issue #439, opened 2023-12-01
Require Admin approval for new caches


turetske:
When a cache is added to the namespace registry, there should be a default admin setting (a JSON) that is set to false. If this evaluates to false when the director queries the registry for that cache, it will return a 404 error. This is to prevent caches being added to the federation without admin approval.

Notes:

We should ensure that advertiseOSDF still works properly as all the OSDF caches have all been approved. Also make sure we can still do end-to-end testing with a federation in a box

turetske:
If existing (non-OSDF) caches do not have this set, assume they haven't been approved.

haoming29:
As a note, #264 will implement this as part of the registry web-ui in7.4

PR #515, opened 2023-12-19
Registry UI APIs

### LINKED ISSUE #264: Namespace web UI
The namespace registry currently works in a "first come, first serve" mode -- anyone registering an unclaimed namespace is allowed to get it.

An actual production deploy will require the origin admin to provide additional information (such as institution name, issuer details -- anything that is in OSG Topology today) and the registry admin to approve.

Extend the registry so:
- The fields to register are taken from the config.  Should be able to specify some types in order to allow value validation.
- The database holds generic key/value pairs associated with the registration (probably a new table?).
- It has a web UI where the form can be filled in.  If need be, the UI can be left to a separate issue (but a RESTful interface is needed).
- Users can view their pending registrations.
- Admins can approve (or edit) registrations.

haoming29:
Closes #264 

* Please refer to Swagger UI / [Swagger documentation](https://github.com/PelicanPlatform/pelican/blob/main/web_ui/frontend/app/api/docs/pelican-swagger.yaml) for available APIs
* This PR didn't include support for custom registration field to scale down the scope (it's already ~2000 lines of change!), which will be a separate PR
* Other follow-up tickets will be added to improve test coverage and improve the robustness
* **Note**: This PR breaks the Web UI for registry as the data type of `admin_metadata` has been changed from string to json. I would assume this will be part of the UI update from @CannonLock
* **Note**: This PR has foreseeable merge conflict with #439 / #462 . Depending on which PR gets merged first, the other PR requires non-trivial conflict resolution. @turetske We can talk about how to fix the conflict.


config/config.go
```diff
@@ -117,11 +121,22 @@ var (
 	transport     *http.Transport
 	onceTransport sync.Once
 
+	// Global CSRF hanlder that shares the same auth key
```
bbockelm, line 124:
```suggestion
	// Global CSRF handler that shares the same auth key
```

config/config.go
```diff
@@ -354,6 +369,36 @@ func setupTransport() {
 	}
 }
 
+func setupCSRFHandler() {
```
bbockelm, line 372: Can you refactor this into a separate file in the `config` directory?  I fear `config.go` is already too large, adding a bunch of web-related imports only makes it more complicated to read/review.

registry/registry_db.go
```diff
@@ -40,20 +40,50 @@ import (
...
+// and the field will also be excluded from field discovery endpoint (OPTION method).
+//
+// We use validator package to validate struct fields from user requests. If a field is
+// required, add `validate:"required"` to that field. This tag will also be used by fields discovery
+// endpoint to tell the UI if a field is required. For other validator tags,
+// visit: https://pkg.go.dev/github.com/go-playground/validator/v10
+type AdminMetadata struct {
+	UserID                string             `json:"user_id" post:"exclude"` // "sub" claim of user JWT who requested registration
+	Description           string             `json:"description"`
+	SiteName              string             `json:"site_name" validate:"required"`
```
bbockelm, line 63: Let's make the SiteName field not-required.

In OSG Topology, the "site name" is an organization within an institution (topology calls this a "facility").  For example, a university may have a "Physics Department" and "CS Department" that run independent resources.  However, the site name is only used to group contacts for services (a shortcut for putting the same person as in charge of every service at the site).

In Pelican, I'd like to see how far we can push things without having a "site name" concept.  We may come back to it -- but I'd like to start off with it being optional.

registry/registry_db.go
```diff
@@ -73,6 +109,23 @@ https://www.alexedwards.net/blog/organising-database-access
 */
 var db *sql.DB
 
+func (st ServerType) String() string {
+	return string(st)
+}
+
+func (rs RegistrationStatus) String() string {
+	switch rs {
+	case 1:
```
bbockelm, line 118: Have the case be the various names (`Pending`, `Approved`, `Denied`) instead of the integer values (1, 2, 3).

registry/registry_validation.go
```diff
@@ -0,0 +1,167 @@
...
+	 *       choices made here.
+	 */
+	key, exists := clientJwks.Key(0)
+	if !exists {
+		return nil, errors.New("There was no key at index 0 in the reuqest pubKey's JWKS. Something is wrong")
+	}
+	return key, nil
+}
+
+func validateInstitution(inst string) bool {
```
bbockelm, line 154: Fairly large change requested:

Please map the institution name to an internal identifier and return the internal identifier here.

For example, the following are entries in the OSG institutional spreadsheet:

```
OSG Identifier                    | Institution Name
--------------------------------- | ---------------------
https://osg-htc.org/iid/05bxb3784 | Academia Sinica
https://osg-htc.org/iid/02d2m2044 | Agricultural Research Service
https://osg-htc.org/iid/05cf8a891 | Albert Einstein College of Medicine
https://osg-htc.org/iid/03thb3e06 | American Museum of Natural History
https://osg-htc.org/iid/00ff4bt20 | Arcadia University
```

We should always display the "Institution name" to the user but, internally, serialize the identifier into the database.  That implies that the `Registry.Institutions` parameter is more complicated than a string slice.

Longer-term (separate PR), we should provide the option of either listing all of the ID<->Institution mappings in the configuration file or providing an API that can be queried.  When running in "OSDF mode", the API would be automatically set to a relevant topology URL.

registry/registry_validation.go
```diff
@@ -0,0 +1,167 @@
...
+			serverError = errors.Wrap(err, "Server encountered an error checking if namespace already exists")
+			return
+		}
+
+		// if not in OSDF mode, this will be false
+		if inTopo {
+			validationError = errors.New("Cannot register a super or subspace of a namespace already registered in topology")
+			return
+		}
+		// If we make the assumption that namespace prefixes are heirarchical, eg that the owner of /foo should own
```
bbockelm, line 84:
```suggestion
		// If we make the assumption that namespace prefixes are hierarchical, eg that the owner of /foo should own
```

registry/registry_db.go
```diff
@@ -270,18 +354,139 @@ func dbGetPrefixJwks(prefix string, approvalRequired bool) (*jwk.Set, error) {
 	return &set, nil
 }
 
+func getNamespaceById(id int) (*Namespace, error) {
+	if id < 1 {
+		return nil, errors.New("Invalid id. id must be a positive number")
```
bbockelm, line 359: Side-note: this check is reasonable for a SQLite backend.  In the future, we may have alternate ID formats.

It's OK to leave for now but would be good to have a mental picture of the ID being somewhat arbitrary.

registry/registry_db.go
```diff
@@ -270,18 +354,139 @@ func dbGetPrefixJwks(prefix string, approvalRequired bool) (*jwk.Set, error) {
...
+	}
+	ns := &Namespace{}
+	adminMetadataStr := ""
+	query := `SELECT * FROM namespace WHERE id = ?`
```
bbockelm, line 363: Nitpick (I know we do this wrong elsewhere, just trying to make future changes simpler):

Instead of doing `SELECT *`, we should name the columns we are selecting from the database.  This way, if we ever do schema evolution and add columns to this table in the future, we don't break the query.

registry/registry_db_test.go
```diff
@@ -343,23 +599,17 @@ func TestCacheAdminTrue(t *testing.T) {
...
-	require.NoError(t, err, "error marshalling json admin data")
 
 	adminTester := func(ns Namespace) func(t *testing.T) {
 		return func(t *testing.T) {
 			err = addNamespace(&ns)
 
 			require.NoError(t, err, "error adding test cache to registry database")
 
-			// This will return a serverCredsError if the admin_approval == false check is triggered, which we don't want to happen
+			// This will return a serverCredsError if the AdminMetadata.Stauts != Approved, which we don't want to happen
```
bbockelm, line 610:
```suggestion
			// This will return a serverCredsError if the AdminMetadata.Status != Approved, which we don't want to happen
```

registry/registry_db.go
```diff
@@ -73,6 +109,23 @@ https://www.alexedwards.net/blog/organising-database-access
...
+func (rs RegistrationStatus) String() string {
+	switch rs {
+	case 1:
+		return "Pending"
+	case 2:
+		return "Approved"
+	case 3:
+		return "Denied"
+	default:
+		return "Unkown"
```
matyasselmeci, line 125:
```suggestion
		return "Unknown"
```

registry/registry_ui.go
```diff
@@ -22,16 +22,131 @@ import (
...
+	DateTime registrationFieldType = "datetime"
+)
+
+var (
+	registrationFields        []registrationField
+	setRegistrationFieldsOnce sync.Once
+)
+
+func init() {
+	setRegistrationFieldsOnce.Do(func() {
```
bbockelm, line 65: I don't think the `once` function is necessary here.  The `init()` function is guaranteed to be invoked once.

`once` objects are useful when you want to do the initialization lazily -- i.e., cases when it's expensive to do the computation so you only want it done when the result is actually needed, not every time the server starts.

registry/registry_ui_test.go
```diff
@@ -251,3 +238,89 @@ func TestGetNamespaceJWKS(t *testing.T) {
 		})
 	}
 }
+
+func TestAdminAuthHandler(t *testing.T) {
+	// Initialize Gin and set it to test mode
+	gin.SetMode(gin.TestMode)
+
+	// Define test cases
+	testCases := []struct {
```
bbockelm, line 247: Very cool way to do the multiple test cases here!

bbockelm (changes requested):
A modest number of TODO items for the review -- the most important to focus on is the institutional validation.

web_ui/frontend/app/api/docs/pelican-swagger.yaml
```diff
@@ -265,19 +398,404 @@ paths:
...
+    get:
+      tags:
+        - "registry_ui"
+      summary: Return a list of all namespaces in the registry
+      description: A public API to get all namespaces in the registry
+      parameters:
+        - name: server_type
+          in: query
+          required: false
+          description: The type of server to filter the results. The value can be either "origin" or "cache"
```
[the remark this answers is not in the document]
haoming29: Unfortunately we can't. OpenAPI2.0 does not support an object for query parameters but string only. 3.0 introduces this flexibility and we can do it in 3.0.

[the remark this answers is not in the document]
bbockelm: Ok, good to know!

bbockelm (approved):
LGTM!

I'm sure there are additional items I've missed but I like the conceptual design, we already went through the first round of issues, and I worry about merge conflicts across such a large change.

PR #520, opened 2023-12-20
Match binary name by prefix, not exact match

### LINKED ISSUE #513: Use prefix matching for behavior based on binary name
For the times when the behavior is different based on the name of the executable (`stashcp` vs `pelican` vs `stash_plugin`, etc.), can we have it be a prefix match instead of an exact comparison? It would be useful to do version comparisons by calling one of the binaries `stashcp-7.4.0` and the other one `stashcp-7.5.0`, for example.

jhiemstrawisc:
This allows us to version the binary by naming it something like <prefix>-7.2.2

Closes #513

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -39,9 +39,9 @@ func handleCLI(args []string) error {
 	// Being case-insensitive
 	execName = strings.ToLower(execName)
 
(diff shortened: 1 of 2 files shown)


matyasselmeci (approved):
Nice, thanks!

PR #525, opened 2023-12-20
Create hidden pelican serve mode

### LINKED ISSUE #172: "pelican serve" mode

turetske:
Close #172 

Adds a hidden command "pelican serve" which starts up multiple modules as a "federation in a box" to be used for development.

The details on how to use it can be found in the pelican google drive under Development Notes.

It currently doesn't allow for enabling web ui, caches, or origins in s3 mode. Separate issues have been created for all three of these enhancements.

### DIFF
--- a/cmd/fed.go
+++ b/cmd/fed.go
@@ -20,6 +20,7 @@ package main
 
 import (
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 var (
@@ -33,13 +34,16 @@ var (
 		 If the director or namespace registry are enabled, then ensure there is a corresponding url in the
 		 pelican.yaml file.
 
-		 NOTE: This currently doesn't guarantee support for the web UIs`,
+		 This feature doesn't currently support the web UIs`,
 		RunE: fedServeStart,
 	}
 )
 
 func init() {
-	serveCmd.Flags().StringSlice("modules", []string{}, "Modules to be started.")
+	serveCmd.Flags().StringSlice("module", []string{}, "Modules to be started.")
+	if err := viper.BindPFlag("Server.Modules", serveCmd.Flags().Lookup("module")); err != nil {
+		panic(err)
+	}
 	serveCmd.Flags().Uint16("reg-port", 8446, "Port for the namespace registry")
 	serveCmd.Flags().Uint16("director-port", 8445, "Port for the director")
 	serveCmd.Flags().Uint16("origin-port", 8443, "Port for the origin")
--- a/cmd/fed.go
+++ b/cmd/fed.go
@@ -1,0 +1,47 @@
+/***************************************************************
+ *
+ * Copyright (C) 2023, Pelican Project, Morgridge Institute for Research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License.  You may
+ * obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ***************************************************************/
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	serveCmd = &cobra.Command{
+		Use:    "serve",
+		Hidden: true,
+		Short:  "Starts pelican with a list of enabled modules",
+		Long: `Starts pelican with a list of enabled modules [registry, director, cache, origin] to enable better
+		 end-to-end and integration testing.
+
+		 If the director or namespace registry are enabled, then ensure there is a corresponding url in the
+		 pelican.yaml file.
+
+		 NOTE: This currently doesn't guarantee support for the web UIs`,
+		RunE: fedServeStart,
+	}
+)
+
+func init() {
+	serveCmd.Flags().StringSlice("modules", []string{}, "Modules to be started.")
+	serveCmd.Flags().Uint16("reg-port", 8446, "Port for the namespace registry")
+	serveCmd.Flags().Uint16("director-port", 8445, "Port for the director")
+	serveCmd.Flags().Uint16("origin-port", 8443, "Port for the origin")
+	serveCmd.Flags().Uint16("cache-port", 8442, "Port for the cache")
+}
--- a/cmd/fed_serve.go
+++ b/cmd/fed_serve.go
@@ -27,119 +27,92 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
-	"time"
 
 	"github.com/pelicanplatform/pelican/config"
 	"github.com/pelicanplatform/pelican/param"
+	"github.com/pelicanplatform/pelican/server_utils"
 	"github.com/pkg/errors"
-	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
-var currentModules = map[string]string{"registry": "", "director": "", "origin": ""}
-var cModsStringList = "registry, director, origin"
-
-func fedCheckWorking(reqUrl string, server string) error {
-	expiry := time.Now().Add(10 * time.Second)
-	success := false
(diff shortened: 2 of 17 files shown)


cmd/fed_serve.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var currentModules = map[string]string{"registry": "", "director": "", "origin": ""}
+var cModsStringList = "registry, director, origin"
+
+func fedCheckWorking(reqUrl string, server string) error {
```
bbockelm, line 43: This appears to largely be a copy/paste of the `serverRunning` function in [xrootd/origin_test.go](https://github.com/PelicanPlatform/pelican/blob/main/xrootd/origin_test.go#L111).  Why not refactor out and make it a common function in the utils module?

cmd/fed.go
```diff
@@ -0,0 +1,47 @@
...
+		 If the director or namespace registry are enabled, then ensure there is a corresponding url in the
+		 pelican.yaml file.
+
+		 NOTE: This currently doesn't guarantee support for the web UIs`,
+		RunE: fedServeStart,
+	}
+)
+
+func init() {
+	serveCmd.Flags().StringSlice("modules", []string{}, "Modules to be started.")
```
bbockelm, line 42: Bind this to a viper config variable ([see this example](https://github.com/PelicanPlatform/pelican/blob/main/cmd/director.go#L88)); that way, the modules to run can be controlled via other configuration means (environment variable, YAML config file).

cmd/fed_serve.go
```diff
@@ -0,0 +1,213 @@
...
+				}
+				moduleMap[mod] = port
+			case "origin":
+				port, err := cmd.Flags().GetUint16("origin-port")
+				if err != nil {
+					return errors.Wrap(err, "Failed to parse origin port")
+				}
+				moduleMap[mod] = port
+			default:
+				moduleMap[mod] = 0 //Should never happen
```
bbockelm, line 97: Throw an error here instead of ignoring it.

cmd/fed_serve.go
```diff
@@ -0,0 +1,213 @@
...
+	return nil
+}
+
+func fedServeStart(cmd *cobra.Command, args []string) error {
+	moduleSlice, err := cmd.Flags().GetStringSlice("modules")
+	if err != nil {
+		return errors.Wrap(err, "Failed to load modules passed via --module flag")
+	}
+	// Faking a set of certain values
+	var moduleMap map[string]uint16 = map[string]uint16{}
```
bbockelm, line 74: Let's convert this to an internal struct.  Something along the lines of:

```
type modulePorts struct {
  RegistryPort uint16
  DirectorPort uint16
  ...
}
```

This introduces some type safety and, I suspect, will result in a cleaner syntax.

cmd/fed_serve.go
```diff
@@ -0,0 +1,213 @@
...
+			return errors.New(fmt.Sprintf("Unknown module %s: Modules must be one of %s", mod, cModsStringList))
+		}
+	}
+	ctx := cmd.Context()
+	return fedServeInternal(ctx, moduleMap)
+}
+
+func fedServeInternal(ctx context.Context, moduleMap map[string]uint16) error {
+
+	hostname, err := os.Hostname()
```
bbockelm, line 109: I think you want to use [Server.Hostname](https://github.com/PelicanPlatform/pelican/blob/main/config/config.go#L587) instead.  When running in containers, the result of `os.Hostname()` is often meaningless.

cmd/fed_serve_test.go
```diff
@@ -0,0 +1,88 @@
...
+	viper.Reset()
+	moduleMap := map[string]uint16{"registry": 8446, "director": 8445, "origin": 8443}
+
+	// Create our own temp directory (for some reason t.TempDir() does not play well with xrootd)
+	tmpPathPattern := "XRootD-Test_Origin*"
+	tmpPath, err := os.MkdirTemp("", tmpPathPattern)
+	require.NoError(t, err)
+
+	// Need to set permissions or the xrootd process we spawn won't be able to write PID/UID files
+	permissions := os.FileMode(0777)
```
bbockelm, line 45: This doesn't smell right!

It sounds like there's a directory within the tempdir needs to be created that is owned (and writable) by xrootd.

World-writable directories are dangerous, even for unit tests.  It looks like this got copied from other tests -- please fix the underlying problem (and create a new GitHub issue to fix the test this was copied from).

cmd/fed_serve_test.go
```diff
@@ -0,0 +1,88 @@
...
+	viper.Set("Server.EnableUI", false)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		err = fedServeInternal(ctx, moduleMap)
+		require.NoError(t, err)
+	}()
+	defer cancel()
+
+	time.Sleep(2 * time.Second)
```
bbockelm, line 75: We shouldn't have a time-based sleep here: it slows down unit tests and can lead to periodic failures (if we land on a slow runner where startup takes 3 seconds, this could crash and burn).

Instead do a fast poll as in `serverRunning`.

cmd/fed_serve_test.go
```diff
@@ -0,0 +1,88 @@
...
+	hostname, err := os.Hostname()
+	require.NoError(t, err)
+
+	curlString := "https://" + hostname + ":8445/.well-known/openid-configuration"
+
+	curl := exec.Command("curl", "-v", "-k", curlString)
```
bbockelm, line 81: We shouldn't make assumptions about `curl` being available.  Switching to a function like `serverRunning` should bring all the HTTP interactions in-process

docs/parameters.yaml
```diff
@@ -291,13 +291,6 @@ type: bool
 default: true
 components: ["origin"]
 ---
-name: Origin.EnableUI
```
bbockelm, line 294: I think removing this is a backward compatibility issue -- what happens in places where this is already set?

It's worth a discussion with @haoming29 -- should the `EnableUI` flag really be just the UI (that is, the piece that is viewed in the browser) or all the API functionality.  If it's the latter, migrating to a common `Server.EnableUI` and deprecating the `Origin.EnableUI` flag is reasonable.  I suspect we might lean toward the former as the API components are becoming critical -- what is the registry without the APIs?

xrootd/resources/xrootd-origin.cfg
```diff
@@ -85,6 +85,6 @@ ofs.ckslib * libXrdMultiuser.so
 xrootd.chksum max 2 md5 adler32 crc32
 xrootd.trace emsg login stall redirect
 pfc.trace info
-pss.setopt DebugLevel 1
+pss.setopt DebugLevel 3
```
bbockelm, line 88: Let's rebase this change away.  I think this debug level should be controlled by the configuration, not hardcoded.

bbockelm (changes requested):
Various refactoring requests inline.

The bigger conceptual question is around the port usage.  What I'm envisioning is having a single `gin` instance serving all the APIs in a single port (this will also be the path toward enabling the UI as part of this mode).  I see two options:

1. Merge now.  It's hidden functionality and gets us partway there.  In 7.5.0, aim for a follow-up that serves all components (except, obviously, xrootd and OA4MP) behind a single port.
2. Hold off.  Address current review comments and start working on the single-port functionality.  This eases the 'glide path' for the larger vision.

Personally, I'd aim toward (1).  What are your thoughts?

bbockelm:
Oh -- and, of course, please fix up the merge conflicts.

Actually, the merge conflicts might be the best argument for merging now even though major improvements are planned in the future.  Large PRs are quite difficult to land!

docs/parameters.yaml
bbockelm: I decided to re-introduce the `Origin.EnableUI` parameter but keep the references pointed to `Server.EnableUI`.

I would like the `Server.EnableUI` to refer to any browser-based interface whatsoever whereas the `$MODULE.EnableUI` would be only the relevant pane in the browser interface.

bbockelm:
All this work was included as part of #570 (including the rebased commits).  Due to the rebase, the merge of that PR did not auto-close this one -- closing it by hand.

PR #527, opened 2023-12-20
Replace `xrd-http` plugin by `xrdcl-pelican` plugin in dev-container

### LINKED ISSUE #485: Switch to xrdcl-pelican
The new [xrdcl-pelican](https://github.com/PelicanPlatform/xrdcl-pelican) plugin is meant to replace `xrdcl-http` as it allows us to put in Pelican-specific logic (e.g., think about retries or overriding the CAs).  It cuts out the dependency on DAVIX and libneon and it is slightly faster to boot.

We'll have the OSG folks bootstrap a copy in their repos for the most common Linux platforms.  Otherwise:
- Linux/arm64: We could checkout and build in a separate step of the dev container build process.  Risks version skew with the OSG and a small increase of the dev container build time.  Should minimize space increase if we split the build into a separate container.
- Mac OS X: We might just need to checkout and rebuild as a separate step in the GHA.  Possibly something that could be cached?

haoming29:
Closes #485 
@bbockelm It seems that testing the production container requires supervisord and it can't run without it, but I can confirm that dev container now uses `xrdcl-pelican` plugin. Although I tried to build plugins in a separate stage, the final size of the image is still ~2.3GB. Not sure what's wrong.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -62,7 +62,7 @@ RUN \
     git clone https://github.com/PelicanPlatform/xrdcl-pelican.git && \
     cd xrdcl-pelican && \
     mkdir build && cd build && \
-    cmake .. && \
+    cmake -DLIB_INSTALL_DIR=/usr/lib64 .. && \
     make && make install
 # Install the S3 and HTTP server plugins for XRootD. For now we do this from source
 # until we can sort out the RPMs.
@@ -70,7 +70,7 @@ RUN \
     git clone https://github.com/PelicanPlatform/xrootd-s3-http.git && \
     cd xrootd-s3-http && \
     mkdir build && cd build && \
-    cmake .. && \
+    cmake -DLIB_INSTALL_DIR=/usr/lib64 .. && \
     make install
 
 FROM dependency-build AS final-stage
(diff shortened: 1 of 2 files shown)


images/Dockerfile
```diff
@@ -122,17 +144,18 @@ ENV JAVA_HOME=/usr/lib/jvm/jre \
...
-    cd xrootd-s3-http && \
-    mkdir build && cd build && \
-    cmake .. && \
-    make install && \
-    # For now, until the RPM is set up, we install the libraries here, but
-    # we need to add to LD_LIBRARY_PATH so XRootD knows where to look
-    echo "/usr/local/lib" > /etc/ld.so.conf.d/xrdplugins.conf && ldconfig
+# Copy built xrdcl-pelican plugin library and config from build
+COPY --from=xrootd-plugin-builder /usr/local/etc/xrootd/client.plugins.d/pelican-plugin.conf /etc/xrootd/client.plugins.d/pelican-plugin.conf
+COPY --from=xrootd-plugin-builder /usr/local/lib/libXrdClPelican-5.so /usr/local/lib/libXrdClPelican-5.so
```
bbockelm, line 149: Please collapse all the `COPY` statements into a single one (each statement adds layers to the resulting image, I believe).

images/Dockerfile
```diff
@@ -122,17 +144,18 @@ ENV JAVA_HOME=/usr/lib/jvm/jre \
...
+# Remove http plugin to use pelican plugin
+RUN rm -f /etc/xrootd/client.plugins.d/xrdcl-http-plugin.conf
+
+# Copy built s3 plugin library from build
+COPY --from=xrootd-plugin-builder /usr/local/lib/libXrdS3-5.so /usr/local/lib/libXrdS3-5.so
+COPY --from=xrootd-plugin-builder /usr/local/lib/libXrdHTTPServer-5.so /usr/local/lib/libXrdHTTPServer-5.so
+# For now, until the RPM is set up, we install the libraries here, but
+# we need to add to LD_LIBRARY_PATH so XRootD knows where to look
+RUN echo "/usr/local/lib" > /etc/ld.so.conf.d/xrdplugins.conf && ldconfig
```
bbockelm, line 158: Instead of doing this, copy all the new plugins into a standard directory (`/usr/lib64`).  When building the plugins, you should be able to add the flag `-DLIB_INSTALL_DIR=/usr/lib64` to CMake to ensure the right configuration file is generated as well.

images/dev.Dockerfile
```diff
@@ -25,14 +25,53 @@ baseurl=https://repo.goreleaser.com/yum/ \n\
 enabled=1 \n\
 gpgcheck=0' > /etc/yum.repos.d/goreleaser.repo
 
-# Install proper version of nodejs so that make web-build works
-RUN dnf module install -y nodejs:20
-
 # Install goreleaser and various other packages we need
-RUN yum install -y goreleaser npm xrootd-devel xrootd-server-devel xrootd-client-devel nano xrootd-scitokens \
```
bbockelm, line 32: For the "dev" container, I'd like to keep an entire, valid development environment.

So, let's not do the split here like we do in the production container.

This means that the development container will indeed grow significantly -- but I'm OK with that.

bbockelm (changes requested):
A few minor comments - should be easy to address.

I wasn't able to reproduce issues with the size.  From a clean checkout, the total size of the production container went _down_ by a few hundred MB's as there are fewer `-devel` packages in the final container.

haoming29:
> A few minor comments - should be easy to address.
> 
> I wasn't able to reproduce issues with the size. From a clean checkout, the total size of the production container went _down_ by a few hundred MB's as there are fewer `-devel` packages in the final container.

Yes! I think the issue is only to my local build and the cause should be that my local `node_modules` folder was included in the final image which bumps the size significantly

bbockelm (approved):
LGTM!

PR #529, opened 2023-12-21
Truncate the plugin output file as soon as possible

### LINKED ISSUE #519: stash_plugin should fsync its output file
Under an out-of-disk error, the output file for the stash_plugin may not make it to disk.  The contents of the output are buffered in memory and may not be flushed to disk until after the `stash_plugin` process exits.  If the flush occurs between when the process exits and when the `condor_starter` reads the file, then an error may have occurred and the contents of some pages lost.

In theory -- and, perhaps in practice -- this could cause the starter to not notice errors reported by the plugin as the buffered pages were lost.

### LINKED ISSUE #528: In the HTCondor transfer plugin, truncate output file as early as possible
HTCondor 23.3.0 and before will reuse the output file name across multiple invocations, meaning that the results from prior file transfers will be present when the plugin runs.  Combined with a bug that causes any exit code other than 1 to be labeled as a success (so a go panic, which is exit code 2, is considered a success), this causes reasonable-looking transfers in the outputs.

Accordingly, if there's a panic during the output file transfer and Pelican was used for the input file transfer, the overall output sandbox transfer will be considered a success.  This can result in truncated or missing outputs.

One way to avoid triggering this is to truncate the output file as early as possible in the plugin.  Then, when a panic occurs, no ClassAds are discovered in the output file and the job is put on hold.

bbockelm:
Additionally, ensure we fsync the final output file when we're finished.

Fixes #519
Fixes #528

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -21,10 +21,12 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"io/fs"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/pelicanplatform/pelican/classads"
@@ -139,7 +141,8 @@ func stashPluginMain(args []string) {
 	if getCaches {
 		urls, err := client.GetCacheHostnames(testCachePath)
 		if err != nil {
-			log.Panicln("Failed to get cache URLs:", err)
+			log.Errorln("Failed to get cache URLs:", err)
+			os.Exit(1)
 		}
 
 		cachesToTry := client.CachesToTry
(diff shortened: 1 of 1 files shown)


cmd/plugin.go
```diff
@@ -185,6 +187,25 @@ func stashPluginMain(args []string) {
...
+	// with missing or truncated output files.
+	//
+	// By moving the truncation of the output file to a very early codepath, we reduce
+	// the chances of hitting this problem.
+	outputFile := os.Stdout
+	if useOutFile {
+		var err error
+		outputFile, err = os.Create(outfile)
+		if err != nil {
+			log.Errorln("Failed to open outfile:", err)
```
matyasselmeci, line 204: Did you want to `os.Exit(1)` here too?

bbockelm: Done!

matyasselmeci (approved):

Issue #309, opened 2023-10-31
Web ui for the pelican cache


turetske:
Create a web_ui for the pelican cache. Currently the web_ui is not spun up for the pelican cache, this should be changed.

-- The periodic advertisement should be moved once this is implemented

haoming29:
I'm a bit confused about "The periodic advertisement should be moved once this is implemented". Where should it be moved to? For Origin, periodic advertisement is a separate function called in `serveOrigin` which I assume would be the same case for cache?

PR #532, opened 2023-12-21
Add storage volume Prometheus metrics from XRootD monitoring packets

### LINKED ISSUE #495: Add a new `xrootd_storage_volume_bytes` Prometheus metric from XRootD monitoring packets
We want to know the storage volume condition of each storage server (origin/cache), grouped by namespace and type (available, total, used). To implement this, we can take advantage of xrootd summary monitoring packet and grab the data from [Oss summary data ](https://xrootd.slac.stanford.edu/doc/dev56/xrd_monitoring.htm#_Toc138968505), particularly `oss.paths.i.free` and `oss.paths.i.tot`. For finding the corresponding namespace, you should look at `oss.paths.i.lp`. 

Note that `oss.paths.i.lp` will only give you the top level namespace. So if you have `/test/foo` as your registered namespace,`oss.paths.i.lp` is `/test`

It would be safe to assume that we only have one path for each datastore although this could be changed in the future.

haoming29:
Closes #495 

A note to reviewer: Although this PR implemented the metric for both origins and caches, we won't receive any metrics from cache yet as we don't have web stuff for cache implemented by the time this PR is ready (it's in a separate ticket here #309). 

I think #493 might suffer from this, as we added caches to director's service discovery endpoint but they are essentially "dead" endpoints. Director Prometheus instance might produce a bunch of "false positive" error messages

### DIFF
--- a/metrics/xrootd_metrics.go
+++ b/metrics/xrootd_metrics.go
@@ -41,7 +41,8 @@ import (
 )
 
 type (
-	UserId struct {
+	SummaryStatType string
+	UserId          struct {
 		Id uint32
 	}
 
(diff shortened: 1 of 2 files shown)


bbockelm (approved):
LGTM.

PR #534, opened 2023-12-21
Better management of token scopes

### LINKED ISSUE #374: Better management of token scopes
With increasingly many tokens added to the system, we should consider centralize the management of their scopes instead of relying on string literals, with which we can also give clear view of what scopes are being used in the system where.

* Either under `config` or `utils` folder, create a file for scope management
* Enumerate scopes we are using, probably as a struct of 
```go
type Scope struct {
    EntityName string
    Action string
}
```
and implement `String()` function for it
* Come up with a defined convention of the format of the scopes, e.g. case-sensitivity, order of name/action, entity prefix (pelican etc)
* Document and enforce such convention when creating and verifying token scopes in the codebase

haoming29:
Closes #374 

* Introduced a new `scopes.yaml` file to configure/document available token scopes. Use `utils.*` constants with type `TokenScope` for a more defined and clear way of assigning token scopes
* Provide a util function `utils.GetScopeString()` to take a `[]TokenScope` and convert to space-separated token claim, which fits the JWT standard.


director/director_api.go
```diff
@@ -78,7 +79,7 @@ func CreateDirectorSDToken() (string, error) {
 	tokenExpireTime := param.Monitoring_TokenExpiresIn.GetDuration()
 
 	tok, err := jwt.NewBuilder().
-		Claim("scope", "pelican.directorSD").
+		Claim("scope", utils.Pelican_DirectorServiceDiscovery.String()).
```
bbockelm, line 82: Here (and throughout PR) -- from the names, it's not clear that these are token scopes.  Options could be:
- Rename the package called `token_scopes` or similar.
- Add a prefix to the variable name (`TokenScope_Pelican_DirectorServiceDiscovery`).

I think I'd lean toward the package rename.

.gitignore
```diff
@@ -3,3 +3,4 @@ README.dev.md
 param/parameters.go
 docs/parameters.json
 param/parameters_struct.go
+utils/token_scopes.go
```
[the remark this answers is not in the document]
bbockelm: Oh!  On second thought, we've found that _not_ including generated files into git is problematic in Go as it breaks others from being able to import Pelican.  So, let's just drop this change to `.gitignore` completely.

director/director_api.go
```diff
@@ -140,7 +141,7 @@ func VerifyDirectorSDToken(strToken string) (bool, error) {
 	scopes := strings.Split(scope, " ")
 
 	for _, scope := range scopes {
-		if scope == "pelican.directorSD" {
```
bbockelm, line 143: I think there's a common pattern of splitting the scope string and then seeing the scope is included.  How about writing a helper function in the `TokenScope` class for that?  Something like (untested code):

```
func (target *TokenScope) HasScope(scopeClaim string) bool {
  for _, scope := range strings.Split(scopeString, " ") {
    if target.String() == scope {
      return true
    }
  }
  return false
}
```

director/origin_api.go
```diff
@@ -75,7 +75,7 @@ func CreateAdvertiseToken(namespace string) (string, error) {
 	}
 
 	tok, err := jwt.NewBuilder().
-		Claim("scope", "pelican.advertise").
+		Claim("scope", utils.Pelican_Advertise.String()).
```
bbockelm, line 78: Let's add a helper function here as well.  Maybe call it `Scope` for adding a single scope to the token and `Scopes` for a slice of `TokenScope` objects?

bbockelm: This would change the syntax here to:

```
	tok, err := jwt.NewBuilder().
		Scope(utils.Pelican_Advertise).
		...
```

bbockelm: The other nice thing about creating a helper function here would be the addition of type safety -- it'll be easy for reviewers to spot mistakes (as we should never have calls of `Claims("scope", ...)`).

utils/token_utils.go
```diff
@@ -184,146 +139,6 @@ func (config *TokenConfig) verifyCreateWLCG() error {
...
-					return errors.New(errMsg)
-				}
-			}
-		}
-	}
-
-	return nil
-}
-
-func CreateEncodedToken(claimsMap map[string]string, profile TokenProfile, lifetime int) (string, error) {
```
bbockelm, line 227: Why are all these being deleted?  Were they never used (and we decided they won't be used)?

bbockelm (changes requested):
A few questions left inline.

utils/token_utils.go
haoming29: This is a fix-up from previous PR #429 where I replaced the existing claim-map based token creation by the OOP based token creation but forget to remove the code. Nothing would be broken, I've updated code and test to use the new `CreateToken` function instead.

director/origin_api.go
haoming29: This should be in the scope for an additional work #559 to replace all existing token creation logic by `utils.CreateToken` function. In this PR, I will add scope type enforcement but won't touch the existing token creation logic, so that we won't repeat ourselves in #559.

director/director_api.go
haoming29: We already have this function in our codebase and it's even more versatile: https://github.com/PelicanPlatform/pelican/blob/1a47c0f35b69f75950954b81977d2bda39b2c800/utils/server_auth.go#L89

However, I think it's time to migrate this function to `token_scopes` package and export it there, so that things are more organized. Also, since `CheckAnyAuth` is already competent in checking scopes, once we finish #559, we should be good to go.

bbockelm: Ok -- we'll do this change in the follow-up (tweaked the title of #559 to reflect this would also handle the token verification case pointed out here).

bbockelm (approved):
Revised PR LGTM.

PR #536, opened 2023-12-22
Have pelican-osdf-compat package replace "stashcache-client"

matyasselmeci:
stashcache-client was the name of the old Python-based stashcp; have the pelican-osdf-compat package provide and replace that package too, since it's apparently still installed in a few places and we want a smooth upgrade path.


bbockelm (approved):
LGTM.

PR #539, opened 2023-12-26
Rename `Federation.NamespaceUrl` to `Federation.RegistryUrl`

### LINKED ISSUE #388: Rename Federation.NamespaceUrl to Federation.RegistryUrl in config
Apparently `Federation.NamespaceUrl` isn't pointing to a namespace but instead to the registry. To be clear on this, we want to rename that config to  `Federation.RegistryUrl`. However, this may be a breaking change to client users, which we should proceed with caution.

haoming29:
Closes #388 

Per request, `Federation.NamespaceUrl` is preserved but marked as deprecated. User can still use this config but a warning message will should up if `Federation.NamespaceUrl` is used instead of `Federation.RegistryUrl`

### DIFF
--- a/cmd/registry_client.go
+++ b/cmd/registry_client.go
@@ -21,7 +21,7 @@
 // You can access it through `./pelican namespace <command>`.
 //
 // Note that you need to have your registry server running either locally,
-// or by setting Federation.NamespaceUrl to the Url of your remote Pelican registry server
+// or by setting Federation.RegistryUrl to the Url of your remote Pelican registry server
 //
 // Example: `./pelican namespace register --prefix /test`
 
@@ -48,14 +48,14 @@ var prefix string
 var pubkeyPath string
 
 func getNamespaceEndpoint() (string, error) {
-	namespaceEndpoint := param.Federation_NamespaceUrl.GetString()
+	namespaceEndpoint := param.Federation_RegistryUrl.GetString()
 	if namespaceEndpoint == "" {
 		return "", errors.New("No namespace registry specified; either give the federation name (-f) or specify the namespace API endpoint directly (e.g., --namespace-url=https://namespace.osg-htc.org/namespaces)")
 	}
 
 	namespaceEndpointURL, err := url.Parse(namespaceEndpoint)
 	if err != nil {
-		return "", errors.Wrap(err, "Unable to parse namespace url")
+		return "", errors.Wrap(err, "Unable to parse namespace registry url")
 	}
 
 	// Return the string, as opposed to a pointer to the URL object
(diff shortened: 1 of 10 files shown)


bbockelm (changes requested):
Two items, one which doesn't require follow-up:

1.  Please avoid spurious whitespace changes (there's a few whitespace-only changes in the parameters YAML file).  If these are necessary (i.e., someone else snuck them in and they are needed to pass the linter), please separate them out into a second commit.  No changes required now, however.
2. I'd like to see a small test that the warning message is actually emitted.  I won't require this for all changes along these lines but mostly want to make sure we have an example of testing log messages

For item (2), relevant examples from Google appear to be:
- https://github.com/sirupsen/logrus#testing
- https://maxchadwick.xyz/blog/testing-log-output-in-go-logrus

haoming29:
> Two items, one which doesn't require follow-up:
> 
> 1. Please avoid spurious whitespace changes (there's a few whitespace-only changes in the parameters YAML file).  If these are necessary (i.e., someone else snuck them in and they are needed to pass the linter), please separate them out into a second commit.  No changes required now, however.
> 2. I'd like to see a small test that the warning message is actually emitted.  I won't require this for all changes along these lines but mostly want to make sure we have an example of testing log messages
> 
> For item (2), relevant examples from Google appear to be:
> 
> * https://github.com/sirupsen/logrus#testing
> * https://maxchadwick.xyz/blog/testing-log-output-in-go-logrus

Sorry about the whitespace changes. I think my VSCode plugin was formatting the yaml file on save and it's buried in my intended changes. This makes me wonder if we want to have a preferred yaml formatter for the project to make things more consistent? I was using [RedHat plugin](https://developers.redhat.com/products/vscode-extensions/overview). Otherwise, I will disable my local formatter.

bbockelm:
> Sorry about the whitespace changes.

It's fine to do fixups -- feel free to keep your plugin enabled -- just please put them as a separate commit.  If you've not used it before, try using `git add --patch` to fine-tune the commits.  The `--patch` flag allows you to manage what's staged for the commit at a per-line level instead of per-file.

> This makes me wonder if we want to have a preferred yaml formatter for the project to make things more consistent?

We have the `check-yaml` precommit hook enabled.  Looks like the one you're using is just pickier.  There's others to choose from (see [this page](https://pre-commit.com/hooks.html), filter on `yaml`); I'm not sure if they can be made to match any VSCode formatter.

bbockelm (approved):
LGTM.  I like the new tests of logging -- hopefully the approach is useful in other places in the codebase.

PR #540, opened 2023-12-26
Unpublish Go package

### LINKED ISSUE #505: Unpublish Go package
Currently, we published our packages at https://pkg.go.dev/github.com/pelicanplatform/pelican but we are not prepared for others to use our package to build their own project, so we want to find a way to "unpublish" it, although it's not possible to completely remove it from Go package manager, we should follow the procedure [here](https://pkg.go.dev/about#removing-a-package) for the best result.

haoming29:
Closes #505 

Note that once this PR is merged, someone needs to "push" the change to Go package repository. On the main branch of the central repository (not your fork), run `GOPROXY=proxy.golang.org go list -m github.com/pelicanplatform/pelican@v1.0.4`

This way, we should be able to retract our modules from Go.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/pelicanplatform/pelican
 
+retract v1.0.4 // Unpublish Go package
+
 go 1.20
 
 require (


bbockelm (approved):
LGTM.

@haoming29 - since you've read up on the process, I'll leave it to you to do the retract step.

PR #542, opened 2023-12-26
Create `tempDir` for all config files to avoid file not found in test

### LINKED ISSUE #541: Registration Test fails in GHA due to credential file not found
We see increasingly many random test failure from `server_ui/register_namespace_test` where the expected credential files are missing. On example is:

In https://github.com/PelicanPlatform/pelican/actions/runs/7293193763/job/19876211750, the windows runner fails due to `Private key file, C:\Users\runneradmin\.config\pelican\certificates\tls.key, contains no private key`

haoming29:
Fixes #541

### DIFF
--- a/server_ui/register_namespace_test.go
+++ b/server_ui/register_namespace_test.go
@@ -49,13 +49,12 @@ type (
 )
 
 func TestRegistration(t *testing.T) {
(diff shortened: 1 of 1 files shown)


bbockelm (approved):
LGTM.

PR #547, opened 2023-12-27
Remove chmod on user-provided xrootd config files

haoming29:
When user provide xrootd config files i.e. `authfile` and `scitokens.cfg` by mounting them through docker/kubernetes in read-only mode, there will be a failure in origin/cache at the start, where Pelican will try to change the user group of both files to the daemon name (xrootd in most case), which is not allowed. 

This chmod is redundant as we eventually copy the two files, make some changes, and save them as final files to xrootd run location, so we only need to be able to read those files.


bbockelm (approved):
LGTM.

PR #549, opened 2023-12-27
Only build and publish containers as well as releasing on tags >=`7.0.0` in GHA

### LINKED ISSUE #548: Change GHA to only build containers and release on tags >= `v7.0.0`
As a follow up work for #505, we need to push a tag to git for the retracted go module version `v1.0.4`, which for now will hit a chain of reactions for GHA. But we don't want to build/release any artifact out of this new tag, so prior doing the push, we want to exclude tags below `v7.0.0` in GHA from building and pushing a new container, as well as cutting a new release.

haoming29:
Closes #548

### DIFF
--- a/.github/workflows/publish-container.yml
+++ b/.github/workflows/publish-container.yml
@@ -4,7 +4,9 @@ on:
   pull_request:
   push:
     tags:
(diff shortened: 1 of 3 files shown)


bbockelm (approved):
LGTM.

PR #553, opened 2023-12-28
Fix mismatched kid between token and public jwks

### LINKED ISSUE #474: Mismatched `kid` in testing with `config.GetIssuerPublicJWKS`
When implementing the PR in #429, if I used `config.GetIssuerPublicJWKS` inside `IssuerCheck`, several of test cases failed due to mismatched `kid` when validating JWT using issuer's key. This shouldn't happen as we assigned `kid` to the private JWK first and then generate the public JWK which then should have the same `kid` as the private. It worth investigating why this isn't working.

haoming29:
Fixes #474 

* We store the private JWK in memory once we get it. We won't "get" from the `issuerKey` file if the in-memory private key is not nill. This is not true for `GetIssuerPublicJwks` as under the hood, it _will_ bypass the in-memory cache and get the private key directly from `issuerKey`. This causes problem in testing because each test function may create their own tempDir and generate a new private key, and the private key we used to sign a token might be outdated and not refreshed. This PR fixes the problem by always looking at the `issuerKey` file, the same behavior as the public jwks look up.
* This PR also replace the code that grab the public key from the raw key of the private JWK by directly calling `GetIssuerPublicJwks`


bbockelm (changes requested):
It looks like this fixes the unit tests -- but at the cost of loading the private key from disk each time `GetIssuerPrivateJWK` is invoked!  That's a bit costly, I suspect.

Two ideas for improvemnet:
1.  Whenever config is invoked, it clears the value of `issuerPrivateJWK`, forcing a reload.
2. Also cache the return value of `GetIssuerPublicJWKS`.  If the caches are reset at the same time, then they should stay consistent.

One observation is this doesn't allow us to easily rotate signing keys without restarting the service.  I'll file a separate issue for that.

haoming29:
@bbockelm I know you are off from technical stuff but just so you know this PR is ready to go. I may find other folks to do the review but I think you might want to look at the changes and see if that fits your thoughts

haoming29:
@jhiemstrawisc requested your review in case Brian is not available for a while. No need to hurry thought, this is for 7.5

config/init_server_creds.go
```diff
@@ -477,14 +478,18 @@ func loadIssuerPublicJWKS(existingJWKS string, issuerKeyFile string) (jwk.Set, e
...
+	key := issuerPrivateJWK.Load()
+	if key == nil {
+		// This returns issuerPrivateJWK if it's non-nil, or find and parse private JWK
+		// located at IssuerKey if there is one, or generate a new private key
+		_, err := loadIssuerPrivateJWK(issuerKeyFile)
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to load issuer private JWK")
+		}
 	}
+	key = issuerPrivateJWK.Load()
```
jhiemstrawisc, line 490: Why the double load here?

director/origin_api_test.go
```diff
@@ -128,8 +128,13 @@ func TestCreateAdvertiseToken(t *testing.T) {
...
+	config.InitConfig()
+	err := config.InitServer(ctx, config.DirectorType)
+	require.NoError(t, err)
+
+	// Launcher will set default values to some of the server urls. Reset here.
+	viper.Set("Federation.RegistryUrl", "")
+	viper.Set("Federation.DirectorURL", "")
 
 	// Test without a namsepace set and check to see if it returns the expected error
 	tok, err := CreateAdvertiseToken("test-namespace")
```
jhiemstrawisc, line 140: I guess this isn't in the purview of this particular PR, but this is an invalid namespace because it doesn't have a preceding `/`. Are there any consequences of testing things at this level of code with invalid namespaces? I guess the token doesn't really care whether the namespace is valid?

web_ui/prometheus_test.go
```diff
@@ -139,206 +144,133 @@ func TestPrometheusProtectionOriginHeaderScope(t *testing.T) {
...
+		assert.NoError(t, err, "Error generating private key")
 
-	jti_bytes = make([]byte, 16)
-	_, err = rand.Read(jti_bytes)
-	if err != nil {
-		t.Fatal(err)
-	}
-	jti = base64.RawURLEncoding.EncodeToString(jti_bytes)
+		// Convert from raw ecdsa to jwk.Key
+		pKey, err := jwk.FromRaw(privateKey)
```
jhiemstrawisc, line 225: How about we change to `privKey` for clarity, since `pkey` could stand for either public or private.

jhiemstrawisc (changes requested):

haoming29:
Hey @bbockelm and @jhiemstrawisc , since this PR has been up for quite a while and I patched in the fix based on the last code review feedback, could we push this through or have another round of review? Thanks!

jhiemstrawisc (approved):
Everything from my last review looks like it has been addressed. I'm going to risk the assumption that I did a thorough job the first time. LGTM!

jhiemstrawisc:
Bypassing branch protections to get this one in.

PR #554, opened 2023-12-28
Fix error in retracting pkg with not including previous versions

haoming29:
I only did it half-right in the previous PR #540 where I created a higher version for go (v.1.0.4) but forget to retract all previous version (v1.0.0-v1.0.3) so that the retract is void. I found this error while already pushed the new 1.0.4 tag.

Added all the previous versions in this PR, I need to build a new tag and push it again. Sorry about adding unhelpful tags to the repo.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/pelicanplatform/pelican
 
-retract v1.0.4 // Unpublish Go package
+// Unpublish Go package as we are not intended to allow users us import our packages for now
+retract [v1.0.0, v1.0.5]
 
 go 1.20
 


bbockelm (approved):
LGTM.

bbockelm:
Feel free to self-merge once you are ready to take another swing at the retract operation!

haoming29:
As a note, I've pushed the tag and updated Go package proxy and confirmed locally all the previous version in Go package registry are retracted. It may take a while for the website to reflect this: https://pkg.go.dev/github.com/pelicanplatform/pelican

PR #555, opened 2023-12-28
Add generated files to git

### LINKED ISSUE #387: Can't easily import pelican pkgs
Not sure if you guys want to support this functionality, but just wanted to bring it up.

I can't import pelican golang pkgs normally due to
needing to generate structs in `generate/param_generator.go`.

Here are the different versions developers can import: https://pkg.go.dev/github.com/pelicanplatform/pelican@v0.0.0-20231116152555-e2c5da2433d3/param?tab=versions

I am adding the latest one in the `go.mod` file:
```
require (
        github.com/pelicanplatform/pelican v0.0.0-20231115165602-38fe2717eba3
)
```

Example error during build:
```
8.540 /go/pkg/mod/github.com/pelicanplatform/pelican@v0.0.0-20231115165602-38fe2717eba3/param/param.go:12:15: undefined: config
```

Currently, I am cloning the repo and making the golang structs:
```
git clone https://github.com/PelicanPlatform/pelican.git &&\
    cd pelican && \
    git checkout tags/v7.2.0 && \
    make generate
```

Then doing a replace in `go.mod` file with directory location:
```
replace github.com/pelicanplatform/pelican => /registration/pelican
```

bbockelm:
If anyone wants to import the Pelican modules externally in the future (we're not there with the client -- but would like to be!), then the generated files need to be part of the release.

This commit adds the generated files and tweaks the parameter struct generator so the results are stable across multiple runs.

Fixes #387


haoming29 (approved):
LGTM!

PR #560, opened 2023-12-29
Return 404 for unregistered prefix

### LINKED ISSUE #294: Namespace registry should return 404 for an unregistered namespace
When doing a curl against a namespace that doesn't exist, we see:

```
$ curl -v https://registry.osg-htc.org/api/v1.0/registry/test2-bbockelm/.well-known/issuer.jwks | jq
 ...
> GET /api/v1.0/registry/test2-bbockelm/.well-known/issuer.jwks HTTP/2
> Host: registry.osg-htc.org
> User-Agent: curl/8.1.2
> Accept: */*
> 
< HTTP/2 500 
< content-type: application/json; charset=utf-8
< date: Sat, 28 Oct 2023 17:28:47 GMT
< content-length: 69
< 
{ [69 bytes data]
100    69  100    69    0     0   1659      0 --:--:-- --:--:-- --:--:--  1815
* Connection #0 to host registry.osg-htc.org left intact
{
  "error": "server encountered an error trying to get jwks for prefix"
}
```

It would be better if this was a 404 response, reflecting the fact that the `/test2-bbockelm` namespace isn't registered.

bbockelm:
When a requested prefix doesn't exist, return a 404 instead of a 500.

This PR also adds minimal unit test coverage for the corresponding DB and API calls as well as a tangentially-related commit to clean up some error messages that I encountered while hand-testing.

Fixes #294

### DIFF
--- a/config/init_server_creds.go
+++ b/config/init_server_creds.go
@@ -48,6 +48,12 @@ var (
 	issuerPrivateJWK atomic.Pointer[jwk.Key]
 )
 
+type (
+	noPrivateKeyError struct {
+		error
+	}
+)
+
 // Return a pointer to an ECDSA private key read from keyLocation.
 //
 // This can be used to load any ECDSA private key we generated for
@@ -79,7 +85,7 @@ func LoadPrivateKey(keyLocation string) (*ecdsa.PrivateKey, error) {
 		}
 	}
 	if privateKey == nil {
-		return nil, fmt.Errorf("Private key file, %v, contains no private key", keyLocation)
+		return nil, noPrivateKeyError{fmt.Errorf("Private key file, %v, contains no private key", keyLocation)}
 	}
 	return privateKey, nil
 }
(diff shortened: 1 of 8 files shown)


haoming29:
@bbockelm Could you rebase against main? I think registry API PR caused some conflicts here.

haoming29:
@bbockelm sorry about the back-and-forth changes due to merging conflict, but in #613 I changed much of how registry response to the GET `issuer.jwks` request. I added the logic to check existence of prefix and return 404 if not, and I modified the approval check to return 403 if cache is not approved, which are similar to what you've changed here. 

I like the other changes to improve test coverage and error message but I would suggest rebasing against main again in case you want things done in a particular way different from what I changed. I will review it ASAP.

haoming29:
@bbockelm This PR seems to be a bit obsolete, with #685 moving towards merging, but I like the extra check in OIDC config as well as registry path parameter. Just wondering how do you plan to proceed with this PR?

For db functions, I thought the intend is to keep them simple an neat, with limited scope focusing only on data retrieval but not request/data validation, which I suppose should be done at controller layer (request handlers). That's why I moved approval check out from db function to handler function in #685, but let me know which way you think is preferred.

haoming29:
Close the PR as most of the issues have been addressed in other PRs

Issue #565, opened 2023-12-29
Reload TLS certificate periodically


bbockelm:
We should assume that Pelican processes are longer-lived than the underlying TLS certificate.  A maintenance goroutine should watch for changes to the certificate or key file and update the HTTPS interface accordingly.

This [StackOverflow](https://stackoverflow.com/questions/37473201/is-there-a-way-to-update-the-tls-certificates-in-a-net-http-server-without-any-d) provides guidance on how to dynamically provide the HTTPS server with a certificate per-request.  We could make a global atomic pointer that holds the current certificate / key that is updated from the goroutine thread.

PR #569, opened 2023-12-30
Reload TLS certificates as needed

bbockelm:
We want to provide for clean updates of the TLS certificate for production services.  This PR does two things:

1.  Makes a copy of the TLS certificates for XRootD, avoiding file permission issues that tend to hit Kubernetes.  Adds a goroutine that periodically updates XRootD's copy of the certificate.  Uses a combination of fsnotify and polling for updates.
2. Using the same update approach, pull in new certificates in the pelican web UI.

Includes unit tests for both items.

### DIFF
--- a/cmd/cache.go
+++ b/cmd/cache.go
@@ -43,7 +43,7 @@ var (
 )
 
 func initCache() error {
-	err := config.InitServer([]config.ServerType{config.CacheType}, config.CacheType)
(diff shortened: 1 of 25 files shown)


bbockelm:
Fixes #565

bbockelm:
With all apologies for the self-merge, I'm going to pull this in so I can make progress on a second round of the `pelican serve` command.

PR #570, opened 2023-12-31
Federation-in-a-box mode / context-based goroutines

bbockelm:
As a follow-up to #525, this refactors the various modules to be launched in a uniform way.


bbockelm:
Based on the size - and amount of merge conflicts this is generating with other PRs - I'm going to go ahead and merge this in.

PR #571, opened 2024-01-01
Add OS X support for XRootD tests

bbockelm:
This PR installs xrootd into the Mac OS X builder along with various missing dependencies (scitokens support, xrdcl-pelican) needed in our environment.

With this change, we can enable end-to-end tests on OS X! 🎉 

Note one subtlety here: the addition of an IPv4-only mode.  It appears that if you land on a Mac OS runner _without_ IPv6, then XRootD fails to startup.  This sets IPv4-only in the tests for OS X and introduces the corresponding parameters.

Also note that, after we build all the xrootd extensions, we fire up a serve with a fake config (and an invalid hostcert) to see if it'll stay live for at least 3 seconds.  This way, bad builds fail early/noisy instead of mysteriously in the middle of the pelican unit tests.


bbockelm:
Going to self-merge this one as a functioning end-to-end test on OS X is necessary for the federation-in-a-box mode.

PR #574, opened 2024-01-02
Stash_plugin uploads now record bytes transferred

joereuss12:
When dostashcpsingle is called and detects a writeback, it will now return the uploaded bytes instead of just an error


bbockelm (approved):
LGTM

PR #575, opened 2024-01-02
Login fixes

### LINKED ISSUE #424: Show error message when authentication failed in UI
Ref: https://github.com/PelicanPlatform/pelican/issues/421#issuecomment-1832302673

Currently there's no error message popping up at the UI if the authentication failed due to any reason. The user only gets "no response" after clicking "Activate" or "Login" button which is very confusing. We should introduce error messages to the activation/login page to let user know what happened if auth fails.

### LINKED ISSUE #556: Incorrect wording on registry configuration screen
The wording on the registry auth bootstrapping screen is wrong:

image

This is the registry, there's nothing to do with "root metrics".

Suggested wording:

> Set Password
> This will become the admin password for this Pelican endpoint

CannonLock:
Closes #556 
Closes #424

### DIFF
--- a/web_ui/authentication.go
+++ b/web_ui/authentication.go
@@ -207,8 +207,16 @@ func loginHandler(ctx *gin.Context) {
 		ctx.JSON(400, gin.H{"error": "Missing user/password in form data"})
 		return
 	}
+	if strings.TrimSpace(login.User) == "" {
+		ctx.JSON(400, gin.H{"error": "User is required"})
+		return
+	}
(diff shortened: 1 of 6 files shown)


haoming29 (changes requested):
When testing locally, it seems that the login page `/view/login` still didn't have prompt/message with auth failure. We want the same behavior as in the initialization code page.

haoming29 (commented):
LGTM! I pushed one commit to improve the backend returned error message.

PR #576, opened 2024-01-02
Fix time zone difference in registry DB test

haoming29:
With what current registry DB test has, if the local machine timezone is different from UTC, tests will report errors. This PR fixes this issue by convert time to unix time first before making the comparison.


bbockelm (approved):
LGTM.

Issue #572, opened 2024-01-02
Remove `utils.LoadDirectorPublicKey`


haoming29:
As a follow-up work for #417, when we successfully turn director into a federation, we no longer need to distinguish between `directorCheck` and `federationCheck` for token validation as both will have the same public JWKS, and `utils.LoadDirectorPublicKey` will be redundant, so we want to remove it.

However, `director.VerifyDirectorTestReportToken` relies on `utils.LoadDirectorPublicKey` to work, and to make the transition smooth, we want to replace `director.VerifyDirectorTestReportToken` by `utils.VerifyToken` to avoid adding extra code to grab federation's public JWKS, and `utils.VerifyToken` will be ready only after #559, so we need to wait for that to be merged in first.

haoming29:
I think the work this ticket described is actually part of #559, as `director.VerifyDirectorTestReportToken` is expected to be removed at the end of #559. Will transit useful items to #559 and close this one.

haoming29:
Closes with #877

PR #577, opened 2024-01-02
Director as the federation

### LINKED ISSUE #417: Add director's public key to federation's discovery URL

haoming29:
Closes #417 

* As #417 required, added a path same as OSDF federation discovery endpoint (`.well-known/pelican-configuration`) to director's web API with the same information as OSDF one
* The existing OpenID style endpoint for director public key discovery remains unchanged (`.well-known/openid-configuration`) for backward compatibility. This is also necessary if we want to strictly follow the token issuer metadata lookup standard from [wlcg](https://github.com/WLCG-AuthZ-WG/common-jwt-profile/blob/master/profile.md#metadata-lookup) or scitokens.
* With above change, Pelican director now is capable of being a federation, and it's repetitive to have separate `directorCheck` in token validation, so related code are removed. However, `utils.LoadDirectorPublicKey` is left as a separate issue to address #572 

This PR will change the order of running Pelican servers. It is **REQUIRED** to run `director` server first before any other servers, if you set `Federation.DiscoveryUrl` to director's URL. 

Please also check that you are not overriding federation discovery url via command line flag `-f` in your other servers.


cmd/fed_serve_test.go
```diff
@@ -65,7 +65,9 @@ func TestFedServePosixOrigin(t *testing.T) {
 	viper.Set("ConfigDir", tmpPath)
 	viper.Set("Xrootd.RunLocation", filepath.Join(tmpPath, "xrootd"))
 	t.Cleanup(func() {
-		os.RemoveAll(tmpPath)
+		if err := os.RemoveAll(tmpPath); err != nil {
+			t.Fatal("Failed to clean up temPath")
```
bbockelm, line 69: Small typo.
```suggestion
			t.Fatal("Failed to clean up temp path")
```

config/config.go
```diff
@@ -737,12 +741,18 @@ func InitServer(ctx context.Context, enabledServices ServerType) error {
...
 	// This populates Server.IssuerUrl, and can be safely fetched using server_utils.GetServerIssuerURL()
-	err = parseServerIssuerURL(enabledServices)
+	err = parseServerIssuerURL(currentServers)
 	if err != nil {
 		return err
 	}
 
-	return DiscoverFederation()
+	// Since director is the federation, we need to wait until director's web engine is running
+	// to do DiscoverFederation on director server
```
bbockelm, line 750: I don't think this logic is right?

If the director is part of a larger federation, then DiscoverFederation may need to be used to discover the registry service.

director/discovery.go
```diff
@@ -27,25 +27,55 @@ import (
...
+	openIdDiscoveryPath     string = "/.well-known/openid-configuration"
+	federationDiscoveryPath string = "/.well-known/pelican-configuration"
+	directorJWKSPath        string = "/.well-known/issuer.jwks"
 )
 
-// Returns links related to director's authentication, including the link
-// to get the public key from the director
-func discoveryHandler(ctx *gin.Context) {
+func federationDiscoveryHandler(ctx *gin.Context) {
 	directorUrl := param.Federation_DirectorUrl.GetString()
```
bbockelm, line 42: Should this fall back to the external web URL if this is the director (probably more of a question in the config.go, not actually putting the logic here?).

launchers/director_serve.go
```diff
@@ -64,5 +64,12 @@ func DirectorServe(ctx context.Context, engine *gin.Engine, egrp *errgroup.Group
 	engine.Use(director.ShortcutMiddleware(defaultResponse))
 	director.RegisterDirector(ctx, rootGroup)
 
+	// Only discover federation when we run in OSDF mode; otherwise, the director is expected to be the federation
```
bbockelm, line 67: This is not quite right: the federation _may_ be the director but we don't have to make this assumption.  So we should still make this call.

However, I think it's safe for the director to assume its key is the federation key.

bbockelm (changes requested):
Most of the code is in the right direction.

However, there's one big different assumption: I assume that the director should always serve the federation metadata -- but not always be the discovery endpoint.  So, we should do federation discovery if that is present in case if there's other metadata (like the registry URL) it needs to look up.

Once that's cleaned up, I think we will be nearly ready to go.

bbockelm (approved):
I'm happy with the revisions -- just needs the merge conflict cleaned up.

PR #580, opened 2024-01-02
Stash plugin calls init-config & added unit test

joereuss12:
Stash plugin will now call init-config on startup so it can get the configured timeouts and log levels as well as any other default/environment configuration. Added a unit test to test functionality of stashPluginMain as well as this new fix.

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -70,6 +70,7 @@ func init() {
 }
 
 func stashPluginMain(args []string) {
+	config.InitConfig()
 	err := config.InitClient()
 	if err != nil {
 		log.Errorln(err)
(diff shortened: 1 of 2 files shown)


joereuss12:
@haoming29 and/or @jhiemstrawisc If you guys have time, can you try to run my test and figure out what error code/syscall error message Mac receives for this: `Failed to sync output file: sync /dev/stdout: operation not supported by device`

The line is question is 
`if errors.As(err, &perr) && errors.As(perr.Unwrap(), &serr) && (int(serr) == 1 || serr == syscall.EINVAL) {`

We should be going through this if statement, just add the condition for `serr` when found. If you guys have any questions or need me to elaborate more let me know.

bbockelm (approved):
LGTM.

Change in the default log level should be fine as long as HTCondor invokes the multi-file plugin (which should be always true here).

Issue #111, opened 2023-09-07
Implement namespace issuer auto-discovery


bbockelm:
Currently, when the director wants to download the public keys for an origin, it queries the namespace issuer directly.  See: https://github.com/PelicanPlatform/pelican/blob/main/director/origin_api.go#L139-L150

Instead, we should implement "openid-configuration"-style auto-discovery: revisit the code [around here](https://github.com/PelicanPlatform/pelican/blob/main/namespace-registry/registry.go#L704-L714) to reimplement `.well-known`-based lookup while keeping backward compatibility -- the current URL can still serve up the public keys.

bbockelm:
Since this one is purely internal - and we're running low on v7.1.0 time, I'm pushing this to the v7.2.0 milestone.

bbockelm:
@jhiemstrawisc - since this has repeatedly gotten pushed back, I'd like to prioritize it for 7.5.0.

PR #583, opened 2024-01-02
Update ns key discovery to follow openid-style lookups

jhiemstrawisc:
Previously, namespace keys were fetched by hardcoding the issuer jwks relative to the namespace.
Now we conform to openid-style lookups by fetching the JSON located at
<registry url>/api/v2.0/registry/metadata/<namespace>/.well-known/namespace-configuration
and following the URL associated with the `jwks_uri` key. Currently this is still the same key path
in all instances that I'm aware of, but in theory this could be expanded to point to other locations.
We can also expand this namespace-configuration JSON to include additional information as needed.

The architecture of this PR is that a GET request to the namespace-configuration url mentioned above
should dynamically generate the JSON with `jwks_uri` key and associated value. The value is used throughout
and assigned the issuer URL in various places where we need to create a token on behalf of themnamespace.

### DIFF
--- a/director/origin_api.go
+++ b/director/origin_api.go
@@ -176,6 +176,11 @@ func VerifyAdvertiseToken(ctx context.Context, token, namespace string) (bool, e
 		return false, err
 	}
 
+	keyLoc, err := GetJWKSURLFromIssuerURL(issuerUrl)
+	if err != nil {
+		return false, err
+	}
+
 	var ar NamespaceCache
 
 	// defer statements are scoped to function, not lexical enclosure,
@@ -202,7 +207,7 @@ func VerifyAdvertiseToken(ctx context.Context, token, namespace string) (bool, e
 	if ar == nil {
 		ar = jwk.NewCache(ctx)
 		client := &http.Client{Transport: config.GetTransport()}
-		if err = ar.Register(issuerUrl, jwk.WithMinRefreshInterval(15*time.Minute), jwk.WithHTTPClient(client)); err != nil {
+		if err = ar.Register(keyLoc, jwk.WithMinRefreshInterval(15*time.Minute), jwk.WithHTTPClient(client)); err != nil {
 			return false, err
 		}
 		namespaceKeysMutex.Lock()
@@ -216,8 +221,8 @@ func VerifyAdvertiseToken(ctx context.Context, token, namespace string) (bool, e
 		}
 
 	}
-	log.Debugln("Attempting to fetch keys from ", issuerUrl)
-	keyset, err := ar.Get(ctx, issuerUrl)
+	log.Debugln("Attempting to fetch keys from ", keyLoc)
+	keyset, err := ar.Get(ctx, keyLoc)
 
 	if err != nil {
 		return false, err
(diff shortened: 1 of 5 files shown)


jhiemstrawisc:
Closes #111

haoming29:
I'm holding off reviewing this PR as there are major changes done in #613 which will cause significant merge conflict with this PR, and with #613 merged in, you might need also to downgrade the API version to be compatible, but I like the idea to have a conformed way to do the look up. Let's talk about how to proceed @jhiemstrawisc

jhiemstrawisc:
@haoming29, sounds good. Let's focus on getting #613 in, and then I can revisit this to see what still makes sense.

director/origin_api.go
```diff
@@ -260,18 +263,49 @@ func VerifyDirectorTestReportToken(strToken string) (bool, error) {
...
+	if registryUrlStr == "" {
+		return "", errors.New("federation registry URL is not set and was not discovered")
+	}
+	registryUrl, err := url.Parse(registryUrlStr)
 	if err != nil {
 		return "", err
 	}
-	namespace_url.Path, err = url.JoinPath(namespace_url.Path, "api", "v2.0", "registry", "metadata", prefix, ".well-known", "issuer.jwks")
+
+	registryUrl.Path, err = url.JoinPath(registryUrl.Path, "api", "v2.0", "registry", "metadata", prefix, ".well-known", "namespace-configuration")
```
haoming29, line 282: I think this should be changed to `url.JoinPath(registryUrl.Path, prefix, ".well-known", "openid-configuration")`

First because this is what's defined in [OIDC RFC](https://openid.net/specs/openid-connect-discovery-1_0.html#rfc.section.4) that the `./well-known/openid-configuration` MUST be concatenated directly to the hostname, and the name must be `/openid-configuration`

Fortunately, per OIDC spec, we _can_ append a path before the `./well-known` for multi-tenant use case where a single issuer can serve multiple tenants based on their requested path prefix. It fits perfect into our scenario that we can append the namespace prefix before `./well-known`. It should look like `https://registry.com:9999/test/prefix/.well-known/openid-configuration`.

To do the route matching, we can first check is the path `HasSuffix` with `/.well-known/openid-configuration`, then check if it doesn't have a prefix `/api` or `/view` or any other static resources. If it does, then it's not meant for this purpose. 

The caveat for this is that we will unfortunately hit the `wildcard hell` at the top level where we need to handle the wildcard matching plus all additional routes under root path in the wildcard handler like we did [here](https://github.com/PelicanPlatform/pelican/blob/3caae35f06cd96ae97c2afca51a470b98c060010/registry/registry.go#L646)

haoming29:
Hey @jhiemstrawisc How are we doing here? It feels like there are lots of related work done to the registry that causes considerable amount of merge conflict to this PR, and I thought it'd be beneficial to have it closed soon.

director/origin_api.go
```diff
@@ -271,18 +274,49 @@ func VerifyDirectorTestReportToken(strToken string) (bool, error) {
...
+
+	registryUrl.Path, err = url.JoinPath(registryUrl.Path, "api", "v1.0", "registry", prefix, ".well-known", "openid-configuration")
 	if err != nil {
-		return "", err
+		return "", errors.Wrapf(err, "failed to construct namespace-configuration lookup URL for prefix %s", prefix)
+	}
+
+	// Get/parse the namespace-configuration JSON to lookup key location
+	client := &http.Client{Transport: config.GetTransport()}
+	originConfig, err := client.Get(registryUrl.String())
```
haoming29, line 300: Do we want to consider backward compatibility and fallback option here? At least for OSDF it doesn't seem to have such an endpoint

haoming29 (approved):
LGTM. Only concern from me is about backward compatibility. I'm taking extra caution because it bit us before, but let me know how you think about it and if Brian has some idea.

Issue #564, opened 2023-12-29
Topology/Authfile generation


bbockelm:
In the OSDF, parts of the namespace are managed via X509 client certificates and not tokens.  The X509-based authentication is authorized via entries in the Authfile.

We consider the X509-based authentication deprecated and don't want to collect X509-based information in the Pelican registry.  However, to migrate the OSDF off the legacy packaging before these legacy authentication cases are retired (which might be quite awhile!), we will need to still support these Authfile entries.

Accordingly, when in "OSDF mode" and generating the cache's Authfile, we should query topology for additional X509-related Authfile entries and merge those authorizations in.  @matyasselmeci already has a Python-based script for performing these queries -- we just need to convert it to golang and embed into Pelican.

matyasselmeci:
This is the script that we currently use: https://github.com/opensciencegrid/xcache/blob/V3-branch/src/authfile-update

matyasselmeci:
By the way this issue says Authfile but OSDF caches/origins also pull their scitokens.conf from Topology. Is there another issue for doing the same generation for scitokens.conf?

bbockelm:
The Authfile covers X509 which I don't plan to port into the core of Pelican.  It's OSDF specific and once it's not in topology, we will drop it.

For scitokens.cfg, the director should be importing info from topology and the caches ought to generate the correct info.  I believe there are some open issues along these lines - all things I'm hoping to resolve before 7.5.

matyasselmeci:
and the cache/origin will download scitokens.cfg from the director instead of Topology? I assume it has to be on disk so xrootd-scitokens will see it...

turetske:
Just to confirm, this task is just for the cache's Authfile, not any other configuration file from from topology.

jhiemstrawisc:
Closed by #738

PR #584, opened 2024-01-03
systemd service files and corresponding configs for Pelican services and Pelican-based OSDF services.

### LINKED ISSUE #331: Add pelican-server RPM
Forked from #246: add an RPM named "pelican-server" that contains systemd unit files for starting Pelican services.

matyasselmeci:
This will resolve #331 

The osdf-cache and osdf-origin services are meant as a replacement for the 'stash-cache-auth' and the 'stash-origin-auth' xrootd instances of xcache, so they listen on the authenticated ports (8443 for the cache and 1095 for the origin).


systemd/osdf-cache.yaml
```diff
@@ -0,0 +1,24 @@
...
+Logging:
+  Level: "Error"
+Federation:
+  DiscoveryUrl: osg-htc.org
+  DirectorUrl: https://director-origins.osg-htc.org
+  TopologyNamespaceURL: https://topology.opensciencegrid.org/osdf/namespaces
+Server:
+  TLSCertificate: /etc/grid-security/xrd/xrdcert.pem
+  TLSKey: /etc/grid-security/xrd/xrdkey.pem
+  TLSCACertificateDirectory: /etc/grid-security/certificates
```
bbockelm, line 11: For OSDF, we shouldn't be using the IGTF distribution but the system CAs.

systemd/osdf-cache.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description = Pelican service osdf-cache
+After = network.target nss-lookup.target
+
+[Service]
+ExecStart = /usr/bin/pelican --config /etc/pelican/osdf-cache.yaml cache serve
```
bbockelm, line 6: Let's use a binary named OSDF.  That turns on a bunch of default configs and other internal code paths.

systemd/osdf-cache.yaml
```diff
@@ -0,0 +1,24 @@
+Debug: false
+Logging:
+  Level: "Error"
+Federation:
+  DiscoveryUrl: osg-htc.org
+  DirectorUrl: https://director-origins.osg-htc.org
+  TopologyNamespaceURL: https://topology.opensciencegrid.org/osdf/namespaces
+Server:
+  TLSCertificate: /etc/grid-security/xrd/xrdcert.pem
```
bbockelm, line 9: Is there not a better system location?

systemd/osdf-cache.yaml
```diff
@@ -0,0 +1,24 @@
...
+  TLSKey: /etc/grid-security/xrd/xrdkey.pem
+  TLSCACertificateDirectory: /etc/grid-security/certificates
+  EnableUI: false
+#  Hostname: f.q.d.n
+Xrootd:
+  Sitename: TOPOLOGY_RESOURCE_NAME
+  ManagerHost: redirector.osgstorage.org
```
bbockelm, line 16: Next three lines are unnecessary if in OSDF mode. Let's remove them.

systemd/osdf-cache.yaml
```diff
@@ -0,0 +1,24 @@
...
+  SummaryMonitoringHost: xrd-report.osgstorage.org
+  DetailedMonitoringHost: xrd-mon.osgstorage.org
+  Mount: "/mnt/osdf"
+  Port: 8443
```
bbockelm, line 20: Can be dropped. Overridden by Cache.Port.

systemd/osdf-director.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description = Pelican service osdf-director
+After = network.target nss-lookup.target
+
+[Service]
+ExecStart = /usr/bin/pelican --config /etc/pelican/osdf-director.yaml director serve
```
bbockelm, line 6: For topology integration to work, needs to be named osdf.

systemd/osdf-director.yaml
```diff
@@ -0,0 +1,17 @@
...
+  TLSCertificate: /etc/grid-security/xrd/xrdcert.pem
+  TLSKey: /etc/grid-security/xrd/xrdkey.pem
+  TLSCACertificateDirectory: /etc/grid-security/certificates
+#  Hostname: f.q.d.n
+Xrootd:
+  Sitename: TOPOLOGY_RESOURCE_NAME
+  ManagerHost: redirector.osgstorage.org
+  SummaryMonitoringHost: xrd-report.osgstorage.org
+  DetailedMonitoringHost: xrd-mon.osgstorage.org
+  Mount: "/mnt/osdf"
```
bbockelm, line 17: Not needed for the director process.

systemd/osdf-director.yaml
```diff
@@ -0,0 +1,17 @@
...
+Federation:
+  DiscoveryUrl: osg-htc.org
+  TopologyNamespaceURL: https://topology.opensciencegrid.org/osdf/namespaces
+Server:
+  TLSCertificate: /etc/grid-security/xrd/xrdcert.pem
+  TLSKey: /etc/grid-security/xrd/xrdkey.pem
+  TLSCACertificateDirectory: /etc/grid-security/certificates
+#  Hostname: f.q.d.n
+Xrootd:
+  Sitename: TOPOLOGY_RESOURCE_NAME
```
bbockelm, line 13: Not used by the director, just things that use xrootd.  Can remove.

systemd/osdf-director.yaml
```diff
@@ -0,0 +1,17 @@
+Debug: false
+Logging:
+  Level: "Error"
```
bbockelm, line 3: Default values, I think.  Can be removed.

systemd/osdf-director.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description = Pelican service osdf-director
+After = network.target nss-lookup.target
+
+[Service]
```
bbockelm, line 5: For all these systemd services - should we provide a sysconfig-style file to make overrides via env var easier?

systemd/osdf-origin.yaml
```diff
@@ -0,0 +1,24 @@
...
+  EnableUI: false
+#  Hostname: f.q.d.n
+Xrootd:
+  Sitename: TOPOLOGY_RESOURCE_NAME
+  ManagerHost: redirector.osgstorage.org
+  SummaryMonitoringHost: xrd-report.osgstorage.org
+  DetailedMonitoringHost: xrd-mon.osgstorage.org
+  Mount: "/mnt/osdf"
+  Port: 1095
+#  Authfile: /run/stash-origin-auth/Authfile
```
bbockelm, line 20: Let's remove commented-out lines (or add a comment about what they're for).

bbockelm (changes requested):
Systemd services look mostly fine - just the question about whether we should also provide a sysconfig-based override.

The bigger question is about the config files. For the most part, what's filled in here is the same as the code defaults - my inclination would be to drop them.

Once you drop the defaults, then there are really only one or two settings.  At that point, is it better to just have a single pelican.yaml/osdf.yaml shared by all services?

Obviously, Pelican needs to add a config.d-style system ASAP.  If we assume one is done in February, does this change any of the approaches here?

matyasselmeci:
The reason I had all those options even though they're the defaults is they're the most likely ones for the admin to change and I don't want them to have to go back and forth to the documentation to figure out which ones they are. I plan to add explanatory comments as well (but I will need help writing some of them).

systemd/osdf-origin.yaml
matyasselmeci: The plan is to do the latter.

systemd/osdf-director.service
matyasselmeci: For systemd services, admins should be making changes by editing the config instead of setting environment variables. But sure, I can add a sysconfig file for knobs that can't be set otherwise; it would be especially useful if we could use that sysconfig file for both the systemd-based and the supervisord-based services.

systemd/osdf-origin.yaml
matyasselmeci: Actually for now I'm going to uncomment them since we're going to need them until #564 is implemented.

systemd/osdf-cache.yaml
matyasselmeci: Not really. There's no 'standard' location for a host cert; Let's Encrypt puts the cert/key in `/etc/letsencrypt/live/$domain` for example. At least `/etc/grid-security/xrd/xrdcert.pem` is backward compatible with xcache-based cache/origin configurations; I can pick something else like `/etc/pelican/{cert,key}.pem` for the registry, director, and non-osdf configs.

systemd/osdf-director.service
matyasselmeci: I added EnvironmentFiles in `/etc/sysconfig/SERVICE`.

matyasselmeci:
I think this is ready for another look. For the service files, I added sysconfig files, restarting on failure, and using a working directory we'll be able to write core files to. (The RPM will handle creating that directory.)

For the configs, I added comments describing many of the options. Biggest change is to have log files (`/var/log/osdf-cache.log`, `/var/log/pelican-registry.log`, etc.) instead of logging to the systemd journal. The osdf-* services now start the `osdf` binary so I was able to drop some options.

bbockelm (approved):
General approach is OK.  I'm still ambivalent on having separate config files per service versus a shared one.  I can see pros and cons for both directions.  I'm going to defer to your experience here.

Approved!  Let's get this in and move on to the RPM piece.

systemd/osdf-director.service
```diff
@@ -0,0 +1,12 @@
...
+EnvironmentFile = -/etc/sysconfig/osdf-director
+ExecStart = /usr/bin/osdf --config /etc/pelican/osdf-director.yaml director serve
+Restart = on-failure
+RestartSec = 20s
```
brianhlin, line 9: I see your commit removing the working dir for the director/registry but it's not clear to me why they don't need it

systemd/osdf-cache.yaml
```diff
@@ -0,0 +1,28 @@
+## Enable Debug to send detailed logs, including XRootD logs, to the log file
+# Debug: false
+Logging:
+  LogLocation: /var/log/osdf-cache.log
+  ## Valid Levels are Trace, Debug, Info, Warning, Error, Fatal and Panic.
+  # Level: "Error"
```
brianhlin, line 6: Not really related to this PR but do logs also go to the journal or just to the location here?

brianhlin (commented):
Minor comments, the big one being the standard cert location.

I wish we could consolidate the service files down to 1 or 2. Maybe we could use the following vars?

```
       ├───────────┼─────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┤
       │ "%i"      │ Instance name                                       │ For instantiated units this is the string between the first "@"      │
       │           │                                                     │ character and the type suffix. Empty for non-instantiated units.     │
       ├───────────┼─────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┤
       │ "%p"      │ Prefix name                                         │ For instantiated units, this refers to the string before the first   │
       │           │                                                     │ "@" character of the unit name. For non-instantiated units, same as  │
       │           │                                                     │ "%N".                                                                │
       ├───────────┼─────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┤
```

And have users run something like `systemctl start pelican@director`? I think the only thing that would prevent that is that the director/registry don't have a working dir specified and the cache/origin do

brianhlin:
Mat and I discussed the split of systemd files and its affect on packaging and appropriate location for OSDF vs Pelican bits (we feel that OSG Software should package OSDF bits in a downstream metapackage). For the time being, this split of packages is ok and should allow for flexible future changes.

Mat convinced me that `pelican@director` isn't the right UX as it's not exactly an instance but I could see an interface like `pelican-director@osdf` or something like that in the future. 

I think the cert location is the only thing that is left to address here before a merge.

systemd/osdf-cache.yaml
matyasselmeci: Only to the log location.

systemd/osdf-director.service
matyasselmeci: I suppose the director/registry can core dump too.

Most service files don't have a WorkingDirectory set; either the services run as root (in which case they can write their core files to `/`), or they run as non-root (in which case they start in their home directories and can write core files there). The reason we have to specify it in this case is that pelican starts as root but then spawns xrootd with reduced privileges (without changing the directory first).

systemd/osdf-cache.yaml
```diff
@@ -0,0 +1,28 @@
...
+  TLSCertificate: /etc/grid-security/xrd/xrdcert.pem
+  TLSKey: /etc/grid-security/xrd/xrdkey.pem
+  ## Use TLSCACertificateDirectory instead of TLSCACertificateFile to support auth from grid clients
+  TLSCACertificateFile: /etc/pki/tls/certs/ca-bundle.crt
+  # TLSCACertificateDirectory: /etc/grid-security/certificates
+  EnableUI: false
+  ## Set Hostname to the external DNS name this can be accessed over, if
+  ## different than the current hostname.
+  # Hostname:
+Xrootd:
```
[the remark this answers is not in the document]
matyasselmeci: re. cert location: I guess your google-fu is better than mine: I didn't find a good standard last time I looked. FWIW, we need a cert/key with permissions for the xrootd user to read.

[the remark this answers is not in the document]
matyasselmeci: more re. cert location: Also the page you linked mentioned a directory but not a file name -- that is, it said the cert should be named after the FQDN; obviously I can't put that in the default config, not knowing what it is.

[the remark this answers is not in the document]
brianhlin: I think I prefer mostly aligning with best practices, e.g. `/etc/pki/tls/certs/pelican.{crt,key}`, than maintaining the old-world order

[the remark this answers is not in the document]
matyasselmeci: Fine. #743

PR #586, opened 2024-01-03
Fixup various cache and testing failures

bbockelm:
This cleans up a deadlock in the `pelican cache serve` mode generated in #570 and fixes some race/ordering conditions I spotted which may be causing flakey unit tests also post the #570 merge.

### DIFF
--- a/cmd/fed_serve_test.go
+++ b/cmd/fed_serve_test.go
@@ -80,6 +80,7 @@ func TestFedServePosixOrigin(t *testing.T) {
 	viper.Set("Origin.EnableVoms", false)
 	viper.Set("TLSSkipVerify", true)
 	viper.Set("Server.EnableUI", false)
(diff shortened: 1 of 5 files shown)


bbockelm:
@haoming29 - I can't promise there aren't other race conditions.  However, this definitely solves two of them in the tests and has been one of the few GHA runs to get a clean bill of health since this afternoon.  So, a quick review would be appreciated.

cmd/fed_serve_test.go
haoming29, line 1: I also added this line in my [director as federation PR](https://github.com/PelicanPlatform/pelican/blob/463e003c0820e7f2678b2357c7c26a0b9d9b79a9/cmd/fed_serve_test.go#L73) Depending on whose gets merged in first, the other need to address conflict.

haoming29 (approved):
LGTM

PR #589, opened 2024-01-03
Only run Origin start script when it's origin

haoming29:
Currently, we can't run any server other than origin as the launcher script will always call `OriginServeFinish` and `LaunchPeriodicAdvertise` which are only for the origin, and calling it for other server will lose any origin-specific config values and will cause program to exit.

### DIFF
--- a/launchers/launcher.go
+++ b/launchers/launcher.go
@@ -133,13 +133,18 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
 	}
 
 	log.Debug("Finishing origin server configuration")
-	if err = OriginServeFinish(ctx, egrp); err != nil {
-		return shutdownCancel, err
+	if modules.IsEnabled(config.OriginType) {
+		if err = OriginServeFinish(ctx, egrp); err != nil {
+			return shutdownCancel, err
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


launchers/launcher.go
```diff
@@ -114,6 +114,11 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
 		if err != nil {
 			return shutdownCancel, err
 		}
+
+		log.Debug("Finishing origin server configuration")
```
bbockelm, line 118: This needs to move back to the original location: the web engine needs to have started up for `OriginServeFinish` to run (as this will potentially perform API requests against the director APIs).

bbockelm (changes requested):
Please move the `OriginServeFinish` back to the original place in the sequence.  You're right in that it should be in a `modules.IsEnabled(config.OriginType)` conditional.

haoming29:
@bbockelm Could you take another look? It's a bit blocking now.

launchers/launcher.go
bbockelm: Please remove this dangling log line.

bbockelm (approved):
Pre-approving.  Please remove the forgotten log line and then merge when you're ready!

PR #594, opened 2024-01-03
Regenerate the authfile and scitokens.cfg periodically

### LINKED ISSUE #562: Periodically reload admin-specified Authfile
On startup, we take the admin-specified Authfile, merge it with the generated configuration, and write out the result into the xrootd configuration.

If the admin updates said Authfile, it'll never get propagated into the xrootd configuration.  Have a maintenance goroutine periodically (or, using https://github.com/fsnotify/fsnotify, on changes) reload the admin-provided Authfile and rewrite the generated Authfile.  Care should be used to do an atomic overwrite of the generated one so the `xrootd` process won't load half-written data.

bbockelm:
To replace existing OSDF caches, which rely on X509 authentication that we don't want to pull into the Pelican codebase, we'll need to merge with their authfiles (which are periodically updated).  This utilizes the Xrootd maintenance routine to regenerate authfiles and scitokens.cfg when an update occurs.

Includes unit tests.

Fixes #562

### DIFF
--- a/cmd/cache_serve.go
+++ b/cmd/cache_serve.go
@@ -171,7 +171,7 @@ func serveCacheInternal(ctx context.Context) error {
 		return err
 	}
 
-	xrootd.LaunchXrootdMaintenance(ctx, 2*time.Minute)
+	xrootd.LaunchXrootdMaintenance(ctx, cacheServer, 2*time.Minute)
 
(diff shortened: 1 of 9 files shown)


haoming29 (approved):
LGTM

PR #600, opened 2024-01-03
Fix broken namespace registration with device code flow

haoming29:
I believe when I did #175, I changed the value for various device based token flow request, which are just empty string and breaks the flow. Revert back to what they are supposed to be in this PR.


bbockelm (approved):
LGTM.

Probably points out this area of the code is under-tested.

PR #613, opened 2024-01-08
Roll back registry api from `v2.0` to `v1.0`

### LINKED ISSUE #566: Version compatibility break for origins
In the work for v7.4.0, we added a new "v2.0" API for the registry.

However, this introduced a compatibility break -- the origin may be registering with an older registry that does not have the v2.0 API.

The origin should either auto-detect the supported API version or gracefully fall back to v1.0 if it gets a 404.

To reproduce, simply run:

```
pelican origin serve -f osg-htc.org -v /tmp/test:/test
```

haoming29:
Fixes #566 

See #566 for reasoning behind this implementation.

### DIFF
--- a/cmd/registry_client.go
+++ b/cmd/registry_client.go
@@ -80,7 +80,6 @@ func registerANamespace(cmd *cobra.Command, args []string) {
 	if err != nil {
 		log.Errorf("Failed to construction registration endpoint URL: %v", err)
 	}
-	// registrationEndpoint := url.JoinPath(namespaceEndpoint, "/api/v1.0/registry/register").String()
 	if prefix == "" {
 		log.Error("Error: prefix is required")
 		os.Exit(1)
--- a/cmd/registry_client.go
+++ b/cmd/registry_client.go
@@ -76,7 +76,7 @@ func registerANamespace(cmd *cobra.Command, args []string) {
 	}
 
 	// Parse the namespace URL to make sure it's okay
-	registrationEndpointURL, err := url.JoinPath(namespaceEndpoint, "api", "v2.0", "registry")
+	registrationEndpointURL, err := url.JoinPath(namespaceEndpoint, "api", "v1.0", "registry")
 	if err != nil {
 		log.Errorf("Failed to construction registration endpoint URL: %v", err)
 	}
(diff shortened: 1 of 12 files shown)


bbockelm (changes requested):
The code itself looks fine but it doesn't appear to be working with the existing osg-htc.org registry.

The following command:
```
pelican serve --module origin -d -f osg-htc.org
```

Ends with:

```
Error: Failed to determine whether namespace is already registered: Failed unmarshal namespace from response
```

haoming29:
@jhiemstrawisc Requested your review so that this can be merged before v7.4 release, given that Brian isn't available at the moment. This will block the release as it the bug breaks backward compatibility.

jhiemstrawisc:
@haoming29, sounds good. Let's focus on getting #613 in, and then I can revisit this to see what still makes sense.

jhiemstrawisc (commented):
LGTM and works with all of the things I could think to double check by hand, but I had one thing I wanted to double check.

In `server_ui/register_namespace.go`, Emma added the `X-Pelican-Prefix` header as a part of the v2.0 API and the metadata endpoint. From what I can tell, we no longer check that anywhere. Can you confirm? If that's the case, can we get rid of it before we release it as a part of 7.4?

haoming29:
> LGTM and works with all of the things I could think to double check by hand, but I had one thing I wanted to double check.
> 
> In `server_ui/register_namespace.go`, Emma added the `X-Pelican-Prefix` header as a part of the v2.0 API and the metadata endpoint. From what I can tell, we no longer check that anywhere. Can you confirm? If that's the case, can we get rid of it before we release it as a part of 7.4?

You are right. The `X-Pelican-Prefix` header was removed and I don't recall I saw it in the metadata endpoint. Although we might want to bring it back when we are ready for an API version upgrade, I link the idea to put the prefix in the header instead of doing a messy URL for various GET requests. @jhiemstrawisc

jhiemstrawisc:
@bbockelm It would be good if we could discuss at a team at some point in the near future how we should view things like adding request/response headers. I know we generally want to be cautious, but I don't know that I have the greatest sense as to when it's okay. My thinking is that relying on a header like `X-Pelican-Prefix` when sending requests from a client to another pelican service makes things less functional with curl because curl users have to know about our special headers. At the same time, Haoming is right that things are starting to get a little more complicated, and determining the intention of various `GET` requests based on the url path has become a much more nuanced task.

jhiemstrawisc (approved):
LGTM. Thanks, Haoming!

haoming29:
@bbockelm I think the merged is still blocked as you requested changes but not approved the new change. Could you approve it at some point?

bbockelm (approved):
Relying on @jhiemstrawisc's review ... just pushing the button to clear things forward.

PR #619, opened 2024-01-08
Add user role to `/auth/whoami` response

haoming29:
Add user role ("admin" for user with admin privilege or "user" if they don't) so that UI can do RBAC.


haoming29:
@CannonLock This PR should give you what you need

web_ui/authentication.go
```diff
@@ -274,6 +324,27 @@ func resetLoginHandler(ctx *gin.Context) {
 	}
 }
 
+func whoamiHandler(ctx *gin.Context) {
```
joereuss12, line 327: Maybe add comments before this function as well, just to remain consistent even though it is pretty self-explanitory

joereuss12 (changes requested):
LGTM other than this one simple change

joereuss12 (approved):
LGTM

PR #620, opened 2024-01-08
Use origin default key loc for cli token creation

### LINKED ISSUE #595: Regression in `pelican origin token create`
I tried creating a token by running `pelican origin token create <values>` and got an error that the origin's private key could not be found. I verified that the key exists and should be findable.

This appears to be related to removing `initServer` from `initOrigin()` in PR #570 because the token creation logic no longer has access to the default location of the key via `viper`.

A temporary workaround is to pass the `--private-key` flag with a path to the origin's private key.

jhiemstrawisc:
Closes #595

### DIFF
--- a/cmd/origin_token.go
+++ b/cmd/origin_token.go
@@ -1,13 +1,16 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


turetske (approved):
LGTM!

PR #624, opened 2024-01-09
Update the config page to consume {Value: <>, Type: <>} ConfigValue[]

CannonLock:


CannonLock:
@haoming29 Lets push that in a new PR for 7.5, then we can lump it with the write api potentially.

haoming29 (approved):

PR #627, opened 2024-01-09
Fixes to origin/director advertise to client

joereuss12:
These fixes are for PR #622 which helps reinstate backwards compatibility. Changes include:
- Returning `UploadFile` to its original state
- Running a PUT on the director to get PUT endpoint instead of a GET
   - Moved this logic to `client/main.go` when we gather ns info
- Moved the gathering of ns info to a separate function

### DIFF
--- a/client/director.go
+++ b/client/director.go
@@ -124,7 +124,9 @@ func CreateNsFromDirectorResp(dirResp *http.Response) (namespace namespaces.Name
 	return
 }
 
-func QueryDirector(source string, directorUrl string) (resp *http.Response, err error) {
+// Make a request to the director for a given verb/resource; return the
+// HTTP response object only if a 307 is returned.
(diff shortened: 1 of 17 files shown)


bbockelm (approved):
Did some further refactoring to reduce a double-query to the director.

Added a commit to allow "fallback reads" so I could test via federation-in-a-box as follows:

```
PELICAN_ORIGIN_ENABLEFALLBACKREAD=true \
   PELICAN_ORIGIN_ENABLEVOMS=false \
   PELICAN_ORIGIN_EXPORTVOLUME=/tmp/test \
   pelican serve --module origin --module registry --module director -d
```

and then doing a `pelican object copy` to/from the instance.  Also checks out with `stashcp` moving files to/from the OSDF.

Now that we can do an upload and download via federation-in-a-box, we should think about adding that to either the unit test or the integration tests.

There may be further unit test cleanup judging from failures on `main`; I'll leave that to @joereuss12

PR #628, opened 2024-01-10
Simple refactoring of director module

bbockelm:
- Do not export internal methods or types
- Rename `advertise` to make it clear it's only for OSDF
- Break out the registration functions from redirect.go

Do not merge until after 7.4.0 release.  There should be no functional changes, just renames and hiding symbols from outside the director package.

### DIFF
--- a/cache_ui/advertise.go
+++ b/cache_ui/advertise.go
@@ -30,8 +30,8 @@ type (
 	}
 )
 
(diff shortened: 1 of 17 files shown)


bbockelm:
Yes -- don't worry about it for now, let's make sure 7.4.0 is branched off first.

bbockelm:
Closing this out -- it was a bunch of automated refactorings/renamings, will be easier to start from scratch at this point.

PR #634, opened 2024-01-10
Fix Config Links

CannonLock:
- Use path base aware link component for page navigation


CannonLock:
@haoming29 The issue was this page also used the Material-UI Link component which is not aware of the basepath that I set.

haoming29 (approved):
The links to various server page in config page works, but I found that the link pelican icon has always leads to the "serve page option" with exception for registry page, the icon of which leads to `/view/registry/index.html`. I think we need a way to return to the "current server page" as well as going back to "serve matrix" page

I guess this could be a separate issue if @bbockelm  is fine with that.

CannonLock:
@haoming29 Thats another issue, and isn't pressing because IIRC the pelican in a box is currently internal. The other urls will get redirected to that servers home.

PR #638, opened 2024-01-10
Stash plugin abort early branch

joereuss12:
If some sort of failure occurs during download, the plugin now will immediately write the output file and exit (1 or 11 depending on if the error is retryable). Also, I was having issues with param_generator making param_struct.go have an extra new line at the bottom whenever it was generated. Go-linter does not like this so I have to `git restore` it every time I want to commit. This should fix that (hopefully it's not just me having this issue).

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -259,14 +259,35 @@ func stashPluginMain(args []string) {
 				resultAd.Set("TransferError", result.Error())
 			} else {
 				errMsg := " Failure "
-				if upload {
+				if upload { // For problem uploading, continue to give partial result
 					errMsg += "uploading "
-				} else {
+					errMsg += transfer.url + ": " + client.GetErrors()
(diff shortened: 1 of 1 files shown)


haoming29:
@joereuss12 Saw you marked this as v7.5. For the extra whitespace at the EDF, I also had the same issue but I think goreleaser also doesn't like it with a dirty commit, and it cause failure of running goreleaser in GHA. Do you think we should cherry pick that small fix and merge in before you roll out 7.4?

joereuss12:
@haoming29 Yeah we are working on that issue of a dirty commit right now before this fix is even in place. I don't think this will make too much of a difference if it is in 7.4 or 7.5. The dirty commit seems like it's unrelated after some digging.

joereuss12:
@haoming29 Nevermind, seems like you are actually correct

joereuss12:
This issue is now solved by #674

PR #639, opened 2024-01-10
Fixes to how param_struct is generated

joereuss12:
I was having issues with param_generator making param_struct.go have an extra new line at the bottom whenever it was generated. Go-linter does not like this so I have to `git restore` it every time I want to commit. This should fix that (hopefully it's not just me having this issue)


turetske (approved):
LGTM!

PR #649, opened 2024-01-12
Take a URL from config value to fetch institution list for Registry 

### LINKED ISSUE #635: Check uniqueness of institution IDs from config

### LINKED ISSUE #648: Enable admin to pass a URL for fetching institution list for registry
Currently, for a list of institution available for user to register to (used by registry UI when user register their namespace for listing options for institutions), we get it from config value `Registry.Institutions`, a yaml array of institutions. 

We want to make it more flexible by allowing admin to feed a URL with a remote endpoint providing such list.

haoming29:
Fixes #635 and Closes #648 

To test locally, you may use the topology URL which gives valid institution list for OSDF: https://topology.opensciencegrid.org/institution_ids

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -606,7 +606,7 @@ components: ["nsregistry"]
 ---
 name: Registry.InstitutionsUrl
 description: >-
-  A url to get a list of available institutions for users to regiter their namespaces to. 
+  A url to get a list of available institutions for users to register their namespaces to.
   The url must accept a GET request with 200 response in JSON/YAML content with the following format:
 
   JSON:
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -130,7 +130,6 @@ type: object
 default: none
 components: ["director"]
 ---
-
 ############################
 #     Log-Level Configs    #
 ############################
@@ -206,7 +205,7 @@ name: Federation.TopologyReloadInterval
 description: >-
   The frequency, in minutes, that topology should be reloaded.
 type: duration
-osdf_default: 10
+osdf_default: 10m
 default: none
 components: ["director", "nsregistry"]
 ---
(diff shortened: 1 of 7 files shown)


docs/parameters.yaml
```diff
@@ -600,10 +599,48 @@ description: >-
...
   ```
 
+  Note that this value will take precedence over Registry.InstitutionsUrl if both are set
 type: object
 default: none
 components: ["nsregistry"]
 ---
+name: Registry.InstitutionsUrl
+description: >-
+  A url to get a list of available institutions for users to regiter their namespaces to.
```
jhiemstrawisc, line 609: `regiter --> register`

registry/registry_ui_test.go
```diff
@@ -267,3 +279,311 @@ func TestPopulateRegistrationFields(t *testing.T) {
...
+		unique := checkUniqueInstitutions([]Institution{{ID: "1"}, {ID: "2"}})
+		assert.True(t, unique)
+	})
+
+	t.Run("duplicated-gives-false", func(t *testing.T) {
+		unique := checkUniqueInstitutions([]Institution{{ID: "1"}, {ID: "1"}})
+		assert.False(t, unique)
+	})
+
+	t.Run("large-entries", func(t *testing.T) {
```
jhiemstrawisc, line 439: What does this test check that the previous doesn't?

registry/registry_ui.go
```diff
@@ -575,3 +682,67 @@ func RegisterRegistryWebAPI(router *gin.RouterGroup) error {
...
+			} else {
+				log.Info("Institution TTL cache is nil, stop clean up process.")
+			}
+			return nil
+		})
+
+		// Try to populate the cache at the server start. If error occured, it's non-blocking
+		cachedInsts, intErr, _ := getCachedInstitutions()
+		if intErr != nil {
+			log.Warning("Failed to populate institution cache. It's non-blocking. Error: ", intErr)
```
jhiemstrawisc, line 734: I think it's redundant to point out this is non-blocking if it's already in a `log.Warning`.

jhiemstrawisc (approved):
There's one typo that should be cleaned up, and two small comments that you can choose to ignore if you want. Otherwise LGTM. Once the typo is fixed, I'll approve and merge.

registry/registry_ui_test.go
haoming29: The idea was to add some relatively "large" number of entries to make sure we are not just reading the first couple of entries to check for duplicates

PR #659, opened 2024-01-12
Add option to Logout for server Web UI

### LINKED ISSUE #630: Provide `logout` option for web UI and ensure a _clear_ logout
With #625 merged in, we now have different user roles based on user's login credentials, and at least for our own demo purposes, it would be convenient to have the option to log out current signed in user to allow us switching between different users elegantly. While we can manually delete cookies, I figured that it's very flaky to do. Some of the user credentials still exists even if cookies are manually removed, which apparently is a bug. Let's fix the bug too.

haoming29:
Closes #630 

I didn't conditionally render the logout button and in registry page people can click it without login, but it will return `401` and cause no hard. Let me know if you want me to add another request to get login status first, although I think the login status should be a global state we manage, so save some requests.

Comes with unit tests

### DIFF
--- a/web_ui/authentication.go
+++ b/web_ui/authentication.go
@@ -180,14 +180,8 @@ func setLoginCookie(ctx *gin.Context, user string) {
 		return
 	}
 
-	ctx.SetCookie("login", string(signed), 30*60, "/api/v1.0",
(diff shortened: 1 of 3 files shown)


CannonLock (approved):
LGTM

PR #660, opened 2024-01-12
Xrootd log levels now configurable

joereuss12:
Moves many of the xrootd log levels to be configurable. Set defaults in `defaults.yaml` to what we previously had set.


config/resources/defaults.yaml
```diff
@@ -17,6 +17,15 @@
...
   Level: "Error"
+  XrootdScitokensTrace: all
+  XrootdPssTrace: all
+  XrootdCmsTrace: all
+  XrootdOfsTrace: all
+  XrootdPfcTrace: info
+  XrootdXrdTrace: all -sched
+  XrootdXrootdTrace: emsg login stall redirect
+  XrootdPssSetOptCache: DebugLevel 3
+  XrootdPssSetOptOrigin: DebugLevel 1
```
matyasselmeci, line 28: The defaults are way too verbose and impact performance. Here are the defaults OSG ships with:

```suggestion
  XrootdScitokensTrace: none
  XrootdPssTrace: off
  XrootdCmsTrace: -all
  XrootdOfsTrace: -all
  XrootdPfcTrace: info
  XrootdXrdTrace: -all
  XrootdXrootdTrace: emsg login stall redirect
  XrootdPssSetOptCache: DebugLevel 1
  XrootdPssSetOptOrigin: DebugLevel 1
```

We tell people that if they want to debug they should set
```
ofs.trace all
xrd.trace all
cms.trace all
scitokens.trace all
pss.setopt DebugLevel 3
```

docs/parameters.yaml
```diff
@@ -148,6 +148,74 @@ type: filename
...
+name: Logging.XrootdPssTrace
+description: >-
+  Logging level of pss within Xrootd configuration values include: all, on, debug
+type: string
+default: all
+components: ["cache"]
+---
+name: Logging.XrootdScitokensTrace
+description: >-
+  Trace level of scitokens debug output within Xrootd configuration. Values include: all, on, debug
```
matyasselmeci, line 176:
```suggestion
  Trace level of scitokens debug output within Xrootd configuration. Values include: all, debug, info, warning, error, none
```
(from https://github.com/xrootd/xrootd/blob/8f8498d66aa583c54c0875bb1cfe432f4be040f4/src/XrdSciTokens/XrdSciTokensAccess.cc#L951-L963)

docs/parameters.yaml
```diff
@@ -148,6 +148,74 @@ type: filename
...
+type: string
+default: all
+components: ["origin", "cache"]
+---
+name: Logging.XrootdCmsTrace
+description: >-
+  Trace level of cms debug output within Xrootd configuration. Values include: all, debug, defer, files, forward, redirect, space, stage
```
matyasselmeci, line 183:
```suggestion
  Trace level of cms debug output within Xrootd configuration. Values are a combination of: all, debug, defer, files, forward, redirect, space, stage; values can be subtracted by prefixing them with `-`
```

docs/parameters.yaml
```diff
@@ -148,6 +148,74 @@ type: filename
...
+type: string
+default: all
+components: ["cache"]
+---
+name: Logging.XrootdOfsTrace
+description: >-
+  Trace level of ofs debug output within Xrootd configuration. Values include: aio, all, chmod, close, closedir, debug, delay, dir, exists,
+  fsctl, getstats, io, mkdir, most, open, opendir, qscan, read, readdir, redirect, remove, rename, sync, truncate, write
```
matyasselmeci, line 191:
```suggestion
  Trace level of ofs debug output within Xrootd configuration. Values are a combination of: aio, all, chmod, close, closedir, debug, delay, dir, exists,
  fsctl, getstats, io, mkdir, most, open, opendir, qscan, read, readdir, redirect, remove, rename, sync, truncate, write; values can be subtracted by prefixing them with `-`
```

docs/parameters.yaml
```diff
@@ -148,6 +148,74 @@ type: filename
...
+description: >-
+  Trace level of pfc debug output within Xrootd configuration. Values include: none, error, warning, info, debug, dump
+type: string
+default: info
+components: ["origin"]
+---
+name: Logging.XrootdXrootdTrace
+description: >-
+  Trace options for Xrootd debug output within Xrootd configuration. Values include: all, auth, debug, emsg, fs, fsaio, fsio, login, mem, off,
+  pgcserr, redirect, request, response, stall
```
matyasselmeci, line 206:
```suggestion
  Trace options for Xrootd debug output within Xrootd configuration. Values are a combination of: all, auth, debug, emsg, fs, fsaio, fsio, login, mem, off,
  pgcserr, redirect, request, response, stall; values can be subtracted by prefixing them with `-`
```

docs/parameters.yaml
```diff
@@ -148,6 +148,74 @@ type: filename
...
+type: string
+default: emsg login stall redirect
+components: ["origin"]
+---
+name: Logging.XrootdXrdTrace
+description: >-
+  Trace options for Xrootd debug output (xrd) within Xrootd configuration. Values include: all, conn, debug, mem, net, none, off, poll, protocol,
+  sched, tls, tlsctx, tlsio, tlssok
```
matyasselmeci, line 214:
```suggestion
  Trace options for Xrootd debug output (xrd) within Xrootd configuration. Values are a combination of: all, conn, debug, mem, net, none, off, poll, protocol,
  sched, tls, tlsctx, tlsio, tlssok; values can be subtracted by prefixing them with `-`
```

docs/parameters.yaml
```diff
@@ -148,6 +148,74 @@ type: filename
...
+description: >-
+  Configures the XRootD client used by the proxy top communicate with the cache (we use to set debug level).
+  Values include: DebugLevel (0, 1, 2, 3)
+type: string
+default: DebugLevel 3
+components: ["cache"]
+---
+name: Logging.XrootdPssTrace
+description: >-
+  Logging level of pss within Xrootd configuration values include: all, on, debug
```
matyasselmeci, line 169:
```suggestion
  Logging level of pss within Xrootd configuration values include: all, on, debug, off
```
("off" is not listed in the docs but the code seems to accept it:
https://github.com/xrootd/xrootd/blob/8f8498d66aa583c54c0875bb1cfe432f4be040f4/src/XrdOuc/XrdOucPsx.cc#L805 )

matyasselmeci (commented):
I have some opinions about the defaults; the code looks fine at a glance but I'm not familiar enough with the codebase so someone else should look at that.

bbockelm (changes requested):
A few thoughts:

1.  With @turetske working on configurations where both the cache and origin are running under the same pelican process, we're finding it's bad to have things named generically "Xrootd".  We should split it out to "Cache" and "Origin" prefixes.
2. This gets really into the nitty-gritty of XRootD logging configuration and exposes its inconsistencies.  To understand what `all -sched` does, you need to be an xrootd expert.  We use simple logging levels elsewhere -- can we keep in that paradigm?

What about a config like this:

```
Logging:
   Level: Error
   CategoryLevels:
     CategoryName1: Debug
     CategoryName2: Info
     CategoryName3: Error
```

Where we then map the different category names to different xrootd configurations.

The downside of the above approach is schema verification is tough.  An alternate would be:

```
Logging:
   Level: Error
   DebugLevel: ["CategoryName1"]
   InfoLevel: ["CategoryName2"]
   ErrorLevel: ["CategoryName3"]
```

I'm leaning toward the former scheme but open to input.

matyasselmeci:
CategoryName as in "pfc", "pss", etc.?

joereuss12:
@bbockelm @matyasselmeci assuming these tests pass, take a look at the changes. Seems to work and is more user friendly. A question I have tho is for the link below, what I should do for Origin_Xrootd since our/osg defaults are very specific? Let me know any feedback and if there are any typos (there was a lot of copy/pasting)

https://github.com/PelicanPlatform/pelican/pull/660/files#diff-7d2d0cf2c84876d1136ec4bb84b2cc80441aa374baa2083b8e91896648808671R215-R222

matyasselmeci:
Does it catch invalid log levels? It would get frustrating to have `CachePfc: degub` be silently accepted but give no additional info

config/resources/defaults.yaml
```diff
@@ -17,6 +17,18 @@
 Debug: false
 Logging:
   Level: "Error"
+  Cache:
+    Scitokens: error
+    XrootdPssSetOptCache: error
+    Pss: error
+    Cms: error
```
matyasselmeci, line 24: Other way around: cms is for origins, not caches

config/resources/defaults.yaml
```diff
@@ -17,6 +17,18 @@
 Debug: false
 Logging:
   Level: "Error"
+  Cache:
+    Scitokens: error
+    XrootdPssSetOptCache: error
+    Pss: error
```
matyasselmeci, line 23: I was thinking of having `Logging.Cache.Pss` control both `pss.trace` and `pss.setopt` for the cache. Also it looks like you're missing this from the origin?

joereuss12:
> Does it catch invalid log levels? It would get frustrating to have `CachePfc: degub` be silently accepted but give no additional info

Yes, the `else` within the mapping should set everything to the default. Should we maybe change it to an `else if <category name> == "error"` and have an `else` statement tell the user that it did not understand the config setting and set the log level to the default setting?

The statement: 
https://github.com/PelicanPlatform/pelican/blob/3ab9112ed80ae36a39d3dad84b8a4e401f6633c7/xrootd/xrootd_config.go#L631-L640

matyasselmeci:
Yes, if the user entered an unrecognized value, they should be alerted to that.

joereuss12:
Sounds good, I will make those changes!

config/resources/defaults.yaml
joereuss12: Oh okay, why is it set within the `xrootd/resources/xrootd-cache.cfg` and not `xrootd-origin.cfg`? Was this a mistake or is this how it is supposed to work?

matyasselmeci: I think that was a mistake. `cms` refers to the `cmsd` daemon, which is the component of xrootd that advertised the hosted paths to a central redirector so caches know where to pull data they don't have from. It needs to run on origins but it doesn't really make sense to run on caches because caches pull data from origins, not each other.

matyasselmeci:
@bbockelm might have ideas but I'd prefer to get this even if "info" doesn't have very good values, and get better values in a later PR.

jhiemstrawisc:
@matyasselmeci I chatted with Brian this morning, and he's short on time for a review. After Joe's most recent changes, are you happy with merging this?

matyasselmeci (approved):
LGTM. Thanks!

PR #663, opened 2024-01-16
Main test now restores current version

joereuss12:
Bug where the test changes the client version and never sets it back, causes issues with other tests using a director


haoming29 (approved):
LGTM!

PR #664, opened 2024-01-16
Fixed issue: auto-unpacking failing if dest is "."

joereuss12:
Within the auto-unpacking code, the dest was reading as just "." so I changed it to recognize that is the current working directory and change it manually


jhiemstrawisc (approved):
LGTM

Issue #654, opened 2024-01-12
Sort out config page based on relevant values


CannonLock:
Only show config parameters relevant to the server(s) running.

PR #669, opened 2024-01-17
Filter config

CannonLock:
Closes https://github.com/PelicanPlatform/pelican/issues/654

### DIFF
--- a/generate/param_generator.go
+++ b/generate/param_generator.go
@@ -154,8 +154,17 @@ func GenParamEnum() {
 	if err != nil {
 		panic(err)
 	}
-	// Create the json file to be generated (for the web ui)
+	// Create the json file to be generated (for the documentation website
 	fJSON, err := os.Create("../docs/parameters.json")
+	if err != nil {
+		panic(err)
+	}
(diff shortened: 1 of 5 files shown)


CannonLock:
@haoming29 Alright this is ready for review, copying the parameters.json was more of a adventure then it needed to be.

generate/param_generator.go
```diff
@@ -154,7 +154,7 @@ func GenParamEnum() {
 	if err != nil {
 		panic(err)
 	}
-	// Create the json file to be generated (for the web ui)
+	// Create the json file to be generated (for the documentation website
```
haoming29, line 157:
```suggestion
	// Create the json file to be generated (for the documentation website)
```

haoming29 (changes requested):
LGTM. One thing I noticed is that after I ran `make web-build`, there's a new file generated at `frontend/public/data/parameters.json` which is not added to `.gitignore`. Is this file supposed to be committed? If so, could you add it to your PR? Or it's a file for local only?

haoming29 (approved):
LGTM with changes

PR #671, opened 2024-01-17
Exit main process if error group goroutine returns error

### LINKED ISSUE #661: Cache: deal with xrootd processes dying
The xrootd process on my cache (started with `osdf cache serve`) crashed:
```
Jan 13 21:01:30 sc-cache.chtc.wisc.edu osdf[7861]: time="2024-01-13T21:01:29-06:00" level=error msg="xrootd process failed unexpectedly: signal: segmentation fault"
```
The osdf process neither exited nor did it attempt to restart xrootd. In case xrootd crashes, Pelican should exit with an error; the service manager (systemd or supervisord) will take care of restarting it if it's configured to do so.

haoming29:
Fixes #661 
* Fix the issue where Pelican won'y exit main process if error group goroutine returns error, unless an error returns in the main goroutine
* Fix the panic at the director when a cache advertise itself to the director, as we didn't have the case for cache when doing version compatibility check
* Improve logging at program exit to be consistent.
* Improve cache error group, remove duplicated exit handler that `root.go` already had.

~~However, this didn't fix the issue where when we exit the program with `ctrl+c` , xrootd reported error as it treated SIGINT similar as SIGSEGV. We might have to refactor some of the code in `daemon/launch_unix.go`. Created a separate issue for this: #672~~

### DIFF
--- a/cmd/cache_serve.go
+++ b/cmd/cache_serve.go
@@ -24,9 +24,6 @@ import (
 	"context"
 	"encoding/json"
 	"net/url"
-	"os"
-	"os/signal"
-	"syscall"
 	"time"
 
 	"github.com/pelicanplatform/pelican/cache_ui"
@@ -87,31 +84,15 @@ func serveCache(cmd *cobra.Command, _ []string) error {
 
 func serveCacheInternal(cmdCtx context.Context) error {
 	// Use this context for any goroutines that needs to react to server shutdown
-	ctx, shutdownCancel := context.WithCancel(cmdCtx)
-
-	err := config.InitServer(ctx, config.CacheType)
+	err := config.InitServer(cmdCtx, config.CacheType)
 	cobra.CheckErr(err)
 
-	egrp, ok := ctx.Value(config.EgrpKey).(*errgroup.Group)
+	egrp, ok := cmdCtx.Value(config.EgrpKey).(*errgroup.Group)
 	if !ok {
 		egrp = &errgroup.Group{}
 	}
 
-	egrp.Go(func() error {
-		log.Debug("Will shutdown process on signal")
-		sigs := make(chan os.Signal, 1)
-		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
-		select {
-		case sig := <-sigs:
-			log.Infof("Received signal %v; will shutdown process", sig)
-			shutdownCancel()
-			return nil
-		case <-ctx.Done():
-			return nil
-		}
-	})
-
-	err = xrootd.SetUpMonitoring(ctx, egrp)
+	err = xrootd.SetUpMonitoring(cmdCtx, egrp)
 	if err != nil {
 		return err
 	}
@@ -132,11 +113,11 @@ func serveCacheInternal(cmdCtx context.Context) error {
 
 	viper.Set("Origin.NamespacePrefix", cachePrefix)
 
-	if err = server_ui.RegisterNamespaceWithRetry(ctx, egrp); err != nil {
+	if err = server_ui.RegisterNamespaceWithRetry(cmdCtx, egrp); err != nil {
 		return err
 	}
 
-	if err = server_ui.LaunchPeriodicAdvertise(ctx, egrp, []server_utils.XRootDServer{cacheServer}); err != nil {
+	if err = server_ui.LaunchPeriodicAdvertise(cmdCtx, egrp, []server_utils.XRootDServer{cacheServer}); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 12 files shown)


matyasselmeci (commented):
Some tweaks; also the changes in `director/redirect.go` seem unrelated to this issue and should be in a separate PR.

cmd/cache_serve.go
```diff
@@ -146,43 +127,44 @@ func serveCacheInternal(ctx context.Context) error {
...
-	go func() {
-		if err := web_ui.RunEngine(ctx, engine, egrp); err != nil {
-			log.Panicln("Failure when running the web engine:", err)
+	egrp.Go(func() error {
+		if err := web_ui.RunEngine(cmdCtx, engine, egrp); err != nil {
+			log.Errorln("Failure when running the web engine:", err)
+			return err
+		} else {
+			return err
 		}
```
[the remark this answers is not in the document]
bbockelm: That refactor has slightly different semantics.

In the original, the `err` object is local to the conditional block.

In your refactor, it uses the `err` object from the containing function, meaning that multiple routines might be using the same object at the same time, resulting in a potential (but admittedly unlikely) race condition).

One could do this:

```
	egrp.Go(func() (err error) {
		if err = web_ui.RunEngine(cmdCtx, engine, egrp); err != nil {
			log.Errorln("Failure when running the web engine:", err)
		}
		return err
	})
```

cmd/cache_serve.go
```diff
@@ -146,43 +127,44 @@ func serveCacheInternal(ctx context.Context) error {
...
-		shutdownCancel()
```
bbockelm, line 157: This cancel (and the equivalent one in launchers) needs to be kept around.

The web engine may fail to startup and this `shutdownCancel()` call may be the thing that triggers the shutdown originally.

cmd/root.go
```diff
@@ -78,18 +78,17 @@ func (i *uint16Value) Type() string {
...
+	egrp, egrpCtx := errgroup.WithContext(context.Background())
 	defer func() {
 		err := egrp.Wait()
 		if err != nil {
-			if err.Error() == "Federation process has been cancelled" {
-				log.Info("Process was shutdown")
-				return
-			}
-			log.Errorln("Error occurred when shutting down process:", err)
+			log.Errorln("Fatal error occured that leads to shutdown of the process:", err)
```
bbockelm, line 85:
```suggestion
			log.Errorln("Fatal error occurred that lead to the shutdown of the process:", err)
```

cmd/cache_serve.go
```diff
@@ -85,33 +82,17 @@ func serveCache(cmd *cobra.Command, _ []string) error {
...
+	err := config.InitServer(cmdCtx, config.CacheType)
 	cobra.CheckErr(err)
 
-	egrp, ok := ctx.Value(config.EgrpKey).(*errgroup.Group)
+	egrp, ok := cmdCtx.Value(config.EgrpKey).(*errgroup.Group)
 	if !ok {
 		egrp = &errgroup.Group{}
 	}
 
-	egrp.Go(func() error {
```
bbockelm, line 100: I don't like removing the signal handler from here.  It means that the only way signals are propagated to shutdown Pelican are when xrootd has already been started, pelican forwards the signal to xrootd, and xrootd decides to shutdown.  That's a fairly brittle combination.

cmd/root.go
```diff
@@ -78,18 +78,17 @@ func (i *uint16Value) Type() string {
...
+		} else {
+			// Use Error instead of Info because our default log level is Error
+			log.Error("Pelican is safely exited")
 		}
 	}()
-	ctx := context.WithValue(context.Background(), config.EgrpKey, &egrp)
+	ctx := context.WithValue(egrpCtx, config.EgrpKey, egrp)
 	return rootCmd.ExecuteContext(ctx)
```
bbockelm, line 92: This is a related bug (not caused by your PR but made it obvious).

If `serveCache` fails with an error (that is, the cache fails to startup), then nothing cancels the context, causing the process to hang indefinitely.

Contrast this to the `serveOrigin` process which wraps everything with a non-nil-error check:

```
func serveOrigin(cmd *cobra.Command, args []string) error {
	cancel, err := launchers.LaunchModules(cmd.Context(), config.OriginType)
	if err != nil {
		cancel()
	}

	return err
}
```

We either want to wrap `cacheServe` similarly or hoist the cancelation all the way up to here.

cmd/cache_serve.go
haoming29: I thought we had this logic in `cmd/root.go` which is global to all commands we run. Isn't here a duplicated code?

haoming29: No you are right. I got myself confused by the one in the `launcher.go` but cache wasn't launched by that route.

cmd/cache_serve.go
turetske: @haoming29 Waiting on @bbockelm to answer this before I do my review, as this was my only concern as well.

bbockelm (approved):
LGTM with latest revisions.

@joereuss12 - I believe you were the 7.4.0 release manager, right?  If so, can you please backport this PR to that branch and do a bugfix release for @matyasselmeci?

joereuss12:
Just cut a bugfix release!

PR #673, opened 2024-01-18
Add `pre-commit` to dev container and GHA

### LINKED ISSUE #614: Add `pre-commit` package to dev container
Depending on what's the usage of our dev container, we might want to install `pre-commit` package on the dev container as this is what we used for commit linter and formatting. For the current dev container, `pre-commit` is missing and all pre-commit hooks are not working by default, if developers are developing with dev container.

Also, it seems that to install and use `pre-commit`, python is a required dependency. Adding python to the dev container may significantly increase the side of the image, and we might need to do some trade-off here.

haoming29:
Closes #614 

We have `pre-commit` configuration in our repo already, but we didn't enforce the installation and usage anywhere in our codebase which causes golinter failure that's supposed to be alerted when we commit the code. This PR installed `pre-commit` in the dev container so that it's enabled by default for people developing using dev container.

Other improvements introduced in this PR:
* Improved GHA linter to incorporate `pre-commit` as part of the linter check
* Runs `pre-commit` against all files in the repo to update the format.


images/dev.Dockerfile
```diff
@@ -140,9 +140,15 @@ RUN if [ "$TARGETARCH" = "amd64" ]; then \
...
+# Install pre-commit
+RUN yum install -y python39 &&\
+    /bin/pip-3.9 install pre-commit
+
+COPY ./images/dev-container-entrypoint.sh /usr/local/bin/
+
 WORKDIR /app
 
-CMD ["/bin/bash"]
+ENTRYPOINT ["dev-container-entrypoint.sh"]
```
jhiemstrawisc, line 151: Why not just do the `pre-commit install` in the Dockerfile and skip having a `CMD` or `ENTRYPOINT`?

jhiemstrawisc (approved):
One small comment that may or may not warrant a change, depending on what you think @haoming29. Otherwise LGTM. Thanks for tidying this up!

images/dev.Dockerfile
haoming29: If you mean by `RUN pre-commit install` at the end of the Dockerfile, then the answer would be No, because `pre-commit install` installs hooks to `.git` folder, which needs to be ran after the container binds the project folder and I think the only way is to add to the entrypoint script so that the command runs when the container has started.

jhiemstrawisc: Ahh, makes sense. Thanks for the clarification.

PR #674, opened 2024-01-18
Have pelican plugin move files in parallel

joereuss12:
Now the plugin can move files just like handle_http does, starts up many workers and has a producer/consumer setup to download files in parallel.

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -210,10 +210,10 @@ func stashPluginMain(args []string) {
 
 	var wg sync.WaitGroup
 
-	workChan := make(chan Transfer)
+	workChan := make(chan Transfer, len(transfers))
 	results := make(chan *classads.ClassAd, len(transfers))
 
-	// start workers
+	// Start workers
 	for i := 1; i <= 5; i++ {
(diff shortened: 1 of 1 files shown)


joereuss12:
This is still a work in progress so setting as a draft.

joereuss12:
Smoke tests seem to show that this is working how it should, ready for review.

bbockelm:
Haven't had time for a full review but to provide some early feedback:

https://github.com/PelicanPlatform/pelican/pull/674/files#diff-417e18b3b05bac37d2ebbc44d3be121f6026e5e4ab13e6400d0eb879027c5160R232-R240

This is approximately the design of the current parallel download when moving a directory.  I want to improve things, however, by having the result of individual transfers be returned on a channel.  That means there's a single for-loop that's both writing new tasks and reading results.

For example of what this would give us: right now, there's a hard synchronization -- all transfers must complete before the overall process can fail.  We know that we instead will want to fail all of them after the first transfer fails in the case of inputs.

jhiemstrawisc (approved):
LGTM

PR #675, opened 2024-01-18
Add support for custom registration fields for registry

haoming29:


docs/parameters.yaml
```diff
@@ -655,6 +655,43 @@ type: object
 default: none
 components: ["nsregistry"]
 ---
+name: Registry.InstitutionsUrl
```
turetske, line 658: Has this been checked with @bbockelm? Last I heard is he wanted institutions to be something set via the web interface, but my information could be out of date.

registry/registry_db.go
```diff
@@ -676,7 +765,7 @@ func deleteNamespace(prefix string) error {
 }
 
 func getAllNamespaces() ([]*Namespace, error) {
-	query := `SELECT id, prefix, pubkey, identity, admin_metadata FROM namespace ORDER BY id ASC`
+	query := `SELECT id, prefix, pubkey, identity, admin_metadata, custom_fields FROM namespace ORDER BY id ASC`
```
turetske, line 768: Will there be any backwards compatibility issues with the getAllNamespacesHandler returning namespaces with custom fields if called by something not expecting the custom fields?

turetske (changes requested):

registry/registry_db.go
haoming29: I can't think of the issue. If an older version of director/client is request all namespaces and we return extra fields they don't expect, then the fields won't be unmarshalled to their structs and no errors should be reported.

registry/registry_db.go
```diff
@@ -701,6 +745,21 @@ func getAllNamespaces() ([]*Namespace, error) {
...
+			if err := json.Unmarshal([]byte(customRegFieldsStr), &ns.CustomFields); err != nil {
+				return nil, err
+			}
+		}
+		// By default, JSON unmarshall will convert any generic number to float
+		// and we only allow integer in custom fields, so we convert them back
+		for key, val := range ns.CustomFields {
+			switch v := val.(type) {
+			case float64:
+				ns.CustomFields[key] = int(v)
```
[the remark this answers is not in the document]
haoming29: Based on [this](https://stackoverflow.com/questions/40821855/do-go-switch-cases-fallthrough-or-not) stackoverflow post, Go can't do automatic fallthrough. Adding `fallthrough` gives me error that type switch can't do fallthrough.

docs/parameters.yaml
haoming29: Sorry I think I based this PR off from my earlier PR that's already merged in main, so there's extra commit diff. But yes, that InstitutionUrl is a requested feature from Brian.

registry/registry_db.go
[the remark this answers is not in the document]
turetske: Ah, the perils of knowing too many languages. My brain mixed them up.

turetske (approved):
LGTM!

PR #678, opened 2024-01-19
Project name added to user-agent

joereuss12:
For downloads, if the `ProjectName` is present in a job ad for the plugin, it is added to the user-agent in the http request header.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -353,7 +353,7 @@ func download_http(sourceUrl *url.URL, destination string, payload *payloadStruc
 	// Start the workers
 	for i := 1; i <= 5; i++ {
 		wg.Add(1)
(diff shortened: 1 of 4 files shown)


joereuss12:
@bbockelm I currently have it so that the user-agent is set to the project name if it is present for downloads. I do have two questions:
1. Should I also add this to the header for uploads too on top of downloads?
2. I cannot seem to find this `ProjectName` as a jobad attribute from when running jobs and even in the condor manual, I'm not sure how it is populated, is this a custom classad?

jhiemstrawisc (commented):
LGTM, and we tested by hand together. I still can't explain the weird issue with the RegEx, but I'm okay with saying "this seems to work and there's a comment explaining why we made the choice."

jhiemstrawisc (approved):

PR #683, opened 2024-01-24
Fix linter failure on main

haoming29:
The latest merge #497 causes linter failure on main branch, due to outdated local branch. This fixes the linter issue.


CannonLock (approved):
LGTM

Issue #681, opened 2024-01-23
Pelican in production Docker image has the wrong version number


haoming29:
From @brianhlin, if you run the version command inside the production Docker container, you'll get:

```console
[root@hostname pelican]# ./osdf-client --version
Version: 7.4.1-next
Build Date: 2024-01-10T22:48:53Z
Build Commit: 3caae35f06cd96ae97c2afca51a470b98c060010
Built By: goreleaser
```

while the tag this container was built for was `7.4.0`.

This is because we use `--snapshot` flag when running goreleaser in the production image build. We should remove this flag to ensure the version matches git tag for production build.

williamnswanson:
It seems that there was something wrong with my fix, as this issue still exists in the 7.5.4 container:

```
sh-4.4# ./osdf-client --version
Version: 7.5.5-next
Build Date: 2024-02-09T14:45:28Z
Build Commit: 71053e4b6852a7eda4422131d06fffcf8b0e981f
Built By: goreleaser
```

brianhlin:
@haoming29 what's the Pelican dev policy for closing issues that have been fixed in `main`? Should they be closed as PRs are merged or do we leave the issues open until there's been a release with the fix?

haoming29:
> @haoming29 what's the Pelican dev policy for closing issues that have been fixed in `main`? Should they be closed as PRs are merged or do we leave the issues open until there's been a release with the fix?

We link the issue with the PR, so it closes when the PR is merged in. This one is special because we re-opened it after the initial fix, but I think we should be good to close this one now.

PR #684, opened 2024-01-24
Remove --snapshot tag from goreleaser command in Dockerfile

williamnswanson:
[For Pelican Issue #681](https://github.com/PelicanPlatform/pelican/issues/681)

### DIFF
--- a/.github/workflows/publish-container.yml
+++ b/.github/workflows/publish-container.yml
@@ -95,3 +95,5 @@ jobs:
           file: ./images/Dockerfile
           push: ${{ github.event_name != 'pull_request' }}
           tags: "${{ steps.generate-tag-list.outputs.taglist }}"
+          build-args: |
+            IS_PR_BUILD=${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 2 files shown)


haoming29:
Sorry about the linter changes and causes failure on your branch. Could you rebase against the latest main? The linter fix is in place.

haoming29 (approved):
LGTM

haoming29:
@williamnswanson  Looks like the build failed due to missing GitHub tokens. Looking at the doc, it seems that running `goreleaser` by default will build and try to release the build, which we don't want. We only want to build the project locally in the container. So we should use `goreleaser build -- clean`

haoming29:
Ahh. That GHA runs on both PR and release with a tag. For PR, we don't have a tag attached to the branch, and goreleaser is unhappy about it. It's going to be a bit tricky for docker build to be compatible with both PR build and release build. You'll need to pass `${{ github.event_name != 'pull_request' }}` from GHA that build the prod image to Dockerfile.

williamnswanson:
Let me know if you want the conditional in the Dockerfile to be arranged differently.

haoming29:
> Let me know if you want the conditional in the Dockerfile to be arranged differently.

LGTM! Thanks for taking care of it.

PR #691, opened 2024-01-25
Attempt to fix duplicated director file transfer tests

### LINKED ISSUE #689: Director issued excessive file transfer test to origins
As the federation runs for a while ( >1h), the origin seems to receive high volume of file transfer tests from the director. It receives 10 folds of tests per 15 second, which is not the intended behavior.

haoming29:
Attempt to fix #689 

The core fix is an additional check here:

```go
if _, ok := healthTestCancelFuncs[sAd]; ok {
			// If somehow we didn't clear the key, we call cancel first before
			// adding a new test cycle
			healthTestCancelFuncs[sAd]()
		}
```


docs/parameters.yaml
```diff
@@ -556,6 +556,20 @@ root_default: /var/cache/pelican/maxmind/GeoLite2-City.mmdb
...
 components: ["director"]
 ---
+name: Director.AdvertisementTTL
+description: >-
+  The time to live (TTL) of director's internal cache to store origins and caches advertisement.
+type: duration
+default: 15m
+components: ["director"]
+---
+name: Director.FileTransferInterval
```
jhiemstrawisc, line 566: I think we should name this something like `Director.OriginHealthTestFreq`, or a name that highlights the service we're testing. While the description here is good, if someone comes across the config option in a config file without context, it might not be clear what this does.

jhiemstrawisc (changes requested):
One small change to a config name, otherwise things look sensible to me. I know we didn't have a super great way to test that this fixes the issues Matyas was having with the director, but I think the extra safety you added by double checking if an uncached server ad still has an uncancelled context floating around is a good thing to double check anyway.

docs/parameters.yaml
haoming29: Agree. For upstream compatibility, let's name it `OriginCacheHealthTestInterval` as we are about to add test to caches.

jhiemstrawisc (approved):
LGTM!

PR #692, opened 2024-01-25
Add config param to change storage server registration retry delay

### LINKED ISSUE #668: Allow changing registration retry delay
If my cache or origin fails to register with the namespace service, I would like to retry (much) less frequently than every 10 seconds because otherwise it is just noise in the logs.

haoming29:
Closes #668


joereuss12 (approved):
LGTM!

PR #695, opened 2024-01-26
Add port to cache namespace

### LINKED ISSUE #662: Allow multiple cache instances to register if they run on different ports
xcache-based osdf caches have two instances running: an unauthenticated instance (on port 8000) and an authenticated instance (on port 8443). When I try to do the same thing using Pelican, I get the following error from one of the instances:

> Failed to register with namespace service: Failed to register prefix /caches/CHTC_STASHCACHE_CACHE: Failed to make request: server encountered an error during key-sign challenge commit failed: The prefix already is registered

Perhaps include the port number in the prefix that gets registered? `/caches/CHTC_STASHCACHE_CACHE:8000` ?

haoming29:
Closes #662

### DIFF
--- a/cmd/cache_serve.go
+++ b/cmd/cache_serve.go
@@ -23,6 +23,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/url"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


haoming29:
Closed upon agreement at standup on 1/26

PR #706, opened 2024-01-29
Fix bad topology reload interval in Registry

### LINKED ISSUE #694: OSDF registry topology fetch has bad time interval
I made a mistake! The registry doesn't reload topology at the correct interval. [This](https://github.com/PelicanPlatform/pelican/blob/4b91a1f809ae911a773c96490fe779126202e491/registry/registry_db.go#L848) line should just load and apply the value from `GetDuration()` instead of multiplying by a minute.

jhiemstrawisc:
Closes #694

### DIFF
--- a/registry/registry_db.go
+++ b/registry/registry_db.go
@@ -845,7 +845,7 @@ func PopulateTopology() error {
 
 func PeriodicTopologyReload() {
 	for {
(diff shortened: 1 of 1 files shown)


joereuss12 (approved):
LGTM

PR #707, opened 2024-01-29
Use `pelican://` protocol for pss.origin setting in the cache

bbockelm:
This enables the `pelican://` protocol for the cache's `pss.origin` setting.

Contains four parts:
- Setting the `pss.origin` xrootd configuration variable directly
- To avoid having to touch system files, creating a separate client plugins directory and point the cache at the new directory via environment variable.
- Extending the "launcher" setup to take environment variable overrides
- Updating the OSX GitHub builder scripts to patch in `pelican://` support for XRootD.  For Linux, I assume we'll pull in the patched version via the OSG build.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,7 @@ repos:
     rev: v3.2.0
     hooks:
     -   id: trailing-whitespace
+        exclude: github_scripts/pelican_protocol.patch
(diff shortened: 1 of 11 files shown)


bbockelm:
Yes -- hand tested on both Linux and Mac (that debugging actually led to the last commit).

turetske (approved):

PR #709, opened 2024-01-29
Fix origin export options to allow Xrootd.Mount/Origin.NamespacePrefix combo

### LINKED ISSUE #621: launching origin fails if Origin.ExportVolume is not specified
I'm trying to launch an origin but `osdf origin serve` fails complaining that Origin.ExportVolume is not defined:
https://github.com/PelicanPlatform/pelican/blob/adfcce273338944cfa86107297534325a24c75f7/launchers/launcher.go#L101-L103

However, the code for setting up an origin also allows using the combination of Xrootd.Mount and Origin.NamespacePrefix instead of Origin.ExportVolume: https://github.com/PelicanPlatform/pelican/blob/adfcce273338944cfa86107297534325a24c75f7/xrootd/xrootd_config.go#L159-L169

Can you make it so that the launcher allows that combination too?

jhiemstrawisc:
To specify the local export and namespace prefix when serving an origin, both `Origin.ExportVolume` or a combination of `Xrootd.Mount` and `Origin.NamespacePrefix` are valid options. There was a bug that prevented the origin from spinning up in the case that `Origin.ExportVolume` was not directly specified, or populated via the command line volume mount flag. This fix allows either combination to work.

Closes #621

### DIFF
--- a/launchers/launcher.go
+++ b/launchers/launcher.go
@@ -98,8 +98,24 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
 		mode := param.Origin_Mode.GetString()
 		switch mode {
 		case "posix":
-			if param.Origin_ExportVolume.GetString() == "" {
-				return shutdownCancel, errors.Errorf("Origin.ExportVolume must be set in the parameters.yaml file.")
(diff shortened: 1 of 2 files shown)


matyasselmeci (approved):
LGTM. Thanks!

PR #713, opened 2024-01-29
Allow for public-read origins

jhiemstrawisc:
Previously, Pelican origins were hard-coded to require tokens. While the authfile could be modified post-fact to allow public reads, this wasn't a capability within Pelican itself, and the director wouldn't return proper information on queries to the namespace.

This commit adds "Origin.EnablePublicReads" to turn on public reads for the origin's namespace prefix. NOTE that pretty soon we're moving away from single-prefix origins. When we get to that point, there will be more work to do here, but for now I think this gets the job done, and is in line with the single-prefix suggestions for origins from https://docs.google.com/document/d/1W-2s5a_wj3Eq7URR4t7cpvQXXY0ixzGP6tUHbtKG_mc/edit


jhiemstrawisc:
@turetske, I still need to add some kind of a test for this, but I wasn't sure the best place to stick one. Do you think `xrootd/origin_test.go` is the right place, or should that stay focused on "does it run" as opposed to "can I get files from it"?

turetske:
> @turetske, I still need to add some kind of a test for this, but I wasn't sure the best place to stick one. Do you think `xrootd/origin_test.go` is the right place, or should that stay focused on "does it run" as opposed to "can I get files from it"?

@jhiemstrawisc I think that's the correct place. The only reason it wasn't "can I get files from it" is because we weren't making them public in the first place.

turetske (changes requested):
If you could add a test that just checks to make sure the authfile generated is what is expected in the xrootd authorization tests, that would be good.

xrootd/authorization_test.go
```diff
@@ -295,9 +295,8 @@ func TestWriteOriginAuthFiles(t *testing.T) {
 
 	originAuthTester := func(server server_utils.XRootDServer, authStart string, authResult string) func(t *testing.T) {
 		return func(t *testing.T) {
-
+			defer viper.Reset()
```
turetske, line 298: We don't want to just defer it, do we? It was at the top to ensure that there is a clean viper to begin with, just in case it wasn't reset in another test.

turetske (changes requested):

xrootd/authorization_test.go
jhiemstrawisc: If we do a `viper.Reset()` at the beginning of this, there's no way for me to set the viper variables that need to be configured for `EmitAuthfile` to get the right output.

I'll add an extra `viper.Reset()` at the beginning of the overall test, then we're guaranteed that the environment between these auth tests is clean.

turetske (approved):
LGTM!

PR #714, opened 2024-01-29
Expanding plugin error reporting from plugin to starter

joereuss12:
This is a work in progress still but current changes include:
- Populating the new `DeveloperData` classad created by Todd M with most of the fields specified in the document of the issue
- Changing some inner-workings of the client, so that we can pass all relevant information back to the plugin
- Classads package can now decipher types of map[string]interface{} for nested classAds

Work still needed:
- Still need to add a few more classads with more info on each attempt
- More testing by hand is needed, to ensure functionality and backwards compatibility
- Testing with Todd M's latest merge to HTCondor to ensure functionality there
- Need to figure out how/what to return for classads on uploads

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -244,8 +244,19 @@ func NewTransferDetails(cache namespaces.Cache, opts TransferDetailsOptions) []T
 }
 
 type TransferResults struct {
-	Error      error
-	Downloaded int64
+	Error           error
+	TransferedBytes int64
+	Attempts        []Attempt
+}
+
+type Attempt struct {
+	Number            int    // indicates which attempt this is
+	TransferFileBytes int64  // how much each attempt downloaded
+	TimeToFirstByte   int64  // how long it took to download the first byte
+	TransferEndTime   int64  // when the transfer ends
+	Endpoint          string // which origin did it use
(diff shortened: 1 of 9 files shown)


client/handle_http.go
```diff
@@ -264,15 +275,14 @@ func GenerateTransferDetailsUsingCache(cache CacheInterface, opts TransferDetail
 	return nil
 }
 
-func download_http(sourceUrl *url.URL, destination string, payload *payloadStruct, namespace namespaces.Namespace, recursive bool, tokenName string) (bytesTransferred int64, err error) {
-
+// func download_http(sourceUrl *url.URL, destination string, payload *payloadStruct, namespace namespaces.Namespace, recursive bool, tokenName string) (bytesTransferred int64, err error, attempts []Attempt) {
```
joereuss12, line 278: Missed this on accident, it is removed

joereuss12:
Currently not able to work on a submit AP however, I am opening this up for review

joereuss12:
Just fixed merge conflicts

jhiemstrawisc (approved):
LGTM. Let's wait for the upstream build dependency to get fixed up and when the test can pass, I'll merge.

PR #716, opened 2024-01-29
Update advertisement

turetske:
Updated the NamespaceAd and the OriginAd to the V2 version of the advertisements.

The director response to querying the cache now returns multiple issuers
There is a V1 to V2 conversion and vice versa as needed to account for backwards compatibility


haoming29 (changes requested):

cmd/cache_serve.go
```diff
@@ -46,26 +46,39 @@ import (
...
+			return respNS, err
+		}
+		respData, err = utils.MakeRequest(directorNSListEndpointURL, "GET", nil, nil)
+		if err != nil {
+			if jsonErr := json.Unmarshal(respData, &respNS); jsonErr == nil { // Error creating json
+				return respNS, errors.Wrapf(err, "Failed to make request: %v", err)
+			}
+			return respNS, errors.Wrap(err, "Failed to make request")
+		}
+	} else {
```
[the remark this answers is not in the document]
turetske: Had an incorrect nesting on the if-else statement. Thanks for the catch.

cmd/cache_serve.go
```diff
@@ -46,31 +47,55 @@ import (
...
-		if jsonErr := json.Unmarshal(respData, &respNS); jsonErr == nil { // Error creating json
-			return respNS, errors.Wrapf(err, "Failed to make request: %v", err)
+		if strings.Contains(err.Error(), "404") {
+			directorNSListEndpointURL, err = url.JoinPath(directorEndpoint, "api", "v1.0", "director", "listNamespaces")
+			if err != nil {
+				return respNS, err
+			}
+			respData, err = utils.MakeRequest(directorNSListEndpointURL, "GET", nil, nil)
+			var respNSV1 []director.NamespaceAdV1
+			if err != nil {
```
haoming29, line 76: If there _is_ an error from `MakeRequest` to `v1.0` endpoint, why do we still want to unmarshal the response data into `respNSV1`? Wouldn't that simply causes an empty `respNSV1` or even json unmarshal error? Or `/v1.0/director/listNamespaces` endpoint _does_ return namespaces even if the response status code != 200?

cmd/cache_serve.go
```diff
@@ -46,26 +46,39 @@ import (
...
 
 	respData, err := utils.MakeRequest(directorNSListEndpointURL, "GET", nil, nil)
 	if err != nil {
-		if jsonErr := json.Unmarshal(respData, &respNS); jsonErr == nil { // Error creating json
+		directorNSListEndpointURL, err = url.JoinPath(directorEndpoint, "api", "v1.0", "director", "listNamespaces")
+		if err != nil {
+			return respNS, err
+		}
+		respData, err = utils.MakeRequest(directorNSListEndpointURL, "GET", nil, nil)
+		if err != nil {
```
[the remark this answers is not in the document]
haoming29: But with your last change, here `if err != nil` means that we got an error from requesting `/v2.0/api/director/listnamespaces` and the error is not `404`. And in line 89, we still try to unmarshal the body into `respNS` which is `[]director.NamespaceAdV2`. Isn't this going to give us an empty `respNS`? or simply cause json unmarshal error?

haoming29 (changes requested):
See comments for changes requested

cmd/cache_serve.go
turetske: Ah, I see, line 77 is a holdover from old code that I just moved within the if/else statement. It might have been incorrect to begin with. I'll remove it.

director/redirect.go
```diff
@@ -440,14 +473,14 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType ServerTy
...
-					sType, ad.Name, namespace.Path)
+					sType, adV2.Name, namespace.Path)
 				ctx.JSON(http.StatusForbidden, gin.H{"error": "Authorization token verification failed. Token missing required scope"})
 				return
 			}
 		}
 	} else {
 		token := strings.TrimPrefix(tokens[0], "Bearer ")
-		prefix := path.Join("caches", ad.Name)
+		prefix := path.Join("caches", adV2.Name)
```
haoming29, line 483:
```suggestion
		prefix := path.Join("/caches", adV2.Name)
```

haoming29: Looking at git blame I think this causes error because of Justin's recent PR #583 that updated `GetNSIssuerURL` to check if the prefix comes with a slash. It returns error if it's not. So not directly related to your change

haoming29 (approved):
Already, after spend some time poking around cache and client, I managed to do the E2E test and things appear to work, so LGTM!

PR #717, opened 2024-01-29
Improve test coverage for various registry UI API functions and bug fixes along the way

### LINKED ISSUE #611: Improve test coverage for registry API
The new registry UI API comes with test suite for db access functions but not API handlers, which added layers of complexity upon the db access functions. There's only a limited amount of test coverage for those functions now and we want to improve the coverage to an amount where each handler should be at least tested against common cases.

haoming29:
Closes #611 

Current test coverage for `registry_ui` is 71%, was 30%


joereuss12 (approved):
LGTM, very thorough and I couldn't think of any more test cases. Glad we are getting more tests!

PR #719, opened 2024-01-30
Overhaul token creation to use util token cfg interface

### LINKED ISSUE #159: Standardize token creation/verification throughout Pelican

jhiemstrawisc:
Our tokens have been all over the place, and this constitutes another step toward standardizing them. Now, all of the tokens we create (aside from a few unit tests where we intentionally create some bad tokens) use the utils package. Not only does this bring us closer to WLCG-conformant tokens, but it also guarantees we always assign a JTI, among other things.

One thing we still need to do is go through and really think about the particular values we use when assigning Audience, Issuer, and Subject fields.

Closes #159


jhiemstrawisc:
@haoming29, this is failing the linter because of the generated `token_scopes.go` file. The problem is the spacing for the block of constants at the start of the file, but it doesn't look like they're currently formatted in the file on main. I think trying to template out the spacing the linter wants would be... not the best use of time. Do you have any idea how to turn off the linter for these generated files?

haoming29:
@jhiemstrawisc Updated our golangci configuration file to exclude generated code and enable full-file run for linter (this is why the issue wasn't being captured locally by `pre-commit` hook)

haoming29:
hooray! It's finally working now.

client/handle_http_test.go
```diff
@@ -389,41 +386,36 @@ func generateFileTestScitoken() (string, error) {
...
+	scopes = append(scopes, modScope)
+
+	fTestTokenCfg := utils.TokenConfig{
+		TokenProfile: utils.WLCG,
+		Lifetime:     time.Minute,
+		Issuer:       issuerUrl,
+		Audience:     []string{param.Origin_Url.GetString()},
+		Version:      "1.0",
+		Subject:      "origin",
+		Claims:       map[string]string{"scope": token_scopes.GetScopeString(scopes)},
```
haoming29, line 409: For all the scope claim, `TokenConfig` has two built-in functions to handle it, one is `AddScopes`, which is to add an array of our pre-defined token scopes in`token_scopes.TokenScope`, the other is `AddRawScope`, which is to add any other scopes we didn't define. I'd suggest we use those two functions wherever possible to be consistent.

director/origin_api_test.go
```diff
@@ -69,48 +69,38 @@ func TestVerifyAdvertiseToken(t *testing.T) {
...
-	assert.NoError(t, err)
-
-	scopelessTok, err := jwt.NewBuilder().
-		Issuer("").
-		Audience([]string{"director.test"}).
-		Subject("origin").
-		Build()
+	//Create token without a scope - should return an error upon validation
+	scopelessTokCfg := utils.TokenConfig{
+		TokenProfile: utils.WLCG,
```
haoming29, line 74: For cases where we haven't used `scitoken` or `WLCG` as the token profile, I'd suggest using `utils.None` as the token profile now unless we want to take this chance to enforce all the token in Pelican to adhere to a certain token profile, which I feel should be documented somewhere to let people know

generate/scope_generator.go
```diff
@@ -98,8 +104,12 @@ func GenTokenScope() {
 	defer f.Close()
 
 	err = tokenTemplate.Execute(f, struct {
-		Scopes []ScopeName
-	}{Scopes: scopes})
+		Scopes        []ScopeName
```
haoming29, line 107: Ah, I think the reason why the linter reported error is probably that you didn't have `pre-commit` installed locally. This error might be buried in other linter errors from the generated code, but one error among them is that this line of code wasn't being `gofmt -ed with -s`. Could you try install `pre-commit` locally?

web_ui/authentication.go
```diff
@@ -151,36 +152,28 @@ func GetUser(ctx *gin.Context) (string, error) {
...
-		ctx.JSON(http.StatusInternalServerError, gin.H{"error": "Failed to build token"})
-		return
+
+	loginCookieTokenCfg := utils.TokenConfig{
+		TokenProfile: utils.WLCG,
+		Lifetime:     30 * time.Minute,
+		Issuer:       param.Server_ExternalWebUrl.GetString(),
+		Audience:     []string{param.Server_ExternalWebUrl.GetString()},
+		Version:      "1.0",
+		Subject:      "user",
```
haoming29, line 163:
```suggestion
		Subject:      user,
```

director/director_api.go
```diff
@@ -76,29 +76,24 @@ func CreateDirectorSDToken() (string, error) {
 	if directorURL == "" {
```
haoming29, line 76: For this one, since the only use for the function is to create a token, let's remove this function and put your code to where the function was referred (which should be only one place)

director/director_api.go
```diff
@@ -155,29 +150,23 @@ func CreateDirectorScrapeToken() (string, error) {
 	// We assume this function is only called on a director server,
```
haoming29, line 150: Same here, if the function only gets referred once, let's remove the function and put logic to where it was being used.

director/origin_api.go
```diff
@@ -77,35 +77,23 @@ func CreateAdvertiseToken(namespace string) (string, error) {
 		return "", errors.New("Director URL is not known; cannot create advertise token")
```
haoming29, line 77: Same here

director/origin_api.go
```diff
@@ -202,32 +190,22 @@ func CreateDirectorTestReportToken(originWebUrl string) (string, error) {
 		return "", errors.New("Director URL is not known; cannot create director test report token")
```
haoming29, line 190: here as well

haoming29 (changes requested):

director/origin_api_test.go
jhiemstrawisc: I do like the idea of conforming to some kind of well-established and well-studied token scheme, even for internal stuff. Where would you recommend documenting this change? The dual of this is that wherever we add a token profile where there wasn't one before, the receiver of the token should also double check that the token conforms to the profile. Let's open that as an issue after this is merged.

generate/scope_generator.go
jhiemstrawisc: Hmm, pre-commit definitely kicks of its list of checks every time I commit. I do use a custom dev container, but my Dockerfile does run `pip3 install pre-commit`

director/origin_api_test.go
haoming29: I would say in `pelican-dev` channel? Agree that we should enforce the check at the issuer end as well.

web_ui/prometheus.go
```diff
@@ -149,14 +149,39 @@ func configDirectorPromScraper(ctx context.Context) (*config.ScrapeConfig, error
...
-	sdToken, err := director.CreateDirectorSDToken()
+
+	promTokenCfg := utils.TokenConfig{
+		TokenProfile: utils.WLCG,
+		Lifetime:     param.Monitoring_TokenExpiresIn.GetDuration(),
+		Issuer:       serverDiscoveryUrl.String(),
+		Audience:     []string{serverDiscoveryUrl.String()},
+		Version:      "1.0",
+		Subject:      "director",
+		// Claims:       map[string]string{"scope": token_scopes.Pelican_DirectorServiceDiscovery.String()},
```
haoming29, line 160: Can you remove the commented code?

haoming29 (changes requested):
Almost there! with two small things to consider

web_ui/prometheus.go
jhiemstrawisc: Good catch.

haoming29 (approved):
LGTM! Thanks for taking care of this!

PR #721, opened 2024-01-30
Enable OSDF defaults for stashcp/stash_plugin as well

### LINKED ISSUE #720: Enable director by default
Right now, in "OSDF" mode, the director is in use by default (example: `osdf object get`).

However, the majority of the load is the lesser-known "STASH" mode.  We should have these two modes use the same configuration with a result of enabling the director service for all `stashcp` usage by default.

bbockelm:
NOTE: This switches the default for the plugin to use the director which will migrate a large amount of load to the director compared to what's currently seen.

Fixes #720

### DIFF
--- a/client/main.go
+++ b/client/main.go
@@ -755,9 +755,6 @@ func DoStashCPSingle(sourceFile string, destination string, methods []string, re
 			dest_url.Path = "/" + path.Join(dest_url.Host, dest_url.Path)
 		} else if dest_url.Scheme == "pelican" {
 			config.SetFederation(config.FederationDiscovery{})
-			viper.Set("Federation.DiscoveryUrl", "")
-			viper.Set("Federation.DirectorUrl", "")
-			viper.Set("Federation.RegistryUrl", "")
 			federationUrl, _ := url.Parse(dest_url.String())
 			federationUrl.Scheme = "https"
 			federationUrl.Path = ""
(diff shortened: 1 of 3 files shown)


joereuss12:
@bbockelm After lots of digging, reading through code, poking Justin for opinions a ton, and trial and errors this is the solution I came up with. Now we should recognize the hostname in pelican:// url's and discover the federation correctly. The problem was when using an `osdf` or `stash` binary with a pelican url, we always used the osdf defaults (even when the hostname is specified to `pelican.example.com`). This should properly unset federation urls and rediscover the federation with the specified hostname in the object url.

joereuss12:
Found some lines that you forgot to delete, otherwise this PR looks good to me. I think I'll have someone else look at this though for some double checking to make sure neither of us missed anything. However, all my tests seem to be passing.

client/main.go
```diff
@@ -294,24 +294,30 @@ func CheckOSDF(destination string, methods []string) (remoteSize uint64, err err
 // FIXME: GetCacheHostnames is not director-aware!
```
jhiemstrawisc, line 294: I think this change makes this FIXME false. Let's get rid of it.

jhiemstrawisc (changes requested):
One small cleanup, otherwise LGTM

bbockelm:
Since it's now been reviewed by @joereuss12 and @jhiemstrawisc -- and I see that @joereuss12 has addressed the last request -- I'm going to use my admin powers to clear up the state and merge the PR.

PR #722, opened 2024-01-30
Layout updates

CannonLock:
Creates a speed dial in the sidebar for quick actions:

- Creating a issue
- Navigating to docs
- Navigating to repo
- Viewing current version with link pointing to release notes

Screenshot 2024-01-30 at 1 42 13 PM (2)

Updates the release action to also update the website tag so the website version tracks the git tags.

- Verification of the action working -> https://github.com/CannonLock/pelican/actions/runs/7716561427/job/21033642592

Creates a `Main` component for commonly used layout box.


web_ui/frontend/components/layout/Sidebar.tsx
```diff
@@ -20,12 +20,141 @@
...
+        <Grow
+            in={open}
+            style={{ transformOrigin: '0 0 0' }}
+            {...(open ? { timeout: 200 * order } : {})}
+        >
+            <Box pl={1} {...boxProps}>
+                <Tooltip title={title} arrow>
+                    <Paper elevation={2} sx={{ borderRadius: "50%", bgcolor: "#ffffff00"}}>
+                        { href != undefined ?
+                            <Link href={href}>
```
haoming29, line 65: Can we open those links in a new tab?

web_ui/frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "src",
-  "version": "0.1.0",
+  "version": "7.2.0",
```
haoming29, line 3:
```suggestion
  "version": "7.5.0",
```

If we want a version to start with, we'd better start with the latest version to release?

.github/workflows/release.yml
```diff
@@ -22,6 +22,15 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 20
+      - name: Update npm version
+        run: |
+          cd web_ui/frontend
+
+          # Get the current tag and set the package.json version to it
+          npm version ${GITHUB_REF_NAME:1}
```
haoming29, line 30: If we bump npm version for frontend, don't we also need to write the updated `package.json` file back to the repo? Otherwise there will be version discrepancy between released artifacts and our codebase.

We should also set `package.json` version in building the production Docker image.

haoming29 (changes requested):
Mostly about NPM version, with minor things to address in the UI. Please see comments for details.

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,53 @@
...
+          branch: ${{ steps.set_branch.outputs.branch }}
+          create_branch: true
+
+  build-website:
+    needs: [pre-release]
+    uses: CannonLock/pelican/.github/workflows/publish-container.yml@main
+
+  goreleaser:
+    needs: [pre-release]
+    uses: CannonLock/pelican/.github/workflows/release.yml@main
```
haoming29, line 53: Shouldn't it use `Pelicanplatform/pelican/...`? But that file may not exist before this PR is merged...

.github/workflows/release.yml
```diff
@@ -22,6 +18,15 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 20
+      - name: Update npm version
```
haoming29, line 21: If we already had `pre-release.yml` and write back update npm version to the repo, I think we could get rid of the process here?

web_ui/frontend/package.json
CannonLock: The action is set to do it automatically, unfortunately the main branch is fated to have a package.json that is forever behind.

.github/workflows/pre-release.yml
CannonLock: Yes, updated.

web_ui/frontend/components/layout/Sidebar.tsx
```diff
@@ -20,12 +20,141 @@
...
+        <Grow
+            in={open}
+            style={{ transformOrigin: '0 0 0' }}
+            {...(open ? { timeout: 200 * order } : {})}
+        >
+            <Box pl={1} {...boxProps}>
+                <Tooltip title={title} arrow>
+                    <Paper elevation={2} sx={{ borderRadius: "50%", bgcolor: "#ffffff00"}}>
+                        { href != undefined ?
+                            <Link href={href} target={"_blank"}>
```
haoming29, line 65: Do you think it's still worth adding `rel="noopener noreferrer"` to this? I just know that most of the latest browsers add `rel="noopener"` when `target="_blank"`

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,52 @@
+# .github/workflows/release.yaml
```
haoming29, line 1:
```suggestion
# .github/workflows/pre-release.yaml
```

.github/workflows/release.yml
haoming29: This one seems to be still here?

CannonLock: Huh, swore I removed that. Fixed.

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,52 @@
...
+        id: set_branch
+        run: |
+          echo "branch=${GITHUB_REF_NAME%.*}.x" >> $GITHUB_OUTPUT
+
+      - uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: "Pre Release: Update npm version"
+          branch: ${{ steps.set_branch.outputs.branch }}
+
+  build-website:
```
haoming29, line 46: One last thing! This name should be `publish-container`

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,52 @@
...
+    needs: [pre-release]
+    uses: PelicanPlatform/pelican/.github/workflows/publish-container.yml@main
+
+  goreleaser:
```
haoming29, line 50: Here should be `release`

haoming29 (changes requested):
One last change and we are good to go

haoming29 (approved):
LGTM!

PR #727, opened 2024-01-30
Integrate Xrootd monitoring shoveler to Pelican 

### LINKED ISSUE #641: Integrate XRootD Monitoring Shoveler into Pelican server instance
Looking at the [shoveler repo](https://github.com/opensciencegrid/xrootd-monitoring-shoveler/tree/main), it seems that it was also written in Go. That makes me wonder if we want to import the source code and start from their just we did in integrating Prometheus, or we want to keep things simple and only launch the shoveler process as we did for xrootd?

@bbockelm any opinion on this?

haoming29:
Closes #641


config/config.go
```diff
@@ -660,17 +660,20 @@ func InitServer(ctx context.Context, currentServers ServerType) error {
 	viper.SetDefault("OIDC.ClientSecretFile", filepath.Join(configDir, "oidc-client-secret"))
 	viper.SetDefault("Cache.ExportLocation", "/")
 	viper.SetDefault("Registry.RequireKeyChaining", true)
+	viper.SetDefault("Shoveler.TokenLocation", "/etc/pelican/shoveler-token")
```
jhiemstrawisc, line 663: Do we ever expect someone running an origin as non-root to have this setup? I don't think they'll have permission to write the token here if that's the case. Maybe you could put this at `filepath.Join(configDir, "shoveler/shoveler-token")`?

docs/parameters.yaml
```diff
@@ -1327,6 +1327,177 @@ default: true
...
+name: Shoveler.URL
+description: >-
+  For amqp and stomp.
+
+  The URL to connect to the shoveler
+type: url
+default: none
+components: ["origin", "cache"]
+---
+name: Shoveler.Exchange
```
jhiemstrawisc, line 1369: I'd suggest that for protocol-specific configurations, you prefix the sub variable. Eg `Shoveler.AMQPExchange`, `Shoveler.StopUsername`

jhiemstrawisc (approved):
I'm approving for now, but I left a bit of minor feedback you can feel free to change if you think appropriate. I mostly skimmed after we hand-tested a bit of this together, but I'm not too worried at this point because this is an optional feature that's off by default.

haoming29:
Good call. I'll address them first thing next Monday

Issue #731, opened 2024-01-31
Have pelican plugin write `TransferErrorData` with specified errror type


joereuss12:
Now with #714 in place, we can interpret nested classad list of classads to write to our output file. This way we can populate Todd M's `TransferErrorData` field with known error types that even condor understands. This should be done after/with #538 so that we can use newly enumerated client error codes for our errors (whether in `TransferErrorData` or in `DeveloperData` if not defined by condor.

bbockelm:
Assigning to Sarthak -- this will require a discussion with Todd Miller (there's also a relevant Google doc he has created with the associated error classifications).

Issue #538, opened 2023-12-23
Enumerate client error codes


bbockelm:
Having a structured set of error codes from the client helps us summarize and aggregate failures and provides some guidance to the user on how to correct the issue.

I propose we start a new YAML file collecting the known Pelican client error codes.  A few thoughts.

1.  It's good to group these up into major and minor codes, much like there's the 3xx, 4xx, and 5xx series of HTTP status codes.  Todd Miller did a [good writeup](https://docs.google.com/document/d/1Gbb7XDOtmLX_xcfmCwP8KAZ_s6PDDGgCamGoDl145Ww) of proposed error categories in HTCondor and those would be a good set to start with.
   * For example, if there's a major code for problems in writing out the file to local disk, the minor subcode might be the POSIX errno.
3. Design a corresponding PelicanClient error object that can wrap other errors throughout the client base.
4. For `pelican object copy` and the like, map from Pelican error code to process exit code.  The help function should document the possible exit codes; this might come from the YAML file.
5. The YAML file should have a document per known error code.  The keys should include:
   * Major / minor code values.
   * Client exit code.
   * Description.
   * Whether the error might be resolved by a retry (could be a boolean).

Before we try and implement the error code functionality, it'd be better to put together a design document trying to enumerate the errors for review.

joereuss12:
Work in progress design doc: https://docs.google.com/document/d/1hP3tyxq1jjH3Tb9Z2M1e_OUT0IGUPfh7-AYZF_l62Ck/edit?usp=sharing

Issue #537, opened 2023-12-23
Expand error reporting from Pelican plugin to the the condor_starter


bbockelm:
The HTCSS dev team is [expanding the data that can be sent from the plugin](https://opensciencegrid.atlassian.net/browse/HTCONDOR-2122).

We should start filling in the data as part of our plugin outputs.  See [this document](https://docs.google.com/document/d/1jN5FTFwhCD9tAHru0ERxRXWwmF3WdtZy8KuF8pOJA5I/edit) for example of the proposed schema.

There's the idea of "Condor error number" versus "Pelican error number".  The condor error number isn't defined yet so skip that.  The Pelican error number is going to require engineering in a separate issue.  So, for first pass, let's skip those attributes.

joereuss12:
@bbockelm after some discussion with Todd M, it turns out that previous document is outdated and here is the updated one (specifically look under `syntax`): 
https://docs.google.com/document/d/1Gbb7XDOtmLX_xcfmCwP8KAZ_s6PDDGgCamGoDl145Ww/edit

Some things he mentioned that is worth highlighting from our discussion:
- There is now a classad list of classads defined as `TransferErrorData` which contains a list of classad fields that the plugin should start sending out, I can begin progress on this right away.
- A particular classad within `TransferErrorData` is called `IntermediateServer*` attributes, which could provide information if something happens when contacting a server that is not the one specified by the `url` classad
- There is also a classad attribute introduced called `DeveloperData` which can send back whatever classad attributes we would like, condor just sends whatever is inside that attribute and does not check types or anything against it, it is just sent back with the classad outfile.
- Within this `TransferErrorData` with multiple transfers, they will be listed in the order they were tried instead of numbering them within the error.

PR #733, opened 2024-01-31
Stash_plugin produce classad output in more cases

joereuss12:
Changes include:
1. A recovery function that will catch any panics from stashPluginMain, this will attempt to write out a classad output file and exit
2. If we fail to create the classad outfile, we have an exit code of 3 to indicate that the problem is the creation of the file
3. If we fail to initialize our config, we also write classad output and exit
4. New function that can be used to write our output and exit

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -61,6 +61,15 @@ type Transfer struct {
 	localFile string
 }
 
+type ExitCode int
+
+const (
+	Success ExitCode = iota
+	Error
+	FailedOutfile = 3
+	Retryable     = 11
+)
+
 func init() {
 	// Define the file transfer plugin command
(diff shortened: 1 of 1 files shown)


joereuss12:
Note: I created issue #731 to add specific error types to the plugin since with #537 starts allowing us to write a classad list of classads I want to make sure everything is working with the new changes in condor. I also think it would be good to go through the enumerate client error codes issue #538 either with or before I do this so that we can utilize all our new error handling with this.

Currently in place is just printing the `TransferError` with a description of the error

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM.

Issue #732, opened 2024-01-31
Pelican client logs verbose exit message


haoming29:
As reported by @bbockelm:

```console
 ./dist/pelican_darwin_arm64/pelican object get -f osg-htc.org /ospool/uc-shared/public/OSG-Staff/validation/test.txt /dev/null; echo $?
null 14.00 b / 14.00 b [==============================================================================================================] Done!
ERROR[2024-01-30T13:20:42-06:00] Pelican is safely exited                     
0
```

For pelican client, nothing extra should be printed/logged at the end of the program.

PR #734, opened 2024-01-31
Fix server exit status code and remove verbose log at the client

### LINKED ISSUE #715: Exit pelican with exit code 1 if errgrp returns with error

haoming29:
Fixes #715 and #732 

For reviewer, the following can be a good verification process of the fix:
* Client should not log "Pelican is safely exited"  at the end of file transfer

```
./pelican object get -f osg-htc.org /ospool/uc-shared/public/OSG-Staff/validation/test.txt /dev/null
```

* Pelican should exit with status code = 1 if there's an error
  * Spin up your origin by appending `; echo $?` at the end of your existing command. i.e `./pelican origin serve; echo $?`
  * Kill XRootD by finding your xrootd process and kill it
  * Pelican should exit with something like:
 ```
ERROR[2024-01-31T19:22:06Z] Fatal error occurred that lead to the shutdown of the process: xrootd process failed unexpectedly: signal: killed 
1
 ```
Note that `1` means the status code is 1.

Also this PR removed the error log "Pelican is safely exited" and use `fmt.Println` instead to not to cause confusion.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -20,10 +20,12 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/pelicanplatform/pelican/config"
+	"github.com/pelicanplatform/pelican/launchers"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


matyasselmeci (approved):
My tests succeeded. LGTM, thanks!

PR #738, opened 2024-01-31
Topology authfile generation

turetske:
OSDF origins and caches now pull permissions from topology. This is to get around x509 legacy issues in OSDF.


jhiemstrawisc (approved):
LGTM, thanks Emma!

PR #743, opened 2024-02-01
OSDF configs: use more system-ish default locations for the TLS cert and key

matyasselmeci:
According to https://www.getpagespeed.com/server-setup/ssl-directory, the correct locations of the cert and key (for RHEL-based systems) is `/etc/pki/tls/certs/$hostname.crt` and `/etc/pki/tls/private/$hostname.key`.

We don't know the hostname when building the RPM so just call them `pelican.crt` and `pelican.key`.

Debian/Ubuntu uses `/etc/ssl` instead of `/etc/pki/tls` but I don't want to make two default configs.

### DIFF
--- a/systemd/osdf-cache.yaml
+++ b/systemd/osdf-cache.yaml
@@ -8,8 +8,8 @@ Federation:
   # We need the "origins" director to find origins instead of the default ("caches" director)
   DirectorUrl: https://osdf-director-origins.osg-htc.org
 Server:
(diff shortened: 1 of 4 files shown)


bbockelm:
> I don't want to make two default configs

If we start to get serious about high-quality Debian native packaging (or get lucky enough to have external contributions), I wouldn't mind having two defaults at all!

However, for now, this is good enough.

brianhlin (approved):
LGTM

Issue #676, opened 2024-01-18
Add support for external field validation/options API for custom registration fields in registry


haoming29:
This should be built on top of #675 

For getting options to a custom registration field with `enum` type, instead of reading static config, we should allow admin to read the options data from a URL for the custom fields with `enum` type, as long as the URL the endpoint returns follow our defined schema. Any such URLs should be cached to avoid DDOS on the third-party API

Similarly, we should also provide an option to allow admin to validate their custom registration fields via a validation URL, which follows our defined data schema. This API can't be cached locally by pelican as it's a dynamic POST request, so we should add in the documentation for users to be cautious about potential traffic increase from using such URL.

haoming29:
Close this ticket for now as we don't have much feedback from users on what they really want for this feature, and it's non-blocking

PR #748, opened 2024-02-02
Remove param docs that are not implemented

haoming29:
Just figured that I put something in the `parameters.yaml` that never get implemented. So just to remove it to not to cause confusion. The issue to implement them is here #676


turetske (approved):
LGTM!

Issue #765, opened 2024-02-06
Upgrade project to go 1.21


bbockelm:
In #590, I want to do some simple refactoring that requires `slices.Sort`, necessitating the upgrade to 1.21.

However, to do that we need to upgrade the various development Docker containers first (since on Ubuntu the GHA builds inside the container).

Upgrades should occur at least in the following locations (and probably a few others that can be found via grep):
- https://github.com/PelicanPlatform/pelican/blob/main/.github/workflows/test.yml#L9
- https://github.com/PelicanPlatform/pelican/blob/main/images/dev.Dockerfile#L61
- https://github.com/PelicanPlatform/pelican/blob/main/images/Dockerfile#L32
- https://github.com/PelicanPlatform/pelican/blob/main/.github/workflows/release.yml#L32

PR #771, opened 2024-02-06
Update project to go version 1.21

williamnswanson:
For Issue #765. Stable go version for 1.21 for dev.Dockerfile found at https://go.dev/dl/

Closes Issue #765


bbockelm:
@williamnswanson - there appears to be a bootstrapping issue.  You may need to do this across two PRs:
1.  PR to update the containers to go 1.21 so it is available at build time.
2. PR to switch the minimum requirement to 1.21.

Your build failure is because the build is using the existing dev container which has go 1.20.

williamnswanson:
@bbockelm Looks like all the checks are sorted now.

bbockelm (approved):
LGTM.

PR #772, opened 2024-02-06
Add knob to self test frequency

### LINKED ISSUE #667: Allow customizing self-test frequency
I would like to run the origin's self-test less frequently than every 15 seconds since it is just noise in the logs.

haoming29:
Closes #667


matyasselmeci (approved):
LGTM. Thanks!

PR #705, opened 2024-01-29
Connection broker

bbockelm:
This PR adds support for a new module, the "connection broker", that allows origins to callback to the cache.

PR #774, opened 2024-02-06
Add rate-limiter to the login endpoint

### LINKED ISSUE #711: Add rate limiting to various login endpoint

haoming29:
Closes #711 

To test, build locally and run your any of your Pelican server and go to `https://<hostname>:<port>/view/login/index.html`

Hit `Confirm` button without typing any username. If you hit it slowly (not double click), then you should get 400 error. If you double-click or triple click, then you should get 429 error indicating too many requests.

### DIFF
--- a/config/resources/defaults.yaml
+++ b/config/resources/defaults.yaml
@@ -33,6 +33,7 @@ Server:
   WebHost: "0.0.0.0"
   EnableUI: true
   RegistrationRetryInterval: 10s
+  UILoginRateLimit: 1
 Director:
   DefaultResponse: cache
   MinStatResponse: 1
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -992,6 +992,13 @@ type: duration
 default: 10s
 components: ["origin", "cache"]
 ---
+name: Server.UILoginRateLimit
+description: >-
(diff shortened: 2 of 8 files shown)


bbockelm:
Could you switch to `golang.org/x/time/rate` perhaps?  I was planning on using it elsewhere:

https://github.com/PelicanPlatform/pelican/pull/705/files#diff-4f90d3c881e24d44c38cce40e200654acfd79622cf17bea2fd7b557aa9c58bc1R49-R62

Avoids the redis dependency.

haoming29:
@bbockelm It should be doable to use `golang.org/x/time/rate` but `gin-rate-limit` is more flexible as it limits the rate by a key per request (IP in our case). I didn't find the built-in rate package to have the equivalent which means we need to build the `key` support ourselves.

bbockelm:
Gotcha.  How much larger does the binary become with this new dependency?

haoming29:
> Gotcha. How much larger does the binary become with this new dependency?

The size of the built binary didn't seem to change. It's `42.1MB` for both `main` branch and my feature branch. But the size seems to be doubled compared to the `7.5.0` release artifact.

joereuss12 (approved):
LGTM, also I really like how you always say how to test it. I had a joy of a time spamming that button

PR #776, opened 2024-02-06
Fixes to xrootd logging config

joereuss12:
Changes include:
- Xrootd logging now uses the standard logrus logging values
- Instead of string comparisons for mapping, we now use log.ParseLevel and compare the levels (which are unsigned ints)
- The program now exits when there is an improper config value and returns a proper error
- Anywhere these configs are listed, they are all listed alphabetically for consistency and an easier time finding each option
- Unit tests for the generation of the xrootd config file

### DIFF
--- a/xrootd/xrootd_config.go
+++ b/xrootd/xrootd_config.go
@@ -135,13 +135,12 @@ type (
 
 	LoggingConfig struct {
 		OriginCms       string
-		OriginPfc       string
-		OriginPss       string
 		PssSetOptOrigin string
 		OriginScitokens string
 		OriginXrd       string
 		OriginXrootd    string
 		CacheOfs        string
+		CachePfc        string
 		CachePss        string
 		PssSetOptCache  string
 		CacheScitokens  string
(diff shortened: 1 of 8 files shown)


joereuss12:
Let me know if I missed anything or need to add anything else! Hopefully this patches everything up.

xrootd/xrootd_config.go
```diff
@@ -699,154 +704,308 @@ func SetUpMonitoring(ctx context.Context, egrp *errgroup.Group) error {
...
+	/////////////////////////ORIGIN/////////////////////////////
+	// Origin Cms
+	// https://xrootd.slac.stanford.edu/doc/dev54/cms_config.htm
+	originCmsConfig, err := log.ParseLevel(param.Logging_Origin_Cms.GetString())
+	if err != nil {
+		return errors.Wrap(err, "Error parsing specified log level for Origin_Cms, proper values include: panic, fatal, error, warn, info, debug, trace")
+	}
+	switch originCmsConfig {
+	case log.PanicLevel:
+		xrdConfig.Logging.OriginCms = "-all"
```
bbockelm, line 718: Use `fallthrough` for duplicate lines.

Doesn't save lines of code -- but does show intent (that things are supposed to be duplicated).

bbockelm: Actually, better yet, let's wrap this all into a function.  Here's an idea:

- Create a struct called `loggingMap` or similar.  Have each of the logging levels be a separate member of the struct.
- Iterate from top to bottom of the struct (in order of the logging levels).  If the string is set, use that in the xrdConfig.  If it's not set, use the prior value.
- Have a single helper function `genLoggingConfig` that does all the translation, including error messages.

`loggingMap` might look like:

```
type loggingMap struct {
    Trace string
    Debug string
    ...
    Panic string
}
```

So, in this case, we might see:

```
xrdConfig.Logging.OriginCms, err = genLoggingConfig(
    param.Logging_Origin_Cms.GetString(),
    loggingMap {
        Trace: "debug",
        Info: "all",
        Error: "-all",
    }
)
if err != nil {
    ...
}
```

That's about as compact as I can think of making this.

bbockelm (changes requested):
Two things:
- See comments for one potential way to rewrite the `mapXrootdLogLevels` to be a bit more readable.
- Talk with @matyasselmeci about the description of the configuration parameters.  I feel they're still not particularly approachable.  How to write them to make sense to someone with little knowledge about XRootD?

joereuss12:
Note: still need to make the config params more approachable and user-friendly, but thought I'd push my work in progress

bbockelm (changes requested):
Very close.

I like the new, much more compact, style of setting the log levels.  It makes the code quite readable.

Only thing I spotted were some logging categories don't really make sense (e.g., setting logging for cache-related modules in the origin config).

xrootd/ca-bundle.crt
haoming29, line 1: Are we intended to include these CAs in git? and why there's a CA update in log config PR?

haoming29 (changes requested):
One small question and we should be good to go

xrootd/ca-bundle.crt
joereuss12: Oh, not sure how I missed these getting added. I'm assuming they should not be there. I can try to remove them

joereuss12:
Okay @haoming29 cleanup has happened. Just waiting for tests to pass to make sure I didn't remove anything important but I think everything should be good to go!

haoming29 (approved):
LGTM with the extra files removed. Thanks for taking care of this!

haoming29:
@joereuss12 You might need to ask @jhiemstrawisc to overwrite Brian's requested change to get this merged in

jhiemstrawisc:
@Haoming, since you've been reviewing for Brian, can you resolve any of the the conversations that are open that have a fix? Once I see everything resolved, I'll merge

haoming29:
> @haoming, since you've been reviewing for Brian, can you resolve any of the the conversations that are open that have a fix? Once I see everything resolved, I'll merge

@jhiemstrawisc  Peek through the conversations and looks to me they have all been taken care of. Resolved conversations accordingly

jhiemstrawisc:
Perfect, thanks. Just wanted the acknowledgement from someone for bookkeeping since I'm not reviewing myself :) Consider it merged! Thanks to everyone who worked on this

PR #781, opened 2024-02-07
Fixed issue with added new line in cache authfile

turetske:
-- Removed extra newline when generating an authfile with a
       provided authfile with a plublic line
    -- Adjusted an authfile test to catch this


jhiemstrawisc (approved):
LGTM. Thanks for looking into this, Emma!

Issue #777, opened 2024-02-07
`pelican object put` does not work with `pelican://` URLs


bbockelm:
`pelican object put` only appears to work with either `osdf://` or non-schema URLs.  This should be covered by a unit test.

PR #784, opened 2024-02-07
Fix bug with pelican object put & expand unit test

### LINKED ISSUE #778: `object get` / `object put` unit tests can trigger os.Exit()
The `object get` and `object put` unit tests invoke cobra `Run` functions which call `os.Exit` directly.  This isn't compatible with the unit test framework as it skips any sort of cleanup -- and doesn't print out what test has failed or why.

joereuss12:
Changes include:
- Fixed typo for bytesTransferred
- Fixed bug in sharing_url_test to skip a terminal verify so it can generate a token (was failing for me)
- Reworked unit tests for object get/put as well as recursive object get/put to not call cobra commands that can call an exit, causing things to not be shut down
  - Moved the tests to handle_http_test.go since it calls the functions within client now
  - Removed object_get_put_test.go because of this
- Fixed bug where you could not do a pelican object put with a pelican:// url
- Removed unused code within object_put.go (left over from some copy/pasting)

Fixes #778 and #777


client/handle_http_test.go
```diff
@@ -551,3 +552,620 @@ func TestFullUpload(t *testing.T) {
...
+
+	// Need to set permissions or the xrootd process we spawn won't be able to write PID/UID files
+	permissions := os.FileMode(0755)
+	err = os.Chmod(tmpPath, permissions)
+	require.NoError(f.T, err)
+
+	viper.Set("ConfigDir", tmpPath)
+
+	// Increase the log level; otherwise, its difficult to debug failures
+	//viper.Set("Logging.Level", "Debug")
```
haoming29, line 589: Remove commented code or uncomment this?

client/handle_http_test.go
```diff
@@ -551,3 +552,620 @@ func TestFullUpload(t *testing.T) {
...
+
+	// Create a token file
+	tokenConfig := utils.TokenConfig{
+		TokenProfile: utils.WLCG,
+		Lifetime:     time.Minute,
+		Issuer:       param.Origin_Url.GetString(),
+		Audience:     []string{param.Origin_Url.GetString()},
+		Subject:      "origin",
+	}
+	tokenConfig.AddRawScope("storage.read:/ storage.modify:/")
```
haoming29, line 689: I think Justin's recent PR added support for WLCG token scopes as enumerated type. Could you update this? Here is a reference: https://github.com/PelicanPlatform/pelican/blob/5b921c42bdb965f6413d2f93fc874a993a611dc3/client/handle_http_test.go#L391

haoming29 (changes requested):
Two small changes requested

haoming29 (approved):
LGTM!

PR #785, opened 2024-02-07
Create non-world-readable logfiles

joereuss12:
When specifying the -l flag, log files now have permissions set to 0640 (not world readable) instead of 0644


matyasselmeci (approved):
LGTM

Issue #782, opened 2024-02-07
Create non-world-readable logfiles


matyasselmeci:
Pelican Service:
- [x] Origin
- [x] Cache

XRootD has some issues with putting sensitive data in logs. Until we get that fixed, if `Logging.LogLocation` is specified, create the log files without world-readable permissions (i.e. `0640` instead of `0644`).

PR #786, opened 2024-02-07
Put default logfiles in /var/log/pelican; create /var/log/pelican non-world-readable in the RPM

matyasselmeci:
This is a workaround for #782 to avoid exposing sensitive data in world-readable logs, though it might be worth keeping around even after that issue is fixed, as an extra layer of safety.


joereuss12 (approved):
LGTM

PR #790, opened 2024-02-08
Added DownloadComponent to install.mdx

Rhymlore:


haoming29 (approved):
Looks like just a update to use component from docs repo to render the download options. Would be nice if you can share a screenshot of how the page looks like after the change

PR #792, opened 2024-02-08
Only create topology table at server start and update topology table 10m instead of 10ms

### LINKED ISSUE #789: ITB registry is killed when trying to update institution or approve an origin without institution
Pelican Version:

Pelican Service:
- [ ] Client
- [ ] Plugin
- [x] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

 - OS: [e.g. Linux, MacOS]

**Additional context**

haoming29:
Should fix #789 

* From ITB registry log, the registry was killed with the following log message

```console
time="2024-02-08T17:08:39Z" level=fatal msg="Failed to create topology table: database is locked (5) (SQLITE_BUSY)"
```

Looks like there's race condition in DB write access, which makes sense because the error occurs every time someone tries to update/create a new registration

Looking at the code, we call `log.fatal` when we failed create the topology table, and we create the table every "topology update" cycle. The update interval was supposed to set to 10min but the value is `10` which was recognized by viper as **10 milliseconds** and it keep creating the table, doing a bunch of topology update, every 10ms.

This also fixes `504` error we sometimes get from topology, as we basically DDOS-ed topology endpoint.

### DIFF
--- a/config/resources/osdf.yaml
+++ b/config/resources/osdf.yaml
@@ -22,7 +22,7 @@ Federation:
   DiscoveryUrl: osg-htc.org
   TopologyURL: https://topology.opensciencegrid.org
   TopologyNamespaceURL: https://topology.opensciencegrid.org/osdf/namespaces
-  TopologyReloadInterval: 10
+  TopologyReloadInterval: 10m
 Registry:
   CacheApprovedOnly: true
   OriginApprovedOnly: true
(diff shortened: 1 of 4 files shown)


CannonLock:
@haoming29 This is pretty funny, as I have been running into issues with topology not responding fast enough all week.

It all comes full circle.

haoming29:
> @haoming29 This is pretty funny, as I have been running into issues with topology not responding fast enough all week.
> 
> It all comes full circle.

Feel bad for whoever is maintaining the topology service, even worse if it won't restart itself after being DDos-ed

turetske (approved):
LGTM!

PR #795, opened 2024-02-09
Fix PSS debug level

bbockelm:
In XRootD, PSS debug level is the integer "4"; this makes the generated config match the XRootD code.

Only backporting to v7.5.x branch because @joereuss12 has a larger overhaul in progress for main.


jhiemstrawisc (approved):
LGTM. I'll patch into a 7.5.4 release

PR #798, opened 2024-02-09
Remove use of global fed metadata within client

joereuss12:
Changes include:
- Created a local metadata object that store federation metadata per url.
- Changed a lot of tests on client commands, we no longer rely on osdf for tests to pass
- Made a function in config that discovers a federation but does not populate global config values

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -275,7 +275,7 @@ func GenerateTransferDetailsUsingCache(cache CacheInterface, opts TransferDetail
 	return nil
 }
 
-func download_http(sourceUrl *url.URL, destination string, payload *payloadStruct, namespace namespaces.Namespace, recursive bool, tokenName string) (transferResults []TransferResults, err error) {
+func download_http(sourceUrl *url.URL, destination string, metadata UrlMetadata, payload *payloadStruct, namespace namespaces.Namespace, recursive bool, tokenName string) (transferResults []TransferResults, err error) {
 	// First, create a handler for any panics that occur
 	defer func() {
 		if r := recover(); r != nil {
@@ -307,7 +307,7 @@ func download_http(sourceUrl *url.URL, destination string, payload *payloadStruc
 	// Check the env var "USE_OSDF_DIRECTOR" and decide if ordered caches should come from director
 	var transfers []TransferDetails
 	var files []string
-	directorUrl := param.Federation_DirectorUrl.GetString()
+	directorUrl := metadata.directorUrl
 	closestNamespaceCaches, err := GetCachesFromNamespace(namespace, directorUrl != "")
 	if err != nil {
 		log.Errorln("Failed to get namespaced caches (treated as non-fatal):", err)
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -275,7 +275,7 @@ func GenerateTransferDetailsUsingCache(cache CacheInterface, opts TransferDetail
 	return nil
 }
 
-func download_http(sourceUrl *url.URL, destination string, metadata UrlMetadata, payload *payloadStruct, namespace namespaces.Namespace, recursive bool, tokenName string) (transferResults []TransferResults, err error) {
+func download_http(pelicanURL PelicanURL, destination string, payload *payloadStruct, namespace namespaces.Namespace, recursive bool, tokenName string) (transferResults []TransferResults, err error) {
 	// First, create a handler for any panics that occur
 	defer func() {
 		if r := recover(); r != nil {
@@ -288,16 +288,16 @@ func download_http(sourceUrl *url.URL, destination string, metadata UrlMetadata,
 		}
 	}()
 
-	packOption := sourceUrl.Query().Get("pack")
+	packOption := pelicanURL.objectUrl.Query().Get("pack")
 	if packOption != "" {
 		log.Debugln("Will use unpack option value", packOption)
 	}
-	sourceUrl = &url.URL{Path: sourceUrl.Path}
+	pelicanURL.objectUrl = &url.URL{Path: pelicanURL.objectUrl.Path}
 
 	var token string
 	if namespace.UseTokenOnRead {
 		var err error
-		token, err = getToken(sourceUrl, namespace, false, tokenName)
+		token, err = getToken(pelicanURL.objectUrl, namespace, false, tokenName)
 		if err != nil {
 			log.Errorln("Failed to get token though required to read from this namespace:", err)
 			return nil, err
@@ -307,7 +307,7 @@ func download_http(sourceUrl *url.URL, destination string, metadata UrlMetadata,
 	// Check the env var "USE_OSDF_DIRECTOR" and decide if ordered caches should come from director
 	var transfers []TransferDetails
 	var files []string
-	directorUrl := metadata.directorUrl
+	directorUrl := pelicanURL.directorUrl
 	closestNamespaceCaches, err := GetCachesFromNamespace(namespace, directorUrl != "")
 	if err != nil {
 		log.Errorln("Failed to get namespaced caches (treated as non-fatal):", err)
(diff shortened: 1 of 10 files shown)


joereuss12:
Still a work in progress. Still need to have the url metadata be stored in a ttl cache and everything could use a little more testing.

client/handle_http.go
```diff
@@ -275,7 +275,7 @@ func GenerateTransferDetailsUsingCache(cache CacheInterface, opts TransferDetail
 	return nil
 }
 
-func download_http(sourceUrl *url.URL, destination string, payload *payloadStruct, namespace namespaces.Namespace, recursive bool, tokenName string) (transferResults []TransferResults, err error) {
+func download_http(sourceUrl *url.URL, destination string, metadata UrlMetadata, payload *payloadStruct, namespace namespaces.Namespace, recursive bool, tokenName string) (transferResults []TransferResults, err error) {
```
bbockelm, line 278: Don't keep the metadata as a separate object.  Make a PelicanURL struct that is composed with a URL.

client/main.go
```diff
@@ -85,6 +85,14 @@ type payloadStruct struct {
 	downloadSize int64
 }
 
+type UrlMetadata struct {
```
bbockelm, line 88: Make a PelicanURL with an embedded URL.

client/main.go
```diff
@@ -467,6 +475,63 @@ func getNamespaceInfo(resourcePath, OSDFDirectorUrl string, isPut bool) (ns name
 	}
 }
 
+func getUrlMetadata(remoteUrl *url.URL, scheme string) (metadata UrlMetadata, err error) {
```
bbockelm, line 478: Make this into a NewPelicanURL method.

(Though might not be exported?  Depending on the interactions, maybe newPelicanURL)

client/main.go
```diff
@@ -556,7 +604,6 @@ func DoPut(localObject string, remoteDestination string, recursive bool) (transf
 	uploadedBytes, err := doWriteBack(localObjectUrl.Path, remoteDestUrl, ns, recursive, "")
 	AddError(err)
 	return uploadedBytes, err
-
```
bbockelm, line 559: Avoid unnecessary churn.

client/main.go
```diff
@@ -589,41 +636,24 @@ func DoGet(remoteObject string, localDestination string, recursive bool) (transf
...
-			federationUrl.Path = ""
-			viper.Set("Federation.DiscoveryUrl", federationUrl.String())
-			err = config.DiscoverFederation()
-			if err != nil {
-				return nil, err
-			}
-		}
-	}
 
+	// This is for condor cases:
```
bbockelm, line 640: Just like you've factored out URL processing for metadata, you can factor out this logic for unsupported schemes. Will make a good unit test too.

client/main.go
```diff
@@ -632,7 +662,8 @@ func DoGet(remoteObject string, localDestination string, recursive bool) (transf
 		remoteObject = "/" + remoteObject
 	}
 
-	directorUrl := param.Federation_DirectorUrl.GetString()
+	//directorUrl := param.Federation_DirectorUrl.GetString()
```
bbockelm, line 665: Avoid commented out code.

config/config.go
```diff
@@ -292,6 +292,75 @@ func GetAllPrefixes() []string {
 	return prefixes
 }
 
+// This function is for discovering federations as specified by a url during a pelican:// transfer.
+// this does not populate global fields and is more temporary per url
+func DisoverUrlFederation(federationDiscoveryUrl string) (FederationDiscovery, error) {
```
bbockelm, line 297: Too much duplication with DidcoverFederation.

Have DiscoverFederation use this function when it is ready.

Needs a unit test.

bbockelm (commented):
First round of feedback on the approach.

bbockelm (changes requested):
Another draft review contained.

Please talk with @haoming29 about how to have the cache expire objects.  You'll need to run the `Start()` method to launch the goroutine and think about what lifetime you'd like it to have.

Do you have precommit setup on your development host?  There's a large number of static check issues that are making it into the PR that should be caught by precommit.

joereuss12:
> Do you have precommit setup on your development host? There's a large number of static check issues that are making it into the PR that should be caught by precommit.

The linter was not working for me for a while, I took a few stabs at fixing it but each time was taking a little to long so I put it off. I just updated it here: #829 and it seems to be working now

client/handle_http_test.go
```diff
@@ -1124,75 +1084,26 @@ func TestRecursiveUploadsAndDownloads(t *testing.T) {
...
-		uploadURL := "pelican:///test/" + dirName
-
-		//////////////////////////////////////////////////////////
-
-		// Upload the file with PUT
-		transferDetailsUpload, err := DoPut(tempDir, uploadURL, true)
+		uploadStr := "osdf:///test/" + dirName
+		uploadURL, err := url.Parse(uploadStr)
 		assert.NoError(t, err)
-		if err == nil && len(transferDetailsUpload) == 2 {
```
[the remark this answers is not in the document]
joereuss12: My thought process as well as discussion with Justin was to not fully rely on `osg-htc.org` for our tests so we should just make sure that everything is correct up to the point of the upload/download. Not super sure if that was the best way to do so or if we should just rely on `osg-htc.org` for `osdf://` cases? The other tests do uploads and downloads and check for sizes of the returned files

client/main.go
```diff
@@ -467,6 +480,91 @@ func getNamespaceInfo(resourcePath, OSDFDirectorUrl string, isPut bool) (ns name
...
+		return fmt.Errorf("Do not understand the destination scheme: %s. Permitted values are %s",
+			scheme, strings.Join(understoodSchemes, ", "))
+	}
+	return nil
+}
+
+func newPelicanURL(remoteUrl *url.URL, scheme string) (pelicanURL PelicanURL, err error) {
+	if remoteUrl.Host != "" {
+		if scheme == "osdf" || scheme == "stash" {
+			remoteUrl.Path, err = url.JoinPath(remoteUrl.Host, remoteUrl.Path)
```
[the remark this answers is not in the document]
joereuss12: This section is used when someone does an osdf or stash url but only puts two `/`'s instead of 3 (hence they have a hostname). What would you like me to do differently? Create our own url like the comment below and modify that url?

client/main.go
```diff
@@ -727,81 +782,50 @@ func DoStashCPSingle(sourceFile string, destination string, methods []string, re
...
 		AddError(err)
 		return transferResults, err
 	}
 
+	// If we continue here, we are doing a get
+
+	// Verify valid scheme
+	err = schemeUnderstood(sourceScheme)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
joereuss12: The user should be told the scheme they tried as well as the permitted ones from the err returned from `schemeUnderstood`: https://github.com/PelicanPlatform/pelican/blob/144ec348c48ce69ca7812cc0bd0550b98d0fc568/client/main.go#L488-L489

client/main.go
```diff
@@ -467,6 +480,91 @@ func getNamespaceInfo(resourcePath, OSDFDirectorUrl string, isPut bool) (ns name
...
+			}
+			// Set our local url metadata
+			pelicanURL.directorUrl = urlFederation.DirectorEndpoint
+			pelicanURL.discoveryUrl = federationUrl.String()
+			pelicanURL.registryUrl = urlFederation.NamespaceRegistrationEndpoint
+
+			// Populate cache with data
+			pelicanURLCache.Set(federationUrl.String(), pelicanURL, ttlcache.DefaultTTL)
+		}
+	}
```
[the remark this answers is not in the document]
joereuss12: Ah, missed that case. Making a fix that will throw an error if the discovery-url is not specified with `-f`

joereuss12:
@bbockelm check out these changes and my comments. Let me know if anything needs tweaking. Also, look to see where the cache is created, where the loader function is (right next to cache creation), and where I start and stop the cache. I think the locations are pretty good but let me know what you think.

bbockelm:
@joereuss12 - now that the big client refactor has merged, please clear up the conflicts here.

PR #802, opened 2024-02-12
cache directorUrl fix for pss.origin

turetske:
Fixed an issue where the pss.origin wasn't being generated properly from the Federation.DirectorUrl

Note: I did do an end-to-end cache test where I curled the cache and was able to get a file served from a local origin.

### DIFF
--- a/xrootd/xrootd_config.go
+++ b/xrootd/xrootd_config.go
@@ -325,7 +325,7 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
 				discoveryUrl.Host = discoveryUrl.Path
 				discoveryUrl.Path = ""
 			} else if discoveryUrl.Path != "" && discoveryUrl.Path != "/" {
-				return "", errors.New("The discoveryUrl Path is non-empty, ensure the Federation.DiscoveryUrl has the format <host>:<port>")
+				return "", errors.New("The Federation.DiscoveryUrl's path is non-empty, ensure the Federation.DiscoveryUrl has the format <host>:<port>")
 			}
 			discoveryUrl.Scheme = "pelican"
 			discoveryUrl.Path = ""
--- a/xrootd/xrootd_config.go
+++ b/xrootd/xrootd_config.go
@@ -316,7 +316,17 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
 	if err != nil {
 		return "", errors.Wrap(err, "Failed to pull information from the federation")
 	}
-	viper.Set("Cache.PSSOrigin", param.Federation_DirectorUrl.GetString())
+	if directorUrlStr := param.Federation_DirectorUrl.GetString(); directorUrlStr != "" {
+		directorUrl, err := url.Parse(param.Federation_DirectorUrl.GetString())
+		if err == nil {
+			log.Debugln("Parsing director URL for 'pss.origin' setting:", directorUrlStr)
+			directorUrl.Scheme = "pelican"
+			viper.Set("Cache.PSSOrigin", directorUrl.String())
+		} else {
+			return "", errors.Wrapf(err, "Failed to parse director URL %s", directorUrlStr)
+		}
+	}
+
 	if discoveryUrlStr := param.Federation_DiscoveryUrl.GetString(); discoveryUrlStr != "" {
 		discoveryUrl, err := url.Parse(discoveryUrlStr)
 		if err == nil {
@@ -332,6 +342,10 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
 		} else {
 			return "", errors.Wrapf(err, "Failed to parse discovery URL %s", discoveryUrlStr)
 		}
+	}
+
+	if viper.GetString("Cache.PSSOrigin") == "" {
+		return "", errors.New("Cache pss.origin url empty")
 	}
 
 	if cacheServer, ok := server.(*cache_ui.CacheServer); ok {
--- a/xrootd/xrootd_config.go
+++ b/xrootd/xrootd_config.go
@@ -341,7 +341,7 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
 		if err == nil {
 			log.Debugln("Parsing director URL for 'pss.origin' setting:", directorUrlStr)
 			if directorUrl.Path != "" && directorUrl.Path != "/" {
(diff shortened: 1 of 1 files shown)


xrootd/xrootd_config.go
```diff
@@ -316,7 +316,17 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
 	if err != nil {
 		return "", errors.Wrap(err, "Failed to pull information from the federation")
 	}
-	viper.Set("Cache.PSSOrigin", param.Federation_DirectorUrl.GetString())
+	if directorUrlStr := param.Federation_DirectorUrl.GetString(); directorUrlStr != "" {
+		directorUrl, err := url.Parse(param.Federation_DirectorUrl.GetString())
+		if err == nil {
```
bbockelm, line 321: Handle the case (generate an error) if the directorUrl.path is not empty or not `/`.

xrootd/xrootd_config.go
```diff
@@ -316,7 +316,17 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
...
+			log.Debugln("Parsing director URL for 'pss.origin' setting:", directorUrlStr)
+			directorUrl.Scheme = "pelican"
+			viper.Set("Cache.PSSOrigin", directorUrl.String())
+		} else {
+			return "", errors.Wrapf(err, "Failed to parse director URL %s", directorUrlStr)
+		}
+	}
+
```
bbockelm, line 329: In the `Federation.DiscoveryUrl` parsing below, similarly generate an error if the path component is not trivial.

xrootd/xrootd_config.go
```diff
@@ -334,6 +344,10 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
 		}
 	}
 
+	if viper.GetString("Cache.PSSOrigin") == "" {
+		return "", errors.New("Cache pss.origin url empty")
```
bbockelm, line 348: It's unlikely the user knows what "Cache pss.origin url empty" is or how to solve it.

How about this:
```suggestion
		return "", errors.New("One of Federation.DiscoveryUrl or Federation.DirectorUrl must be set to configure a cache")
```

xrootd/xrootd_config.go
bbockelm: Also, I think we should switch the logic order here.  If DirectorUrl is set, it is more specific so it should override the DiscoveryUrl.

bbockelm (approved):
Some minor changes requested.

Did you test with the director URL case that the auto-generated federation metadata in the director is sufficient?

bbockelm (changes requested):
Whoops, meant to hit "request changes" along with the previous comment.  Trying again.

turetske:
> Some minor changes requested.
> 
> Did you test with the director URL case that the auto-generated federation metadata in the director is sufficient?

Not sure what this means, specifically. Such as if the Federation.Director isn't set but the Federation.DiscoveryUrl is?

xrootd/xrootd_config.go
turetske: We don't want an error in this case. The discoveryUrl.Host is gotten from discoveryUrl.Path. So the host shouldn't be trivial.

turetske: Ah, wait, I see the way to check

turetske:
Unsure why the origin test is failing, it works locally and I didn't touch any code that would affect the origin at all. Trying to look into it, but will take some time since it's only failing on GitHub.

xrootd/xrootd_config.go
```diff
@@ -316,14 +316,16 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
...
+
 	if discoveryUrlStr := param.Federation_DiscoveryUrl.GetString(); discoveryUrlStr != "" {
 		discoveryUrl, err := url.Parse(discoveryUrlStr)
 		if err == nil {
 			log.Debugln("Parsing discovery URL for 'pss.origin' setting:", discoveryUrlStr)
 			if len(discoveryUrl.Path) > 0 && len(discoveryUrl.Host) == 0 {
 				discoveryUrl.Host = discoveryUrl.Path
 				discoveryUrl.Path = ""
+			} else if discoveryUrl.Path != "" && discoveryUrl.Path != "/" {
+				return "", errors.New("The discoveryUrl Path is non-empty, ensure the Federation.DiscoveryUrl has the format <host>:<port>")
```
bbockelm, line 328:
```suggestion
				return "", errors.New("The Federation.DiscoveryUrl's path is non-empty, ensure the Federation.DiscoveryUrl has the format <host>:<port>")
```

xrootd/xrootd_config.go
```diff
@@ -334,6 +336,24 @@ func CheckCacheXrootdEnv(exportPath string, server server_utils.XRootDServer, ui
 		}
 	}
 
+	if directorUrlStr := param.Federation_DirectorUrl.GetString(); directorUrlStr != "" {
+		directorUrl, err := url.Parse(param.Federation_DirectorUrl.GetString())
+		if err == nil {
+			log.Debugln("Parsing director URL for 'pss.origin' setting:", directorUrlStr)
+			if directorUrl.Path != "" && directorUrl.Path != "/" {
+				return "", errors.New("The directorUrl Path is non-empty, ensure the Federation.DirectorUrl has the format <host>:<port>")
```
bbockelm, line 344:
```suggestion
				return "", errors.New("The Federation.DirectorUrl's path is non-empty, ensure the Federation.DirectorUrl has the format <host>:<port>")
```

bbockelm:
> Not sure what this means, specifically. Such as if the Federation.DirectorUrl isn't set but the Federation.DiscoveryUrl is?

Other way around.  What happens if the `Federation.DirectorUrl` is set but the `Federation.DiscoveryUrl` is not?

turetske:
> > Not sure what this means, specifically. Such as if the Federation.DirectorUrl isn't set but the Federation.DiscoveryUrl is?
> 
> Other way around. What happens if the `Federation.DirectorUrl` is set but the `Federation.DiscoveryUrl` is not?

@bbockelm Ah, I tested both. Both work.

bbockelm (approved):
Two small error message nitpicks.

Ready to merge once those are taken care of and you verify it's been tested with and without the director URL.

(Actually, instead of only hand testing, it might be quicker/better to add a small unit test that the generated xrootd config is what you expect)

PR #806, opened 2024-02-13
Rebuild OSG's xrootd RPM from source inside the dev container

matyasselmeci:
This downloads the latest xrootd source RPM from the OSG 23 development repos, and uses rpmbuild to build it locally for the current architecture.  The resulting RPMs are put into /usr/local/src/rpmbuild/RPMS/$arch/* (where $arch is the architecture, either x86_64, noarch, or aarch64) and a yum repo is built from it. The repo has highest priority so the locally built RPMs are picked over what's in EPEL.

To save time and space, the xrootd-compat libraries are not built, and neither are the docs.  Even with that, though, it's a good 20 minutes to build on ARM.  Caching the build results might be a good idea.

### DIFF
--- a/.github/workflows/publish-container.yml
+++ b/.github/workflows/publish-container.yml
@@ -93,3 +93,5 @@ jobs:
           tags: "${{ steps.generate-tag-list.outputs.taglist }}"
           build-args: |
             IS_PR_BUILD=${{ github.event_name == 'pull_request' }}
+          cache-from: type=registry,ref=hub.opensciencegrid.org/pelican_platform/pelican:buildcache
+          cache-to: type=registry,ref=hub.opensciencegrid.org/pelican_platform/pelican:buildcache,mode=max,ignore-error=true
--- a/.github/workflows/publish-dev-container.yml
+++ b/.github/workflows/publish-dev-container.yml
@@ -48,5 +48,5 @@ jobs:
           file: ./images/dev.Dockerfile
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          cache-from: type=registry,ref=hub.opensciencegrid.org/pelican_platform/pelican-dev:buildcache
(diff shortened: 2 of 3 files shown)


matyasselmeci:
I think I need #807 for the image build to succeed.

matyasselmeci:
I marked this as draft so it doesn't get accidentally merged until we come up with a solution for the increase in CI runtime.

haoming29:
> @haoming29 - can you take this across the line by setting up a GHA cache containing the built RPMs?

Will take a look!

brianhlin:
@haoming29 thanks for tagging me! The registry build cache support is new since I last looked at all of this stuff but it looks really useful

haoming29:
@matyasselmeci @bbockelm After some tweaking, now both dev container and production container will cache the image to the harbor registry. However, since pushing images to harbor requires login, which needs harbor credentials via GitHub secrets. For GHA in PRs, the secrets in Pelican repo is not available to GHA runners because the owner of the PR is Mat or someone else. Therefore, the cached image won't be pushed to harbor during PR, but will when this PR is merged in (or in release process)

matyasselmeci:
I think that's fine, as long as the cached image is pulled for PRs.

brianhlin:
Yeah, I think that's preferable. You don't want random folks submitting PRs that push artifacts into our registries, even if GitHub does its best to stop drive bys

matyasselmeci:
The action looks good to me; I can't push the approve button because I own the PR.

haoming29:
I can help push the button if I can get a hint on how to verify it works. Since this is to build xrootd from source, will running xrootd gives me a particular version number?

matyasselmeci:
You won't see it in the program version itself but if you run `rpm -q xrootd-server`, you should see `.osg.el8` in the version of the RPM.

haoming29 (approved):
run `rpm -q xrootd-server` gives me `xrootd-server-5.6.7-1.2.osg.el8.aarch64` which looks good to me. I'll hit approve for now since Brian's out for a couple of days. Let's see if the GHA cache actually works.

PR #812, opened 2024-02-13
Remove bucket requirement for S3 origins

### LINKED ISSUE #811: S3 origins should work without specifying a bucket
In the case of AWS open data, we don't want to have to enumerate all of the buckets... Instead, for a given service url like `s3.amazonaws.com`, we should be able to set up an origin that exports everything hosted by the service.

jhiemstrawisc:
It was a mistake to require that S3 origins be configured with a bucket. In the case of AWS public data, for example, it would be nice to have a single origin that exports all public buckets from `s3.amazonaws.com`. Under the old scheme, this would require an origin per bucket -- whoops!

# Testing notes for this one:
- Spin up a registry/director.

## Testing a public origin with a single bucket
- Serve a public origin (`Origin.EnablePublicReads: true`) that exports a single bucket from AWS open data -- we do still want this to work, even if we no longer require it:
`pelican origin serve --mode s3 --port 8444 --service-name "s3.amazonaws.com" --region "us-east-1" --bucket "noaa-wod-pds" --service-url https://s3.us-east-1.amazonaws.com -d`
- Now curl a test file from the origin: `curl -v -k https://<origin hostname>:<origin port>/s3.amazonaws.com/us-east-1/noaa-wod-pds/MD5SUMS`. You should see a small md5 sum file print to terminal

## Testing a public origin with no bucket specified
Now we basically want to serve the same origin, but without pinning the bucket. We can still use the same bucket for testing, but the difference is that if this works with the unpinned `noaa-wod-pds` bucket, it will also work for any bucket under `s3.amazonaws.com/us-east-`.
 - Serve a public origin (`Origin.EnablePublicReads: true`):
`pelican origin serve --mode s3 --port 8444 --service-name "s3.amazonaws.com" --region "us-east-1" --service-url https://s3.us-east-1.amazonaws.com -d`
- Now curl a test file from the origin: `curl -v -k https://<origin hostname>:<origin port>/s3.amazonaws.com/us-east-1/noaa-wod-pds/MD5SUMS`. You should see a small md5 sum file print to terminal

Both of those origin configurations should work.

Closes #811

### DIFF
--- a/cmd/origin.go
+++ b/cmd/origin.go
@@ -162,7 +162,7 @@ func init() {
 	originServeCmd.MarkFlagsMutuallyExclusive("volume", "bucket-access-keyfile")
 	originServeCmd.MarkFlagsMutuallyExclusive("volume", "bucket-secret-keyfile")
 	// We don't require the bucket access and secret keyfiles as they're not needed for unauthenticated buckets
-	originServeCmd.MarkFlagsRequiredTogether("service-name", "region", "bucket", "service-url")
+	originServeCmd.MarkFlagsRequiredTogether("service-name", "region", "service-url")
 	originServeCmd.MarkFlagsRequiredTogether("bucket-access-keyfile", "bucket-secret-keyfile")
 
 	// The port any web UI stuff will be served on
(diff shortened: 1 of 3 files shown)


turetske (approved):

jhiemstrawisc:
Awesome, thanks Emma!

Issue #810, opened 2024-02-13
Mark DNS timeouts as retryable


bbockelm:
Pelican Version: v7.5.1

The following was observed in a condor hold message:

> Attempt #1: Get "https://osdf-director.osg-htc.org/ospool/PROTECTED/USER/USER_FILE.sif": dial tcp: lookup osdf-director.osg-htc.org: i/o timeout (10.6s since start)

This appears to be a DNS issue with the local site and certainly transient; it should be marked as retryable.  The error should be targeted to DNS timeouts and not failures.

Since this was observed by a user, it's reasonable to backport.

PR #813, opened 2024-02-13
Failure to get metadata is now retryable error

joereuss12:
Adds and error to config so that if there is a timeout getting the metadata response, we mark as retryable and exit(11). In addition, found that we were marking things as retryable even when the client has no errors in its checked error stack array, so check for that now and return false for retryable if that array is empty.

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -109,10 +109,10 @@ func stashPluginMain(args []string) {
 
 	var isConfigErr = false
 	config.InitConfig()
-	err := config.InitClient()
-	if err != nil {
-		log.Errorf("Problem initializing the Pelican client config: %v", err)
-		err = errors.Wrap(err, "Problem initializing the Pelican Client configuration")
+	configErr := config.InitClient()
+	if configErr != nil {
+		log.Errorf("Problem initializing the Pelican client config: %v", configErr)
+		configErr = errors.Wrap(configErr, "Problem initializing the Pelican Client configuration")
 		isConfigErr = true
 	}
 
@@ -178,7 +178,12 @@ func stashPluginMain(args []string) {
 
 		// Set as failure and add errors
 		resultAd.Set("TransferSuccess", false)
-		resultAd.Set("TransferError", err.Error())
+		resultAd.Set("TransferError", configErr.Error())
+		if client.ShouldRetry(configErr) {
+			resultAd.Set("TransferRetryable", true)
+		} else {
+			resultAd.Set("TransferRetryable", false)
+		}
 		resultAds = append(resultAds, resultAd)
 
 		// Attempt to write our file and bail
@@ -273,7 +278,7 @@ func stashPluginMain(args []string) {
 	ctx, cancel := context.WithCancel(context.Background())
 	egrp, _ := errgroup.WithContext(ctx)
 	defer func() {
-		err = egrp.Wait()
+		err := egrp.Wait()
 		if err != context.Canceled {
 			log.Errorln("Error when shutting down worker:", err)
 		}
@@ -318,6 +323,7 @@ func stashPluginMain(args []string) {
 
 	// Ensure all our workers are shut down.
 	cancel()
+	var err error
 	if waitErr := egrp.Wait(); waitErr != nil && waitErr != context.Canceled {
 		log.Errorln("Error when shutting down worker:", waitErr)
 		success = false
@@ -354,7 +360,11 @@ func writeClassadOutputAndBail(exitCode int, resultAds []*classads.ClassAd) {
 	}
 
 	// We'll exit 3 in here if anything fails to write the file
-	writeOutfile(nil, resultAds, outputFile)
+	_, retryable := writeOutfile(nil, resultAds, outputFile)
+
+	if retryable {
+		exitCode = 11
+	}
 
 	log.Errorln("Failure with pelican plugin. Exiting...")
 	os.Exit(exitCode)
--- a/config/config.go
+++ b/config/config.go
@@ -162,9 +162,9 @@ func (e *MetadataErr) Is(target error) bool {
 }
 
 func (e *MetadataErr) Wrap(err error) *MetadataErr {
-	return &MetadataErr {
+	return &MetadataErr{
 		InnerErr: err,
-		Msg: fmt.Sprintf("%s", e.Msg),
+		Msg:      e.Msg,
 	}
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -88,6 +88,10 @@ type (
 	ServerType int // ServerType is a bit mask indicating which Pelican server(s) are running in the current process
 
 	ContextKey string
+
+	MetaDataErr struct {
+		Err string
+	}
 )
 
 const (
@@ -137,6 +141,15 @@ var (
 	// Pelican version
 	version string = "dev"
 )
+
+func (e *MetaDataErr) Error() string {
+	return e.Err
+}
+
+func (e *MetaDataErr) Is(target error) bool {
+	_, ok := target.(*MetaDataErr)
+	return ok
+}
 
 func init() {
 	validate = validator.New(validator.WithRequiredStructEnabled())
@@ -381,7 +394,10 @@ func DiscoverFederation() error {
 
 	result, err := httpClient.Do(req)
 	if err != nil {
-		return errors.Wrapf(err, "Failure when doing federation metadata lookup to %s", discoveryUrl)
+		errMsg := errors.Wrapf(err, "Failure when doing federation metadata lookup to %s", discoveryUrl)
+		return &MetaDataErr{
+			Err: errMsg.Error(),
+		}
 	}
 
 	if result.Body != nil {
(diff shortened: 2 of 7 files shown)


joereuss12:
Marking as a draft so that I can add #810 in this

joereuss12:
Can check this works by modifying this line in config/config.go to be a really short timeout:
https://github.com/PelicanPlatform/pelican/blob/1fb0a529676e558a8ffc4da84e0e7f0cb2ab9a0a/config/config.go#L373

as well as modify:
https://github.com/PelicanPlatform/pelican/blob/1fb0a529676e558a8ffc4da84e0e7f0cb2ab9a0a/config/resources/defaults.yaml#L79-L80
to a shorter timeout

It should deal with both of those cases reporting a retryable error (exit code 11).

joereuss12:
Needs to be back-ported to v7.5.x

config/config.go
```diff
@@ -88,6 +88,10 @@ type (
 	ServerType int // ServerType is a bit mask indicating which Pelican server(s) are running in the current process
 
 	ContextKey string
+
+	MetaDataErr struct {
```
bbockelm, line 92: Rename this symbol to `Metadata`.

config/config.go
```diff
@@ -136,6 +140,15 @@ var (
 	PelicanVersion string
 )
 
+func (e *MetaDataErr) Error() string {
+	return e.Err
+}
+
+func (e *MetaDataErr) Is(target error) bool {
```
bbockelm, line 147: This makes the error a single-use: all metadata errors are considered the "same".

Have `MetadataErr` wrap the underlying error.  This allows things like stack traces to propagate through and logic on the original error cause.  That is, add an `Unwrap` error that returns the original cause (see https://pkg.go.dev/errors).

config/config.go
```diff
@@ -367,7 +380,17 @@ func DiscoverFederation() error {
 
 	result, err := httpClient.Do(req)
 	if err != nil {
-		return errors.Wrapf(err, "Failure when doing federation metadata lookup to %s", discoveryUrl)
+		errMsg := errors.Wrapf(err, "Failure when doing federation metadata lookup to %s", discoveryUrl)
+		// If DNS timeout, we want this to be retryable so return a specific error
+		if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
```
bbockelm, line 385: If we make the `MetadataErr` wrappable, we could have a global var, MetadataTimeoutErr:

```
var MetadataTimeoutErr *MetadataErr := &MetadataErr{err: "Timeout when querying metadata"}
```

And then here, we could have
```
var netErr net.Error
if errors.As(err, netError) {
    err = MetadataTimeoutErr.Wrap(err)
} else {
    NewMetadataError(err, "Error occurred when querying for metadata")
}
```

Then, we can simplify for the retryable test:

```
if errors.Is(err, config.MetadataTimeoutErr) {
   ...
}
```

bbockelm (changes requested):
At some point we need to torch the global error registry and use the normal golang error mechanism.  That day is not today, however.

The big request here is to make the metadata err object more flexible, allowing it to wrap errors and just wrapping the timeout in the end, not replacing it.

config/config.go
```diff
@@ -367,7 +389,13 @@ func DiscoverFederation() error {
 
 	result, err := httpClient.Do(req)
 	if err != nil {
-		return errors.Wrapf(err, "Failure when doing federation metadata lookup to %s", discoveryUrl)
+		var netErr net.Error
+		if errors.As(err, &netErr) {
```
bbockelm, line 393: This doesn't limit to only timeouts.  `*net.Error` appears to have a separate `Timeout` method.

```suggestion
		if errors.As(err, &netErr) && netErr.Timeout() {
```

By taking the underlying error object and turning it to a string, you loose the error hierarchy.  You really want:

```
MetadataTimeoutErr.Wrap(err)
```

and to add an `Unwrap` method.

joereuss12:
@bbockelm is this more of what you are looking for?

cmd/plugin.go
```diff
@@ -358,7 +363,11 @@ func writeClassadOutputAndBail(exitCode int, resultAds []*classads.ClassAd) {
 	}
 
 	// We'll exit 3 in here if anything fails to write the file
-	writeOutfile(resultAds, outputFile)
+	_, retryable := writeOutfile(resultAds, outputFile)
+
+	if retryable {
```
bbockelm, line 368: Hm - this seems brittle.  If we accidentally set retryable to true on success, it'll make the exit code 11, right?

config/config.go
```diff
@@ -134,8 +139,37 @@ var (
...
 	// Pelican version
 	PelicanVersion string
+
+	MetadataTimeoutErr *MetadataErr = &MetadataErr{Msg: "Timeout when querying metadata"}
 )
 
+// This function creates a new MetadataError by wrapping the previous error
+func NewMetadataError(err error, msg string) *MetadataErr {
+	return &MetadataErr{
+		Msg: fmt.Sprintf("%s: %v", msg, err),
```
bbockelm, line 149: I think the err shouldn't be concatenated here.  Otherwise it's done twice later on.  What does it look like in your testing?

config/config.go
```diff
@@ -134,8 +139,37 @@ var (
...
+	}
+}
+
+func (e *MetadataErr) Error() string {
+	return fmt.Sprintf("%v; %s", e.InnerErr, e.Msg)
```
bbockelm, line 154: Standard go style is colon-separated, not semicolon.

config/config.go
```diff
@@ -134,8 +139,37 @@ var (
...
+}
+
+func (e *MetadataErr) Is(target error) bool {
+	_, ok := target.(*MetadataErr)
```
bbockelm, line 158: Missing string comparison for Msg.

config/config.go
```diff
@@ -88,6 +88,11 @@ type (
 	ServerType int // ServerType is a bit mask indicating which Pelican server(s) are running in the current process
 
 	ContextKey string
+
+	MetadataErr struct {
+		Msg      string
```
bbockelm, line 93: Do not export internal variables.

bbockelm (changes requested):
Down to a few minor items.

cmd/plugin.go
```diff
@@ -192,6 +192,11 @@ func stashPluginMain(args []string) {
 		// Set as failure and add errors
 		resultAd.Set("TransferSuccess", false)
 		resultAd.Set("TransferError", client.GetErrors())
+		if client.ErrorsRetryable() {
+			resultAd.Set("TransferRetryable", true)
```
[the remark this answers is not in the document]
joereuss12: In this case there was an error because we failed to run `config.InitClient` which is what the bool `isConfigErr` indicates. I just deal with the error after parsing our arguments so we can get information on the output file and such. The client should know about the error since we explicitly add the error as well:
https://github.com/PelicanPlatform/pelican/blob/e84f775368c5d981dbfb096ef7351b65de0a610c/cmd/plugin.go#L117-L125

cmd/plugin.go
joereuss12: This function is used only when we need to bail early (such as with a panic or failing to InitClient). Therefore, we should only ever be failing here so success should never be true. Our exit codes out of this function should be either exit(1) or exit(11). I at least can't think of a case where we'd be successful but end up within this function but correct me if I'm wrong with any of this.

config/config.go
joereuss12: For my hand testing triggering that error it seems to work okay:
```
TransferError = "Attempt #1: Problem initializing the Pelican Client configuration: Error occured when querying for metadata: Get \"https://osg-htc.org/.well-known/pelican-configuration\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```
Note: it's a timeout error above because I just triggered that error with the timeout instead of the other. No duplication seems to be showing up though.

joereuss12:
@bbockelm check my comments and correct me if I am wrong with any of them or if you still want adjusting. Next round on revisions are in and everything seems to be working how it should.

cmd/plugin.go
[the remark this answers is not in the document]
bbockelm: Ah, ok, makes sense.

cmd/plugin.go
bbockelm: Makes sense, thanks.

joereuss12:
Okay, should be all good now. I think this struggle to make proper errors like this is good since now I will have a good template to move forward with and know what mistakes not to make. Ready for review again.

config/config.go
```diff
@@ -134,8 +139,45 @@ var (
...
+		return fmt.Sprintf("%s: %v", e.msg, e.innerErr)
+	} else {
+		return e.msg
+	}
+}
+
+func (e *MetadataErr) Is(target error) bool {
+	// We want to verify we have a timeout error
+	if _, ok := target.(*MetadataErr); ok {
+		return e.msg == target.Error()
```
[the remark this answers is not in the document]
joereuss12: Ah yeah, you're right. In the earlier comment I was not sure because `target` is of type `error` but we literally convert it right there lol. Fixed now!

joereuss12:
@bbockelm is this good to merge? I think we need your approval since you requested changes.

bbockelm (approved):
LGTM.

bbockelm:
@joereuss12 - this is good to go once the merge conflicts are cleaned up.

Issue #808, opened 2024-02-13
Cleanup schema for file transfer plugin output


bbockelm:
Pelican Version: 7.5.1

The schema for file transfer plugin output is a bit of a mess.  Here's the results of a successful transfer (newlines added for readability):

```
[
   TransferFileBytes = 1730318336;
   TransferTotalBytes = 1730318336;
   DeveloperData = [
      PelicanClientVersion = "7.5.1";
      Attempts = "1";
      TransferFileBytes0 = "1730318336";
      TimeToFirstByte0 = "6003863488";
      Endpoint0 = "sc-cache.chtc.wisc.edu:8443";
      TransferEndTime0 = "1707831751";
      ServerVersion0 = "XrootD/v5.6.6";
   ];
   TransferEndTime = 1707831751;
   TransferLocalMachineName = "spark-a029";
   TransferFileName = "/var/lib/condor/execute/osg01/glide_BkCx62/execute/dir_1450435/USER_FILE.sif";
   TransferSuccess = true;
   TransferStartTime = 1707831751;
   TransferProtocol = "stash";
   TransferUrl = "osdf:///ospool/PROTECTED/USER/USER_FILE.sif";
   TransferType = "download";
]
```

Notes:
- Timestamps need to be integers.  This affects `TransferFileBytes` and `TransferEndTime`
- Counts need to be integers.  This affects `Attempts.
- The `TransferProtocol` does not match the `TransferUrl`.
- I don't believe the transfer start and end times here.  If I look at the timestamps of the input and output files -- and try a few hand transfers from the host, I believe it was a 6 second transfer.
- I've no clue what units `TimeToFirstByte` is in.  Use [Prometheus rules](https://prometheus.io/docs/practices/naming/#base-units) of utilizing base units and make it a floating point.
- Since the transfer end times are rounded to the nearest integer, we should add a new field into the developer data that is the transfer time.

PR #814, opened 2024-02-13
Cleanup scheme for ft plugin output

joereuss12:
All changes fix issues listed in #808

### DIFF
--- a/classads/classads.go
+++ b/classads/classads.go
@@ -70,9 +70,15 @@ func (c *ClassAd) String() string {
 			for key, value := range v {
 				buffer.WriteString(key)
 				buffer.WriteString(" = ")
-				buffer.WriteString("\"")
-				buffer.WriteString(fmt.Sprintf("%v", value))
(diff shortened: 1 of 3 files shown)


joereuss12:
Marked as draft since it still needs some testing

bbockelm (changes requested):
Big picture change is to keep the time in the native go data structure for as long as possible.  Converting to integers too early loses too much data.

joereuss12:
@bbockelm Next round of revisions are in, I moved the converting to `.Unix()` for the time stamps to be a tad later in the code. Let me know if you would like it to be even later than when I moved it.

jhiemstrawisc (approved):

joereuss12:
@bbockelm same here too, are we good to merge? I think github wants your stamp of approval

jhiemstrawisc:
Going to bypass branch protections on this one to get it in before Joe has to chase more merge conflicts. Thanks for taking care of this one, @joereuss12!

PR #815, opened 2024-02-14
Stage lotman

jhiemstrawisc:
The goal here is to get LotMan staged into the prod/dev containers.


jhiemstrawisc:
There's no command line utility for lotman. I think success in this case is "the containers build and the shared objects are in `/usr/lib64`"

turetske (approved):
LGTM!

PR #816, opened 2024-02-14
Add cache launcher

### LINKED ISSUE #427: End-to-End Cache Test
Create an end to end cache test in the same vein as the end to end origin test.

### LINKED ISSUE #522: Add Cache to "pelican serve" mode
Currently the pelican serve mode only can run a director, registry, and/or an origin. We want to expand this to allow for a cache.

turetske:
Split out variables that were shared between the cache and origin that needed to be separate for the variables to run in the same process.

Moved the cache to the launcher to be in line with the other servers.

An end-to-end cache test with fed serve was added.

Close #427 
Close #522


bbockelm (commented):
Not through with the review but wanted to checkpoint all my comments so far.

registry/registry_ui_test.go
```diff
@@ -96,6 +96,8 @@ func TestListNamespaces(t *testing.T) {
 	setupMockRegistryDB(t)
 	defer teardownMockNamespaceDB(t)
 
+	viper.Set("Origin.Port", 8443)
+	viper.Set("Server.WebPort", 8444)
```
jhiemstrawisc, line 100: We're moving toward the randomized ports in our tests, so set these to 0. If you need to create tokens, there should be examples about how to get the appropriate issuer in `handle_http_test.go`

server_ui/advertise.go
```diff
@@ -100,11 +100,16 @@ func advertiseInternal(ctx context.Context, server server_utils.XRootDServer) er
 	if err != nil {
 		return errors.Wrap(err, fmt.Sprintf("%s failed to get namespaceAds from the director", server.GetServerType()))
 	}
+	serverUrl := param.Origin_Url.GetString()
+	webUrl := param.Server_ExternalWebUrl.GetString()
+	prefix := param.Origin_NamespacePrefix.GetString()
+	if server.GetServerType().IsEnabled(config.CacheType) {
+		serverUrl = param.Origin_Url.GetString()
```
jhiemstrawisc, line 107: Should this be `Origin_Url`? We just set that above, but this looks conditional to caches.

systemd/osdf-cache.yaml
```diff
@@ -24,6 +24,6 @@ XRootD:
   ## Authfile is the xrootd authorization file for VOMS/X.509 auth
   ## XXX remove this once #564 (https://github.com/PelicanPlatform/pelican/issues/564) is done
   Authfile: /run/stash-cache-auth/Authfile
-  RunLocation: /run/pelican/osdf-cache/xrootd
 Cache:
   DataLocation: "/mnt/osdf"
+  RunLocation: /run/pelican/osdf-cache/xrootd/cache
```
jhiemstrawisc, line 29: Why not just `/run/pelican/rootd/cache`? Seems redundant to specify cache twice in the path

jhiemstrawisc (changes requested):

docs/parameters.yaml
```diff
@@ -600,13 +612,39 @@ type: string
...
+  For non-root servers, if $XDG_RUNTIME_DIR is not set, a temporary directory will
+  be created (and removed on shutdown)
+type: filename
+root_default: /run/pelican/xrootd/cache
+default: $XDG_RUNTIME_DIR/pelican/cache
+components: ["cache"]
+---
+name: Cache.NamespacePrefix
+description: >-
+  The filepath prefix at which a cache is registered in the registry. Note that this value is not expected to be set by the user.
```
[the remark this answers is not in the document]
turetske: We need to register this as a viper variable. Previously the cache prefix was saved to Origin.NamespacePrefix, but if serving both a cache and an origin in the same process, we need a separate value for the Cache or it would just overwrite the Origin's.

launchers/launcher.go
```diff
@@ -199,14 +200,39 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
...
+	if modules.IsEnabled(config.CacheType) && modules.IsEnabled(config.OriginType) {
+		log.Debug("Advertise Origin")
+		if err := server_ui.Advertise(ctx, servers); err != nil {
+			return shutdownCancel, err
+		}
+	}
+
+	if modules.IsEnabled(config.CacheType) {
+		// Give five seconds for the origin to finish advertising to the director
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
turetske: Ah. Except this isn't waiting for the director to be up and running. The director already is up an running. This is waiting for the origin to finish its advertisement to the director. There's no easy way to check for that.

registry/registry_ui_test.go
turetske: Okay, but if my container only forwards specific ports, how can the tests ensure they use the ports my container forwards and has open? This seems like an issue for testing in the local environment.

launchers/launcher.go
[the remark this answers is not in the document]
jhiemstrawisc: What about querying the director at the endpoint that would redirect to that origin? Once the advertisement is complete, the director will respond with a 307 instead of a 404. I think `WaitUntilWorking` should be toolable for that.

launchers/launcher.go
```diff
@@ -209,14 +209,22 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
...
+		if err = server_ui.Advertise(ctx, servers); err != nil {
 			return shutdownCancel, err
 		}
 	}
 
 	if modules.IsEnabled(config.CacheType) {
 		// Give five seconds for the origin to finish advertising to the director
-		time.Sleep(5 * time.Second)
+		if modules.IsEnabled(config.OriginType) {
+			time.Sleep(5 * time.Second)
```
jhiemstrawisc, line 220: Adding a place holder here even though we just discussed in person.

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM. Great job getting finagling all of this together!

PR #820, opened 2024-02-15
gorelease with snapshotting unless there are tags for Issue #681

williamnswanson:
Change release-container action to have build use goreleaser with snapshotting if  input.tags is empty, and only use goreleaser without snapshotting if input.tags contains a tag.


haoming29 (approved):
Let's give it another shot. The test failure should be unrelated to the change

PR #821, opened 2024-02-15
Cache now gets the namespaceads from the director every time it re-adv…

turetske:
…ertises

	-- Added a server function to get namespace ads from the director
	-- Does nothing if server is an origin
	-- During advertise, the server calls this function
	
	
To test:

1. Spin up a local director and registry.
2. Serve an origin with a namespace <ns1>
3. Serve a cache and check /run/pelican/xrootd/cache/scitokens-cache-generated.cfg to ensure the origin is there.
4. Serve another origin with a namespace <ns2>
5. Wait 2 minutes for the cache advertisement to refresh.
6. Recheck the scitokens-cache-generated.cfg to ensure the new origin is there

Note: There is a known issue with duplicated base paths in scitokens that is not introduced in this PR.

### DIFF
--- a/cache_ui/advertise.go
+++ b/cache_ui/advertise.go
@@ -19,9 +19,17 @@
 package cache_ui
 
 import (
+	"encoding/json"
+	"net/url"
+	"strings"
+
 	"github.com/pelicanplatform/pelican/common"
 	"github.com/pelicanplatform/pelican/config"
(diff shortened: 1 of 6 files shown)


haoming29 (approved):
LGTM! I can confirm that the scitoken config file was updated after a new origin joins the federation.

PR #822, opened 2024-02-15
Fix Prometheus token expires

### LINKED ISSUE #787: Prometheus token not renewed
Pelican Services:
- [X] Origin
- [X] Cache

After my director was running for an hour or so, I started seeing messages like this in the logs of my origin and cache:

```
DEBUG[2024-02-07T20:15:33-06:00] Token validation failed with server issuer: Invalid JWT: "exp" not satisfied 
INFO[2024-02-07T20:15:33-06:00] Authentication failed. Didn't pass the chain of checking:
Token validation failed with federation issuer:  "exp" not satisfied
Token validation failed with server issuer:  Invalid JWT: "exp" not satisfied 
INFO[2024-02-07T20:15:33-06:00] Served Request                                client=192.168.0.12 daemon=gin fields.time="875.167µs" method=GET resource=/metrics status=403
```

Seeing that it's the `/metrics` endpoint, I'm guessing the Prometheus token isn't being periodically renewed.

haoming29:
Fixes #787 

The cause of the issue was that only the self-scraper config and service discovery config are reloaded. The origin/cache scraper never gets reloaded.

This PR also reduced the token refresh time to 5m to ensure we won't wait until another 59m if the token update wasn't successful.

To test, set the following config for faster expire/refresh cycle:

```yaml
Monitoring:
  TokenExpiresIn: 2m
  TokenRefreshInterval: 1m
```

Also make sure you have `Federation.DiscoveryUrl`  set to your director address.

I'd suggest spin up your servers individually in separate terminals in the order of `director`, `registry`, and `origin`. It also works for fed-in-a-box, but you'll see much more verbose log.

And you should see `200` for all `/metrics` request at the **origin** server (your log level should be at at least `INFO`):

```console
INFO Served Request  daemon=gin fields.time=1.991334ms method=GET resource=/metrics status=200
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -795,9 +795,9 @@ func InitServer(ctx context.Context, currentServers ServerType) error {
 	tokenExpiresIn := param.Monitoring_TokenExpiresIn.GetDuration()
 
 	if tokenExpiresIn == 0 || tokenRefreshInterval == 0 || tokenRefreshInterval > tokenExpiresIn {
-		viper.Set("Monitoring.TokenRefreshInterval", time.Minute*59)
+		viper.Set("Monitoring.TokenRefreshInterval", time.Minute*5)
 		viper.Set("Monitoring.TokenExpiresIn", time.Hour*1)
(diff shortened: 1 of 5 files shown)


jhiemstrawisc (approved):
LGTM

PR #826, opened 2024-02-19
Add user role icon to the side bar in UI

### LINKED ISSUE #629: Display user role in UI
With #625 merging in, we are able to tell the role of the currently logged in user (whether they are `admin` or `user`). To help user identify the privilege they have, we can have a user icon somewhere in the web UI for all servers to display the role of currently logged in user.

haoming29:
Closes #629 

With admin privilege:
Screenshot 2024-02-19 at 9 47 47 AM

Without:
Screenshot 2024-02-19 at 9 47 54 AM

I kind of hate fetching `/whoami` one more time during rendering (I believe we already did 2 or 3 times per page) but not doing that requires us using context or some state management package which seems outside of the scope for this ticket.

### DIFF
--- a/web_ui/frontend/components/layout/Sidebar.tsx
+++ b/web_ui/frontend/components/layout/Sidebar.tsx
@@ -18,16 +18,16 @@
 
 "use client"
 
-import Image from 'next/image'
 import { useRouter } from 'next/navigation'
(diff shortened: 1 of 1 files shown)


turetske:
@CannonLock Can you look at this soon?

CannonLock (approved):
LGTM. I think it would be better to not show anything if they aren't logged in but we should just fix that when we do the big login overhaul next release.

PR #827, opened 2024-02-19
Remove nested gin status code writer in `cliRegisterNamespace`

### LINKED ISSUE #514: Don't send http response in `keySignChallenge`, `keySignChallengeInit`, and `keySignChallengeCommit`
Right now, we randomly call `ctx.JSON` or `ctx.AbortWithJSON`, or something similar when we catch an error in `keySignChallenge`, `keySignChallengeInit`, and `keySignChallengeCommit`. However, `keySignChallengeInit`, and `keySignChallengeCommit` are helper functions used in `keySignChallenge`, and `keySignChallenge` is the helper function used in `cliRegisterNamespace` which is the actual gin http handler for namespace registration. We also return error in the above functions and double-checked in their parent functions and so on... It's a bit scary to have nested responses in a single request, which will cause confusion at the client.

So we want to only populate errors in the child functions and handle http response in `NamespaceRegister` function. If we want to specify different status code based on the error, we should provide a new return variable like `errorCode`, in each child function to indicate.

haoming29:
Fixes #514 

Easy test is to run `pelican namespace register --prefix <>` and make sure you got 200 as expected. Or you may run `--prefix` with a bad namespace to see if it returns error.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -123,6 +123,21 @@ type registrationData struct {
 	DeviceCode       string          `json:"device_code"`
 	Prefix           string          `json:"prefix"`
 }
+type permissionDeniedError struct {
+	Message string
+}
+
+type badRequestError struct {
+	Message string
+}
+
+func (e permissionDeniedError) Error() string {
+	return e.Message
+}
+
+func (e badRequestError) Error() string {
+	return e.Message
+}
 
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
Love the added comments. Works when I test it and otherwise LGTM

PR #828, opened 2024-02-20
Add logrotate file

### LINKED ISSUE #728: Add log rotation for pelican/osdf service log files
#584 creates log files in /var/log; these should be rotated. Add logrotate config files for the pelican services created in that PR.

matyasselmeci:
This rotates at 500 MB, 10 times, but given that distros by default run logrotate only once a day, this just means that light days won't rotate out heavy days; we still need to be conservative about log sizes so a sudden spike won't use up all the disk.

Closes #728


brianhlin (approved):
LGTM!

PR #831, opened 2024-02-20
Add parameters file validation

CannonLock:


CannonLock:
Proof it works: https://github.com/PelicanPlatform/pelican/actions/runs/7978563040/job/21783950012?pr=831

docs/parameters.yaml
```diff
@@ -128,6 +128,7 @@ description: >-
   will result in the IP address "123.234.123.234" being mapped to Madison, WI, and IP addresses in the range ABCD::0000-FFFF will be mapped
   to a field in Kansas.
 type: object
+objectStructure: GeoIPOverride
```
haoming29, line 131: I would suggest for the object parameters, we use parameter name for their object type instead of having a separate field that's mostly the same name as the parameter name, unless it's needed for something else.

docs/parameters.yaml
```diff
@@ -706,6 +707,7 @@ description: >-
   for detail
 type: int
 default: 1000
+components: ["director"]
```
haoming29, line 710: Thanks for filling this out!

haoming29 (changes requested):
Small change requested. Let me know if that makes sense!

docs/parameters.yaml
CannonLock: @haoming29 So are you saying in this case it would be

type: "GeoIPOverride"

haoming29: Nah, I meant to remove `objectStructure` field from `parameters.yaml` and in the python script, use `GeoIPOverride` as the item for `VERIFIED_OBJECT_STRUCTURES`. Similarly, we can do `Registry.Institutions`, `Issuer.OIDCAuthenticationRequirements` as the other items. Does that make sense?

CannonLock: Great idea, I changed it to read the object name from the end of the period delimited string so we don't have to duplicate things. This also allows another area to use the  Institutions object without additional work.

haoming29 (approved):
LGTM. Thanks for adding this check!

PR #823, opened 2024-02-16
Add project to Prometheus file transfer metric

haoming29:
Closes #797 

To test, rename your `pelican` binary to `osdf` and set `Federation.DiscoveryURL` to `osg-htc.org`, and run your origin.

Login to the web UI, then go to `https://<origin-host>:<origin-port>/metrics` endpoint. Wait for a couple of minutes, refresh the page, and at the bottom of the page, you should be able to see the following metrics.

Note that now the `proj` label is available which is the `User-Agent` header from the client requesting the object.

```console
# TYPE xrootd_transfer_bytes counter
xrootd_transfer_bytes{ap="https",dn="",org="",path="/.well-known",proj="",role="",type="read"} 18709
xrootd_transfer_bytes{ap="https",dn="",org="",path="/.well-known",proj="",role="",type="readv"} 0
xrootd_transfer_bytes{ap="https",dn="",org="",path="/.well-known",proj="",role="",type="write"} 0
xrootd_transfer_bytes{ap="https",dn="",org="",path="/.well-known",proj="Go-http-client/1.1",role="",type="read"} 216
xrootd_transfer_bytes{ap="https",dn="",org="",path="/.well-known",proj="Go-http-client/1.1",role="",type="readv"} 0
xrootd_transfer_bytes{ap="https",dn="",org="",path="/.well-known",proj="Go-http-client/1.1",role="",type="write"} 0
xrootd_transfer_bytes{ap="ztn",dn="origin",org="https://0be1a304b5d7:8443",path="/pelican",proj="",role="",type="read"} 19654
xrootd_transfer_bytes{ap="ztn",dn="origin",org="https://0be1a304b5d7:8443",path="/pelican",proj="",role="",type="readv"} 0
xrootd_transfer_bytes{ap="ztn",dn="origin",org="https://0be1a304b5d7:8443",path="/pelican",proj="",role="",type="write"} 0
# HELP xrootd_transfer_operations_count Number of transfer operations performed
# TYPE xrootd_transfer_operations_count counter
xrootd_transfer_operations_count{ap="https",dn="",org="",path="/.well-known",proj="",role="",type="read"} 106
xrootd_transfer_operations_count{ap="https",dn="",org="",path="/.well-known",proj="",role="",type="readv"} 0
xrootd_transfer_operations_count{ap="https",dn="",org="",path="/.well-known",proj="",role="",type="write"} 0
[...]

PR #833, opened 2024-02-20
Remove debugging log from xrootd metrics handler

haoming29:
In #823, the debug message was kept in the commit. This PR removes such debug log.


bbockelm (approved):
LGTM.

Issue #753, opened 2024-02-03
Web-based configuration changes


bbockelm:
We should allow edits of the Pelican configuration through the web UI.

Beyond the obvious browser UI changes to allow admins to edit and hit "save", the other parts of this infrastructure should be:

1. Add a generated configuration file - clearly stating it is automatically overwritten and named accordingly - that is loaded LAST after all the existing configuration.
2.  Add a SetConfig API to set multiple variables (the backend equivalent of a "save").  When invoked, it should atomically write out a new set of generated config overrides.
3.  After the SetConfig API is invoked, we should have Pelican restart itself.  A CancelWithCause should be invoked with a well-known cause. The very top level handler should inspect the cause of the cancel and, if it's a config restart, re-exec the process. To have the UI know when the settings are in effect, we should have an API that indicates the time of last start and poll for it to change.

Restarting the server is a bit ugly.  However, our param subsystem isn't currently thread-safe (fixable) and we would need to review it use to see if some of the values are cached long-term (hard).  The really tough aspect is XRootD, however: we'd need to track which variables are xrootd-related and be able to bounce those services.

That might be doable in the future: for now, let's simply restart.

PR #834, opened 2024-02-20
Web-based config change - backend support

haoming29:
This should close part of the ticket #753 with all necessary backend support added.

Test is a bit tricky and a bit complicated:

1. Build and spin up one of your server (I used registry as the example below but it work for all servers)
2. Login to your server web ui **with your admin account** (You may test with a general user account but the API will return 403)

If you are using [postman](https://www.postman.com/downloads/), here is the instruction:
1. After login, `inspect` the current page in your browser (cmd+opt+i for Chrome) 
2. Go to `Application` tab
3. Find on the left side `Cookie` and go to the domain where your server is hosted (likely `localhost:xxxx`

Screenshot 2024-02-20 at 7 29 36 PM

4. Copy the value of the `login` cookie 
5. Go to postman and create a new HTTP session with `PATCH` action and `https://localhost:<port>/api/v1.0/config`
6. Paste the copied value as a new cookie named `login` in the http session.
7. For request body, go to `Body` tab under the request URL and select `raw` with `JSON`
8. Input your test request in JSON form
9. Hit `Send`
10. If succeeded, your service should be automatically restarted, and when you go to the config page of the UI, the changes should be reflected.

Or if you prefer using `curl`, here is the instruction:
1. Similar to the Postman section, grab the `login` cookie and paste it to your curl request as the cookie value.
2. You may find it very difficult to send JSON body via curl, but you can prepare the data in [online JSON serializer](https://codebeautify.org/json-serialize-online) first

### DIFF
--- a/web_ui/frontend/app/api/docs/pelican-swagger.yaml
+++ b/web_ui/frontend/app/api/docs/pelican-swagger.yaml
@@ -375,6 +375,66 @@ paths:
                   Value: ""
         "401":
           description: Unauthorized
+    patch:
+      tags:
+        - "common"
+      summary: Update the server configuration with user-provided changes
+      description:
+        "`Authentication Required` `Admin Previlege Required`
+
+
+        The server will restart promptly after a  successful request. You may query against
+        `/health` endpoint to check if the server is back online.
+
+
+        Note that the endpoint will update the value of **all parameters provided**, no matter the value is in default or not (`0`, `null`,`undefined`, etc).
+
+
+        This means you want to eliminate parameters from the request body that are untouched."
+      parameters:
+        - in: body
+          name: config
+          required: true
+          description:
+            The config value to update.
+            It must be a JSON object with keys/values that match config parameters' name and data type.
+            Parameter names are case-insensitive.
+          schema:
+            type: object
+            example:
+              Logging:
+                Cache:
+                  Scitokens: "info"
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      responses:
+        "200":
+          description: Successfully updated the config. Server is restarting.
+          schema:
+            type: object
+            $ref: "#/definitions/SuccessModel"
+        "400":
+          description: Bad request. The requested changes of config value are invalid.
+          schema:
+            type: object
+            $ref: "#/definitions/ErrorModel"
+        "401":
+          description: Unauthorized. You need you login to take this action.
+          schema:
+            type: object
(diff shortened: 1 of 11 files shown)


web_ui/frontend/app/api/docs/pelican-swagger.yaml
```diff
@@ -375,6 +375,66 @@ paths:
...
+      produces:
+        - application/json
+      responses:
+        "200":
+          description: Successfully updated the config. Server is restarting.
+          schema:
+            type: object
+            $ref: "#/definitions/SuccessModel"
+        "400":
+          description: Bad request. The requested changes of config value are invalid.
```
turetske, line 419:
```suggestion
          description: Bad request. The requested changes to config values are invalid.
```

web_ui/frontend/app/api/docs/pelican-swagger.yaml
```diff
@@ -375,6 +375,66 @@ paths:
...
+          schema:
+            type: object
+            $ref: "#/definitions/ErrorModel"
+        "401":
+          description: Unauthorized. You need you login to take this action.
```
turetske, line 424:
```suggestion
          description: Unauthorized. You need to login to take this action.
```

web_ui/frontend/app/api/docs/pelican-swagger.yaml
```diff
@@ -375,6 +375,66 @@ paths:
...
+          schema:
+            type: object
+            $ref: "#/definitions/ErrorModel"
+        "403":
+          description: Forbidden. You need admin previlege to take this action.
```
turetske, line 429:
```suggestion
          description: Forbidden. You need admin privileges to take this action.
```

web_ui/ui.go
```diff
@@ -71,6 +73,52 @@ func getConfigValues(ctx *gin.Context) {
 	ctx.JSON(200, configWithType)
 }
 
+func updateConfigValues(ctx *gin.Context) {
+	updatedConfig := param.Config{}
+	updatedConfigMap := map[string]interface{}{}
+
+	// Check if the request data is valid config params
```
turetske, line 80:
```suggestion
	// Check if the request data is a valid config
```

turetske (changes requested):
Looks good except for some minor grammar/spelling fixes

turetske (approved):

PR #840, opened 2024-02-22
Added support for Grafana

### LINKED ISSUE #839: Add support for Grafana visualization of Prometheus metrics
Pelican Service:
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

_A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]_

We want to support Grafana visualization of Prometheus metrics but we are still having some distance before getting there:
* Add knob to turn off authentication requirement for prometheus promQL. We want to have this flexibility
* Provide a recipe for generating valid token to access promQL endpoint.
* Reverse proxy Prometheus requests to their default routes at a different port:
  * This is the most tricky part. When I was testing locally, Grafana's built-in Prometheus exporter reports 404 to all the requests to our Prometheus instance, as it assumes that the base URL for Prometheus is at `http//localhost:9090`. It will naively append `/api/ v1/` to all the requests which results in a 404.

haoming29:
Closes #839 

The PR finished the first and third point in the ticket. A token generation recipe should best be located in the documentation world rather than in this PR. It would also be beneficial to have instructions on how to set up a Grafana visualization of Pelican server metrics, but that should also go to the docs world.

To test, I managed to run a local Grafana and build a dashboard for origin server for reference:

Screenshot 2024-02-22 at 8 47 29 AM

### DIFF
--- a/config/resources/defaults.yaml
+++ b/config/resources/defaults.yaml
@@ -65,6 +65,7 @@ Monitoring:
   TokenExpiresIn: 1h
   TokenRefreshInterval: 5m
   MetricAuthorization: true
+  PromQLAuthorization: true
   AggregatePrefixes: ["/*"]
 Shoveler:
   MessageQueueProtocol: amqp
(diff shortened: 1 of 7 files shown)


jhiemstrawisc (approved):
LGTM. I haven't tested the actual integration with Grafana bits, but otherwise this all seems reasonable

PR #841, opened 2024-02-22
Add user info and client IP address to namespace registration from CLI and server auto-registration

### LINKED ISSUE #612: Populate `AdminMetadata.UserID` when user registers ns via CLI with `--with-identity` flag
For users who register their namespace via `pelican namespace register --identity`, they are authenticated against CILogon too, and we will have a `Identity` JSON string from CILogon documenting various user information, including the same `sub` field we used when user register a namespace through UI. We should somehow unmarshall that JSON string and read `sub` field from it, add it to `UserID` field of the registration, so that things are synced.

haoming29:
Closes #612 

This PR adds the following to a namespace registration via `./pelican namespace register` and server auto-registration:

* `AdminMetadata.Description` noted that the registration is from CLI or auto-register which may contain missing fields. If the registration is from CLI with `--with-identity` flag, then it will likely show user email as part of the description
* `AdminMetadata.UserID` is populated by the `sub` field of CILogon user info endpoint response when the registration is from CLI with `--with-identity` flag. This is consistent with the web-based registration
* `AdminMetadata.SiteName` is populated by the client IP address

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -318,6 +318,37 @@ func keySignChallengeCommit(ctx *gin.Context, data *registrationData) (bool, map
 		ns.Pubkey = string(pubkeyData)
 		ns.Identity = data.Identity
 
+		if data.Identity != "" {
+			idMap := map[string]interface{}{}
+			err := json.Unmarshal([]byte(data.Identity), &idMap)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


turetske (approved):
LGTM!

Issue #744, opened 2024-02-02
Cleanup use of federation metadata in client / avoid race conditions


bbockelm:
When cleaning up #721, it became obvious that the client code structure is wrong in its handling of metadata.

The client uses the global discovery metadata.  However, the "correct" director, for example, may change URL-by-URL -- and if we are downloading multiple URLs in parallel, there's going to be race conditions setting/unsetting the global director location for each URL we parse.

Instead of reaching into the global we should create a URL-like structure that holds the metadata along with the object name.  We should also keep a ttlcache of metadata for different federations to prevent frequent lookups if we are parsing many URLs within the process.

PR #843, opened 2024-02-22
Refactor client to have an asynchronous core

bbockelm:
This overhauls the existing client module to have a completely asynchronous core, allowing things like object listings to proceed independently while downloads are ongoing.

It removes global state, allowing more than one client process to run in parallel.  Progress bars are refactored out of the guts of the client and moved to the command line utilities.


joereuss12 (changes requested):
A few items:
1. I ran into an issue when running: `pelican object get/put osdf:///...`, just that prefix/url combo does not seem to work which is most likely an issue with url metadata discovery. While this probably gets fixed by #744, I'm not sure if you want it to be broken with this PR. Not entirely sure why the unit tests are passing though but when I run it by hand the last debug message I get is: `DEBUG[2024-02-26T16:57:12Z] Federation URL is unset; skipping discovery` and nothing else.

2. I'm a fan of a good amount of comments, mostly because it helps me read thru the code a lot faster and minimizes having to relearn what snippets of code are doing each time I read thru. So I mentioned a few areas that I feel could use a little more comments.

3. Overall I really like these changes, it'll probably take me a while to get used to them but I do really like it and it is something I feel the client really needed.

PR #844, opened 2024-02-22
Reorganize project to have less intertwined internal dependency

### LINKED ISSUE #559: Replace existing ad-hoc token creation and verification by a utils.CreateToken / utils.VerifyToken function
As one more step to finish #159 , we want to replace existing token creation logic in the codebase by one standard `utils.CreateToken` function to improve consistency.

Looking at `utils.CheckAnyAuth`, I think a couple of issues need to be addressed before the replacement:
* Rename the function to `utils.VerifyToken`
* Potentially rename the file name from `server_auth` to `token_validation`
* Besides the boolean return value, add a new return value to reflect the error or failure reason (if any) in the chain of checking
  * I'm thinking we can have a struct like 
  ```go
    type TokenValidationError struct {
        Type ValidationErrorType // string enum type, value = ValidationError or InvalidToken
        Message string // the failure reason/error message
    }
  ```
For `ValidationErrorType`, `ValidationError` is any error happened when validating the token, and  `InvalidToken` is due to incorrect/invalid token. We might want to distinguish this so that when we send the token validation failed http response back, we can give a clearer error message or even error code (400/500/401) 

* You might want to wait for #374 to be merged in first to as that PR changed the way `utils.CreateToken` works with scopes. #374 should be merged pretty soon.

haoming29:
This PR originally was to work on #559, but quickly I realized that we still have many instances in the codebase where dependency is messed up and put us at the fringe of cyclic import. So this PR took a digression to focus on cleaning up unnecessary import of server packages to other internal packages, as well as pulling commonly used code to correct places.

The long term goal should be that each server package should _not_ import other server package, e.g. `origin_ui` package should not import `director` package, etc. The next PR to close #559 should move us forward to this goal even closer.

The result dependency graph is not perfect but has a better shape:
[graph-cleaned.pdf](https://github.com/PelicanPlatform/pelican/files/14378593/graph-cleaned.pdf)


bbockelm:
@haoming29 - can you clear up the merge conflicts?

haoming29:
> @haoming29 - can you clear up the merge conflicts?

Done

registry/client_commands.go
```diff
@@ -230,7 +231,7 @@ func NamespaceDelete(endpoint string, prefix string) error {
 	// TODO: We might consider moving widely-useful functions like `GetRegistryIssuerURL`
 	//       to a more generic `pelican/utils` package so that they're easier to find
 	//       and more likely to be used.
-	issuerURL, err := director.GetNSIssuerURL(prefix)
+	issuerURL, err := server_utils.GetNSIssuerURL(prefix)
```
jhiemstrawisc, line 234: It looks like you can remove the previous TODO. I think this function has drifted both in name and location.

jhiemstrawisc (changes requested):
One small change request, otherwise looks good to me. At some point soon, I think the team should have a quick sync up to talk about all the utility functions we have so that we can settle on best practices. Things are moving around frequently enough that it's getting hard to keep track of whether particular functions already exist.

jhiemstrawisc (approved):
LGTM. Thanks Haoming!

PR #848, opened 2024-02-23
Add API at the director to return the longest prefix match given a path

### LINKED ISSUE #847: Director should be able to return the longest prefix matched namespace prefix given a path
Pelican Service:
- [x] Director

We want to know what namespace prefix is for a specific object in the federation or directory. At it seems that the director is the best place for now to answer this question. So Director should be able to return the longest prefix matched namespace prefix given a path. This also helps @jhiemstrawisc on his Lotman integration

haoming29:
Closes #847 

The first commit and third in this PR tackles on the massive public functions inside of the director and change them to private. The most related changes to close #847 is the second commit


jhiemstrawisc (approved):
LGTM

PR #849, opened 2024-02-23
Add support for throttling plugin for xrootd cache

joereuss12:
Creates a config knob that if set, enables the throttling plugin for xrootd. Also adds a simple unit test to ensure it is set in config. Settled on the name Cache.Concurrency but that can be easily changed.

### DIFF
--- a/xrootd/xrootd_config_test.go
+++ b/xrootd/xrootd_config_test.go
@@ -23,6 +23,7 @@ package xrootd
 import (
 	"bytes"
 	"context"
+	"io"
 	"io/fs"
 	"os"
 	"path"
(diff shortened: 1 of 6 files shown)


xrootd/xrootd_config_test.go
```diff
@@ -61,10 +78,63 @@ func TestXrootDCacheConfig(t *testing.T) {
...
+
+		// Verify the output
+		file, err := os.Open(configPath)
+		assert.NoError(t, err)
+		defer file.Close()
+
+		content, err := io.ReadAll(file)
+		assert.NoError(t, err)
+		assert.NotContains(t, string(content), "xrootd.fslib throttle default")
+		assert.NotContains(t, string(content), "throttle.throttle concurrency 10")
```
matyasselmeci, line 134:
```suggestion
		assert.NotContains(t, string(content), "throttle.throttle concurrency")
```
I suspect if the text is wrongly in there, then the number is not likely to be 10.

matyasselmeci (approved):
LGTM with one comment. Pre-approving.

joereuss12:
@matyasselmeci do you want me to go ahead and merge? Or do you want to wait for Brian's input?

matyasselmeci:
Let's merge it so I can try it out on sc-cache. (I don't need a release, I'll just build from main.) We have time to change the name before the release if Brian wants us to.

Issue #818, opened 2024-02-14
Cache Authfile creates the basepath multiple times -- possible an origin advertisement issue


turetske:
Pelican Version: 7.5.6

Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [x] Director
- [x] Origin
- [x] Cache
- [ ] Other (please give the detail)

**Describe the bug**
When serving a cache and an origin, if the cache waits a bit after the origin has been served, the sci-tokens authfile creates repeated paths.

Example:

```
audience_json = ["https://70fa1e4d6777:8443"]

[Issuer https://70fa1e4d6777:8443]
issuer = https://70fa1e4d6777:8443
base_path = /turetsky-test0, /turetsky-test0, /turetsky-test0, /turetsky-test0, /turetsky-test0
```

**To Reproduce**
Serve a pelican origin:

```
./pelican origin serve -p <port> -v <volume>:<namespace>
```

Wait a bit, ~10 minutes

Serve a pelican cache:

```
./pelican cache serve -p <port>
```

Then view the cache's generated scitokens file.

**Expected behavior**

Even if the origin advertises multiple times, the namespace shouldn't be repeated, either in the director or the cache. So it should look like:

```
audience_json = ["https://70fa1e4d6777:8443"]

[Issuer https://70fa1e4d6777:8443]
issuer = https://70fa1e4d6777:8443
base_path = /turetsky-test0
```

**Desktop (please complete the following information):**
 - OS: Linux

**Additional context**
While this is showing up in the cache, I suspect it's an issue with the director and the periodic re-advertisement. Either in the cache or the origin.

Since duplicating the namespace doesn't break anything, everything still functions as normal, it is a growing memory issue.

turetske:
Assigning this to Haoming if only because I'm pretty sure his current PR is already closing it.

haoming29:
Yes. I confirmed that with the fix in #852, authfile in cache has no duplicated namespaces

PR #852, opened 2024-02-23
Fix duplicated server advertisement at the director

### LINKED ISSUE #851: Director registers duplicated server advertisement and issued multiple file transfer test on a single origin
Pelican Version:
main branch as in a176ae18544219513c0b8f8e638030d06a5f76b2

Pelican Service:
- [x] Director

**Describe the bug**

The director registers duplicated server advertisement at each time a new server advertisement comes from the same origin  and issued multiple file transfer test on a single origin.

**To Reproduce**

* Run the federation-in-a-box, go to `http://localhost:8444` and login. Select director
* Initially there's only one server available as origin
* Wait for 1 minute, refresh the UI
* The same origin appears twice in the origin list

**Expected behavior**

Same origin should only be registered once.

**Screenshots**

 - OS: [e.g. Linux, MacOS]

**Additional context**

haoming29:
Fixes #851 

The cause of the issue is that in #705, when we added `brokerUrl`, we used `*url.URL` instead of `url.URL`. When broker service is turned off, `brokerUrl` is empty. When the director reads an empty url string, parses it as an empty url instance, and _pass the pointer to the empty url instance_ to the `serverAd`, the `serverAd` records the **address** of the empty `url.URL` instance, instead of just an empty url.

This causes bug because in the following registration from the same server, that address of empty url will change, resulting in a different `serverAd` than before, and all the map structure we have took it as a new key.

### DIFF
--- a/common/director.go
+++ b/common/director.go
@@ -68,9 +68,9 @@ type (
 	ServerAd struct {
 		Name               string
 		AuthURL            url.URL
-		BrokerURL          *url.URL // The URL of the broker service to use for this host.
-		URL                url.URL  // This is server's XRootD URL for file transfer
-		WebURL             url.URL  // This is server's Web interface and API
+		BrokerURL          url.URL // The URL of the broker service to use for this host.
+		URL                url.URL // This is server's XRootD URL for file transfer
+		WebURL             url.URL // This is server's Web interface and API
 		Type               ServerType
 		Latitude           float64
 		Longitude          float64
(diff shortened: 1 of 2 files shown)


turetske:
This might also fix: https://github.com/PelicanPlatform/pelican/issues/818. Can you test and see?

haoming29:
> This might also fix: #818. Can you test and see?

Yes, I can confirm this will effectively fix #818

bbockelm (approved):
Whoa, this appears to be enormously brittle.

This appears to "stop the bleeding" now but I think I'll file a separate ticket to make this more robust.

I'm not convinced we should be keying off the entire ServerAd, especially as we'll want to create dynamic portions of the ad later (representing, say, server load or number of errors).

PR #860, opened 2024-02-28
Added error when discovery-url contains a path

joereuss12:
When a discovery-url contains a path (when set in config or with `-f`), we will throw an error when we call DiscoverFederation on that url. If it is contained within a `pelican://` url however, we cannot really catch that specific error since we will assume any paths added to the discovery-url to be the namespace-prefix and so on. Therefore, I added a few notes to the documentation alerting that this will fail.

Note: I realize the client-usage docs are out of date but I added this for when we refresh the docs

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -291,6 +291,8 @@ func TestDiscoverFederation(t *testing.T) {
 		viper.Set("Federation.DiscoveryUrl", server.URL+"/this/is/some/path")
 		err := DiscoverFederation()
 		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "Invalid federation discovery url is set. No path allowed for federation discovery url. Provided url: ",
+			"Error returned does not contain the correct error")
 		viper.Reset()
 	})
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -19,6 +19,7 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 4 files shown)


bbockelm (changes requested):
Please add a test for setting the federation URL to `osg-htc.org` (or similar -- just the hostname but not specifying the scheme).

Go parses this form as a path with no hostname and, I suspect, this change breaks that case (which is the default for OSDF).

joereuss12:
@bbockelm I think I need the okay for this one since you requested changes.

config/config_test.go
```diff
@@ -261,3 +262,61 @@ func TestEnabledServers(t *testing.T) {
...
+		w.WriteHeader(http.StatusOK)
+		_, err = w.Write(responseJSON)
+		assert.NoError(t, err)
+	}))
+	defer server.Close()
+	t.Run("testInvalidDiscoveryUrlWithPath", func(t *testing.T) {
+		viper.Set("tlsskipverify", true)
+		viper.Set("Federation.DiscoveryUrl", server.URL+"/this/is/some/path")
+		err := DiscoverFederation()
+		assert.Error(t, err)
```
jhiemstrawisc, line 293: It's nice to have our tests not only assert that there _is_ an error, but that the correct error is returned. It's also nice to look at these types of tests as an index of "if I do XXX I get an error that says YYY". Can modify this to use something like `assert.ErrorIs(t TestingT, err, target error, msgAndArgs ...interface{})`?

jhiemstrawisc (changes requested):
Just a few small things.

jhiemstrawisc (approved):
LGTM. Thanks Joe!

jhiemstrawisc:
Bypassing branch protections to avoid waiting for the extra review since I took point on this one.

Issue #859, opened 2024-02-28
Cache takes `Origin.Port` instead of `Cache.Port` as the port number for xrootd


haoming29:
Pelican Version:
_Run `./pelican --version` to get version number._

Pelican Service:
- [ ] Cache

**Describe the bug**
_A clear and concise description of what the bug is._

Setting `Cache.Port` on a cache won't work but `Origin.Port` works. 

**To Reproduce**
_Steps to reproduce the behavior._

Spin up your fed-in-a-box with origin, then spin up a cache in the same container. Change `PELICAN_SERVER_WEBPORT` to non-8444 to avoid web port collision and `PELICAN_CACHE_PORT` to something different from `8443`. However, cache won't start up still with the error `Port 8443 already in use`

**Expected behavior**
_A clear and concise description of what you expected to happen._

**Screenshots**
_If applicable, add screenshots to help explain your problem._

**Desktop (please complete the following information):**
 - OS: [e.g. Linux, MacOS]
 - Browser: [e.g. chrome, safari, if the bug relates to any web user interface]

**Additional context**
_Add any other context about the problem here._

PR #861, opened 2024-02-28
Self-test for cache

### LINKED ISSUE #544: Periodic self tests for cache
Add periodic self tests for cache, similar to what origin does.

haoming29:
Closes #544 

To test, run your federation with cache at `debug` level and wait for a couple of seconds, there should be a debug message popping up:

```
Self-test monitoring cycle succeeded at Wed Feb 28 17:50:29 UTC 2024 
```

To enable direct-write-to-cache, this PR migrated XRootD cinfo struct to Pelican and auto-generate a cinfo file along with the test file to allow cache to discover the directly added file 

This PR also introduced two params to turn off the self-test for cache and change the test frequency:

```
Cache.SelfTest
Cache.SelfTestInterval
```

Finally, this PR fixes two bugs that 1. #859  `Origin.Port` instead of `Cache.Port` was assigned to the xrootd port for cache. 2. `Origin.Url` was used to advertise cache but that takes `Origin.Port` only


cache_ui/cinfo.go
```diff
@@ -0,0 +1,122 @@
...
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"hash/crc32"
+
+	"github.com/pkg/errors"
+)
+
+type (
+	CInfo struct {
```
bbockelm, line 13: Do not export the structs
```suggestion
	cInfo struct {
```

cache_ui/cinfo.go
```diff
@@ -0,0 +1,122 @@
...
+		Store            Store
+		cksum            uint32
+		buffSynced       []byte
+		cksumSyncedAstat uint32
+	}
+
+	Store struct {
```
bbockelm, line 20: Same as above - do not export.
```suggestion
	store struct {
```

cache_ui/cinfo.go
```diff
@@ -0,0 +1,122 @@
...
+		CreationTime int64 // Unix time in seconds
+		NoCkSumTime  int64 // Unix time in seconds
+		AccessCnt    uint64
+		Status       uint32 // Checksum status
+		AStatSize    int32  // Should always be unset (0) unless you know what you are doing
+	}
+)
+
+const (
+	DefaultCinfoVersion = 4
```
bbockelm, line 32:
```suggestion
	defaultCinfoVersion = 4
```

cache_ui/self_monitor.go
```diff
@@ -0,0 +1,236 @@
...
+		return err
+	}
+
+	gid, err := config.GetDaemonGID()
+	if err != nil {
+		return err
+	}
+
+	basePath := param.Cache_DataLocation.GetString()
+	pelicanMonPath := path.Join(basePath, "/pelican")
```
bbockelm, line 39: Use `filepath` for files on disk, `path` for URLs:

```suggestion
	pelicanMonPath := filepath.Join(basePath, "/pelican")
```

On Unix, they are the same but this will reduce the work we would have to do for any future Windows port.

cache_ui/self_monitor.go
```diff
@@ -0,0 +1,236 @@
...
+	extFilePath := path.Join(selfTestDir, testFileName)
+	fileCinfoPath := path.Join(monitoringPath, testFileCinfoName)
+
+	file, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE, 0600)
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed to create self-test file %s", filePath)
+	}
+	defer file.Close()
+
+	cinfoFile, err := os.OpenFile(fileCinfoPath, os.O_WRONLY|os.O_CREATE, 0600)
```
bbockelm, line 98: There's a race condition here -- xrootd, when doing its normal directory walk, may try to parse the file after it's created but before it's been written to.

Create a temporary file (`blah.cinfo.tmp`), populate it, then do an `os.Rename` to put it in place atomically.

cache_ui/self_monitor.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if _, err := file.Write(testFileBytes); err != nil {
+		return "", errors.Wrapf(err, "Failed to write test content to self-test file %s", filePath)
+	}
+
+	if _, err := cinfoFile.Write(cinfoBytes); err != nil {
+		return "", errors.Wrapf(err, "Failed to write cinfo content to self-test cinfo file %s", fileCinfoPath)
+	}
+
+	if err = os.Chown(filePath, uid, gid); err != nil {
```
bbockelm, line 112: Use [File.Chown](https://pkg.go.dev/os#File.Chown) instead.  That will chown the open file handle.

cache_ui/self_monitor.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return baseUrl.String(), nil
+}
+
+func downloadTestFile(ctx context.Context, fileUrl string) error {
+	req, err := http.NewRequestWithContext(ctx, "GET", fileUrl, nil)
+	if err != nil {
+		return errors.Wrap(err, "Failed to create GET request for test file transfer download")
+	}
+	// req.Header.Set("Authorization", "Bearer "+tkn)
```
bbockelm, line 135: Please create a follow-up issue to upgrade the test to add authorization.

bbockelm (changes requested):
Quite minor changes requested.

I suggest we start without exporting the cInfo / store data structure.  I would eventually like to use them for writes to the cache from the EP -- but until we _have to_ do the export, we should keep it private.

haoming29:
@jhiemstrawisc might need your superpower to merge this PR given Brian's requested changes which I should have all tackled

jhiemstrawisc (approved):
LGTM and it feels good to have more test coverage in the cache. Thanks Haoming!

jhiemstrawisc:
Bypassing branch protections to merge this one since I took point on this one.

PR #882, opened 2024-03-01
Fixes to pelican plugin staging

joereuss12:
Changes:
- Found that how condor gives hooks a classad is in a different format than the plugin for some reason. For the hook the classad is split up by newlines but for the plugin it is by ';'. Therefore, made a new function for plugin stage to utilize.
- Added function within the plugin to force logging to go to stderr (even though that is the default) as well as a file called `.PelicanPlugin.log`. This extra file is something you can specify in a condor submit file in `transfer_output_files` to get it back but only works if the job does not go on hold/if the plugin does not fail (not super helpful yet)


joereuss12:
Note: waiting until I write a unit test for review

joereuss12:
Since I have to wait until #798 is merged for a test to work. Marking as ready for review so we can get the bug fixes and refactoring in

jhiemstrawisc (approved):
Overall seems good to me. Comments were very helpful to figure out what was going on, thanks!

PR #884, opened 2024-03-02
Add a local cache module

bbockelm:
This adds a "local cache" module which provides simple caching functionality via a Unix domain socket.

Functionality:
- The local cache invokes the built-in client and downloads to a dedicated directory.
- Subsequent requests for the same object should come from the local directory.
- Only full-file requests are supported.
- Only supports token-based authorization; authorization policies are synchronized from the director.

PR is currently a draft; needs a good amount of testing and bugfixes.


bbockelm:
Marking as ready for review.  Since original draft was posted, this has gained a reasonably test suite and the first round of bugfixes.

(Due to OSG Hub outages, it's not possible for unit tests to pass currently -- will continue to clean those up but it should be ready for a first-pass of a review.)

joereuss12:
One thing I noticed when writing a unit test for StatHttp():
https://github.com/PelicanPlatform/pelican/blob/459b7a60c57ec53840813820ace99c10f393f0e6/client/main.go#L198-L222

In this code snippet, you don't list `stash` as a understood scheme yet check for the `stash` scheme later on in the code. Do we want to allow `stash` as a valid scheme here?

tannenba:
Nice !!!  Some quick questions: Looking at the PR, it appears one can also set the size of the cache and a high/low water mark for when to do garbage collection ... is this using LRU or?  If this module is activated, how does it impact meta data reported back to the stash_plugin...  e.g. are there new/different error codes?  can the client report back if the file was pulled from the local cache or not?  Also I am reading it correctly that all authorization rules are honored as before, i.e. user X cannot access files limited only to user Y?  How is that enforced, i.e. what prevents user X from simply doing a chdir to the directory containing the data cache and reading files?

bbockelm:
> In this code snippet, you don't list stash as a understood scheme yet check for the stash scheme later on in the code. Do we want to allow stash as a valid scheme here?

This appears to be an issue in the current code.  I've fixed it as part of the commit series; it should be symmetric as far as I can tell.

bbockelm:
> is this using LRU or? 

Yes, this uses LRU.

> If this module is activated, how does it impact meta data reported back to the stash_plugin... e.g. are there new/different error codes? can the client report back if the file was pulled from the local cache or not?

The client treats the local cache as if it were any other cache, reporting-wise, meaning you get back the "cache hostname" from the unix socket URL (e.g., it's empty as the URL is of the form `unix:///foo/bar`).  I actually see that as a weakness: if you want to understand failures, you probably want to see the upstream caches used by the local cache, not the fact the local cache was used.

However, this PR is already obnoxiously large -- I'm OK with getting this in now and then tuning the plugin outputs.

> Also I am reading it correctly that all authorization rules are honored as before, i.e. user X cannot access files limited only to user Y?

Correct -- well, that is, the authorization rules are honored by the cache if you talk to the cache.

> How is that enforced, i.e. what prevents user X from simply doing a chdir to the directory containing the data cache and reading files?

If the user can read the cache's files directly, it can bypass the access control.  That leaves two options:
1. ("EP running as root") Make sure the cache is running as a different Unix user.
2. ("EP launched by glidein") Make sure the job is running inside a container.

I feel that's sufficient for now.  However, since the cache is really designed to be transient (it deletes the cache directory's contents on startup if it exists), some follow-up work could be to encrypt at rest with an in-memory-only encryption key.  In that case, one would have to ptrace the cache process to steal the keys -- a higher level of sophistication.

local_cache/cache_test.go
```diff
@@ -0,0 +1,535 @@
...
+func (ft *fedTest) spinup(t *testing.T, ctx context.Context, egrp *errgroup.Group) {
+
+	modules := config.ServerType(0)
+	modules.Set(config.OriginType)
+	modules.Set(config.DirectorType)
+	modules.Set(config.RegistryType)
+	// TODO: the cache startup routines not sequenced correctly for the downloads
+	// to immediately work through the cache.  For now, unit tests will just use the origin.
+	viper.Set("Origin.EnableFallbackRead", true)
+	/*
```
haoming29, line 68: Could you make a note on the commented out code? Or is it something to clean up

local_cache/cache_test.go
```diff
@@ -0,0 +1,535 @@
...
+	"github.com/pelicanplatform/pelican/utils"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
+)
+
+type (
+	fedTest struct {
```
haoming29, line 53: Seeing another fedTest instance in client, I think it's a good time for us to pull them into test_utils package to reduce repetition

haoming29 (changes requested):
Minor changes requested. I'm not super familiar with client side logic so @joereuss12 might also want to take a look, just to ensure.

client/fed_test.go
```diff
@@ -0,0 +1,847 @@
+//go:build !windows
+
+/***************************************************************
+ *
+ * Copyright (C) 2023, University of Nebraska-Lincoln
```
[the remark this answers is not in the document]
bbockelm: (I'm going to mark these as resolved as I make the changes on my local copy, then push in bulk... so don't be surprised if you don't see new commits right away)

Fixed!

joereuss12:
The client logic LGTM. I mentioned this in the created issue but the duplicate test is deleted in one of my PR's #878. Also with #890 tested off of this branch, it seems statHttp seems to be working how it should as well.

docs/parameters.yaml
```diff
@@ -622,6 +631,57 @@ default: path
...
+description: >-
+  The location of the socket used for client communication for the local cache
+type: filename
+default: $PELICAN_LOCALCACHE_RUNLOCATION/cache.sock
+components: ["localcache"]
+---
+name: LocalCache.Size
+description: >-
+  The maximum size of the local cache.  If not set, it is assumed the entire device can be used.
+type: string
```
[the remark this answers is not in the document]
bbockelm: Actually, it is correct as written (currently); users can specify suffixes to indicate units (20MB, 45GB) which requires the string.

I will add a note in the description along these lines.

haoming29 (changes requested):
A couple of tiny changes requested and some questions. I added the comments in the individual commits and didn't expect they are not showing up in the conversation tab. You should be able to find them in the commits history.

haoming29:
@bbockelm I figured that you'll need to hit the comment icon in the commit history above ⬆️ instead of in the commit tab. Sorry for the hassle!

web_ui/authentication.go
```diff
@@ -153,16 +153,12 @@ func GetUser(ctx *gin.Context) (string, error) {
...
-	loginCookieTokenCfg := token.TokenConfig{
-		TokenProfile: token.WLCG,
-		Lifetime:     30 * time.Minute,
-		Issuer:       param.Server_ExternalWebUrl.GetString(),
-		Audience:     []string{param.Server_ExternalWebUrl.GetString()},
-		Version:      "1.0",
-		Subject:      user,
-	}
-	loginCookieTokenCfg.AddScopes(scopes)
+	loginCookieTokenCfg := token.NewWLCGToken()
```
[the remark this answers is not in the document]
haoming29: As discussed with Brian, this should be a good follow-up ticket

client/handle_http.go
```diff
@@ -1120,25 +1184,124 @@ func runTransferWorker(ctx context.Context, workChan <-chan *clientTransferFile,
...
+				size, err = strconv.ParseInt(contentLengthStr, 10, 64)
+				if err != nil {
+					log.Errorln("problem converting content-length to an int:", err)
+				}
+			}
+			headChan <- struct {
+				idx  int
+				size uint64
+				err  error
+			}{idx, uint64(size), nil}
```
haoming29, line 1236: Copy comment from hidden commit history:

> Do we want to return error or nil here if there is an error converting Content-Length header?

client/handle_http.go
```diff
@@ -1120,25 +1184,124 @@ func runTransferWorker(ctx context.Context, workChan <-chan *clientTransferFile,
...
+	for ctr := 0; ctr != len(attempts); ctr++ {
+		result := <-headChan
+		if result.err != nil {
+			if result.err != context.Canceled {
+				log.Debugf("Failure when doing a HEAD request against %s: %s", attempts[result.idx].Url.String(), result.err.Error())
+			}
+		} else {
+			finished[result.idx] = true
+			if result.idx == 0 {
+				cancel()
```
haoming29, line 1249: Copy comment from hidden commit history:

> Trying to understand the logic here. Are we cancelling the reminding request if we get back from the first attempt? Wouldn't that cause the remaining attempts (no matter they succeeded or not) to have finished[idx] = false? Not sure if this will cause unwanted side effect

client/handle_http.go
bbockelm: Good idea.  An error here will sort the cache toward the back (but not be fatal to the transfer).

haoming29 (approved):
LGTM with the revision!

PR #885, opened 2024-03-04
Clean up goroutines when fatal error occurs at server start

### LINKED ISSUE #868: Leak of xrootd/cmsd processes
Pelican Version: 7.5.7

Pelican Service: Origin (probably cache too)

**Describe the bug**
When @brianhlin and @biozit were installing the Pelican origin, both encountered issues with xrootd process leaks.

That is, after an initial failure in startup, the xrootd process was not killed.  Subsequent restarts of pelican from within the same container would then rapidly fail as the new xrootd process was unable to bind to the new port.

**To Reproduce**
@biozit's problem occurred when port 8444 was firewalled off.  This caused the startup health check to fail, the server to shutdown, and xrootd to remain lingering.

Here's a patch that creates an error at the same point in the startup:
```
--- a/launchers/launcher.go
+++ b/launchers/launcher.go
@@ -22,7 +22,6 @@ import (
        "context"
        "fmt"
        "net"
-       "net/http"
        "os"
        "os/signal"
        "syscall"
@@ -192,11 +191,13 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
                shutdownCancel()
                return nil
        })
-
+/*
        if err = server_utils.WaitUntilWorking(ctx, "GET", param.Server_ExternalWebUrl.GetString()+"/api/v1.0/health", "Web UI", http.StatusOK); err != nil {
                log.Errorln("Web engine startup appears to have failed:", err)
                return shutdownCancel, err
        }
+*/
+       return shutdownCancel, errors.New("test shutdown")
 
        if modules.IsEnabled(config.OriginType) {
                log.Debug("Finishing origin server configuration")
```

Run a origin with this setup.  It should crash nearly immediately and you'll notice an xrootd process still running afterward.

haoming29:
Fixes #868 

There are two issues causing the bug:
1. We return immediately if `err := rootCmd.ExecuteContext(ctx); err != nil`, not waiting for error group to clean up.
2. We didn't kill daemons if the `ctx` parameter is cancelled, which is the case when there's error from server goroutines (where it will call shutdownCancel). We only did it either when receiving OS signal or the daemon expires.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -87,7 +87,7 @@ func Execute() error {
 	ctx := context.WithValue(egrpCtx, config.EgrpKey, egrp)
 	exeErr := rootCmd.ExecuteContext(ctx)
 	if exeErr != nil {
-		log.Errorln("Fatal error occurred at the start of the program. Cheanup started:", exeErr)
+		log.Errorln("Fatal error occurred at the start of the program. Cleanup started:", exeErr)
 	}
 	// Wait until all goroutines in errgroup finish their clean up
 	egrpErr := egrp.Wait()
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -103,7 +103,6 @@ func Execute() error {
 		log.Errorln("Fatal error occurred that lead to the shutdown of the process:", egrpErr)
 		return egrpErr
 	} else {
-		fmt.Println("Pelican clean up finished")
 		return exeErr
 	}
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -87,7 +87,7 @@ func Execute() error {
 	ctx := context.WithValue(egrpCtx, config.EgrpKey, egrp)
 	exeErr := rootCmd.ExecuteContext(ctx)
 	if exeErr != nil {
-		log.Errorln("Fatal error occurred at the start of the program. Chean up started:", exeErr)
+		log.Errorln("Fatal error occurred at the start of the program. Cheanup started:", exeErr)
 	}
 	// Wait until all goroutines in errgroup finish their clean up
 	egrpErr := egrp.Wait()
@@ -98,7 +98,7 @@ func Execute() error {
 		fmt.Println("Restarting server...")
 		return restartProgram()
 	}
-	// Other errors we from the errogroup
+	// Other errors we got from the errogroup
 	if egrpErr != nil {
 		log.Errorln("Fatal error occurred that lead to the shutdown of the process:", egrpErr)
 		return egrpErr
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -85,25 +85,27 @@ func (i *uint16Value) String() string { return strconv.FormatUint(uint64(*i), 10
 func Execute() error {
 	egrp, egrpCtx := errgroup.WithContext(context.Background())
 	ctx := context.WithValue(egrpCtx, config.EgrpKey, egrp)
-	err := rootCmd.ExecuteContext(ctx)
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -85,25 +85,26 @@ func (i *uint16Value) String() string { return strconv.FormatUint(uint64(*i), 10
 func Execute() error {
 	egrp, egrpCtx := errgroup.WithContext(context.Background())
 	ctx := context.WithValue(egrpCtx, config.EgrpKey, egrp)
-	err := rootCmd.ExecuteContext(ctx)
-	if err != nil {
-		log.Errorln("Fatal error occurred at the start of the program:", err)
-		return err
+	exeErr := rootCmd.ExecuteContext(ctx)
+	if exeErr != nil {
+		log.Errorln("Fatal error occurred at the start of the program. Chean up started:", exeErr)
```
jhiemstrawisc, line 90: `Chean up` --> `Cleanup`

cmd/root.go
```diff
@@ -85,25 +85,26 @@ func (i *uint16Value) String() string { return strconv.FormatUint(uint64(*i), 10
...
 		return nil
-	} else if err == launchers.ErrRestart {
+	} else if egrpErr == launchers.ErrRestart {
 		fmt.Println("Restarting server...")
 		return restartProgram()
 	}
-	if err != nil {
-		log.Errorln("Fatal error occurred that lead to the shutdown of the process:", err)
-		return err
+	// Other errors we from the errogroup
```
jhiemstrawisc, line 101: `we from` --> `we got from`

jhiemstrawisc (changes requested):

jhiemstrawisc:
Do you know of an easy way I can test that this fixes the xrootd process that occasionally doesn't get cleaned up? I know that when some tests fail, xrootd doesn't get killed, but I don't know which tests to look at (or how to attempt forcing this behavior).

haoming29:
> Do you know of an easy way I can test that this fixes the xrootd process that occasionally doesn't get cleaned up? I know that when some tests fail, xrootd doesn't get killed, but I don't know which tests to look at (or how to attempt forcing this behavior).

Brian has a pretty nice setup in the ticket to force the exit of the program at start #868

jhiemstrawisc:
Okay, confirmed that XRootD doesn't linger after trying Brian's patch. I'll merge once the typos are cleaned up.

cmd/root.go
```diff
@@ -85,25 +85,26 @@ func (i *uint16Value) String() string { return strconv.FormatUint(uint64(*i), 10
 func Execute() error {
 	egrp, egrpCtx := errgroup.WithContext(context.Background())
 	ctx := context.WithValue(egrpCtx, config.EgrpKey, egrp)
-	err := rootCmd.ExecuteContext(ctx)
-	if err != nil {
-		log.Errorln("Fatal error occurred at the start of the program:", err)
-		return err
+	exeErr := rootCmd.ExecuteContext(ctx)
+	if exeErr != nil {
+		log.Errorln("Fatal error occurred at the start of the program. Cheanup started:", exeErr)
```
jhiemstrawisc, line 90: Whoops 😂  `Cheanup` --> `Cleanup`

jhiemstrawisc (changes requested):

cmd/root.go
haoming29: My bad 🫠

jhiemstrawisc (approved):

PR #889, opened 2024-03-05
Add LotMan configuration to the cache

### LINKED ISSUE #817: Create basic set of CRUD actions for LotMan in the cache
[LotMan](https://github.com/pelicanplatform/lotman) is going to be a solution for storage management in distributed high-throughput data environments like the cache. In order to modify lot information in the cache, we should expose a set of RESTful CRUD endpoints through Pelican.

jhiemstrawisc:
This PR adds a variety of ways to interface with Lotman in the cache. 

Note that as of my opening this PR, the CRUD actions exposed via gin don't have tests. While the token verification is very nuanced and undoubtedly contains mistakes, I propose we wait until the next release cycle to work on these tests because there's a cyclic import issue I need to figure out (launchers imports lotman to register everything, and lotman tests would need to import launchers to spin up the entire federation). For now, the API endpoints can be turned off, and the lots are configurable via the pelican config.

Closes #817


jhiemstrawisc:
@Cannon, let me know if you need anything else to define the Lots list in your object UI.

turetske (changes requested):
First major issue. Because you branched off of add-cache-launcher at the beginning before it was finished, there are multiple duplicated commits at the beginning of this pull request.  This will also cause conflicts with the finished version that is currently in the code. I suggest creating a new branch off of main and applying your specific commits (the last 9 as opposed to all 12) to the branch and recreating the PR otherwise merging this in will be a mess.

haoming29:
> First major issue. Because you branched off of add-cache-launcher at the beginning before it was finished, there are multiple duplicated commits at the beginning of this pull request. This will also cause conflicts with the finished version that is currently in the code. I suggest creating a new branch off of main and applying your specific commits (the last 9 as opposed to all 12) to the branch and recreating the PR otherwise merging this in will be a mess.

Since the cache launcher has been merged, it would be easier to rebase against main to forward commits to your new changes, but you might need to fix the merge conflicts in rebasing, if that's too much hassle, then Emma's idea should help

turetske:
> > First major issue. Because you branched off of add-cache-launcher at the beginning before it was finished, there are multiple duplicated commits at the beginning of this pull request. This will also cause conflicts with the finished version that is currently in the code. I suggest creating a new branch off of main and applying your specific commits (the last 9 as opposed to all 12) to the branch and recreating the PR otherwise merging this in will be a mess.
> 
> Since the cache launcher has been merged, it would be easier to rebase against main to forward commits to your new changes, but you might need to fix the merge conflicts in rebasing, if that's too much hassle, then Emma's idea should help

@haoming29 If he had just branched off of my branch, I'd agree with you, but since I think he applied specific commits from my PR rather than branching, I don't this it's possible. This is one of the downsides of working with forked repos rather than on the main repo, it's not easy to branch off of other peoples branches.

I just rebased and those 3 commits are still there. Without any of the potential later fixes I added. This will also add duplicate commits that say the same thing.

turetske (changes requested):

CannonLock (approved):
LGTM, I will tack on a UI in a separate PR.

lotman/lotman.go
```diff
@@ -0,0 +1,655 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+***************************************************************/
+
+package lotman
```
turetske, line 21: Can you add a description of the what lotman is/does? This will be very opaque to someone who isn't a part of our project without it.

lotman/lotman.go
```diff
@@ -0,0 +1,655 @@
...
+
+type (
+	Int64FromFloat struct {
+		Value int64
+	}
+
+	LotPaths struct {
+		Path      string `json:"path" mapstructure:"Path"`
+		Recursive bool   `json:"recursive" mapstructure:"Recursive"`
+		LotName   string `json:"lot_name,omitempty"`
```
turetske, line 79: Can `omitempty` be changed to `omit_empty` for consistency in json value names.

lotman/lotman.go
```diff
@@ -0,0 +1,655 @@
...
+	}
+
+	return &owners, nil
+}
+
+// Initialize the LotMan library and bind its functions to the global vars
+// We also perform a bit of extra setup such as setting the lotman db location
+func InitLotman() bool {
+	// If we've already tried to init the library, return the result of that attempt
+	if lotmanInitTried {
```
turetske, line 307: Why would InitLotman be called multiple times? I wonder if Lotman is enabled, InitLotman should be called once at startup.

lotman/lotman.go
```diff
@@ -0,0 +1,655 @@
...
+
+	defaultInitialized := false
+	rootInitialized := false
+
+	err = param.Lotman_Lots.Unmarshal(&initializedLots)
+	if err != nil {
+		log.Warningf("Error while unmarshaling Lots from config: %v", err)
+	}
+
+	federationIssuer := param.Federation_DiscoveryUrl.GetString()
```
turetske, line 359: I think this federation Issuer fall back to the DirectorUrl may be something that's useful enough to be a utility function.

lotman/lotman.go
```diff
@@ -0,0 +1,655 @@
...
+	var lot Lot
+	err := json.Unmarshal(outputBuf, &lot)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error unmarshalling lot JSON: %v", err)
+	}
+	return &lot, nil
+}
+
+func UpdateLot(lotUpdate *LotUpdate, caller string) error {
+	success := InitLotman()
```
turetske, line 603: Can you do a header for this? What does updating a lot entail? What specifically is being updated?

lotman/lotman.go
```diff
@@ -0,0 +1,655 @@
...
+	i.Value = int64(f)
+	return nil
+}
+
+func (i Int64FromFloat) MarshalJSON() ([]byte, error) {
+	return json.Marshal(i.Value)
+}
+
+func cArrToGoArr(cArr *unsafe.Pointer) []string {
+	ptr := uintptr(*cArr)
```
turetske, line 182: I assume it's converting a C Array to a Go Slice. Can you add a simple header explaining what this is doing?

docs/parameters.yaml
```diff
@@ -1698,3 +1704,63 @@ description: >-
...
+  - /usr/lib64/libLotMan.so
+  - /usr/local/lib64/libLotMan.so
+  - /opt/local/lib64/libLotMan.so
+type: filename
+default: none
+components: ["cache"]
+---
+name: Lotman.EnableAPI
+description: >-
+  Whether Lotman should enable its CRUD web endpoints. If true, administrators with an appropriately signed token interface
```
turetske, line 1733:
```suggestion
  Whether Lotman should enable its CRUD web endpoints. If true, administrators with an appropriately signed token can interface
```

turetske (changes requested):
I have a few clarification comments requests.

lotman/lotman.go
```diff
@@ -0,0 +1,655 @@
...
+	if ret != 0 {
+		trimBuf(&errMsg)
+		return nil, errors.Errorf("Failed to determine %s's parents: %s", lotName, string(errMsg))
+	}
+
+	parents := cArrToGoArr(&cParents)
+
+	owners := []string{}
+	internalOwners := []string{}
+	for _, parent := range parents {
```
turetske, line 280: These parents are only the immediate parents. Don't you need to also get the parents of the parents, etc? I feel like this needs to be a worklist algorithm, not just a loop through the parents.

lotman/lotman.go
```diff
@@ -0,0 +1,655 @@
...
+		LotmanGetLotOwners(parent, true, &cOwners, &errMsg)
+		if ret != 0 {
+			trimBuf(&errMsg)
+			return nil, errors.Errorf("Failed to determine appropriate owners of %s's parents: %s", lotName, string(errMsg))
+		}
+
+		internalOwners = append(internalOwners, cArrToGoArr(&cOwners)...)
+	}
+
+	// Deduplicate the owners
```
turetske, line 291: Rather than making this a slice that you need to deduplicate, you can mimic a set by creating a go map from string to an empty struct and setting the owners as keys.

lotman/lotman_ui.go
```diff
@@ -0,0 +1,573 @@
...
+
+	// Since we're updating the lot, assume it already exists (lotman will yell if it doesn't).
+	// Then we can make sure the token is signed by any owners of the lot's parents
+
+	authzCallers, err := GetAuthorizedCallers(lot.LotName)
+	if err != nil {
+		return false, errors.Wrap(err, "Error getting authorized callers")
+	}
+
+	// Now we have a list of owners who can modify the lot. We need to check that the token is signed by one of them
```
turetske, line 315: Since owner is an overloaded term, maybe manager?

lotman/lotman_ui.go
```diff
@@ -0,0 +1,573 @@
...
+}
+
+func VerifyUpdateLotToken(lot *LotUpdate, strToken string) (bool, error) {
+	tokenApproved := false
+	log.Debugf("Attempting to update lot %s", lot.LotName)
+
+	// Since we're updating the lot, assume it already exists (lotman will yell if it doesn't).
+	// Then we can make sure the token is signed by any owners of the lot's parents
+
+	authzCallers, err := GetAuthorizedCallers(lot.LotName)
```
turetske, line 310: Just to confirm. There will never be a case where any owners of a lot's parents *wouldn't* be an authorize caller and any non-parent own *would*?

And I'm assuming the root lot is its own parent?

lotman/lotman_ui.go
```diff
@@ -0,0 +1,573 @@
...
+	nsIssuerUrl, err := server_utils.GetNSIssuerURL(namespace)
+	if err != nil {
+		return false, errors.Wrapf(err, errMsgPrefix+"no issuer could be found for namespace %s", namespace)
+	}
+	lot.Owner = nsIssuerUrl
+
+	return true, nil
+}
+
+func VerifyDeleteLotToken(lotName string, strToken string) (bool, error) {
```
turetske, line 260: All my comments below apply to here. In fact, this is a lot of duplicated code. Really the only difference between this and VerifyUpdateDeleteLotToken is the first debug message and the scope. I would make this one function and just pass the scope you are trying to match rather than duplicate the whole thing.

lotman/lotman_ui.go
```diff
@@ -0,0 +1,573 @@
...
+}
+
+// Verify that a token received is a valid token. Upon verification, we set the lot's parents/owner to the
+// appropriate values. Returns true if the token is valid, false otherwise.
+func VerifyNewLotToken(lot *Lot, strToken string) (bool, error) {
+	tokenApproved := false
+
+	// Get the path associated with the lot (right now we assume/enforce a single path) and try
+	// to deduce a namespace prefix from that. We'll use that namespace prefix's issuer as the
+	// owner of the lot. If there is no associated isuer, we assign ownership to the federation
```
turetske, line 80: Like we discussed, owner in this case is an overloaded term. Maybe call it the data-owner? Or some other name. And all data-owners are lot-managers of any child lot?

lotman/lotman_ui.go
```diff
@@ -0,0 +1,573 @@
...
+	} else {
+		// We found a different logical parent
+		lot.Parents = goLots
+	}
+
+	var tok jwt.Token
+	if len(lot.Parents) == 0 && lot.Parents[0] == "root" {
+		// We check that the token is signed by the federation
+		// First check for discovery URL and then for director URL, both of which should host the federation's pubkey
+		issuerUrl := param.Federation_DiscoveryUrl.GetString()
```
turetske, line 111: See my earlier comment about making this issuerURL retrieval and utility function.

lotman/lotman_ui.go
```diff
@@ -0,0 +1,573 @@
...
+		tok, err = jwt.Parse([]byte(strToken), jwt.WithKeySet(*kSet), jwt.WithValidate(true))
+		if err != nil {
+			return false, errors.Wrap(err, "Error parsing token")
+		}
+	} else {
+
+		// For each parent that might be here, get all owners and check that the token is signed by one of them
+		owners := []string{}
+		internalOwners := []string{}
+		for _, parent := range lot.Parents {
```
turetske, line 134: Do you want owners parents of parents to have the ability to manage lower lots? Because currently, this doesn't allow for that, but maybe you only want the assigned user of the lot (still testing out different terminology in the comments) to be able to create sub-lots.

If you do want the more, then I would again suggest a worklist algorithm. And since this would be the same as the one needed in lotman.go, I suggest creating a separate function to get all the ancestors.

lotman/lotman_ui.go
```diff
@@ -0,0 +1,573 @@
...
+			LotmanGetLotOwners(parent, true, &cOwners, &errMsg)
+			if ret != 0 {
+				trimBuf(&errMsg)
+				return false, errors.Errorf("Error getting lot JSON: %s", string(errMsg))
+			}
+			internalOwners = append(internalOwners, cArrToGoArr(&cOwners)...)
+		}
+
+		// Handle possible duplicates
+		occurred := map[string]bool{}
```
turetske, line 145: See again my comment about using a map here to mimic a set.

lotman/lotman_ui.go
```diff
@@ -0,0 +1,573 @@
...
+	namespaceHeader := resp.Header.Values("X-Pelican-Namespace")
+	xPelicanNamespaceMap := utils.HeaderParser(namespaceHeader[0])
+	namespace := xPelicanNamespaceMap["namespace"]
+
+	// Get the issuer URL for that namespace
+	nsIssuerUrl, err := server_utils.GetNSIssuerURL(namespace)
+	if err != nil {
+		return false, errors.Wrapf(err, errMsgPrefix+"no issuer could be found for namespace %s", namespace)
+	}
+	lot.Owner = nsIssuerUrl
```
turetske, line 255: So, an owner of the data in a lot will be an issuerURL and that issuer will be the same issuer as the issuer of the namespace prefix?

turetske (changes requested):
Review focused on the token logic

lotman/lotman.go
jhiemstrawisc: This can't be changed because `omitempty` is a special flag that prevents the JSON from injecting a key that would otherwise have a null value. See the [here](https://pkg.go.dev/encoding/json).

lotman/lotman.go
jhiemstrawisc: Most of the linking stuff here was based on choices I made when integrating Lotman with HTCondor (integration efforts ongoing). I guess there's probably no immediate reason we'd end up calling a Lotman function if we don't initialize in the cache. I'll remove the checks.

lotman/lotman_ui.go
jhiemstrawisc: While this only gets the lot's immediate parents, we then recursively get all owners of those parents two lines down when we call `LotmanGetLotOwners`

lotman/lotman.go
jhiemstrawisc: Just below, we call `LotmanGetLotOwners()` in recursive mode, which recursively gets all owners of each parent.

lotman/lotman_ui.go
jhiemstrawisc: To the extent possible, I'd like to keep the language in Pelican aligned with the language as it exists in Lotman. We can definitely come back to change things on the Lotman side of things to make this all more clear at a later point, but that feels like scope creep for this PR.

lotman/lotman_ui.go
jhiemstrawisc: Correct on both counts.

turetske (changes requested):
Will approve once the parameter validation is fixed.

jhiemstrawisc:
@CannonLock Is the parameter addition is something you'll need to handle, or can you give me a quick overview of how I can do it?

CannonLock:
@jhiemstrawisc Will push a commit on top of this in ~1 hours to add UI.

CannonLock:
@jhiemstrawisc You will have to merge with main for the parameters check to pass.

turetske (approved):
Approving since Cannon's change is in

jhiemstrawisc:
Okay, merging then! Thanks everyone 😄

PR #891, opened 2024-03-05
Refactor integration test scripts to be more robust

haoming29:
When fixing test failure in #885 I spot that the `get_put_test.sh` leaves zombie pelican and xrootd processes if the test fails, and the grep condition for output logs seem not to capture the correct server response, causing false negative (or positive, depending on how you look at it) test result. So this PR attempts to fix the problems in the test script.

### DIFF
--- a/github_scripts/get_put_test.sh
+++ b/github_scripts/get_put_test.sh
@@ -18,34 +18,71 @@
 # This tests the functionality of `pelican object get` and `pelican object put` with the
 # "federation in a box"
 
+set -e
+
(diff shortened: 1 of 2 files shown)


haoming29:
I'm not sure what's going on with this windows test but I don't see any changes in this PR will affect unit tests

joereuss12:
Yeah I keep running into that same test being flaky for me only on windows for some reason. I'll run it again and see if it will pass and if not I'll probably just review anyway and ignore it

joereuss12 (approved):
LGTM

PR #892, opened 2024-03-05
Expanded unit tests for plugin

joereuss12:
The test for StashPluginMain() now utilizes fed in a box and can successfully download a file. Also added unit test for the WriteOutfile function.

### DIFF
--- a/cmd/plugin_test.go
+++ b/cmd/plugin_test.go
@@ -21,15 +21,27 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/pelicanplatform/pelican/classads"
 	"github.com/pelicanplatform/pelican/config"
+	"github.com/pelicanplatform/pelican/launchers"
+	"github.com/pelicanplatform/pelican/param"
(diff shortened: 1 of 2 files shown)


cmd/plugin_test.go
```diff
@@ -66,9 +188,32 @@ func TestStashPluginMain(t *testing.T) {
...
+		viper.Set("Origin.EnablePublicReads", true)
+		fed := FedTest{T: t}
+		fed.Spinup()
+		defer fed.Teardown()
+
+		testFileContent := "test file content"
+		// Drop the testFileContent into the origin directory
+		tempFile, err := os.Create(filepath.Join(fed.OriginDir, "test.txt"))
+		assert.NoError(t, err, "Error creating temp file")
+		defer os.Remove(tempFile.Name())
```
haoming29, line 203: `Teardown` already removes all files under `f.OriginDir` so no need to do it separately

cmd/plugin_test.go
```diff
@@ -66,9 +188,32 @@ func TestStashPluginMain(t *testing.T) {
...
+		_, err = tempFile.WriteString(testFileContent)
+		assert.NoError(t, err, "Error writing to temp file")
+		tempFile.Close()
```
haoming29, line 206: You might want to do `defer tempFile.Close` after checking error creating the file

haoming29 (approved):
LGTM! With some small changes requested but feel free to fix them or leave them as is depending on how you think

PR #893, opened 2024-03-05
Cache can now filter to only serve a list of provided namespaces

### LINKED ISSUE #312: Allow cache to filter out origins when advertising
Currently the cache allows for all possible origins the director has advertised to it when it creates a cacheAd to the director.

Ideally, it should be able to filter or determine which origins it will actually cache and which it won't.

This should probably be some sort of config parameter which either is a list of origins it accepts or a list of origins it doesn't accept and then, when cache_serve advertises to the director, it filters the origins based on that list before creating the advertisement.

turetske:
-- Will default to all namespaces if not set
	-- Filtering is done at the cache side

closes #312


docs/parameters.yaml
```diff
@@ -681,6 +681,15 @@ type: int
 default: none
 components: ["cache"]
 ---
+name: Cache.AcceptedNamespaces
+description: >-
+  A list of namespaces the cache is allowed to pull from. If the list is empty, it's assumed that
```
jhiemstrawisc, line 686: I'd also say "If the list is empty or this option is unset".

docs/parameters.yaml
```diff
@@ -681,6 +681,15 @@ type: int
 default: none
 components: ["cache"]
 ---
+name: Cache.AcceptedNamespaces
```
jhiemstrawisc, line 684: How about `Cache.PermittedNamespaces`?

cache_ui/advertise.go
```diff
@@ -103,6 +140,10 @@ func (server *CacheServer) GetNamespaceAdsFromDirector() error {
 		}
 	}
 
+	if viper.IsSet("Cache.AcceptedNamespaces") {
```
jhiemstrawisc, line 143: If I set this with an empty list flow:
```
Cache:
  AcceptedNamespaces: []
```

What happens?

jhiemstrawisc (changes requested):

cache_ui/advertise.go
```diff
@@ -49,6 +51,41 @@ func (server *CacheServer) CreateAdvertisement(name string, originUrl string, or
...
+			ns := ad.Path
+			sentinel := true
+			for sentinel {
+				_, exists := server.namespaceFilter[ns]
+				if exists {
+					filteredAds = append(filteredAds, ad)
+					break
+				}
+
+				splitIndex := strings.LastIndex(ns, "/")
```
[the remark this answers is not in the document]
turetske: Good point. I'll put in something to check for both.

cache_ui/advertise_test.go
```diff
@@ -0,0 +1,118 @@
...
+			desc:          "no-filters-set",
+			expectedNumNS: 6,
+		},
+	}
+	viper.Reset()
+	defer viper.Reset()
+
+	nsAds := []common.NamespaceAdV2{
+		{
+			Path: "ns1",
```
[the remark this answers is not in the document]
turetske: Good point, will do.

cache_ui/advertise.go
```diff
@@ -52,21 +51,37 @@ func (server *CacheServer) CreateAdvertisement(name string, originUrl string, or
...
+	 */
 	server.namespaceFilter = make(map[string]struct{})
-	if viper.IsSet("Cache.AcceptedNamespaces") {
-		nsList := param.Cache_AcceptedNamespaces.GetStringSlice()
-		for _, ns := range nsList {
-			server.namespaceFilter[ns] = struct{}{}
+	nsList := param.Cache_PermittedNamespaces.GetStringSlice()
+	// Ensure that each permitted namespace starts with a "/"
+	for _, ns := range nsList {
+		if ns[0] != '/' {
```
jhiemstrawisc, line 63: Maybe use `strings.HasPrefix`? If someone provides a bad configuration, ns might be empty. I think `strings.HasPrefix` handles being passed an empty string.

cache_ui/advertise.go
```diff
@@ -52,21 +51,37 @@ func (server *CacheServer) CreateAdvertisement(name string, originUrl string, or
...
+			ns = "/" + ns
 		}
+		server.namespaceFilter[ns] = struct{}{}
 	}
 }
 
 func (server *CacheServer) filterAdsBasedOnNamespace(nsAds []common.NamespaceAdV2) []common.NamespaceAdV2 {
+	/*
+	* Filters out ads based on the namespaces listed in server.NamespaceFiler
```
jhiemstrawisc, line 72: Typo in `server.NamespaceFiler`

cache_ui/advertise.go
```diff
@@ -52,21 +51,37 @@ func (server *CacheServer) CreateAdvertisement(name string, originUrl string, or
...
+	* Note that this does a few checks for trailing and non-trailing "/" as it's assumed that the namespaces
+	* from the director and the ones provided might differ.
+	 */
 	filteredAds := []common.NamespaceAdV2{}
 	if len(server.namespaceFilter) > 0 {
 		for _, ad := range nsAds {
 			ns := ad.Path
 			sentinel := true
+			//If the final character isn't a '/', add it to the string
+			if ns[len(ns)-1] != '/' {
```
jhiemstrawisc, line 82: Maybe `strings.HasSuffix`?

jhiemstrawisc (changes requested):
Just a few small things, otherwise I think this is looking good.

jhiemstrawisc (approved):
LGTM. Thanks!

PR #894, opened 2024-03-05
Add Metalink for redirect to origin

### LINKED ISSUE #357: Implement Metalink for origin redirects in director
Right now, when we request a cache from the director, we get back a metalink-conformant list with ranked priority. However, when we request an origin, we get only whichever happens to be first in the ttlcache.

We should make the origin redirects more like the cache redirects to account for the fact that there might be multiple origins corresponding to a namespace (as is the case, for example, with `/chtc/PUBLIC`)

haoming29:
Closes #357


jhiemstrawisc (approved):
LGTM!

PR #896, opened 2024-03-05
Xrootd tests now do not generate files outside tmp

joereuss12:
Fixed the bug where xrootd tests would create files outside of tmpdir. Found the leak was doing a viper.reset() and resetting the cache and origin run locations so the config was being written locally. Should be fixed now!


haoming29 (approved):
Changes look good to me! Thanks fox fixing it

PR #903, opened 2024-03-06
Replace `pelican` with server name for production image

haoming29:
I think I missed changing the docker image tag from `pelican_platform/pelican` to `pelican_platform/<server-name>`. This PR attempts to address the issue.

Update:
This PR also fixed the bug where pelican arguments can't be passed to tini

### DIFF
--- a/.github/workflows/build-prod-container.yml
+++ b/.github/workflows/build-prod-container.yml
@@ -124,11 +124,12 @@ jobs:
 
           echo $GITHUB_TAG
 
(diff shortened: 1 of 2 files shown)


brianhlin (approved):
LGTM! The `--` syntax is fairly standard, everything before it is for tini options and everything after is the subsequent command + its options

haoming29:
> LGTM! The `--` syntax is fairly standard, everything before it is for tini options and everything after is the subsequent command + its options

Thanks for the explanation!

PR #904, opened 2024-03-06
Improve installation docs and include running Pelican server with docker image

haoming29:
This PR revised Install page of Pelican documentation. It adds a page to explain how to run Pelican servers with Docker images. 

Still waiting for example to run cache as docker container, and instructions to run windows binary. Appreciate if anyone knows how to do any of them.

### DIFF
--- a/docs/pages/install.mdx
+++ b/docs/pages/install.mdx
@@ -1,96 +1,48 @@
 import DownloadsComponent from "/components/DownloadsComponent";
 
-# Installing the Pelican Platform
+# Install Pelican
 
-This document explains how a user can download and install Pelican client or Pelican server components.
+This document lists the operating system requirements to install Pelican and explains how a user can download Pelican client or Pelican server components.
 
-## Installation Steps:
+> Pelican also has Docker images for users who want to serve **Pelican servers**. Refer to [Run Pelican Server using Docker image](https://docs.pelicanplatform.org/install/docker) for instructions.
 
+## Supported Operating Sytems
 
-### 1. Install the Pelican Platform Binary
+Pelican supports the following operating systems:
+
+* [Red Hat, RHEL, CentOS, Fedora, or openSUSE](https://docs.pelicanplatform.org/install/rhel)
+* [Debian or Ubuntu](https://docs.pelicanplatform.org/install/debian)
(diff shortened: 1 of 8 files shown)


docs/pages/install.mdx
```diff
@@ -1,96 +1,48 @@
...
 
-This document explains how a user can download and install Pelican client or Pelican server components.
+This document lists the operating system requirements to install Pelican and explains how a user can download Pelican client or Pelican server components.
 
-## Installation Steps:
+> Pelican also has Docker images for users who want to serve **Pelican servers**. Refer to [Run Pelican Server using Docker image](./install/docker.mdx) for instructions.
 
+## Supported Operating Sytems
+
+Pelican supports the following operating systems:
```
jhiemstrawisc, line 11: I'd be careful in making this claim. While we can technically run the pelican binary on Windows, most of the functionality is turned of or compiled out. It's not as bad with MacOS, but we're still largely untested there as well.

docs/pages/install/rhel.mdx
```diff
@@ -0,0 +1,96 @@
+# Install Pelican on CentOS, Fedora, or openSUSE
```
jhiemstrawisc, line 1: I'd also include Alma/Rocky here (since they're also supposed to be RHEL variants)

jhiemstrawisc (changes requested):

docs/pages/install.mdx
haoming29: Do you think is is for windows only or the claim that we _support_ the OS? If it's former, then I can add a note to Windows item suggesting that we only support client functionality. This would be the same thing for macOS where I don't think xrootd works well with mac and people should be required to build xrootd from source

haoming29: I've bugged @joereuss12 to run the `.exe` program we generated and see if that works

haoming29:
@jhiemstrawisc Are we good with Joe's testing? Any other comments/thoughts?

jhiemstrawisc:
I think Joe's testing shows we're okay. We can always come back and change it if a user proves us wrong.

haoming29:
Latest commit is to reflect changes in #945

jhiemstrawisc (approved):
LGTM. Thanks Haoming!

Issue #858, opened 2024-02-28
Origins should support multiple exports


jhiemstrawisc:
Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [X ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

**Is your feature request related to a problem? Please describe.**
Origins should be able to export multiple prefixes with multiple auth configurations. We have internal docs related to the syntax at https://docs.google.com/document/d/1W-2s5a_wj3Eq7URR4t7cpvQXXY0ixzGP6tUHbtKG_mc/edit#heading=h.y539i3hp19ku

PR #907, opened 2024-03-07
Configure 'sec.protbind' for public origins

jhiemstrawisc:
Legacy caches won't work with public pelican origins without the ability to fallback to xroot protocol. This adds `sec.protbind * none` to the origin config if the origin is run with `Origin.EnablePublicReads: true`. Note that when we have multi-exports origins configured from #858, we'll have to handle this more carefully.

So that the reviewer doesn't have to hunt, the addition to the struct in `xrootd/xrootd_config` is `EnablePublicReads`. I had to adjust spacing to make the linter happy, and it buried the change.

To test, try running an origin with `Origin.EnablePublicReads` set to true and then to false. You should see the `sec.protbind * none` line appear in the xrootd config for public, and be absent for non-public.

### DIFF
--- a/param/parameters_struct.go
+++ b/param/parameters_struct.go
@@ -30,9 +30,9 @@ type Config struct {
 		EnableVoms bool
 		ExportLocation string
 		Port int
+		RunLocation string
 		SelfTest bool
 		SelfTestInterval time.Duration
-		RunLocation string
 		Url string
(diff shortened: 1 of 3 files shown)


bbockelm (approved):
LGTM.

Please backport so we can get this out to the OSDF quickly.

Issue #322, opened 2023-11-03
Handle the display of complex configuration types


CannonLock:
Current config view does not provide a easy interface for viewing the duration type Go objects. When we add the write interface next milestone lets improve this.

CannonLock:
Clearing milestone, will keep in mind for when the config gets a write endpoint.

bbockelm:
Since a write endpoint is critical-path for 7.6.0, I'm adding a milestone back to this issue.

bbockelm:
@CannonLock - what's the status of this one?

PR #909, opened 2024-03-07
Write config UI

CannonLock:
- Adds ability to edit the config
- Type checking where possible


CannonLock:
Closes #753 and #322

web_ui/frontend/app/config/main.d.ts
haoming29, line 1: This is file required as a placeholder?

docs/parameters.yaml
haoming29, line 1: Not related to code changes in this PR but I figured that `Xrootd.ScitokensConfig` should be for `origin` and `cache` only.

haoming29 (approved):
Two minor changes requested and are non-blocking so feel free to tackle them or leave them for now but otherwise looks pretty nice to me!

I'm thinking it would be very nice if we could add "Cancel" button next to save, so that users mistakenly changing a value have the chance to ignore the change. Could be just a button to refresh the page if that's easier to implement.

As we talked in person, currently if a user changes a value and re-do the change, frontend will still send that change back to backend, which won't be a big issue now but we can fix them in another PR.

Screenshot 2024-03-07 at 12 56 23 PM

PR #922, opened 2024-03-08
Fix production cache container failure at start up due to outdated xrootd

### LINKED ISSUE #917: ITB team reports wrong xrootd in production container
Our integration team is reporting that they have to manually install the correct version of XRootD in our production container so that `pelican://`-style urls are recognized by xrootd. This throws a wrench in their ability to automatically deploy Pelican origins and caches. 

 @williamnswanson reports that he's able to get the required version of xrootd by running 
```
yum install -y https://repo.opensciencegrid.org/osg/23-main/osg-23-main-el8-release-latest.rpm && yum install -y xrdcl-pelican --enablerepo=osg-testing && yum update -y xrdcl-pelican '*xrootd*' --enablerepo=osg-testing
```
from within the container.

haoming29:
Fixes #917 

Also fixes the bug where we check director's endpoint to ensure director is working but we are checking against `Server.ExternalWebUrl`, which is wrong.

Building the container and running locally, I can confirm this fix will enable cache to run.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -57,6 +57,7 @@ RUN useradd -o -u 10940 -g 10940 -s /bin/sh xrootd
 
 # Install dependencies
 RUN yum -y update \
+    && yum -y install https://repo.opensciencegrid.org/osg/23-main/osg-23-main-el8-release-latest.rpm\
     && yum -y install tini xrootd xrootd-client xrdcl-http xrootd-server xrootd-scitokens xrootd-voms curl java-17-openjdk-headless \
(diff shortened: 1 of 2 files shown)


turetske (approved):
LGTM!

haoming29:
> @haoming29 , can you take a look at line 214 in launcher.go? We also do a Server_ExternalWebUrl there if we're configured in both cache and origin mode.

Good catch, that should be fixed too

matyasselmeci (approved):

PR #923, opened 2024-03-08
Hash external repository to invalidate build cache when there's an update

### LINKED ISSUE #914: Invalidate docker build cache when upstream repo changes that we rely on
Currently we have docker build cache for building both the production and dev containers, but if the upstream repo have updates, which we rely on, docker won't detect the change and still using cached file from the repo. Let's have a way to invalidate the cache if the repos have changes. Probably can be done by saying the latest commit hash from the upstream repo as part of the input file when building the container.

haoming29:
Fixes #914


brianhlin (approved):
LGTM

PR #924, opened 2024-03-08
Consolidate two `namespaceExists` DB function

### LINKED ISSUE #824: Inconsistent use of `namespaceExists()`-type functions in the registry db
As mentioned in #804, we're inconsistent about how we use these functions, and there are two nearly identical functions floating around -- `namespaceExistsByPrefix()` and `namespaceExists()`, which also takes a prefix. We should clean this up and settle on one or the other.

haoming29:
Closes #824


jhiemstrawisc (approved):
LGTM

PR #926, opened 2024-03-10
Airplane mode

bbockelm:
Admittedly this is really a corner case but I got tired of the tests failing when I was on an airplane.


turetske (approved):
LGTM

PR #931, opened 2024-03-11
Fix Prometheus handler not called

haoming29:
In the merge from #877 and #840, git seems to skip one important line that causes the Prometheus endpoint to return 200 with empty response body. This PR fixes this issue and added test to ensure


turetske (approved):
LGTM!

PR #933, opened 2024-03-11
Allow admin to configure filtering servers for redirecting traffic at the director

### LINKED ISSUE #825: Allow director admin to pause redirecting traffic to a cache or origin
Pelican Service:
- [x] Director

Add support to pause redirecting traffic to a cache or origin in the federation. This should be done through director's web UI.

haoming29:
Closes #825 

This PR also added a config parameter `Director.FilteredServers` to allow server admin to configure and persist the filtering. They can also add/remove filtering rules through the web UI but those changes are in-memory and will be reset or overwritten by the config value at the server restart. This will give admins the best flexibility in holding redirect traffic to a server without restarting the director.


director/director.go
```diff
@@ -35,14 +35,19 @@ type (
...
+		Name               string            `json:"name"`
+		AuthURL            string            `json:"authUrl"`
+		BrokerURL          string            `json:"brokerUrl"`
+		URL                string            `json:"url"`    // This is server's XRootD URL for file transfer
+		WebURL             string            `json:"webUrl"` // This is server's Web interface and API
+		Type               common.ServerType `json:"type"`
+		Latitude           float64           `json:"latitude"`
+		Longitude          float64           `json:"longitude"`
+		EnableWrite        bool              `json:"enableWrite"`
+		EnableFallbackRead bool              `json:"enableFallbackRead"`
```
jhiemstrawisc, line 47: Can you update these `Enable` variables to the new internal reps? They should be `EnableWrites` and `EnableDirectReads`, respectively. We should probably keep the json as it is, unless these are totally new and we don't need to worry about the wrong key being unmarshalled by an older server.

director/director.go
```diff
@@ -92,15 +97,21 @@ func listServers(ctx *gin.Context) {
...
-			Status:    healthStatus,
+			Name:               server.Name,
+			BrokerURL:          server.BrokerURL.String(),
+			AuthURL:            server.AuthURL.String(),
+			URL:                server.URL.String(),
+			WebURL:             server.WebURL.String(),
+			Type:               server.Type,
+			Latitude:           server.Latitude,
+			Longitude:          server.Longitude,
+			EnableWrite:        server.EnableWrite,
```
jhiemstrawisc, line 110: Same with these

director/director.go
```diff
@@ -148,11 +159,61 @@ func queryOrigins(ctx *gin.Context) {
 	ctx.JSON(http.StatusOK, res)
 }
 
+func handleFilterServer(ctx *gin.Context) {
+	sn := strings.TrimPrefix(ctx.Param("name"), "/")
+	if sn == "" {
+		ctx.JSON(http.StatusBadRequest, gin.H{"error": "name is a required path parameter"})
+		return
+	}
+	filtered, ft := checkFilter(sn)
```
jhiemstrawisc, line 168: Personal preference, so feel free not to implement, but renaming `ft` to `filterType` makes it so I don't have to hunt for what this is supposed to be the first time I see it.

director/director.go
```diff
@@ -148,11 +159,61 @@ func queryOrigins(ctx *gin.Context) {
...
+	// If we previously temporarily allowed a server, we switch to permFiltered (reset)
+	if ft == tempAllowed {
+		filteredServers[sn] = permFiltered
+	} else {
+		filteredServers[sn] = tempFiltered
+	}
+	ctx.JSON(http.StatusOK, gin.H{"message": "success"})
+}
+
+func handleAllowServer(ctx *gin.Context) {
```
jhiemstrawisc, line 185: I'd prefer `handleAllowServer` --> `handlePermittedServer`, but again that's preference. I respect whatever judgement call you make!

Also, can you add a comment header with this functions intended behavior?

director/director.go
```diff
@@ -148,11 +159,61 @@ func queryOrigins(ctx *gin.Context) {
 	ctx.JSON(http.StatusOK, res)
 }
 
+func handleFilterServer(ctx *gin.Context) {
```
jhiemstrawisc, line 162: Can you add a comment header with the function's intended behavior and a quick note about where this gin context is usually coming from?

director/director.go
```diff
@@ -148,11 +159,61 @@ func queryOrigins(ctx *gin.Context) {
...
+	ctx.JSON(http.StatusOK, gin.H{"message": "success"})
+}
+
 func RegisterDirectorWebAPI(router *gin.RouterGroup) {
 	directorWebAPI := router.Group("/api/v1.0/director_ui")
 	// Follow RESTful schema
 	{
 		directorWebAPI.GET("/servers", listServers)
+		directorWebAPI.PATCH("/servers/filter/*name", web_ui.AuthHandler, web_ui.AdminAuthHandler, handleFilterServer)
+		directorWebAPI.PATCH("/servers/allow/*name", web_ui.AuthHandler, web_ui.AdminAuthHandler, handleAllowServer)
```
jhiemstrawisc, line 216: Nice, love the use of `PATCH` here!

director/director_api.go
```diff
@@ -60,6 +61,32 @@ func listServerAds(serverTypes []common.ServerType) []common.ServerAd {
 	return ads
 }
 
+// Check if a server is filtered from "production" servers by
+// checking if a serverName is in the filteredServers map
+func checkFilter(serverName string) (bool, filterType) {
+	filteredServersMutex.RLock()
+	defer filteredServersMutex.RUnlock()
+
+	status, ok := filteredServers[serverName]
```
jhiemstrawisc, line 70: I'd change `ok` to `exists`

director/director_api.go
```diff
@@ -118,3 +145,17 @@ func ConfigTTLCache(ctx context.Context, egrp *errgroup.Group) {
 		return nil
 	})
 }
+
+// Populate internal filteredServers map by Director.FilteredServers
+func ConfigFilterdServers() {
+	filteredServersMutex.Lock()
+	defer filteredServersMutex.Unlock()
+
+	if len(param.Director_FilteredServers.GetStringSlice()) == 0 {
```
jhiemstrawisc, line 154: If it's helpful, I recently added an `IsSet` interface for most of our internal param types:

https://github.com/PelicanPlatform/pelican/blob/cae5ec86cd8dc4d6564035d3880a85f6b854418f/param/parameters.go#L77-L79

director/director_test.go
```diff
@@ -45,24 +46,30 @@ func TestListServers(t *testing.T) {
...
+		Name:               mockOriginServerAd.Name,
+		BrokerURL:          mockOriginServerAd.BrokerURL.String(),
+		AuthURL:            mockOriginServerAd.AuthURL.String(),
+		URL:                mockOriginServerAd.URL.String(),
+		WebURL:             mockOriginServerAd.WebURL.String(),
+		Type:               mockOriginServerAd.Type,
+		Latitude:           mockOriginServerAd.Latitude,
+		Longitude:          mockOriginServerAd.Longitude,
+		EnableWrite:        mockOriginServerAd.EnableWrite,
+		EnableFallbackRead: mockOriginServerAd.EnableFallbackRead,
```
jhiemstrawisc, line 58: Same with these.

director/director_test.go
```diff
@@ -45,24 +46,30 @@ func TestListServers(t *testing.T) {
...
+		Name:               mockCacheServerAd.Name,
+		BrokerURL:          mockCacheServerAd.BrokerURL.String(),
+		AuthURL:            mockCacheServerAd.AuthURL.String(),
+		URL:                mockCacheServerAd.URL.String(),
+		WebURL:             mockCacheServerAd.WebURL.String(),
+		Type:               mockCacheServerAd.Type,
+		Latitude:           mockCacheServerAd.Latitude,
+		Longitude:          mockCacheServerAd.Longitude,
+		EnableWrite:        mockCacheServerAd.EnableWrite,
+		EnableFallbackRead: mockCacheServerAd.EnableFallbackRead,
```
jhiemstrawisc, line 71: And these

docs/parameters.yaml
```diff
@@ -884,6 +884,13 @@ type: bool
 default: true
 components: ["director"]
 ---
+name: Director.FilteredServers
+description: >-
+  A list of server names to not to redirect client requests to.
```
jhiemstrawisc, line 889: Are these names supposed to be host names or prefixes? You might also add a note that this is intended to put certain servers into downtime to give administrators a clear reason to use this config mechanism.

web_ui/frontend/app/api/docs/pelican-swagger.yaml
```diff
@@ -329,6 +333,31 @@ definitions:
         type: number
         description: The longitude of the server based on its IP address
         default: 0
+      enableWrite:
```
jhiemstrawisc, line 336: These too

jhiemstrawisc (changes requested):

web_ui/frontend/app/api/docs/pelican-swagger.yaml
haoming29: If we changed the json names, we will have backward compatibility issue as the data was no longer the same. Do you think we should do a separate API or keep json name untouched for now?

director/director.go
haoming29: I'd like to stick with Allow if that's not a big issue to make it consistent with the URL name, unless we'd like to change that too

web_ui/frontend/app/api/docs/pelican-swagger.yaml
jhiemstrawisc: Ah, I didn't look closely enough and didn't realize these were going to affect the underlying JSON.

jhiemstrawisc (approved):
LGTM

PR #937, opened 2024-03-12
Remove `generate` package when releasing

### LINKED ISSUE #459: Exclude `generate` package when building for release
As we discussed in #442, currently we include `generate` package and the code it generates in the release artifact. However, the `generate` package itself won't be used by the user as an executable, so we want to exclude them when shipping the code as binaries.

* Modify GHA release action to exclude `./generate` package when building for release

haoming29:
Closes #459 

I tested locally that removing the whole `generate` package locally won't cause build to fail, and the docker container I built with that runs as expected.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,9 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "1.21"
+      - name: Remove Generate folder
+        run: |
+          rm -rf generate
       - name: Run GoReleaser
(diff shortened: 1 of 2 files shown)


matyasselmeci (approved):
LGTM. Note that because images are layered, deleting files in a separate RUN command does not actually make images smaller.

haoming29:
> LGTM. Note that because images are layered, deleting files in a separate RUN command does not actually make images smaller.

That should be fine. I think we want to remove it so that people won't get chance to run the generate command themselves

PR #938, opened 2024-03-12
Fix logging color miss and log location not working

### LINKED ISSUE #918: Pelican log doesn't have color coding for tty terminal
When running a Pelican instance in a TTY terminal , the log messages are supposed to have color coding; however, at some point this feature is gone. We should investigate and fix it to make log messages more readable

haoming29:
Fixes #918

### DIFF
--- a/config/logging.go
+++ b/config/logging.go
@@ -19,13 +19,10 @@
 package config
 
 import (
-	"io"
-	"os"
 	"regexp"
 	"sync/atomic"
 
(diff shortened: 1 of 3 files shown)


haoming29:
The fix is invalid with how logrus was set up. Logrus won't create a new entry if the level of the log is smaller than the logging threshold and we reply heavily on intercepting all logs for the code to work. This fix is not effective.

PR #939, opened 2024-03-13
Fix log color loss and `LogLocation` not working

### LINKED ISSUE #918: Pelican log doesn't have color coding for tty terminal

haoming29:
Fixes #918 

Second attempt to fix the logging bug where log color are lost in a TTY terminal and `LogLocation` is not working.

This PR also fixes the bug where `Logging.Level` was not respected.

@turetske  Marking this as critical and putting it under 7.6 because the bug causes trouble at deployment/monitoring side. We can backport it in 7.6.1 if this can't be merged in before 7.6 release.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -681,15 +681,6 @@ func InitConfig() {
 			cobra.CheckErr(err)
 		}
 	}
-	if param.Debug.GetBool() {
-		SetLogging(log.DebugLevel)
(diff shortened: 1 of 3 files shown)


joereuss12 (approved):
LGTM

haoming29:
Should be fine just merged in. It will be cherry-picked to version branch

PR #942, opened 2024-03-14
Fix pelican plugin bug `no valid classads...`

joereuss12:
This fixes the bug we were getting with the pelican plugin where we would see the plugin exiting 0 but sending out `no valid classads`. The issue was an error type being written out and bypassing strconv.QuoteToASCII which would remove any inside quotes.

Also found a spot where we were getting a nil pointer dereference because we were accessing data before an error check, fixed as well


turetske (approved):
LGTM

PR #947, opened 2024-03-14
Embed `serve` as default command to run the server container

### LINKED ISSUE #945: Default container run command to include `serve`
Currently we only provide `./pelican <servername>` as the default run argument for server container, which means admins need to have `serve` command feed in to start the container. We want to instead have `./pelican <servername> serve` as the default command embedded.

haoming29:
Closes #945

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -191,7 +191,6 @@ COPY --from=xrootd-plugin-builder /usr/include/nlohmann /usr/include/nlohmann
 # Copy the JSON schema validator library
 COPY --from=xrootd-plugin-builder /usr/lib64/libnlohmann_json_schema_validator.a /usr/lib64/
 
-# Eventually add more entrypoint commands here
 COPY images/entrypoint.sh /entrypoint.sh
 
 # Copy here to reduce dependency on the pelican-build stage in the final-stage and x-base stage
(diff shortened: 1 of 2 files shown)


brianhlin (changes requested):
If you specify the `CMD` instead of appending this to `ENTRYPOINT`, you'll get the default behavior we want but still allow people to specify other commands when starting the container. See https://docs.docker.com/reference/dockerfile/#cmd

Also, I /think/ you could do this in the `final-stage` layer so that you don't have to repeat it across every single one of the separated images

brianhlin:
> @brianhlin I ended up doing ENTRYPOINT [ "/entrypoint.sh" ,"pelican", "director"] with CMD ["serve"] so that users only need to specify serve [args...] if they prefer running other command.

I think this limits your flexibility a bit too much, though I'll admit that I'm not too familiar with each host's commands other than `serve` or `token`. I don't think it's too much to ask users to specify `serve <additional options>` if they want to do something outside of the defaults.

> I didn't think of a way to put ENTRYPOINT into final-stage as the entrypoint has the server name attached and that's what distinguish different server images right?

Yeah, that's one benefit of putting the `serve` in the `CMD` instead so that you could have per-image `ENTRYPOINT` but a shared default `CMD`

haoming29:
> I don't think it's too much to ask users to specify `serve <additional options>` if they want to do something outside of the defaults.

Oops. This is what I did lol. I meant that if I only put `ENTRYPOINT [ "/entrypoint.sh"]` with `CMD ["pelican", "director", "serve"]`, then users must specify `pelican director serve -p xxx` for additional arguments

haoming29:
OK so I figured that if I have `CMD` set prior to `ENTRYPOINT`, it is ignored. So I have put to `CMD` in each product-stage. But I do removed `WORKDIR` from product-stage

brianhlin (commented):
> OK so I figured that if I have CMD set prior to ENTRYPOINT

Ah bummer! This LGTM, though

brianhlin (approved):

PR #949, opened 2024-03-15
First draft of enum client error codes

joereuss12:
Very first draft for enumerating client error codes which contains these features:
- a YAML file containing all the current classifications of error codes (to be expanded as more come up)
- a generate function that generates error_codes.go, this will ensure the mapping from the YAML to the .go file is consistant and that if we introduce a new error code it must be documented with the proper features
- Replacement of just 1 error type in the client. I wrapped the client's `SlowTransferError` with the new error object. Still not completely sure if I should just replace this error with the new PelicanError type or just wrap it. For now, I just wrapped it
- Changed just object_get.go for now to show how we will handle errors like this, just returning whatever the PelicanError error code is otherwise acting as normal if error returned is not a PelicanError type.


joereuss12:
Ready for any and all feedback on current approach! The tagged issue contains a link to the current design doc as well for the issue but please feel free to review and let me know any thoughts.

generate/error_code_generator.go
```diff
@@ -0,0 +1,190 @@
...
+
+type PelicanError struct {
+	errorType string
+	exitCode  int
+	code      int
+	retryable bool
+	err       error
+}
+
+{{- range $idx, $pelicanError := .PelicanErrors}}
```
jhiemstrawisc, line 158: The beginning `-` you have here and in the `end` statements are why you're not getting newlines between functions in the generated code. It specifies "remove all whitespace on the side of the statement that has -"

docs/error_codes.yaml
```diff
@@ -0,0 +1,147 @@
...
+  the server replied with an authorization error when the file was requested or sent.
+  Basically a HTTP/403 error
+retryable: false
+
+############################
+#   Specification Errors   #
+############################
+---
+type: Specification
+code: 500
```
jhiemstrawisc, line 98: Hmm. It's definitely going to cause confusion that this `500` signifies an HTTP `404`. I don't know how we handle this kind of naming issue, but we should think hard about it before these become permanent.

client/handle_http_test.go
```diff
@@ -34,11 +34,13 @@ import (
 	"testing"
 	"time"
 
+	//"github.com/pkg/errors"
```
jhiemstrawisc, line 37: Remove commented import

client/handle_http_test.go
```diff
@@ -206,7 +208,8 @@ func TestSlowTransfers(t *testing.T) {
 
 	// Make sure the errors are correct
 	assert.NotNil(t, err)
-	assert.IsType(t, &SlowTransferError{}, err)
+	require.IsType(t, &error_codes.PelicanError{}, err)
+	assert.Contains(t, err.Error(), "Transfer.SlowTransferError Error: Error code 602:")
```
jhiemstrawisc, line 212: Is this representative of the error code an end user might see, or are we only checking part of the full error for the test? If the former, I worry it's not very descriptive and it has a confusing format.

docs/error_codes.yaml
```diff
@@ -0,0 +1,147 @@
...
+description: >-
+  The client attempted to contact the cache but failed to do so
+retryable: true
+---
+type: Contact.Origin
+code: 303
+clientExitCode: 6
+description: >-
+  The client attempted to contact the origin but failed to do so
+retryable: false
```
jhiemstrawisc, line 78: I guess clients don't really reach out to the registry, but should we think about propagating some kind of error from the Director in the event that the Director can't get something from the registry on behalf of a requested namespace/object?

generate/error_code_generator.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+}
+{{- end}}
+
+// function that maps the error to the exit code
+func (e *PelicanError) ExitCode() int {
+	return e.exitCode
+}
+
+func (e *PelicanError) Error() string {
```
jhiemstrawisc, line 175: Can you add header comments to each of these functions indicating any intended input, output and behavior?

jhiemstrawisc (commented):
I know we're still in draft mode for this, but I left a few comments/questions I have at this point.

docs/error_codes.yaml
joereuss12: Yeah, I feel this could maybe work for the errors we get that are `no namespace found for path' and such. I'll add this error in. Again, this is a first draft so as I go through and categorize the errors I may find some that I did miss

joereuss12: I guess what I said could be a specification error too tho... Either way I will add it in!

client/handle_http_test.go
joereuss12: No, that is not the full error, just checking to see if this error contains the changes. The full error is this:
`Transfer.SlowTransfer Error: Error code 602: cancelled transfer, too slow.  Detected speed: 10.2 kB/s, total transferred: 153.6 kB, total transfer time: 15.000997891s`

Just can't really do much of the other check since we don't know the detected speed or the other numbers.

joereuss12:
Yeah waiting for input from @bbockelm before I move any further. Just want approval before any major refactoring takes place

docs/error_codes.yaml
```diff
@@ -0,0 +1,154 @@
...
+description: >-
+  The client started transferring file(s) but it got cancelled by Pelican as it seemed to have stopped
+  transferring data
+retryable: true
+---
+type: Transfer.SlowTransfer
+code: 602
+clientExitCode: 11
+description: >-
+  The client started transferring data but the transfer takes longer than the configured timeout
```
bbockelm, line 153: Inconsistent voice.
```suggestion
  The client started transferring data but the transfer was slower than the minimum configured rate
```

bbockelm (changes requested):
Some minor requests inline.

docs/error_codes.yaml
```diff
@@ -0,0 +1,151 @@
...
+############################
+#      Contact Errors      #
+############################
+---
+type: Contact
+code: 3000
+clientExitCode: 6
+description: >-
+  The client attempted to contact the server at its resolved address and failed to do so
+retryable: false
```
jhiemstrawisc, line 57: In the context of these errors, we should think carefully about how we define "retryable". I'd love to see the definition of that somewhere in this doc.

If we resolved the hostname for these servers and tried to contact them but couldn't, it could mean a blip in the network wherever this job is run and that retrying after a few seconds might work. It could also be that the site has a network issue and that the request is retryable, but only from a different site. These are different definitions of retryable. Do we have a way of distinguishing? If not, how do we decide these constitute a non-retryable error?

docs/error_codes.yaml
```diff
@@ -0,0 +1,151 @@
...
+############################
+#   Specification Errors   #
+############################
+---
+type: Specification
+code: 5000
+clientExitCode: 8
+description: >-
+  If the client successfully contacted the server and received a definitive response that the
+  desired file was not present or could not be created. Usually the submitters fault
```
jhiemstrawisc, line 109: Do we have a way of baking into our error codes this notion of "something the submitter might be able to fix on their own"? The major number in each error code broadly defines the type of error, and the last number defines a subtype within the top-level category, but we have a lot of remaining real estate in the middle two numbers.

docs/error_codes.yaml
```diff
@@ -0,0 +1,151 @@
...
+retryable: false
+---
+type: Specification.FileNotFound
+code: 5001
+clientExitCode: 8
+description: >-
+  If the client successfully contacted the server but the desired file does not exist for download.
```
jhiemstrawisc, line 116: Can you modify the other descriptions to also include a period if they don't have one?

generate/error_code_generator.go
```diff
@@ -0,0 +1,196 @@
...
+	"strings"
+	"text/template"
+
+	"gopkg.in/yaml.v3"
+)
+
+// TODO: might need to change this
+type ErrorType struct {
+	Raw       string
+	Display   string
```
jhiemstrawisc, line 35: Can you provide definitions for "raw" and "display"? IMO, the names for the others are explanatory enough, but I don't know what these two variables are supposed to hold.

generate/error_code_generator.go
```diff
@@ -0,0 +1,196 @@
...
+
+	var values []interface{}
+
+	for {
+		var value map[string]interface{}
+		if err := decoder.Decode(&value); err != nil {
+			if err == io.EOF {
+				break
+			}
+			panic(fmt.Errorf("document decode failed"))
```
jhiemstrawisc, line 63: Might be nice to print the error here so we can debug in the case of failure.

generate/scope_generator.go
```diff
@@ -39,7 +39,7 @@ var requiredScopeKeys = [3]string{"description", "issuedBy", "acceptedBy"}
 
 func handleCaseConversion(s string) string {
 	var camelCase string
-	nextCap := false
+	nextCap := true // default as true so we capitalize the first letter
```
jhiemstrawisc, line 42: What happens to the other generators that use this function? Does switching this to false have ramifications for them?

jhiemstrawisc (changes requested):

generate/scope_generator.go
joereuss12: So I looked into that when I changed this. The only other reference to this function capitalized the first letter by hand outside of it. This way there is no need for it to do that.

What I did forget to do was remove the (now redundant) line that did the capitalization manually. Making that change now

docs/error_codes.yaml
joereuss12: Sure! We can do this for now and if we find something better to do for use of those digits we can always change it since it is still early.

docs/error_codes.yaml
joereuss12: Good idea! Added a comment explaining retryable. Basically, it is used in the condor sense as it could work if we retry the entire job. The job can either succeed by:
1. Landing at the same site that is now in a better state
2. Landing at a different site

jhiemstrawisc (approved):
LGTM. Thanks for these!

joereuss12:
@bbockelm Do you think you can give this the okay?

bbockelm (approved):

PR #958, opened 2024-03-17
Add support for HTTP backend

bbockelm:
Invokes the HTTP plugin for XRootD.

Note the first commit in the series does some cleanup on the returned `OriginExports` to not make it a pointer to a slice (which really shouldn't be necessary for a returned value); this cleanup causes the majority of the churn here.


jhiemstrawisc (commented):
Overall I like the approach -- and thanks for the cleanup! My only question is whether we want to think about supporting multi-export HTTPs origins right out of the gate. I think we could expose different sub paths of the `HTTPServiceUrl` at different federation prefixes if we hooked this into the origin exports slice.  

The reason I punted with the S3 bits included time constraints and the fact that I'm waiting for Rich's changes to the underlying plugin configuration. I don't think he's touching the underlying HTTP plugin at the moment.

jhiemstrawisc (approved):
LGTM. I'm excited to try out the HTTPS backend :)

PR #962, opened 2024-03-19
Fix bug with version being "dev"

joereuss12:
Goreleaser was not setting version correctly for the 'config' package. Should work properly now.


turetske (approved):
LGTM

PR #966, opened 2024-03-20
Update pelican client documentation

joereuss12:
More detailed and up to date client documentation as well as a quick-start guide


joereuss12:
Assigning all of you guys as reviewers to give feedback if you have time. I also slacked it in general for feedback as well.

joereuss12:
Edited names of files to be more consistent with current naming conventions

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,78 @@
+
+# Quick Start Guide
+The Pelican platform provides a platform that makes deploying data and accessing data easy. It provides an open-source software for federating dataset repositories together and delivering objects to computing capacity. For more information on Pelican, visit the [About Pelican](./index.mdx) page.
```
jhiemstrawisc, line 3: Let's go with something like:
`The developers of the Pelican Platform believe data is the lifeblood of science. That's why Pelican provides a suite of tools that make it easy for researchers to federate their data, giving them the ability to share objects from a broad range of data repositories and allowing them to connect their data with distributed computational capacity, such as the OSPool. For more information on Pelican...`

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,78 @@
...
+```
+which should output a path to the executable. If there is no output to this command, refer to the [Pelican installation docs](./install.mdx) to acquire a working installation.
+
+### Useful Terminology
+**Objects:**
+Pelican should be thought of as a tool that works with federated *objects* as opposed to *files*. The reason for this is that calling something a file carries with it the connotation that the file is mutable, i.e. its contents can change without requiring a new name. Objects in a Pelican federation, however, should be treated as **immutable**, especially in any case where objects are pulled through a cache (which will be the case for almost all files in the [Open Science Data Federation](https://osg-htc.org/services/osdf), or OSDF). This is because the underlying cache mechanism, powered by [XRootD](https://xrootd.slac.stanford.edu/), will deliver whatever object it already has access to; if an object name changes at the origin, the cache will remain unaware and continue to deliver the old object. In the worst case, when the cache only has a partial object, it may attempt to combine its stale version with whatever exists at the origin. Use object names wisely!
```
jhiemstrawisc, line 18: We might consider updating this a bit to include the fact that *files* carry the connotation of a filesystem (which has a lot of extra stuff like owners, permissions, etc). Pelican doesn't work with any of that.

Also, the note about the object name changing at the origin isn't true (sorry, I think I wrote that initially). If the name changes at the origin, we don't have any issues because we've created a unique object. However, if the actual *contents* of the object change without updating the name, that's when we run into issues. Can you tidy up that distinction?

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+
+**Federations:**
+Objects in Pelican belong to *federations*, which are aggregations of data that are exposed to other individuals in the federation. Each Pelican federation constitutes its own global namespace of objects and each object within a federation has its own path, much like files on a computer. Fetching any object from a federation requires at minimum two pieces of information: a URL indicating the object's federation and the path to the object within that federation (there is the potential that some objects require access tokens as well, but more on that later). For example, the OSDF’s federation URL is https://osg-htc.org and an example object from the federation can be found at the following path:
+```console
+/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
```
jhiemstrawisc, line 25: Let's add something like `Combining these two pieces of information, we say that the object's URL is https://osg-htc.org/ospool/uc-shared/public/OSG-Staff/validation/test.txt`

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,78 @@
...
+```bash
+/demo/testfile.txt
+```
+This is a simple example where the namespace prefix would be `/demo` and the object is named `/testfile.txt`.
+
+However, things can get a little tricky when you add a longer namespace prefix and/or a longer filepath within that namespace where the object lives. For example:
+```bash
+/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
+In this example, the namespace prefix is `/ospool/uc-shared/public/`, and the actual object is named `/OSG-Staff/validation/test.txt`. While the namespace prefix always comes before the object name, it can get a little confusing once paths and names get longer.
```
jhiemstrawisc, line 37: If we say it's confusing, we should also provide information to help disambiguate the two concepts. I'm not sure the best place to point people to, but one option is to point them to their federation's registry (eg https://osdf-registry.osg-htc.org/view/registry/), which should have a list of the Pelican namespaces available in their fed. This at least gives them a way to figure out what belongs to the namespace and what belongs to the object name.

@haoming29 @CannonLock do you have a better resource to help users figure this stuff out?

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,78 @@
...
+## Your First Pelican Client Command
+Accessing objects from a Pelican federation requires the use of the Pelican client commands. Here we will be using the `pelican object get` command to get an object from a federation. This command is formatted like so:
+```console
+pelican object get pelican://<federation-url>/<namespace-prefix>/<path/to/federation/object> </local/path/to/destination>
+```
+We can break down the components of this command before you try it yourself:
+- `pelican object get`: This is the client command we are utilizing. `object` because we want to interact with Pelican objects and `get` because we want to GET the file from our federation.
+- `pelican://`: This is the URL scheme for accessing this object. Pelican accepts different URL schemes which are detailed in the [client usage](./client_usage.mdx#the-different-pelican-url-schemes) page.
+- `<federation-url>`: The URL indicating the object's location.
+- `<namespace-prefix>`: The namespace at which this object is located.
```
jhiemstrawisc, line 49: Not a request for a change, per say, but I think we should think carefully about `namespace prefix`. WRT to origin configuration, we've changed the name to `FederationPrefix`, and I wonder if we should start consolidating the language before we wind up with a messy plate of name spaghetti. For now, let's keep it as is, but let's also bring this up soon.

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,78 @@
...
+- `<path/to/federation/object>`: The file path to the object we would like to get.
+- `</local/path/to/destination>`: The local destination where we want the file to download to (can be a local or absolute path)
+
+**Try The Command Yourself**
+Now it is time for you to get your first object from a federation. Before getting your first object, be sure you are in a directory/location where you want to download this test file. To get the object, run the command below:
+```console
+pelican object get pelican://osg-htc.org/ospool/uc-shared/public/OSG-Staff/validation/test.txt downloaded-test.txt
+```
+>**Note:** with pelican:// URLs like the one above, it is important to include the correct amount of slashes (`/`). This is because pelican URLs rely on the hostname being the federation URL therefore, they must only contain two slashes after `pelican`. For more information, visit the [Client-Usage](./client_usage.mdx#the-different-pelican-url-schemes) page
```
jhiemstrawisc, line 58: `amount --> number`

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,232 @@
...
+
+### The `pelican://` URL Scheme:
+When using the `pelican://` scheme, pelican does not know what federation to communicate with and therefore, the user must specify their federation URL either as the host-name in the URL itself or within the `-f` flag. The examples below show how you can format your request URLs in both of these ways.
+
+### The` osdf:///` URL Scheme:
+When using the `osdf:///` scheme, pelican will load in the federation metadata automatically with the Open Science Data Federation (OSDF)  defaults. This scheme should only be used if your federation or data can be accessed through the OSDF. When using this scheme, you should not specify the federation URL at all, as Pelican will handle this automatically. For example:
+```bash
+pelican object get osdf:///<namespace-prefix></path/to/file> <local/path/to/file>
+```
+This scheme has three slashes (`///`) after the `osdf` because the hostname is left empty to be automatically populated therefore, just start the URL with the namespace prefix. Pelican also supports the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.
```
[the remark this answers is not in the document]
joereuss12: Nope! That's why this sentence is there: `Pelican also supports the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.`

We still allow users to make the mistake of having only two `//` after osdf: and everything should work fine

docs/pages/quickstart_guide.mdx
joereuss12: Yeah if anyone has any tips on how to figure out the separation between namespace prefix and object path for users I'd love to know so I can include it. I just wasn't sure if we had anything for them to use.

docs/pages/quickstart_guide.mdx
joereuss12: Sounds good, good point!

joereuss12:
There's still some unaddressed comments that we should consider but need additional feedback. Just committed the latest edits!

docs/pages/quickstart_guide.mdx
haoming29: I think the best way is to look up the registry web UI, but that would be a non-trivial task for non-OSDF users, as the only way we know for sure to get the registry URL is by looking at the federation metadata (for the client user), which is not a good user experience. 

Since the director hosts the federation discovery URL, I think it would be nice if the director web UI can have a pointer to the registry web UI. This way it's more feasible for the client user to explore the federation. I will open a new ticket for this. For now, we can say contact your federation admin for how to access the registry (which also doesn't' quite help as we don't ask for federation/director admin to provide their contact and display it in the director UI... Life is hard.)

haoming29: Or even better, why not have an endpoint at the director that will redirect user to the registry instead? Something like `https://director.org/registry`

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,231 @@
...
+```
+which should output a path to the executable. If there is no output to this command, refer to the [Pelican installation docs](./install.mdx) to acquire a working installation.
+- You are roughly familiar with Pelican client commands and terminology as demonstrated with the [Quick Start Guide](./user-quickstart-guide.mdx)
+
+
+### Useful Terminology
+
+**Notes on Federations:**
+Below are some notes on federations on top of their terminology. Here is the example URL we used previously:
+```console
```
turetske, line 24: Previously where? Not in this document. The quick start guide? If so, it should be something like "previously in the quick start guide" or something similar.

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+>**Note:** When you set a federation discovery URL within a `pelican://` URL, within configuration files, or with the `-f` flag, do not expect the client to work if the discovery URL contains a path (e.g. `osg-htc.org/PathToMyFederation`). The client will assume that the path provided is the namespace prefix for the object which will end in failure.
+
+### The` osdf:///` URL Scheme:
+When using the `osdf:///` scheme, pelican will load in the federation metadata automatically with the Open Science Data Federation (OSDF)  defaults. This scheme should only be used if your federation or data can be accessed through the OSDF. When using this scheme, you should not specify the federation URL at all, as Pelican will handle this automatically. For example:
+```bash
+pelican object get osdf:///<namespace-prefix></path/to/file> <local/path/to/file>
+```
+>This scheme has three slashes (`///`) after the `osdf` because the hostname is left empty to be automatically populated therefore, just start the URL with the namespace prefix. Pelican also supports the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.
```
turetske, line 48:
```suggestion
>This scheme has three slashes (`///`) after the `osdf` because the hostname is left empty to be automatically populated. Therefore, just start the URL with the namespace prefix. Pelican also supports the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.
```

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+### Useful Terminology
+**Objects:**
+Pelican should be thought of as a tool that works with federated *objects* as opposed to *files*. The reason for this is that calling something a file carries with it the connotation of a filesystem. Filesystems have extra features such as owners, permissions, etc. and Pelican does not work with any of that. In addition, a *file* gives the connotation that the file is mutable, i.e. its contents can change without requiring a new name. Objects in a Pelican federation, however, should be treated as **immutable**, especially in any case where objects are pulled through a cache (which will be the case for almost all files in the [Open Science Data Federation](https://osg-htc.org/services/osdf), or OSDF). This is because the underlying cache mechanism, powered by [XRootD](https://xrootd.slac.stanford.edu/), will deliver whatever object it already has access to; if an object's contents change at the origin, the cache will remain unaware and continue to deliver the old object. In the worst case, when the cache only has a partial object, it may attempt to combine its stale version with whatever exists at the origin. Use object names wisely!
```
turetske, line 18:
```suggestion
Pelican should be thought of as a tool that works with federated *objects* as opposed to *files*. The reason for this is that calling something a file carries with it the connotation of a filesystem. Filesystems have extra features such as owners, permissions, etc. which Pelican does not implement. In addition, a *file* gives the connotation that the file is mutable, i.e. its contents can change without requiring a new name. Objects in a Pelican federation, however, should be treated as **immutable**, especially in any case where objects are pulled through a cache (which will be the case for almost all files in the [Open Science Data Federation](https://osg-htc.org/services/osdf), or OSDF). This is because the underlying cache mechanism, powered by [XRootD](https://xrootd.slac.stanford.edu/), will deliver whatever object it already has access to; if an object's contents change at the origin, the cache will remain unaware and continue to deliver the old object. In the worst case, when the cache only has a partial object, it may attempt to combine its stale version with whatever exists at the origin. Use object names wisely!
```

turetske (changes requested):

docs/pages/quickstart_guide.mdx
joereuss12: IMO, I think since @haoming29 made an issue to improve this, we should just maybe keep this and say it can be tricky. I can possibly add a note for now saying we are working on improving ways for the user to explore the federation and learn the differences or something. Let me know if this approach is okay with you guys. Check out the change once it is committed and see what you think.

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,86 @@
+
+# Quick Start Guide
+The developers of the Pelican Platform believe data is the lifeblood of science. That's why Pelican provides a suite of tools that make it easy for researchers to federate their data, giving them the ability to share objects from a broad range of data repositories and allowing them to connect their data with distributed computational capacity, such as the OSPool. For more information on Pelican, visit the [About Pelican](./index.mdx) page.
+
+### Assumptions
+
+Before using the Pelican client to interact with objects, this guide makes several assumptions:
+
+- You are on a computer where you have access to a terminal. The Pelican client is a command line tool.
+- You've already installed the version of Pelican appropriate for your system, and Pelican is accessible via your path. To test this on Linux, you can run
```
turetske, line 10:
```suggestion
- You've already installed the version of Pelican appropriate for your system and Pelican is accessible via your path. To test this on Linux, you can run
```

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,86 @@
...
+```
+This is a simple example where the namespace prefix would be `/demo` and the object is named `/testfile.txt`.
+
+However, things can get a little tricky when you add a longer namespace prefix and/or a longer filepath within that namespace where the object lives. For example:
+```bash
+/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
+In this example, the namespace prefix is `/ospool/uc-shared/public/`, and the actual object is named `/OSG-Staff/validation/test.txt`. While the namespace prefix always comes before the object name, it can get a little confusing once paths and names get longer.
+
+**How to differenciate namespace prefix and object name?**
```
turetske, line 43:
```suggestion
**How to differentiate namespace prefix and object name?**
```

docs/pages/quickstart_guide.mdx
```diff
@@ -0,0 +1,86 @@
...
+We can break down the components of this command before you try it yourself:
+- `pelican object get`: This is the client command we are utilizing. `object` because we want to interact with Pelican objects and `get` because we want to GET the file from our federation.
+- `pelican://`: This is the URL scheme for accessing this object. Pelican accepts different URL schemes which are detailed in the [client usage](./client_usage.mdx#the-different-pelican-url-schemes) page.
+- `<federation-url>`: The URL indicating the object's location.
+- `<namespace-prefix>`: The namespace at which this object is located.
+- `<path/to/federation/object>`: The file path to the object we would like to get.
+- `</local/path/to/destination>`: The local destination where we want the file to download to (can be a local or absolute path)
+
+**Try The Command Yourself**
+Now it is time for you to get your first object from a federation. Before getting your first object, be sure you are in a directory/location where you want to download this test file. To get the object, run the command below:
```
turetske, line 62: Can they not just use something like "my-directory/downloaded-test.txt" as the destination path name if they aren't in the directory they want to be in? If they can't, possibly specify that this is a requirement for whatever reasons?

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,231 @@
...
+When using the `osdf:///` scheme, pelican will load in the federation metadata automatically with the Open Science Data Federation (OSDF)  defaults. This scheme should only be used if your federation or data can be accessed through the OSDF. When using this scheme, you should not specify the federation URL at all, as Pelican will handle this automatically. For example:
+```bash
+pelican object get osdf:///<namespace-prefix></path/to/file> <local/path/to/file>
+```
+>This scheme has three slashes (`///`) after the `osdf` because the hostname is left empty to be automatically populated. Therefore, just start the URL with the namespace prefix. Pelican also supports the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.
+
+### The `stash:///` URL Scheme:
+This URL scheme comes from legacy code called `stashcp` that Pelican was built off of and is still accepted for backwards compatibility. This URL scheme functions identically to the `osdf:///` URL scheme, loading in OSDF defaults. Here is an example:
+```bash
+pelican object get stash:///<namespace-prefix></path/to/file> <local/path/to/file>
```
turetske, line 53: This is a follow up to Justin's earlier comment about namespace-prefix. This is mostly a placeholder to remind you to do the change in this file in multiple locations.

turetske (changes requested):

docs/pages/quickstart_guide.mdx
joereuss12: They can, but for this example I think we should just keep it as simple/basic as possible since this is just a getting started guide. In the regular client usage and in other examples it is mentioned they can specify the path they would like to use for the destination.

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+**Tokens and JWT:**
+Some namespace prefixes are public, like `/ospool/uc-shared/public`, while others are protected (i.e. they require authorization). Objects in public namespaces can be downloaded by anybody, but downloading objects from protected namespaces requires you prove to the origin supporting that namespace that you are allowed to access the object. In Pelican, this is done using signed JSON Web Tokens, or *JWT*s for short. In many cases, these tokens can be generated automatically.
+
+## The Different Pelican URL Schemes
+When running client commands with a request URL, there are three different URL schemes a user can use: `pelican`, `osdf`, and `stash`. This page uses examples with the `pelican://` URL scheme. Below is a description of the three URL schemes and what to expect when they are used:
+
+### The `pelican://` URL Scheme:
+When using the `pelican://` scheme, pelican does not know what federation to communicate with and therefore, the user must specify their federation URL either as the host-name in the URL itself or within the `-f` flag. The examples below show how you can format your request URLs in both of these ways.
```
jhiemstrawisc, line 36: This is our first chance to introduce the powerful concept of `pelican://` urls. I think we should avoid mentioning the `-f` flag here, because if the user is using a `pelican://` url, we want them to be including the federation hostname.

This may not be the place to do so, but we should also explain what protocol this scheme entails. Something like:

"URLs that begin with `pelican://` allow you to interact with objects from any federation simply by knowing the hostname of the federation's discovery url and the name of the object in that federation. These URLs take the form 
```
pelican://<fed discovery URL>/<path to object>
```
When a Pelican client encounters a Pelican URL, it uses the hostname in the federation discovery portion of the URL to find other services in the specified federation that will help the client discover the object's location."

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+### The `osdf:///` URL Scheme:
+When using the `osdf:///` scheme, pelican will load in the federation metadata automatically with the Open Science Data Federation (OSDF)  defaults. This scheme should only be used if your federation or data can be accessed through the OSDF. When using this scheme, you should not specify the federation URL at all, as Pelican will handle this automatically. For example:
+```bash
+pelican object get osdf:///<namespace-prefix></path/to/file> <local/path/to/file>
+```
+>This scheme has three slashes (`///`) after the `osdf` because the hostname is left empty to be automatically populated therefore, just start the URL with the namespace prefix. Pelican also supports the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.
+
+### The `stash:///` URL Scheme:
```
jhiemstrawisc, line 47: Unless anyone else objects, let's get rid of documenting stash-protocol URLs.

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+To use the pelican client to get public objects from a federation, use Pelican's `object get` sub-command
+
+### Getting an Object Using the pelican:// URL Scheme:
+To use the `pelican://` URL scheme, you need to specify the federation URL within the request URL. To do this, you would format the `object get` command like so:
+```bash
+pelican object get pelican://<federation-url></namespace-prefix></path/to/file> <local/path/to/file>
+```
+Or, you can use the `-f` flag with the command, and format the request URL like so:
+```bash
+pelican object get -f <federation-url> pelican://</namespace-prefix></path/to/file> </local/path/to/file>
```
jhiemstrawisc, line 64: I don't like the idea of presenting this as an option to users. Can we say you can either do:
```

pelican object get -f <fed url> /namespace/prefix/object /local/path/to/file
```

OR

```
pelican object get pelican://<fed url>/namespace/prefix/object /local/path/to/file
```

To the extent possible, let's try to make sure users think of pelican-schemed URLs as a way to provide the federation discovery URL and object name in a single vessel, not to split things up the confusing way we do with `osdf:///`

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+More specifically, it breaks into these components:
+- Federation URL: `osg-htc.org`
+- Namespace: `/ospool/uc-shared/public`
+- Object name: `/OSG-Staff/validation/test.txt`
+- Destination: `downloaded-testfile.txt`
+
+
+## Get A Protected Object From Your Federation
+
+Protected namespaces require that a Pelican client prove it is allowed to access objects from the namespace before the object can be downloaded. In many cases, Pelican clients can do this automatically by initiating an OpenID-Connect (OIDC) flow that uses an external log-in service (such as [CILogon](https://www.cilogon.org)) through your browser. In other cases, a token must be provided to the Pelican client manually.
```
jhiemstrawisc, line 87: It's not quite right to say that Pelican clients can do this automatically. Let's say "In many cases, Pelican clients can guide users through the process of acquiring a token by initiating an OpenID-Connect (OIDC) flow that ...". As written, it sounds like the client will automagically show up with a token on its own, but the truth is the user has to be around to enter authentication information.

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+### For Issuers That Support CILogon Code Flow
+
+Some origins support authentication with CILogon's OIDC client. In these cases, the Pelican client is capable of handling authentication with the origin automatically, once you logged in via CILogon. To download protected objects from origins that support CILogon, run the same command as for downloading a public object:
```
jhiemstrawisc, line 91: Same comments about tidying up the language around the "automatically" phrasing.

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+```
+
+For example, if a token is saved in a file named `my-token`, it can be used to get the object `/ospool/PROTECTED/auth-test.txt` by running:
+```bash
+# Get the object with the path to token file provided via -t flag
+$ pelican object get -f https://osg-htc.org /ospool/PROTECTED/auth-test.txt downloaded-auth-test.txt -t my-token
+```
+
+## PUT an object to Your Federation
+Another powerful Pelican client command is the `pelican object put` command. This command does a simple PUT request to add your object to your federation. Since the vast majority of federations will not allow public writes, we will assume for the example that we will need a token to perform these requests. Here is how you can use `pelican object put`:
```
jhiemstrawisc, line 137: Public writes actually aren't possible in Pelican, so we can just say "Putting files into a federation requires a token"

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+and to do an `object put`:
+
+```bash
+pelican object copy <path/to/local/file> pelican://<federation-url></namespace-prefix></path/to/destination> -t </path/to/token/file>
+```
+
+## Utilizing Queries with your URL
+The Pelican client has the ability to select various amount of queries specified in the remote path to an object. The queries the current Pelican client accepts are: ?pack, ?recursive, and ?directread
+
+### ?pack Query
```
jhiemstrawisc, line 158: Instead of beginning a header with the query, let's say something about what the section tells you how to do. For this one, maybe `Packing Objects with the ?pack Query`. The next header could be `Recursive Downloads with the ?recursive Query`

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+Pelican accepts the following values for the `pack` query:
+- `pack=auto`:
+	- For downloading, auto-detect the file format and unpack (throws error if it is not any detected format).
+	- For uploading, compress using `.tar.xz`.
+- `pack=tar`, `pack=tar.gz`, `pack=tar.xz`, `pack=zip` :
+	- For downloading, throws an error if the specified object is not in the specified format (`tar`, `tar.gz`, `tar.xz`, `zip`, respectively).
+	- For uploading, create the object in the specified format (`tar`, `tar.gz`, `tar.xz`, `zip`, respectively).
+
+### ?recursive Query
+The ?recursive query can be utilized if the desired remote object is a directory. When this query is enabled, Pelican will recursively download all files and subdirectories within the specified directory in the URL. To use this query, run:
```
jhiemstrawisc, line 177: I'm conflicted here, because we should be saying "objects/namespace paths" and not "files/directories", but I don't know that `?recursive` works for anything _other_ than posix backends right now. Maybe we can say something like "When this query is enabled, Pelican will recursively try to download everything beneath the specified path." Of course, that's not entirely true either...

EDIT:
I read a bit further to the point where you describe the behavior of the `-r` flag in the client, and I like the way you've phrased things there when you say "... indicates to Pelican that all sub paths at the level of the provided namespace should be copied recursively."

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+>**Note:** This query functions the same as specifying the `-r` flag described below.
+
+### ?directread Query
+The ?directread query is used if you would like to download an object without utilizing a cache. This way, the object will come directly from the origin each time and never use the cache. To use this query, you can run:
+```bash
+pelican object get pelican://<federation-url></namespace-prefix></path/to/file>?directread <local/path/to/file>
+```
+This query does not make sense to use for uploads since uploads go directly to the origin anyway. If you use this query by mistake, you should not run into any issues and your upload will function as normal.
+
+> **Note about Queries:** the ?recursive and the ?directread queries do **not** require any sort of values assigned to them (i.e. pelican://some/object?recursive=true). If a value is assigned to these queries, that value will be ignored and Pelican will act as if there was no value assigned to that query (i.e. pelican://some/object?recursive=false acts the same as pelican://some/object?recursive meaning the recursive query **will** be set even if the value is set to false).
```
jhiemstrawisc, line 196: These should be `e.g.`

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+
+## Additional Flags
+
+The Pelican client support a variety of command line flags that modify the client's behavior:
```
jhiemstrawisc, line 201: `support --> supports`

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+## Aliases of The Pelican Binary
+
+The Pelican binary can change its behavior depending on what it is named. This feature serves two purposes; it allows Pelican to use a few convenient default settings in the case that the federation being interacted with is the OSDF, and it allows Pelican to run in legacy `stashcp` and `stash_plugin` modes.
+
+### Prefixing The Binary With OSDF
+
+When the name of the Pelican binary begins with `osdf`, Pelican will assume that all objects are coming from the OSDF which allows it to make several assumptions. The most immediate effect for users is that you no longer need to specify a federation URL, ONLY with no URL scheme, `osdf:///`, or `stash:///` URL schemes. The command to download a public file from above can then be simplified to:
+
+```bash
+osdf object copy osdf:///ospool/uc-shared/public/OSG-Staff/validation/test.txt downloaded-testfile.txt
```
jhiemstrawisc, line 231: This can be simplified further to `osdf object copy /ospool/uc-shared/public/OSG-Staff/validation/test.txt downloaded-testfile.txt`, can't it? The `osdf://` scheme does the same thing when using the regular pelican client, so there's not a whole lot of benefit to naming the binary `osdf` unless we tell the user they can stop using ANY url-scheme stuff.

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,93 @@
...
+```bash
+/demo/testfile.txt
+```
+This is a simple example where the namespace prefix would be `/demo` and the object is named `/testfile.txt`.
+
+However, things can get a little tricky when you add a longer namespace prefix and/or a longer filepath within that namespace where the object lives. For example:
+```bash
+/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
+In this example, the namespace prefix is `/ospool/uc-shared/public/`, and the actual object is named `/OSG-Staff/validation/test.txt`. While the namespace prefix always comes before the object name, it can get a little confusing once paths and names get longer.
```
jhiemstrawisc, line 35: Let's remove the second sentence in this line. It's not a great look to write in our own docs that we make confusing software

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+**Federations:**
+Objects in Pelican belong to *federations*, which are aggregations of data that are exposed to other individuals in the federation. Each Pelican federation constitutes its own global namespace of objects and each object within a federation has its own path, much like files on a computer. Fetching any object from a federation requires at minimum three pieces of information: a URL indicating the object's federation, a namespace prefix, and the path to the object within that federation (there is the potential that some objects require access tokens as well, but more on that later). For example, the OSDF’s federation URL is https://osg-htc.org, an example namespace-prefix from the federation is:
```
jhiemstrawisc, line 41: The more I think about it, the more I think we should start framing this differently. Getting an object requires 2 pieces of information -- the federation's root (i.e. the discovery URL) and the path to the object within that federation. If we were to explain how to get a file off of a computer's filesystem, we wouldn't say  you need to know the name of the directory and the name of the file, we'd just say filepath. 

Getting into details about how we distinguish between federation prefixes and object names is probably doing client users a disservice because that doesn't mean much for anyone who's not serving an origin or running a registry. At the end of the day, the client users only care about "which federation" and "what object".

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,93 @@
...
+/ospool/uc-shared/public
+```
+and the location of the object can be found at:
+```bash
+/OSG-Staff/validation/test.txt
+```
+
+Combining these three pieces of information, we say the the **object's URL** is:
+```bash
+https://osg-htc.org/ospool/uc-shared/public/OSG-Staff/validation/test.txt
```
jhiemstrawisc, line 52: This would have to be `pelican://` to be true.

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+## Your First Pelican Client Command
+Accessing objects from a Pelican federation requires the use of the Pelican client commands. Here we will be using the `pelican object get` command to get an object from a federation. This command is formatted like so:
+```console
+pelican object get pelican://<federation-url>/<namespace-prefix>/<path/to/federation/object> </local/path/to/destination>
+```
+We can break down the components of this command before you try it yourself:
+- `pelican object get`: This is the client command we are utilizing. `object` because we want to interact with Pelican objects and `get` because we want to GET the file from our federation.
+- `pelican://`: This is the URL scheme for accessing this object. Pelican accepts different URL schemes which are detailed in the [client usage](./client_usage.mdx#the-different-pelican-url-schemes) page.
+- `<federation-url>`: The URL indicating the object's location.
```
jhiemstrawisc, line 63: The URL indicating which federation you'd like to interact with.

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,93 @@
...
+/OSG-Staff/validation/test.txt
+```
+
+Combining these three pieces of information, we say the the **object's URL** is:
+```bash
+https://osg-htc.org/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
+
+## Your First Pelican Client Command
+Accessing objects from a Pelican federation requires the use of the Pelican client commands. Here we will be using the `pelican object get` command to get an object from a federation. This command is formatted like so:
```
jhiemstrawisc, line 56: Technically you can't get objects from a pelican federation with anything that speaks HTTP, not just Pelican clients. Maybe rephrase?

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,93 @@
...
+```console
+pelican object get pelican://<federation-url>/<namespace-prefix>/<path/to/federation/object> </local/path/to/destination>
```
jhiemstrawisc, line 58: Again, I think we should start framing this just as the object path within the federation.

jhiemstrawisc (changes requested):

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,237 @@
...
+```console
+/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
+**Note:** All object paths in a federation begin with a preceding `/`, and no relative paths are allowed.
+
+**Tokens and JWT:**
+Some namespace prefixes are public, like `/ospool/uc-shared/public`, while others are protected (i.e. they require authorization). Objects in public namespaces can be downloaded by anybody, but downloading objects from protected namespaces requires you prove to the origin supporting that namespace that you are allowed to access the object. In Pelican, this is done using signed JSON Web Tokens, or *JWT*s for short. In many cases, these tokens can be generated automatically.
+
+## The Different Pelican URL Schemes
+When running client commands with a request URL, there are three different URL schemes a user can use: `pelican`, `osdf`, and `stash`. This page uses examples with the `pelican://` URL scheme. Below is a description of the three URL schemes and what to expect when they are used:
```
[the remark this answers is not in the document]
joereuss12: No I agree, since we are working out the old 'stash' we should probably not document it

docs/pages/client_usage.mdx
joereuss12: I think I am just going to remove the examples with `-f` and just leave it as a note. I was not too fond of having it either, takes up space and can be confusing/not useful to mention

docs/pages/client_usage.mdx
joereuss12: You are correct, having no scheme will do the same as having an `osdf://` scheme. I'll leave the example scheme blank to show users it can be simplified even further.

docs/pages/getting_started_access_data.mdx
joereuss12: That sounds good to me, especially since we can kind of think of namespaces/namespace prefixes as "directories" in a sense (using posix terms) and since that seems to be the way we are moving. Sounds a lot less confusing to me as well.

joereuss12:
I left the examples containing </namespace/prefix>/<path/to/object> however when we have our discussion on namespaces, this can be subject to change.

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+### Assumptions
+
+Before using the Pelican client to interact with objects from your federation, this guide makes several assumptions:
+
+- You are on a computer where you have access to a terminal. The Pelican client is a command line tool.
+- You've already installed the version of Pelican appropriate for your system, and Pelican is accessible via your path. To test this on Linux, you can run
+```bash
+which pelican
+```
+which should output a path to the executable. If there is no output to this command, refer to the [Pelican installation docs](./install.mdx) to acquire a working installation.
```
aowen-uwmad, line 16: Add two spaces to the front of each line, so that the content is formatted as contents of the line item.

(suggestion markdown doesn't render properly..)

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+- You are roughly familiar with Pelican client commands and terminology as demonstrated with the [Getting Started: Accessing Data](./getting_started_access_data.mdx) page.
+
+
+### Useful Terminology
+
+**Notes on Federations:**
```
aowen-uwmad, line 22:
```suggestion
**Note on Federations:**
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+```console
+/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
+This is the full object path that we will be using in examples below.
+
+**Tokens and JWT:**
+Some namespace prefixes are public, like `/ospool/uc-shared/public`, while others are protected (i.e. they require authorization). Objects in public namespaces can be downloaded by anybody, but downloading objects from protected namespaces requires you prove to the origin supporting that namespace that you are allowed to access the object. In Pelican, this is done using signed JSON Web Tokens, or *JWT*s for short. In many cases, these tokens can be generated automatically.
+
+## The Different Pelican URL Schemes
+When running client commands with a request URL, there are three different URL schemes a user can use: `pelican` and `osdf`. This page uses examples with the `pelican://` URL scheme. Below is a description of the three URL schemes and what to expect when they are used:
```
aowen-uwmad, line 33:
```suggestion
When running client commands with a request URL, there are two different URL schemes a user can use: `pelican` and `osdf`. This page uses examples with the `pelican://` URL scheme. Below is a description of the two URL schemes and what to expect when they are used:
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+### The `pelican://` URL Scheme:
+URLs that being with `pelican://` allow you to interact with objects from any federation simply by knowing the hostname of the federation's discovery url and the name of that object in that federation. These URLs take the form:
```
aowen-uwmad, line 36:
```suggestion
URLs that begin with `pelican://` allow you to interact with objects from any federation simply by knowing the hostname of the federation's discovery URL and the name of that object in that federation. These URLs take the form:
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+pelican://<federation discovery URL>/<path/to/object>
+```
+When a Pelican client encounters a Pelican URL, it uses the hostname in the federation discovery portion of the URL to find other services in the specified federation that will help the client discover the object's location
+
+>**Note**: Federation discovery URLs can also be specified with the `-f` flag rather than the hostname. Just be sure not to include the URL in both the `-f` flag and in the hostname of the object path, as this will lead to errors.
+
+>**Note:** Federation discovery URLs cannot contain a path component (e.g. https://my-federation.com/path/component). Clients that encounter a discovery URL with a path component will incorrectly assume the path is part of the namespace prefix.
+
+### The `osdf:///` URL Scheme:
+When using the `osdf:///` scheme, pelican will load in the federation metadata automatically with the Open Science Data Federation (OSDF)  defaults. This scheme should only be used if your federation or data can be accessed through the OSDF. When using this scheme, you should not specify the federation URL at all, as Pelican will handle this automatically. For example:
```
aowen-uwmad, line 47:
```suggestion
When using the `osdf:///` scheme, Pelican will load in the federation metadata automatically using the defaults for accessing the Open Science Data Federation (OSDF). This scheme should only be used if your data can be accessed through the OSDF. When using this scheme, you should not specify the federation URL at all, as Pelican will handle this automatically. For example:
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+```bash
+pelican object get osdf:///<namespace-prefix></path/to/file> <local/path/to/file>
+```
+>This scheme has three slashes (`///`) after the `osdf` because the hostname is left empty to be automatically populated therefore, just start the URL with the namespace prefix. Pelican also supports the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.
```
aowen-uwmad, line 51:
```suggestion
>This scheme has three slashes (`///`) after the `osdf` because the hostname is left empty to be automatically populated therefore, just start the URL with the namespace prefix. Pelican currently recognizes the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+## Get a Public Object from your Federation
+
+To use the pelican client to get public objects from a federation, use Pelican's `object get` sub-command
+
+### Getting an Object Using the pelican:// URL Scheme:
+To use the `pelican://` URL scheme, you need to specify the federation URL within the request URL. To do this, you would format the `object get` command like so:
+```bash
+pelican object get pelican://<federation-url></namespace-prefix></path/to/file> <local/path/to/file>
+```
+You can try this yourself by getting the public file that was mentioned earlier from the OSDF. Using the `object get` sub command, and indicating to pelican that you're pulling from the OSDF:
```
aowen-uwmad, line 62:
```suggestion
You can try this yourself by getting the public file that was mentioned earlier from the OSDF. Using the `object get` sub command, and providing the federation URL for the OSDF:
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+<ExportedImage src={"/pelican/cilogon.png"} alt={"Screenshot of CILogon success page"} />
+
+Finally, if the login is successful, Pelican will automatically fetch the token from the CILogon service and continue with the download.
+
+### For Issuers Without CILogon Support
+
+Some origins do not support authentication though CILogon. In this case, users must supply their own JWT that's signed by the origin.
+
+> We are working hard to document how to generate a token from protected origins, before it’s ready. If you have any questions or need help getting a token from the origin, please email us at [help@pelicanplatform.com](mailto:help@pelicanplatform.com) or open a discussion on our [GitHub discussions page](https://github.com/PelicanPlatform/pelican/discussions) and we are happy to assist you.
```
aowen-uwmad, line 112: "before it's ready" - this phrase feels odd to me. Do you mean "but it's not currently ready"? I'm not quite sure what is trying to be communicated here.

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+```
+
+For example, if a token is saved in a file named `my-token`, it can be used to get the object `/ospool/PROTECTED/auth-test.txt` by running:
+```bash
+# Get the object with the path to token file provided via -t flag
+$ pelican object get -f https://osg-htc.org /ospool/PROTECTED/auth-test.txt downloaded-auth-test.txt -t my-token
+```
+
+## PUT an object to Your Federation
+Another powerful Pelican client command is the `pelican object put` command. This command does a simple PUT request to add your object to your federation. Putting files into a federation always requires a token therefore, for the example, we will need a token to perform these requests. Here is how you can use `pelican object put`:
```
aowen-uwmad, line 129:
```suggestion
Another powerful Pelican client command is the `pelican object put` command. This command does a simple PUT request to add your object to your federation. Putting files into a federation always requires a token therefore, for the example, we will need a token to perform these requests (see the [previous section](#get-a-protected-object-from-your-federation) for more information). Here is how you can use `pelican object put`:
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+```bash
+pelican object put <path/to/local/file> pelican://<federation-url></namespace-prefix></path/to/file> -t </path/to/token/file>
+```
+> **Note:** you can also specify the federation url here with the `-f` flag, just be sure not to include it in the request URL as the host name if you decide to do so.
+
+## Pelican Object Copy
+As well with Pelican's `object get` and `object put` commands, there is also an older command called `pelican object copy`. This functions the same as the `pelican object get/put` except that it works for both gets and puts. For example, to do an `object get`:
```
aowen-uwmad, line 137: Do you recommend that users not use the `copy` command? If so, I suggest adding such a statement here.

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+pelican object copy pelican://<federation-url></namespace-prefix></path/to/file> <local/path/to/file>
+```
+and to do an `object put`:
+
+```bash
+pelican object copy <path/to/local/file> pelican://<federation-url></namespace-prefix></path/to/destination> -t </path/to/token/file>
+```
+
+## Utilizing Queries with your URL
+The Pelican client has the ability to select various amount of queries specified in the remote path to an object. The queries the current Pelican client accepts are: ?pack, ?recursive, and ?directread
```
aowen-uwmad, line 148:
```suggestion
The Pelican client has the ability to select various amount of queries specified in the remote path to an object. The queries the current Pelican client accepts are: `?pack`, `?recursive`, and `?directread`.
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+Pelican accepts the following values for the `pack` query:
+- `pack=auto`:
+	- For downloading, auto-detect the file format and unpack (throws error if it is not any detected format).
+	- For uploading, compress using `.tar.xz`.
+- `pack=tar`, `pack=tar.gz`, `pack=tar.xz`, `pack=zip` :
+	- For downloading, throws an error if the specified object is not in the specified format (`tar`, `tar.gz`, `tar.xz`, `zip`, respectively).
+	- For uploading, create the object in the specified format (`tar`, `tar.gz`, `tar.xz`, `zip`, respectively).
+
+### Recurisve Downloads and Uploads with the ?recursive Query
+The ?recursive query can be utilized if the desired remote object is a directory. When this query is enabled, it indicates to Pelican that all sub paths at the level of the provided namespace should be copied recursively. To use this query, run:
```
aowen-uwmad, line 169:
```suggestion
The `?recursive` query can be utilized if the desired remote object is a directory. When this query is enabled, it indicates to Pelican that all sub paths at the level of the provided namespace should be copied recursively. To use this query, run:
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+To upload, you can run something similar but with an `object put`:
+```bash
+pelican object put <local/path/to/directory> pelican://<federation-url></namespace-prefix></path/to/directory>?recursive
+```
+
+>**Note:** This query functions the same as specifying the `-r` flag described below.
+
+### Bypass Caches for Downloads with the ?directread Query
+The ?directread query is used if you would like to download an object without utilizing a cache. This way, the object will come directly from the origin each time and never use the cache. To use this query, you can run:
```
aowen-uwmad, line 182:
```suggestion
The `?directread` query is used if you would like to download an object without utilizing a cache. This way, the object will come directly from the origin each time and never use the cache. To use this query, you can run:
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+```bash
+pelican object get pelican://<federation-url></namespace-prefix></path/to/file>?directread <local/path/to/file>
+```
+This query does not make sense to use for uploads since uploads go directly to the origin anyway. If you use this query by mistake, you should not run into any issues and your upload will function as normal.
+
+> **Note about Queries:** the ?recursive and the ?directread queries do **not** require any sort of values assigned to them (e.g. pelican://some/object?recursive=true). If a value is assigned to these queries, that value will be ignored and Pelican will act as if there was no value assigned to that query (e.g. pelican://some/object?recursive=false acts the same as pelican://some/object?recursive meaning the recursive query **will** be set even if the value is set to false).
```
aowen-uwmad, line 188:
```suggestion
### Note about Queries
The `?recursive` and the `?directread` queries do **not** require any sort of values assigned to them (e.g., `pelican://some/object?recursive=true`). If a value is assigned to these queries, that value will be ignored and Pelican will act as if there was no value assigned to that query (e.g. `pelican://some/object?recursive=false` acts the same as `pelican://some/object?recursive` meaning the recursive query **will** be set even if the value is set to false).
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+- **-h or --help:** Gives additional information on how to use the command as well as lists these flags with short descriptions for the `object copy` command.
+- **--methods:** Takes a comma seperated list of methods to try for downloads/uploads, the default is just http.
+- **-r or --recursive:** Takes no argument and indicates to Pelican that all sub paths at the level of the provided namespace should be copied recursively. This option is only supported if the origin supports the WebDav protocol.
+- **-t or --token:** Takes a path to a file containing a signed JWT, and is used to download protected objects.
+
+## Aliases of The Pelican Binary
+
+The Pelican binary can change its behavior depending on what it is named. This feature serves two purposes; it allows Pelican to use a few convenient default settings in the case that the federation being interacted with is the OSDF, and it allows Pelican to run in legacy `stashcp` and `stash_plugin` modes.
+
+### Prefixing The Binary With OSDF
```
aowen-uwmad, line 218:
```suggestion
### Prefixing The Binary Name With `osdf`
```

docs/pages/client_usage.mdx
```diff
@@ -0,0 +1,229 @@
...
+When the name of the Pelican binary begins with `osdf`, Pelican will assume that all objects are coming from the OSDF which allows it to make several assumptions. The most immediate effect for users is that you no longer need to specify a federation URL, ONLY with no URL scheme or an `osdf:///` URL scheme. The command to download a public file from above can then be simplified to:
+
+```bash
+osdf object copy /ospool/uc-shared/public/OSG-Staff/validation/test.txt downloaded-testfile.txt
+```
+>**Note:** When using client commands with the OSDF binary, be careful when using it with `pelican://` URL schemes. When using the `pelican://` URL scheme, you are still required to provide a federation URL no matter what binary name you are using.
+
+### Naming The Binary `stashcp` Or `stash_plugin`
+
+The Pelican Project grew out of a command line tool called `stashcp` with an associated HTCondor plugin called `stash_plugin`, which were also used for interacting with objects in the OSDF. To support these legacy tools, Pelican has been built to behave similarly as `stashcp` and `stash_plugin` did whenever the Pelican binary is renamed to match the names of these tools.
```
aowen-uwmad, line 229:
```suggestion
The Pelican Platform grew out of a command line tool called `stashcp` with an associated HTCondor plugin called `stash_plugin`, which were also used for interacting with objects in the OSDF. To support these legacy tools, Pelican has been built to behave similarly as `stashcp` and `stash_plugin` did whenever the Pelican binary is renamed to match the names of these tools.
```

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,89 @@
...
+### Assumptions
+
+Before using the Pelican client to interact with objects, this guide makes several assumptions:
+
+- You are on a computer where you have access to a terminal. The Pelican client is a command line tool.
+- You've already installed the version of Pelican appropriate for your system, and Pelican is accessible via your path. To test this on Linux, you can run
+```console
+which pelican
+```
+which should output a path to the executable. If there is no output to this command, refer to the [Pelican installation docs](./install.mdx) to acquire a working installation.
```
aowen-uwmad, line 15: Add two spaces at the start of each line to make the markdown formatting recognize the content as a part of the previous line item, and not as a separate paragraph.

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+**Objects:**
+Pelican should be thought of as a tool that works with federated *objects* as opposed to *files*. The reason for this is that calling something a file carries with it the connotation of a filesystem. Filesystems have extra features such as owners, permissions, etc. which Pelican does not implement. In addition, a *file* gives the connotation that the file is mutable, i.e. its contents can change without requiring a new name. Objects in a Pelican federation, however, should be treated as **immutable**, especially in any case where objects are pulled through a cache (which will be the case for almost all files in the [Open Science Data Federation](https://osg-htc.org/services/osdf), or OSDF). This is because the underlying cache mechanism, powered by [XRootD](https://xrootd.slac.stanford.edu/), will deliver whatever object it already has access to; if an object's contents change at the origin, the cache will remain unaware and continue to deliver the old object. In the worst case, when the cache only has a partial object, it may attempt to combine its stale version with whatever exists at the origin. Use object names wisely!
```
aowen-uwmad, line 22:
```suggestion
Pelican should be thought of as a tool that works with federated *objects* as opposed to *files*. The reason for this is that calling something a file carries with it the connotation of a filesystem. Filesystems have extra features such as owners, permissions, etc., which Pelican does not implement. In addition, a *file* gives the connotation that the file is mutable, i.e., its contents can change without requiring a new name. Objects in a Pelican federation, however, should be treated as **immutable**, especially in any case where objects are pulled through a cache (which will be the case for almost all files in the [Open Science Data Federation](https://osg-htc.org/services/osdf), or OSDF). This is because the underlying cache mechanism, powered by [XRootD](https://xrootd.slac.stanford.edu/), will deliver whatever object it already has access to; if an object's contents change at the origin, the cache will remain unaware and continue to deliver the old object. In the worst case, when the cache only has a partial object, it may attempt to combine its stale version with whatever exists at the origin. Use object names wisely!
```

docs/pages/getting_started_access_data.mdx
```diff
@@ -0,0 +1,89 @@
...
+/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
+
+Combining these two pieces of information, we say the the **object's URL** is:
+```bash
+pelican://osg-htc.org/ospool/uc-shared/public/OSG-Staff/validation/test.txt
+```
+
+## Your First Pelican Client Command
+You can easily access objects from a Pelican federation with the use of the Pelican client commands. Here we will be using the `pelican object get` command to get an object from a federation. This command is formatted like so:
```
aowen-uwmad, line 52:
```suggestion
You can easily access public objects from a Pelican federation with the use of the Pelican client commands. Here we will be using the `pelican object get` command to get an object from a federation. This command is formatted like so:
```

To get protected objects requires tokens, which I hesitate to describe as "easy".

haoming29:
Closed with #1381 open

PR #968, opened 2024-03-20
Add support for custom xrootd config file

### LINKED ISSUE #921: Allow custom XRootD configuration
Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [x] Origin
- [x] Cache

It became apparent in a recent origin deployment that we were missing a single line in the xrootd origin config. The fix required a very simple patch that couldn't be delivered because of a kink in our release infrastructure, preventing the origin from utilizing legacy OSDF caching infrastructure (which still makes up the vast majority of OSDF caches). 

Since xrootd configurations are something we automatically generate/overwrite every time we start an xrootd server, there's no way for knowledgeable administrators to ~~get them selves in trouble~~ fix their own issues by passing custom configuration.

We already handle merging authfiles and scitokens files. Let's figure out how to do something similar so that if I provide an XRootD config file, it gets applied on top of the one we generate. Let's also make this feature poorly documented and hard to find so that people aren't likely to use it without looping us in -- it would be bad if everyone started fixing actual issues with the xrootd configuration and didn't tell us about the issue in the first place.

haoming29:
Closes #921


docs/parameters.yaml
```diff
@@ -1548,6 +1548,20 @@ deprecated: true
 replacedby: ["Cache.RunLocation", "Origin.RunLocation"]
 components: ["origin", "cache"]
 ---
+name: Xrootd.ConfigFile
```
jhiemstrawisc, line 1551: @turetske, I wanted to get your feedback here. Should we immediately split this out into something like `Origin.XrootdConfigFile` and `Cache.XrootdConfigFile` so that we can do custom stuff for fed in a box, or do you not anticipate this being an issue?

xrootd/resources/xrootd-origin.cfg
```diff
@@ -93,3 +93,6 @@ xrd.trace {{.Logging.OriginXrd}}
 cms.trace {{.Logging.OriginCms}}
 xrootd.tls all
 scitokens.trace {{.Logging.OriginScitokens}}
+{{if .Xrootd.ConfigFile}}
+continue {{.Xrootd.ConfigFile}}
```
jhiemstrawisc, line 97: Up to you, but it might be nice to do some validation on this beforehand so we can move the error domain out of XRootD and into Pelican. I know how much we all love debugging XRootD's cryptic ramblings ;) A basic check might look for:
```
- If a file path is provided, does it exist?
- If a file path is provided, does Pelican have permissions to read it?
```

A more advanced check might provide explicit information about anything that the provided config file overwrites (as opposed to adds to).

jhiemstrawisc: Another thing that would be helpful is to add a debug statement somewhere that mentions a custom config is being used and the path to the config. That way more of the useful information makes it into the log and requires less actual knowledge of the `pelican.yaml`.

haoming29: I like the error checking idea and I indeed __love__ debugging with XRootD.

haoming29: > A more advanced check might provide explicit information about anything that the provided config file overwrites (as opposed to adds to).

This is going to be very tricky as XRootD has its own policy at each directive level on overwrite/appending, which is not explicitly exposed anywhere (besides the doc)

docs/parameters.yaml
jhiemstrawisc: @turetske a friendly ping on this

And now that I think about it more, will we _ever_ set up a sentinel file for a cache? Maybe this should only be for origins?

jhiemstrawisc (changes requested):
See comment about adding "absolute" to description of `XRootD.Configfile`.

docs/parameters.yaml
haoming29: I think the ticket was to add sentinel file to the origin, but it won't hurt to do the same thing for the cache. I will open a ticket for this.

haoming29: > @turetske, I wanted to get your feedback here. Should we immediately split this out into something like `Origin.XrootdConfigFile` and `Cache.XrootdConfigFile` so that we can do custom stuff for fed in a box, or do you not anticipate this being an issue?

I would argue that this is a config we intended to hide from general admins, and there's not much need to put it at origin/cache level.

jhiemstrawisc (approved):
LGTM

PR #969, opened 2024-03-20
Parse cache hit/miss into Prometheus metric

### LINKED ISSUE #944: Parse cache hit rate/times from xrootd monitoring packets into Prometheus metric

haoming29:
Closes #944 

Reading from cache data from summary packet, specifically `rd.hits` and `rd.miss`, I got 0 for both values even if requested multiple files through my cache-origin federation (with `EnableFallbackRead` set to false). Skimming through xrootd source code, it seems that the two values are only updated for `rmc` type cache (read? memory cache), but I only got `pfc` type cache in the monitoring packets.

@bbockelm Is this value the one we want to parse? Or it's something else?

### DIFF
--- a/metrics/xrootd_metrics.go
+++ b/metrics/xrootd_metrics.go
@@ -27,6 +27,7 @@ import (
 	"math"
 	"net"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


haoming29:
Close this PR as the implementation is not what we want. Will open a separate PR once that's done.

PR #972, opened 2024-03-21
Import parameters.json into website

### LINKED ISSUE #971: Missing parameters.json file in containers

CannonLock:
[Unify Build Process](https://github.com/PelicanPlatform/pelican/pull/972/commits/9a2bda35ea9c99799ce2ea98fbd3539d0f8b9763) 

Adds check that go generate has been run on the latest commit
Moves make web-build into goreleaser before which will make how we build dev binaries more similar to how we build production binaries
Rewrites the dockerfile moving the web build into the gorelease section
Doesn't needlessly update parameters.json on go generate ./..

Closes #971

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,6 +24,7 @@ before:
   hooks:
     - go mod tidy
     - go generate ./...
+    - make web-build
 builds:
   - env:
       - CGO_ENABLED=0
--- a/.github/workflows/go-generate-check.yml
+++ b/.github/workflows/go-generate-check.yml
@@ -1,0 +1,27 @@
+name: Check Go Generate Has been Run
+on:
+  pull_request:
+jobs:
+  validate-generated-files:
+    name: Check Go Generate
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.21"
+
+      - name: Run Go Generate
+        run: go generate ./...
+
+      - name: Validate Generated Files are up to date
+        run: |
+          if git diff --quiet; then
+            echo "No changes found."
+          else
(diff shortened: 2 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -25,6 +25,7 @@ repos:
     -   id: trailing-whitespace
         exclude: github_scripts/pelican_protocol.patch
     -   id: end-of-file-fixer
+        exclude: ".*/parameters.json"
```
haoming29, line 28: Instead of doing this, I'd prefer we add the new line at the end of the generated `parameter.json`

.goreleaser.yml
```diff
@@ -24,6 +24,7 @@ before:
   hooks:
     - go mod tidy
     - go generate ./...
+    - make web-build
```
haoming29, line 27: This would build the frontend every time we run `goreleaser ...` which can slow down the development quite a bit. Can we conditionally run this hook based on env or command line argument?

web_ui/frontend/public/data/parameters.json
haoming29, line 1: Is there any chance that we can copy/paste this from `docs/parameters.json` when running `make web-build` and exclude this file from git? I'm a bit worried about the duplicated file here

haoming29 (changes requested):
Some questions/concerns left in comments

.goreleaser.yml
CannonLock: Addressed this concern by only updating parameter.json file if necessary. This way `make web-build` only runs when there are web changes.

web_ui/frontend/public/data/parameters.json
CannonLock: Ended up just deleting the parameters.json files and adding there generation to goreleaser.

haoming29 (approved):
LGTM!

PR #973, opened 2024-03-21
Allow configurable cache watermarks

### LINKED ISSUE #665: Cache: allow configuring disk usage low/high water marks
Allow configuring the disk usage high water mark/low water mark since the appropriate values vary depending on the speed of the storage medium. The high water mark is the amount of disk used before the cache starts deleting files; the low water mark is the amount of disk used below which the cache stops deleting files.

This is the directive `pfc.diskusage <LOW> <HIGH>` listed in the [xrootd proxy configuration doc](https://xrootd.slac.stanford.edu/doc/dev56/pss_config.htm); LOW and HIGH are either a fraction of 1.0, or an absolute value suffixed with `k`, `g`, `m`, or `t`. The default value is
```
pfc.diskusage 0.90 0.95
```

haoming29:
Closes #665

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -857,26 +857,26 @@ type: int
 default: 8442
 components: ["cache"]
 ---
-name: Cache.HighWaterMark
-description: >-
-  A value of cache disk usage that triggers the purging of cached files.
-
-  The value should be either a percentage number (in string) of total avaibale disk space (default is 95),
-  or a number suffixed by k, m, g, or t. In which case, they must be absolute sizes in k (kilo-),
-  m (mega-), g (giga-), or t (tera-) bytes, respectively.
-type: string
-default: "95"
-components: ["cache"]
----
 name: Cache.LowWatermark
 description: >-
   A value of cache disk usage that stops the purging of cached files.
 
-  The value should be either a percentage number (in string) of total avaibale disk space (default is 90),
+  The value should be either a decimal fraction of a percentage number of total avaibale disk space (default is 0.90),
   or a number suffixed by k, m, g, or t. In which case, they must be absolute sizes in k (kilo-),
   m (mega-), g (giga-), or t (tera-) bytes, respectively.
 type: string
-default: "90"
+default: 0.90
+components: ["cache"]
+---
+name: Cache.HighWaterMark
+description: >-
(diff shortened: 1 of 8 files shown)


bbockelm (changes requested):
I like the implementation but have questions about the interface.

- Does it match what we did for local_cache?
- Is it better/more friendly to have percentages (90%) instead of floats (0.90).  Providing percentages may be more user friendly as the number provided is unambiguous.

May be a good idea to have a quick chat with someone from the facilitation team to get their feedback on the subject.

haoming29:
> I like the implementation but have questions about the interface.
> 
> * Does it match what we did for local_cache?
> * Is it better/more friendly to have percentages (90%) instead of floats (0.90).  Providing percentages may be more user friendly as the number provided is unambiguous.
> 
> May be a good idea to have a quick chat with someone from the facilitation team to get their feedback on the subject.

Agree. Percentage (integer) should be more user friendly and have better consistency with `local cache`

docs/parameters.yaml
```diff
@@ -857,6 +857,28 @@ type: int
 default: 8442
 components: ["cache"]
 ---
+name: Cache.LowWatermark
+description: >-
+  A value of cache disk usage that stops the purging of cached files.
+
+  The value should be either a percentage integer of total avaibale disk space (default is 90),
```
turetske, line 864:
```suggestion
  The value should be either a percentage integer of total available disk space (default is 90),
```

docs/parameters.yaml
```diff
@@ -857,6 +857,28 @@ type: int
...
+  m (mega-), g (giga-), or t (tera-) bytes, respectively.
+type: string
+default: 90
+components: ["cache"]
+---
+name: Cache.HighWaterMark
+description: >-
+  A value of cache disk usage that triggers the purging of cached files.
+
+  The value should be either a percentage integer of total avaibale disk space (default is 95),
```
turetske, line 875:
```suggestion
  The value should be either a percentage integer of total available disk space (default is 95),
```

turetske (changes requested):

turetske (approved):
Looks good to me.

haoming29:
@bbockelm GitHub requires your approval before merging. Feel free to taking another look.

PR #977, opened 2024-03-22
Fix bug where project name not set in user-agent

joereuss12:
Project name, if set, will now show up in the user-agent header for requests. This PR also removes some old unused legacy code of the "payload" object (the only piece of the object still used was the project name). Also includes a unit test checking that the client properly sets this header.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -135,7 +135,6 @@ type (
 		upload     bool
 		packOption string
 		attempts   []transferAttemptDetails
-		accounting payloadStruct
 		project    string
 		err        error
 	}
@@ -736,7 +735,7 @@ func (te *TransferEngine) runJobHandler() error {
 //
 // The returned object can be further customized as desired.
 // This function does not "submit" the job for execution.
-func (tc *TransferClient) NewTransferJob(remoteUrl *url.URL, localPath string, upload bool, recursive bool, options ...TransferOption) (tj *TransferJob, err error) {
+func (tc *TransferClient) NewTransferJob(remoteUrl *url.URL, localPath string, upload bool, recursive bool, project string, options ...TransferOption) (tj *TransferJob, err error) {
 
 	id, err := uuid.NewV7()
 	if err != nil {
@@ -755,6 +754,7 @@ func (tc *TransferClient) NewTransferJob(remoteUrl *url.URL, localPath string, u
 		upload:        upload,
 		uuid:          id,
 		token:         tc.token,
+		project:       project,
 	}
 	tj.ctx, tj.cancel = context.WithCancel(tc.ctx)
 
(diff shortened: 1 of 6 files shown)


haoming29:
Running the code locally, I still get `grab` as the `appinfo` field in xrootd monitoring packet, but the client-side tests look good to me

```
DEBUG[2024-03-25T14:48:58Z] MonPacket: Received an appinfo packet        
DEBUG[2024-03-25T14:48:58Z] GetSIDRest inputs: http/unknown.1:32828963699019@6ee28c5df997
grab 
```

joereuss12:
> Running the code locally, I still get `grab` as the `appinfo` field in xrootd monitoring packet, but the client-side tests look good to me
> 
> ```
> DEBUG[2024-03-25T14:48:58Z] MonPacket: Received an appinfo packet        
> DEBUG[2024-03-25T14:48:58Z] GetSIDRest inputs: http/unknown.1:32828963699019@6ee28c5df997
> grab 
> ```

Are there multiple of those appinfo fields returned? When I run it I get quite a few returned to me with some being what is expected. I can play around with this more tho if that is not the case

joereuss12:
Yeah, I don't think it will show up unless the project name is set correctly. You can set an env variable called `_CONDOR_JOB_AD` to a file or having a .job.ad file that contains `ProjectName = <project name>`. However, I think we should have something show if it is not set, at least with the Pelican version no matter what.

joereuss12:
@haoming29 Give this a go now. I was seeing grab too locally but with this one line change it seems to be giving me the correct User-Agent now.

local_cache/local_cache.go
```diff
@@ -506,7 +506,7 @@ func (sc *LocalCache) runMux() error {
 
 			sourceURL := *sc.directorURL
 			sourceURL.Path = path.Join(sourceURL.Path, path.Clean(req.request.path))
-			tj, err := sc.tc.NewTransferJob(&sourceURL, localPath, false, false, client.WithToken(req.request.token))
+			tj, err := sc.tc.NewTransferJob(&sourceURL, localPath, false, false, "", client.WithToken(req.request.token))
```
haoming29, line 509: It would be a nice idea here to say the "project" is the localcache

haoming29 (approved):
The revised version gave me `pelican-client/7.6.1-next project/some project name` as the user agent header read by the xrootd, which should be what we expect. Give the green light here. For the comment on the localcache, feel free address it or not.

Thanks for fixing this!

client/handle_http.go
```diff
@@ -1472,8 +1472,8 @@ func downloadHTTP(ctx context.Context, te *TransferEngine, callback TransferCall
 	// Set the headers
 	req.HTTPRequest.Header.Set("X-Transfer-Status", "true")
 	req.HTTPRequest.Header.Set("TE", "trailers")
-	if payload != nil && payload.ProjectName != "" {
-		req.HTTPRequest.Header.Set("User-Agent", getUserAgent(payload.ProjectName))
+	if project != "" {
+		req.HTTPRequest.Header.Set("User-Agent", getUserAgent(project))
```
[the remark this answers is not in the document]
joereuss12: For sure! I think that's a good idea.

local_cache/local_cache.go
joereuss12: Can do!

joereuss12:
Alright, made those suggestions as well as allowed the ProjectName to be surrounded by "" or not and still work. Just going to let the tests run just in case before I merge.

PR #978, opened 2024-03-22
Authenticated self-tests for cache

### LINKED ISSUE #879: Authenticated self-tests for cache
In #861 , we implemented self-tests for cache, but it exercises the tests under a public export path. Let's upgrade this test to an authed test once #816 is merged in and we have a local issuer for cache token verification.

haoming29:
Closes #879 

To make the authed cache work, I need to add a local issuer for the cache for token authentication. The cache issuer only has path for `/pelican/monitoring`


turetske (approved):
LGTM!

PR #987, opened 2024-03-25
Periodic cleanup of origin test files

### LINKED ISSUE #975: Periodically remove origin self-tests files in zombie state
As ITB team reported, the temp files for origin self-tests piled up very quickly if there's an error getting or deleting the test file. Let's make sure there is no temp files that are in zombie state and never get cleaned up.

haoming29:
Closes #975


origin/origin_api.go
```diff
@@ -57,3 +62,54 @@ func ConfigureXrootdMonitoringDir() error {
...
+			selfPrevFile := ""
+			if err != nil {
+				return err
+			}
+			for _, entry := range entries {
+				if !entry.IsDir() {
+					fn := entry.Name()
+					if strings.HasPrefix(fn, "director") {
+						// Always leave the first item
+						if dirPrevFile != "" {
```
turetske, line 89: How do we know the first file is the most recent/the one that we want to keep for the test files?

turetske (changes requested):
Just a question I want to understand better.

origin/origin_api.go
haoming29: Ahh I'm sorry. What I meant was to leave the "last" item in the directory. This is because the `os.ReadDir` sorts the files, and for all the test files, we have a timestamp at the end of the file, which is string-comparable, and we leave the last item as the latest test file, so that there's won't be a race condition where the just-arrived test file was deleted.

haoming29: I'll revise the comment to be more self-explanatory.

PR #992, opened 2024-03-26
More verbose error message at director registration failure

### LINKED ISSUE #976: More verbose response error messages when director registration failed
When an origin or cache fail to register at the director (which it does for every 15s) due to token issues, the only error message showing up at the origin/cache end is 

`XRootD server advertise failed: Error during director registration: Authorization token verification failed.`

which doesn't help a lot for origin/cache admins to debug and report the issue. We should give more verbose message back in the response without leaking any credentials.

haoming29:
Closes #976 

Update: while debugging, I figured that there is still space for improvement for the error message, will mark it ready for review once that's addressed.


turetske (approved):
LGTM!

PR #1000, opened 2024-03-27
Fix director default response and add HEAD as allowed verb

### LINKED ISSUE #754: `HEAD` doesn't work in director against the origins endpoint
We often use `HEAD` to determine if a given file exists.

However, none of our APIs wire up this particular verb. 
 [Example](https://github.com/PelicanPlatform/pelican/blob/main/director/redirect.go#L619-L620)

We should add support for HEAD.  Make sure to do this for the "shortcut middleware" as well that handles the default redirects.

For now, this can just reply with a redirect to the appropriate origin.

jhiemstrawisc:
Short and sweet, this should allow us to issue HEAD requests against the API the `/api/v1.0/director/<origin or object>` endpoints. The tagged issue also mentioned making this compatible with our shortcut middleware in the director, but it looks like it already was -- the middleware was agnostic of verbs that weren't `PUT`. My addition to the middleware here explicitly ties `GET/HEAD` requests to the appropriate redirect functions so that other verbs we don't anticipate don't creep in.

One thing to note is that if a client who issues a `HEAD` for some resource chooses to follow the redirect, XRootD appears to hang with an open connection. Since that's in XRootD land, I'm considering it outside the scope of this small fix.

Closes #754

### DIFF
--- a/launchers/launcher.go
+++ b/launchers/launcher.go
@@ -110,7 +110,7 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
 
 	if modules.IsEnabled(config.DirectorType) {
 
-		viper.Set("Director.DefaultResponse", "cache")
+		viper.Set("Director.DefaultResponse", param.Director_DefaultResponse.GetString())
 
 		viper.Set("Federation.DirectorURL", param.Server_ExternalWebUrl.GetString())
 
--- a/launchers/launcher.go
+++ b/launchers/launcher.go
@@ -109,9 +109,7 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
 	}
 
 	if modules.IsEnabled(config.DirectorType) {
-
-		viper.Set("Director.DefaultResponse", param.Director_DefaultResponse.GetString())
-
+		viper.SetDefault("Director.DefaultResponse", "cache")
 		viper.Set("Federation.DirectorURL", param.Server_ExternalWebUrl.GetString())
 
 		if err = DirectorServe(ctx, engine, egrp); err != nil {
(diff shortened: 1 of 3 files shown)


haoming29:
@jhiemstrawisc Congrats on opening the #1000 PR! ⭐

launchers/launcher.go
```diff
@@ -110,7 +110,7 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (context.Canc
 
 	if modules.IsEnabled(config.DirectorType) {
 
-		viper.Set("Director.DefaultResponse", "cache")
+		viper.Set("Director.DefaultResponse", param.Director_DefaultResponse.GetString())
```
haoming29, line 113: What's the intension here to set the value by the value itself?

haoming29 (changes requested):
One question about viper.Set but otherwise looks good to me!

launchers/launcher.go
jhiemstrawisc: Good point 😆 I noticed that because we had `viper.Set("Director.DefaultResponse", "cache")`, changing the default response in my config had no effect. What I didn't stop to think about was that I could just delete this line all together.

jhiemstrawisc: Actually, on second thought... I should probably set this as a default with viper. I don't know that everything that reaches this code will necessarily ingest our `config/resources/defaults.yaml`, especially various tests.

haoming29: I think tests should be fine, the ones that could be affected by this should be director-specific tests and e2e tests, which should either set this value explicitly, or call InitConfig to read `default.yaml`.

jhiemstrawisc: I made the change before I saw this comment. If you think I should remove the line entirely, I can do that too.

haoming29 (approved):
LGTM!

PR #1002, opened 2024-03-27
Pelican client advertises timeout to cache

joereuss12:
The client now sets a X-Pelican-Timeout header for download requests which should get passed along to the cache


bbockelm (approved):
LGTM.

This does not set things for uploads or stats which, I believe, makes sense.

joereuss12:
> LGTM.
> 
> This does not set things for uploads or stats which, I believe, makes sense.

yeah I figured since neither of those really talk to a cache and go straight to the origin. Going to go ahead and merge.

PR #1003, opened 2024-03-27
Add the option for a sentinel file for the origin

### LINKED ISSUE #995: Add support for the XRootD sentinel file
@biozit reports one useful feature from the legacy OSDF is the ability to have a "sentinel file".

The sentinel file is a file that must exist in order for the service/container to startup; if missing, then the service should shutdown with a recognizable error.  This is useful in cases where the admin may forget to mount the filesystem inside the origin (leading to a confusing set of responses due to an empty directory) or if there's a broken filesystem used by the cache

Fabio writes:

> This option checks the existence of a specific file on a XrootD volume; if the file is not present, the XrootD won't start.
> It was tested using this option:
>    `oss.space public $(rootdir) chkmount thisfileshoulbeheretoxrootdststart -`
> https://xrootd.slac.stanford.edu/doc/dev54/ofs_config.htm#_Toc89982406.

I believe we could implement either at the XRootD or the Pelican layer.  Doing this in Pelican may provide the ability to do a higher-quality error message.

In terms of configuration -- I would suggest a sentinel could be used for each of the origin's exports.

haoming29:
Closes #995

### DIFF
--- a/server_utils/origin.go
+++ b/server_utils/origin.go
@@ -20,6 +20,8 @@ package server_utils
 
 import (
 	"fmt"
+	"os"
+	"path"
 	"path/filepath"
 	"reflect"
 	"strings"
@@ -33,13 +35,14 @@ import (
 	"github.com/pelicanplatform/pelican/server_structs"
 )
 
-var originExports *[]OriginExports
+var originExports *[]OriginExport
 
 type (
-	OriginExports struct {
+	OriginExport struct {
 		StoragePrefix    string
 		FederationPrefix string
 		Capabilities     server_structs.Capabilities
+		SentinelFile     string
 	}
 )
 
@@ -115,12 +118,12 @@ func StringListToCapsHookFunc() mapstructure.DecodeHookFuncType {
 // convert those values (such as Origin.FederationPrefix, Origin.StoragePrefix, etc.) into the OriginExports
 // struct and return a list of one. Otherwise, we'll base things off the list of exports and ignore the single-prefix
 // style of configuration.
-func GetOriginExports() (*[]OriginExports, error) {
+func GetOriginExports() (*[]OriginExport, error) {
 	if originExports != nil {
 		return originExports, nil
 	}
 
-	originExports = &[]OriginExports{}
+	originExports = &[]OriginExport{}
 
 	viper.SetDefault("Origin.StorageType", "posix")
 	StorageType := param.Origin_StorageType.GetString()
@@ -152,7 +155,7 @@ func GetOriginExports() (*[]OriginExports, error) {
 				}
 
 				reads := param.Origin_EnableReads.GetBool() || param.Origin_EnablePublicReads.GetBool()
-				originExport := OriginExports{
+				originExport := OriginExport{
 					FederationPrefix: federationPrefix,
 					StoragePrefix:    storagePrefix,
 					Capabilities: server_structs.Capabilities{
@@ -175,7 +178,7 @@ func GetOriginExports() (*[]OriginExports, error) {
 
 			log.Warningln("Passing export volumes via -v at the command line causes Pelican to ignore exports configured via the yaml file")
 			log.Warningln("However, namespaces exported this way will inherit the Origin.Enable* settings from your configuration")
-			log.Warningln("For finer-grained control of each export, please configure them in your pelican.yaml file")
+			log.Warningln("For finer-grained control of each export, please configure them in your pelican.yaml file via Origin.Exports")
 			return originExports, nil
 		}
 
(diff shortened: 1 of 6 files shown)


haoming29:
This PR added an additional field `SentinelFile` to the `Origin.Exports` parameter. @CannonLock you might need to update the UI as well.

CannonLock:
@haoming29 I will patch on the ui change after the name is decided.

haoming29:
Trying to understand the intension for S3 here. Will the user provide a path to the sentinel file for either S3 and POSIX? Or only a file name? The latter is what I assumed in this PR and I actually checked and require that the parameter does not contain a directory.

jhiemstrawisc:
I think you'll have to think about a good way to incorporate the origin's configured storage type. In the case of an S3 sentinel object, we might require that for a given `s3endpoint`, `bucket` and `object`, we can fetch `<s3endpoint>/<bucket>/<object>` if the origin is using path-style requests or `<bucket>.<s3endpoint>/<object>` if we're using virtual hosting.

Of course, if you want to check that an authenticated S3 object exists _before_ starting xrootd, you'll also need to handle creating the signed AWSv4 URLs with the credentials Pelican knows about, and that's not a small task. You could either write all of the functionality to do this in Pelican (and wind up duplicating half the magic of the S3 plugin) or you could use the AWS SDK and add a lot of bloat...

Maybe the approach is to start XRootD first and immediately try to pull the file/object. If the origin can't get something from itself, we assume the sentinel doesn't exist.

haoming29:
@jhiemstrawisc That clarifies things a lot! I like your suggestion as to start xrootd first then pull a file and see if we can make it for S3 backend. For POSIX backend, I think it's easy enough to `stat` the file before starting the xrootd and that way we eliminate other possibilities that can cause the error getting the file through the origin (auth/token/etc). I will incorporate the comments above and update the code.

CannonLock:
Is the parameter naming finalized here yet?

haoming29:
> SentinelLocation

Let's do `SentinelLocation` for now. I'll revise this PR today but I think `SentinelLocation` sounds good to me

CannonLock:
@haoming29 In testing the UI I hit the following error that killed the web app. 

```
Fatal error occurred at the start of the program. Cleanup started: fail to open SentinelLocation test for StoragePrefix sdfsdf. Directory check failed: stat sdfsdf/test: no such file or directory
```

Notably I used garbage in the config values so not sure if this is really a "bug"

```
origin:
    exports:
        - capabilities:
            - PublicReads
            - Writes
            - Listings
            - DirectReads
            - Reads
          federationprefix: /testtimeusa
          storageprefix: /test0123/cannonlock-test/98y230498q2yt0qwu8gf89q3gt2
        - capabilities:
            - DirectReads
          federationprefix: sdfsdf
          sentinellocation: test
          storageprefix: sdfsdf
```

haoming29:
@CannonLock Do you have a file `test` under `/sdfsdf`?

CannonLock:
@haoming29 No, more so just bummed this took out the web app. Would be nice if we could keep the webapp running through this type of thing so they could continue to edit the config file/read the logs right in the webapp... Just a idea for the future.

haoming29:
> @haoming29 No, more so just bummed this took out the web app. Would be nice if we could keep the webapp running through this type of thing so they could continue to edit the config file/read the logs right in the webapp... Just a idea for the future.

Good point. This would be a corner case where we expect user to have a test file in place before running Pelican. If they decided to change the test file on the go, they should be prepared to have the new test file before making the change. It would be nice if we can have a tooltip/description to emphasize this

server_utils/origin.go
```diff
@@ -246,6 +249,23 @@ func GetOriginExports() (*[]OriginExports, error) {
 	return originExports, nil
 }
 
+func CheckSentinelLocation(exports *[]OriginExport) (ok bool, err error) {
+	for _, export := range *exports {
+		if export.SentinelLocation != "" {
+			sentinelPath := path.Clean(export.SentinelLocation)
+			if path.Base(sentinelPath) != sentinelPath {
+				return false, fmt.Errorf("invalid SentinelLocation path for StoragePrefix %s, file must not contain a directory. Got %s", export.StoragePrefix, export.SentinelLocation)
```
bbockelm, line 257:
```suggestion
				return false, errors.Errorf("invalid SentinelLocation path for StoragePrefix %s, file must not contain a directory. Got %s", export.StoragePrefix, export.SentinelLocation)
```

server_utils/origin.go
```diff
@@ -246,6 +249,23 @@ func GetOriginExports() (*[]OriginExports, error) {
...
+			}
+			fullPath := filepath.Join(export.StoragePrefix, sentinelPath)
+			_, err := os.Stat(fullPath)
+			if err != nil {
+				return false, errors.Wrap(err, fmt.Sprintf("fail to open SentinelLocation %s for StoragePrefix %s. Directory check failed", export.SentinelLocation, export.StoragePrefix))
```
bbockelm, line 262:
```suggestion
				return false, errors.Wrapf(err, "fail to open SentinelLocation %s for StoragePrefix %s. Directory check failed", export.SentinelLocation, export.StoragePrefix)
```

bbockelm (approved):
Two small nitpicks on coding style.  After that, I think this is pre-approved and feel free to merge!

PR #1007, opened 2024-03-28
Cache hit/miss bytes Prometheus metric

### LINKED ISSUE #944: Parse cache hit rate/times from xrootd monitoring packets into Prometheus metric

haoming29:
Closes #944 

This PR also fixes the issue where handling summary monitoring packet with cache reports error due to misformat xml packet
This PR fixes the bug where xrootd summary packet was not monitored if the binary is not `OSDF`

Although this PR collected byte hit/miss/bypass from xrootd g-stream monitoring packet, it appears to me that we only get byte hit number increasing while not for byte miss nor byte bypass. I tried locally to move a brand new file (that the cache doesn't have) and check the `.cinfo` file which should be the source of the g-stream, but the byte miss is 0 in that file. Same thing with the case where no file at the cache side and origin has `direct reads` turned on, the only number got updated in byte hit.

Will open a separate ticket for further investigating this issue.

### DIFF
--- a/metrics/xrootd_metrics.go
+++ b/metrics/xrootd_metrics.go
@@ -22,6 +22,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"encoding/json"
 	"encoding/xml"
 	"fmt"
 	"math"
@@ -87,10 +88,10 @@ type (
 	}
 
 	XrdXrootdMonHeader struct {
-		Code byte
-		Pseq byte
-		Plen uint16
-		Stod int32
+		Code byte   // = | d | f | g | i | p | r | t | u | x
+		Pseq byte   // packet sequence
+		Plen uint16 // packet length
+		Stod int32  // Unix time at Server start
 	}
 
 	XrdXrootdMonMap struct {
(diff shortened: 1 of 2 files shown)


metrics/xrootd_metrics.go
```diff
@@ -747,9 +787,53 @@ func HandlePacket(packet []byte) error {
...
+		providerID := (gs.SID >> XROOTD_MON_PIDSHFT) & XROOTD_MON_PIDMASK
+		detail := NullTermToString(packet[24:])
+		if providerID == 'C' { // pfc: Cache monitoring  info
+			log.Debug("HandlePacket: Received g-stream packet is from cache")
+			strJsons := strings.Split(detail, "\n")
+			aggCacheStat := make(map[string]*CacheAccessStat)
+			for _, js := range strJsons {
+				cacheStat := CacheGS{}
+				if err := json.Unmarshal([]byte(js), &cacheStat); err != nil {
+					return errors.Wrap(err, "fail to parse cache stat json. Raw data is "+string(js))
```
turetske, line 810:
```suggestion
					return errors.Wrap(err, "failed to parse cache stat json. Raw data is "+string(js))
```

turetske (changes requested):

turetske (approved):

PR #1011, opened 2024-03-29
Allow recursive uploads/downloads in plugin

joereuss12:
Recursive uploads/downloads of directories work with the pelican plugin now! Tested with condor, it seems downloads work great however some functionality is not there for uploads when doing a transfer_output_remap. Going ahead to push this for now since it seems to be working for us. To do this, run:

stash_plugin osdf:///<some/path/to/directory>?recursive=true .

I decided to hold off on a unit test since it is not much different from our current recursive uploads/downloads tests and this can be a test ran in the condor test infastructure in the future.

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -418,6 +418,12 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
 				workChan = nil
 				break
 			}
+			// Check we have valid query parameters
+			err := checkValidQuery(transfer.url)
+			if err != nil {
+				return err
+			}
+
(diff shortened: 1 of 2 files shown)


bbockelm (changes requested):
I'd really like a simple unit test here.  There's always something that could go wrong with URL parsing or the logic around destination directory that could go wrong.

jhiemstrawisc (approved):
LGTM. Thanks Joe!

jhiemstrawisc:
Since I reviewed Joe's changes and he provided unit tests for the pieces Brian asked about, I'm going to merge this in.

PR #1012, opened 2024-03-29
Add support for director/federation contact info as a configuration

### LINKED ISSUE #983: Ask director (federation) admin to provide a support contact and display it in the director UI

haoming29:
Closes #983 

@CannonLock this PR introduced a new object type parameter called `Director.SupportContact`. You may need to patch in additional UI form for this. Thanks!

### DIFF
--- a/docs/pages/serving_a_federation.mdx
+++ b/docs/pages/serving_a_federation.mdx
@@ -16,7 +16,7 @@ If you haven't installed Pelican, follow the instructions to [install pelican](/
 
 ### Prepare TLS Credentials
 
-Pelican servers use `https` for serving its web UI, handling file transfer requests and internal http requests. `https` requires a set of credential files in place to work. For details on how to set up the TLS credentials, refer to the [documentation in Serve an Origin](./serving_an_origin.mdx#prepare-tls-credentials).
+Pelican servers use `https` for serving its web UI, handling objects transfer requests and internal http requests. `https` requires a set of credential files in place to work. For details on how to set up the TLS credentials, refer to the [documentation in Serve an Origin](./serving_an_origin.mdx#prepare-tls-credentials).
 
 ## Serve a Registry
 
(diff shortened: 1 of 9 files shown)


CannonLock:
@haoming29 Can this have multiple values?

haoming29:
> @haoming29 Can this have multiple values? If we only want one contact we should not use the object field here and instead have string fields.

Good point. Yeah I think at the end of the day it's simply two string config params (just happen to have the shared parent node). I'll redo the params and I'll let you know when it's ready.

haoming29:
@CannonLock Just redid the config and we should be good with string values.

docs/pages/serving_a_federation.mdx
```diff
@@ -140,14 +140,28 @@ Then pass the file location to `Director.MaxMindKeyFile`.
...
+By default, Pelican director hosts the federation discovery endpoint. Therefore, for origins and caches in your federation, as well as Pelican clients that want to access files in your federation, their `Federation.DiscoveryUrl` value can be set to the director's public URL, i.e. `https://<your-director-hostname>:<director-web-port>` or `https://<your-director-hostname>` if the port number is simply `443`.
+
+Make sure your director is up and running before starting your origins and caches.
+
+### Set Support Contact Information (Recommended)
+> `Director.SupportContact` is only available for Pelican >=7.7.0
+
+Since the director is the entrypoint of your federation, and client users uses director's public URL as the `Federation.DiscoveryUrl`, it is convenient for users to look up information related to a federation or seek help at the director's web UI. Therefore, we recommend director admin to setup contact information at the director by `Director.SupportContactEmail` and `Director.SupportContactUrl` configuration parameters. If set, the information will be displayed in the director's web UI. See [Configuration](./parameters.mdx#Director-SupportContact) page for details.
```
jhiemstrawisc, line 156: The director is your federation's entrypoint, and it is where other services in the federation can learn about each other. Because of this, it is also used to host contact information for federation administrators. Setting values for the configuration options `Director.SupportContactEmail` and `Director.SupportContactUrl` causes the director to display the information through its web UI so that users know who/how to contact you for support. See ... for more details.

docs/parameters.yaml
```diff
@@ -1012,6 +1012,23 @@ type: stringSlice
 default: none
 components: ["director"]
 ---
+name: Director.SupportContactEmail
+description: >-
+  An Email address to receive issues and help request for the federation the director is hosting. The values will
```
jhiemstrawisc, line 1017: `and help requests`

jhiemstrawisc (approved):

PR #1013, opened 2024-03-29
Add "login required" endpoints behind "admin" check and fix bug in admin auth handler

### LINKED ISSUE #1004: Users with role "user" can view config
Pelican Version: Latest

Pelican Service:
- [X ] Registry

**Describe the bug**

If I login through CIlogon and gain the role of a user I can view the config. 

```{"authenticated":true,"role":"user","user":"http://cilogon.org/serverA/users/46022246"}```

**To Reproduce**

Go to the registry page: https://osdf-registry.osg-htc.org/view/registry/

Sign in via CIlogon with a non admin account and view the config.

**Expected behavior**

I expect the config to return a 403 which will redirect the user.

haoming29:
Fixes #1004 

This PR also fix a bug in admin auth handler where if the user is not a admin, the handler didn't abort the request, causing the response to pass.

This is a critical bug with security concern and we should consider backport to 7.6 and 7.5.

Improved testing to cover both cases.

### DIFF
--- a/web_ui/authentication.go
+++ b/web_ui/authentication.go
@@ -212,14 +212,15 @@ func AdminAuthHandler(ctx *gin.Context) {
 	user := ctx.GetString("User")
 	// This should be done by a regular auth handler from the upstream, but we check here just in case
 	if user == "" {
-		ctx.JSON(http.StatusUnauthorized, gin.H{"error": "Login required to view this page"})
+		ctx.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "Login required to view this page"})
+		return
 	}
 	isAdmin, msg := CheckAdmin(user)
 	if isAdmin {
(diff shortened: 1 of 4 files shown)


jhiemstrawisc:
Is there any hand test I should run for this to make sure things are working as expected?

haoming29:
> Is there any hand test I should run for this to make sure things are working as expected?

Try `/api/v1.0/config` without logging in, logging as a non-admin user (I recommend run the test against registry), both of which should give you 401/403 error (instead of 200 for now).

Another endpoint to check is the `/api/v1.0/registry_ui/namespace/100` with a `DELETE` verb at the registry, and you should get the same result as above

jhiemstrawisc (approved):
LGTM and we ran through the hand tests together. Thanks!

PR #1022, opened 2024-04-01
Pin external c++ dependencies at a specific commit

### LINKED ISSUE #881: We should avoid building xrootd-s3-http from main
Right now, both our prod and dev containers build xrootd-s3-http from its main branches. That means any breaking changes in the plugin will break Pelican until Pelican can be updated to handle the changes.

We should instead work out a tagging scheme for the plugin and build based on tags. This allows us to develop xrootd-s3-http more freely.

haoming29:
Closes #881 

This PR pins commit when building from source in dev/production containers for s3 plugin, xrdcl plugin, and lotman


jhiemstrawisc (approved):
LGTM

PR #1028, opened 2024-04-02
Create serving_a_cache.mdx

turetske:
Added preliminary cache documentation

### DIFF
--- a/docs/pages/serving_a_cache.mdx
+++ b/docs/pages/serving_a_cache.mdx
@@ -73,7 +73,9 @@ Once you have your cache set up, follow the steps below to test if your cache ca
 1. Have data available and accessible via an origin. Since a cache only works with data available via an origin that’s in the same Pelican federation, the following assumes that you have a an accessible namespace within the same Pelican federation. For more information on how to set that up, refer to [Serving an Origin](./serving_an_origin.mdx). The following also assume that the object to test is public file. To set up a public file through a Pelican origin, refer to the [Configuration documentation](./parameters.mdx#Origin-Exports).
 2. Curl the director for the test file and ensure that the url used is the cache that you set up. Assuming your directory is `/tmp/demo`, run the following command to curl your known public file from the cache.
 
-```$ curl -v https://<cache-url>/<namespace/path-to-public-file>```
+```
+$ curl -v https://<cache-url>/<namespace/path-to-public-file>
+```
 
 Where:
 * `<cache-url>` is the of your cache
--- a/docs/pages/serving_a_cache.mdx
+++ b/docs/pages/serving_a_cache.mdx
@@ -1,6 +1,6 @@
 # Serve a Pelican Cache
 
-A [Pelican](http://pelicanplatform.org/) *Cache* creates a cache that connects to a Pelican data federation to allow faster data access. It stores data that is accessed via a pelican origin in a cache until it gets evicted. The cache has the potential to both be faster and physically closer to your location than your origin. This allows for faster data access when running complex workflows.
+A [Pelican](http://pelicanplatform.org/) *Cache* creates a cache that connects to a Pelican data federation to allow faster data access. It stores data that is accessed via a Pelican origin until it gets evicted. The cache has the potential to both be faster and physically closer to your location than the Pelican origin you access the data from. This allows for faster data access when running complex workflows.
 
 This document contains instructions on how to serve a Pelican cache.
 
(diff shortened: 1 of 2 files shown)


haoming29:
Could you fix the linter issue?

docs/pages/serving_a_cache.mdx
haoming29, line 1: Could you add `"serving_a_cache":"Serving a Cache",`  to `_meta.json` after the key `serving_an_origin` ? This way we can keep the docs in our order

docs/pages/serving_a_cache.mdx
```diff
@@ -0,0 +1,107 @@
+# Serve a Pelican Cache
+
+A [Pelican](http://pelicanplatform.org/) *Cache* creates a cache that connects to a Pelican data federation to allow faster data access. It stores data that is accessed via a pelican origin in a cache until it gets evicted. The cache has the potential to both be faster and physically closer to your location than your origin. This allows for faster data access when running complex workflows.
```
haoming29, line 3:
```suggestion
A [Pelican](http://pelicanplatform.org/) *Cache* creates a cache that connects to a Pelican data federation to allow faster data access. It stores data that is accessed via a Pelican origin until it gets evicted. The cache has the potential to both be faster and physically closer to your location than the Pelican origin you access the data from. This allows for faster data access when running complex workflows.
```

docs/pages/serving_a_cache.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+### Open Firewall Port for Pelican Cache
+
+The Pelican cache listens to two TCP ports for file transfers and Web UI. By default, the file transfer port is at `8442` and the Web UI and APIs port is at `8444`. If your server has firewall policy in place, please open the two ports for both incoming the outgoing TCP requests to allow the Pelican cache to function as expected.
+
+You may change the port numbers through the [configuration file](./parameters.mdx) with parameter [`Cache.Port`](./parameters.mdx#Cache-Port) and [`Server.WebPort`](./parameters.mdx#Server-WebPort) respectively.
+
+> If it is not possible for you to expose any port through the firewall, Pelican has a special feature called _Connection Broker_, where it allows you to serve a Pelican cache without a public-accessible port and any TLS credential files in place. However, this is an experimental feature and requires the Pelican federation you are joining to be compatible. If you are interested in learning more about _Connection Broker_, please contact help@pelicanplatform.org for further instructions.
```
haoming29, line 26: IIRC, connection broker only works for the origin, not for the cache

docs/pages/serving_a_cache.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+### Prepare TLS Credentials
```
haoming29, line 28: Can you point this to the related origin documentation instead? so that we don't have duplicated pieces. Like we did here: https://docs.pelicanplatform.org/serving_a_federation#prepare-tls-credentials

docs/pages/serving_a_cache.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+> If you are running Pelican docker image to serve a cache, please refer to the [docker image documentation](./install/docker.mdx#run-pelican-cache-server).
+
+### Find a federation to join
+
+Before serving a cache, you need to find a Pelican federation to join in. If you are unfamiliar with the term **federation**, refer to [Useful Terminology](./client-usage.mdx#useful-terminology) before proceeding.
+
+If you don't have a federation in mind, the Open Science Data Federation (OSDF) is an example Pelican federation that you can join in for testing purposes. If you are interesting in serving an OSDF cache, refer to the [OSDF website](https://osg-htc.org/) for details.
+
+The federation discovery URL for OSDF is `osg-htc.org`. You may use this as your `<federation>` argument in the next section when launching your origin.
```
haoming29, line 66:
```suggestion
The federation discovery URL for OSDF is `osg-htc.org`. You may use this as your `<federation>` argument in the next section when launching your cache.
```

docs/pages/serving_a_cache.mdx
```diff
@@ -0,0 +1,107 @@
...
+* **-h or --help**: Output documentation on the `serve` command and its arguments.
+* **-p or --port**: Set the port at which the Pelican admin website should be accessible.
+
+* **--config**: Set the location of the configuration file.
+* **-d or --debug**: Enable the debugging mode, allowing for more verbose log
+* **-l or --log**: Set the location of the file where log messages should be redirected to and not output the the console.
+
+There are other configurations available to modify via the configuration file. Refer to the [Parameters page](./parameters.mdx) for details.
+
+## Test CacheFunctionality
```
haoming29, line 95:
```suggestion
## Test Cache Functionality
```

docs/pages/serving_a_cache.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+Once you have your cache set up, follow the steps below to test if your cache can access a file through a Pelican federation. 
+
+1. Have data available that is accessible via an origin. Since a cache only works with data available via an origin that’s in the same data federation, the following assumes that you have a an accessible namespace within the same data federation. For more information on how to set that up, refer to [Serving an Origin](./serving_an_origin.mdx). Ensure there is some public file available.
```
haoming29, line 99:
```suggestion
1. Have data available and accessible via an origin. Since a cache only works with data available via an origin that’s in the same Pelican federation, the following assumes that you have a an accessible namespace within the same Pelican federation. For more information on how to set that up, refer to [Serving an Origin](./serving_an_origin.mdx). The following also assume that the object to test is public file. To set up a public file through a Pelican origin, refer to the [Configuration documentation](./parameters.mdx#Origin-Exports).
```

docs/pages/serving_a_cache.mdx
```diff
@@ -0,0 +1,107 @@
...
+2. Curl the director for the test file and ensure that the url used is the cache that you set up. Assuming your directory is `/tmp/demo`, run the following command to curl your known public file from the cache.

+```
$ curl -v https://<cache-url>/<namespace/path-to-public-file>
```

Where:
```
Comment (haoming29) [inline docs/pages/serving_a_cache.mdx:101]: You want to have a new line before and after ``` so that the code block can be separate from the text

File: docs/pages/serving_a_cache.mdx [hunk 8]
```diff
@@ -0,0 +1,107 @@
...
+1. Have data available that is accessible via an origin. Since a cache only works with data available via an origin that’s in the same data federation, the following assumes that you have a an accessible namespace within the same data federation. For more information on how to set that up, refer to [Serving an Origin](./serving_an_origin.mdx). Ensure there is some public file available.
+2. Curl the director for the test file and ensure that the url used is the cache that you set up. Assuming your directory is `/tmp/demo`, run the following command to curl your known public file from the cache.

+```
$ curl -v https://<cache-url>/<namespace/path-to-public-file>
```

Where: 
+* `<cache-url>` is the of your cache
+* `<namespace/path-to-public-file> is the namespace and the path of the known public file
```
Comment (haoming29) [inline docs/pages/serving_a_cache.mdx:103]:
```suggestion
* `<namespace/path-to-public-file>` is the namespace and the path of the known public file  
```

docs/pages/serving_a_cache.mdx
```diff
@@ -0,0 +1,107 @@
...


+Check that the curl output contains the data from the public file.
```
haoming29, line 104:
```suggestion
* Check that the curl output contains the data from the public file.
```

docs/pages/serving_a_cache.mdx
haoming29, line 99: If the testing uses OSDF as an example, can we point the user to the public file we are using for the client doc? i.e. `/ospool/uc-shared/public/OSG-Staff/validation/test.txt`

I'd curious if a non-osg cache can serve an osg public file

haoming29 (changes requested):

docs/pages/serving_a_cache.mdx
turetske: It shouldn't. It won't serve an origin that isn't known to the director. So they must be part of the same federation. Which is why I didn't specify the OSDF public file, because these instructions should be generic for all federations, not just the OSDF.

docs/pages/serving_a_cache.mdx
```diff
@@ -92,16 +66,19 @@ This section documents the additional arguments you can pass to the command abov
...

-```
$ curl -v https://<cache-url>/<namespace/path-to-public-file>
```

Where: 
+1. Have data available and accessible via an origin. Since a cache only works with data available via an origin that’s in the same Pelican federation, the following assumes that you have a an accessible namespace within the same Pelican federation. For more information on how to set that up, refer to [Serving an Origin](./serving_an_origin.mdx). The following also assume that the object to test is public file. To set up a public file through a Pelican origin, refer to the [Configuration documentation](./parameters.mdx#Origin-Exports).
+2. Curl the director for the test file and ensure that the url used is the cache that you set up. Assuming your directory is `/tmp/demo`, run the following command to curl your known public file from the cache.
+
+```$ curl -v https://<cache-url>/<namespace/path-to-public-file>```
```
Comment (haoming29) [inline docs/pages/serving_a_cache.mdx:76]: There should be a new line between ``` and the command (and after the command) to make this a code block instead of inline. My markdown renderer still render this into a inline code.
It's like this (desired):
```
$ curl
```

VS this (current):
```$ curl ```

haoming29 (changes requested):
One last thing

haoming29 (approved):
LGTM!

haoming29:
The parameter validation error was fixed by Cannon in #1036 so we should be good to merge

PR #1032, opened 2024-04-02
Changes to dir listings in Pelican client

joereuss12:
This removes the need for namespaces to advertise "dirlisthost" since we can get that information with a PROPFIND request to the origin through a director redirect.

### DIFF
--- a/client/director.go
+++ b/client/director.go
@@ -52,7 +52,6 @@ func CreateNsFromDirectorResp(dirResp *http.Response) (namespace namespaces.Name
 	namespace.Path = xPelicanNamespace["namespace"]
 	namespace.UseTokenOnRead, _ = strconv.ParseBool(xPelicanNamespace["require-token"])
 	namespace.ReadHTTPS, _ = strconv.ParseBool(xPelicanNamespace["readhttps"])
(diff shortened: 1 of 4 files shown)


bbockelm (commented):
Two things to consider as you work on this:

1.  Compatibility issues: What happens if you have a new client and an old director?  At first blush, it appears things would break.  You'll need to fall back gracefully to the old mechanism for finding the host.
2. Scale issues.  How can you make sure you only do the lookup of the host once per session.  Before, the director's namespace was queried once and then the interaction was with the origin.  Now, the director is queried for each stat and each directory listing.

bbockelm:
@joereuss12 - what's the motivation around removing the webdav client and doing our own response parsing?

joereuss12:
> @joereuss12 - what's the motivation around removing the webdav client and doing our own response parsing?

At first it was slightly unintentional but I am not entirely sure how to combine the webdav client with PROPFIND since PROPFIND will return the contents of the directory we request so it seems to be redundant with the webdav client. However, looking more into the source code for gowebdav, it seems that their "client.readdir" is already doing some sort of PROPFIND so I may need to rethink this approach.

joereuss12:
Closing because #1111 solves this issue and is a better approach

Issue #1033, opened 2024-04-02
v7.6.2 Registry repeatedly fails to start, until it doesn't


williamnswanson:
Pelican Version:
v.7.6.2

Pelican Service:
- [ ] Client
- [ ] Plugin
- [X] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

**Describe the bug**

>/tiger-osg-config$ kubectl -n osg logs osdf-registry-749c579558-9zmw8
>Starting Pelican...
>Running osdf with arguments: registry serve
>Logging.LogLocation is set to /var/log/pelican/osdf-registry.log. All logs are redirected to the log file.
>2024/04/02 19:42:08 goose: no migrations to run. current version: 20240212194951
>Error: Received bad status code in reply to server ping: 404. Expected 200,

On starting the v7.6.2 osdf-registry container, the container fails, but kubernetes re-starts it over and over. When this happened in the itb registry it only happened a few times and then settled down and served properly. It happened again when updating the prod registry, but while it settled down once, after accidentally restarting the container it continued to fail.

For the ITB osdf pelican regsitry, the last line about bad status codes did not appear on the successful run.

**To Reproduce**
Start/restart a v.7.6.2 registry container in kubernetes?

**Expected behavior**
The container should either always fail to start or always succeed, without needing to fail several times to serve properly.

**Desktop (please complete the following information):**
 - osdf-registry:v7.6.2 container

williamnswanson:
@brianhlin The ticket for the v7.6.2 registry issues we were seeing.

PR #1034, opened 2024-04-02
Add server urls to error messages for `WaitUntilWorking`

haoming29:
As @williamnswanson noticed in #1033, at a failure of `WaitUntilWorking` test, we don't know the server URL that Pelican attempts to check. This PR add server URL to various error/log messages. Come with tests.


joereuss12 (approved):
LGTM

PR #1036, opened 2024-04-03
Remove Testing Code

CannonLock:
Added a 'cache' string [here](https://github.com/PelicanPlatform/pelican/pull/889/files#diff-e64181cce5ce490155a33cdffaa48b7eea8d1f07e4c7f0aef3ae447f720e7641) so that I could view the cache only `Lotman` configuration. 

This is to remove that leaked testing code.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -874,6 +874,7 @@ description: >-
   number of connections to this value.
 type: int
 default: none
+components: ["cache"]
 ---
 name: Cache.EnableLotman
 description: >-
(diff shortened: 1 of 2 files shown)


haoming29 (approved):
LGTM!

haoming29:
The parameter validation test failed due to no key exists for `Cache.Concurrency` which is an int type, not sure why this would cause error?

CannonLock:
@haoming29 Looks like the `Cache.Concurrency` component list got dropped somewhere? I just tacked it on here as these are both minor changes.

haoming29:
> @haoming29 Looks like the `Cache.Concurrency` component list got dropped somewhere? I just tacked it on here as these are both minor changes.

Thanks for the patch!

PR #1038, opened 2024-04-03
Refactor registry namespace check to be compatible with OSDF topology

### LINKED ISSUE #1029: Topology Namespaces don't include their public key information
Pelican Service:
- [x ] Registry

The pelican registry cannot get public keys for namespaces registered with Topology. When getNamespaceJwksByPrefix is run, it will return a "namespace with prefix %q not fount in database" error because it doesn't check topology. The reason it doesn't check topology is because topology doesn't contain any public key information.

This needs to be fixed in some way. The information in Topology has auth and non-auth endpoints, but the public key location is not clear. There are two options for a solution depending on whether the public key information for topology namespaces is available and it uses jwks for authentication or if there is a different way to do so.

If a public key is available:

   1. When the director queries topology for its information and creates OriginAds, make sure the public key information is included.
   2. Alter the Topology table to include the public key information
   3. Alter the getNamespaceJwksByPrefix function to check topology for the prefix before checking the namespace table

If a different authentication/authorization method is used:
    1. Store that authentication method in the the originAds
    2. Alter the topology table to include the needed information
    3. Anything that uses the getNamespaceJwksByPrefix needs to check if the namespace is from Topology and, if so, implement the other auth schema for verification with topology namespaces

haoming29:
Closes #1029 

## Other fixes
* Added a `Registry` health component to the UI to should if registering namespace succeeded or not
* Better status tracking/updates of the health component overall
* Better error messages and plumbing

The updated policy is as follows (only for OSDF):

Given that a namespace exists in the topology, say `/foo`:
* If an origin server has prefix `/foo` or `/foo/**/**`, and the admin starts the server without registering the namespace at registry website, the auto-registration in the origin will _fail_, with an error message to ask them register the namespace at the registry web UI:
```
The namespace <namespace> already exists in the OSDF topology. To register a Pelican equivalence, you need to present your identity. If you are registering through Pelican CLI, try again with the flag '--with-identity' enabled. If this is an auto-registration from a Pelican origin or cache server, register your namespace or server through the Pelican registry website at <registry url> instead.
```
* If the admin attempts to register a namespace `/foo` or `/foo/**/**` via Pelican CLI (`pelican namespace register`) but without the `--with-identity` flag, the registration will also _fail_, with the same error message as above
* If the admin register the namespace /foo` or `/foo/**/**` via Pelican CLI  with `--with-identity` flag OR via the registry web UI. The registration will succeed. However, there will be a note in the registration `Description` field as `[ Attention: Prefix exists in OSDF topology ]` to warn the registry admin to pay extra attention when reviewing it.

This PR also removes topology check from `checkNamespaceExists` and function alike to maintain consistency.

**Pitfall**: This should fix all the incompatibility between Pelican namespace and OSDF legacy namespaces. However, one pitfall is that the director also fetches namespace from both OSDF and Pelican servers, and there can be two situations:
* The Pelican origin server that serves the OSDF namespace prefix has the **exactly the same** origin server advertisement as we parsed from the topology. This way, the two server advertisement will race against each other every time we update topology/origin server advertises itself to the director. The consequence is unknown
* The Pelican origin server has different fields in the origin advertisement, and we record both in the director TTL cache. Then we essentially have two servers serving the same namespace, which shouldn't be an issue, just make things a bit intertwined.

-----

## How to test

Here's the instruction to test the fix e2e:

* Configuration file (Note the `FederationPrefix`, `DbLocation`, and `Loggin.Level`

```
Origin:
  Exports:
    - StoragePrefix: /tmp/stash
      FederationPrefix: /chtc/itb/<your mock name>
      Capabilities: ["Reads", "PublicReads", "DirectReads"]
Federation:
  DiscoveryUrl: "https://<your container hostname>:8444"
  TopologyNamespaceUrl: "https://topology.opensciencegrid.org/stashcache/namespaces.json"
TLSSkipVerify: true
Logging:
  Level: "Error"
Registry:
  InstitutionsUrl: "https://topology.opensciencegrid.org/institution_ids"
  DbLocation: "<new location>/<new file>.sqlite"
```

* Build Pelican web UI by running `make web-build` in your Pelican root folder
* Run Pelican binary with `osdf` alias.
  * First, run your federation in a box with the director and the registry
  ```
   ./osdf serve --module director,registry
  ```
   * Second, run your origin. Note that you need to change the origin web port to not have a conflict
   ```
   OSDF_SERVER_WEBPORT=<YOUR PORT> ./osdf origin serve
   ```
    * You will see an error message (non-blocking though) from your origin terminal, something like:
    ```
    ERROR[2024-04-04T18:31:17Z] Failed to register with namespace service: Failed to register prefix /chtc/itb/hmtest2: failed to register the namespace: You don't have permission to register the prefix: The namespace /chtc/itb/hmtest2 already exists in the OSDF topology. To register a Pelican equivalence, you need to present your identity. If you are registering through Pelican CLI, try again with the flag '--with-identity' enabled. If this is an auto-registration from a Pelican origin or cache server, register your namespace or server through the Pelican registry website at https://6ee28c5df997:8444 instead.: The POST attempt to https://6ee28c5df997:8444/api/v1.0/registry resulted in status code 403; will automatically retry in 10 seconds 
    ```
  * As the error suggests, you will need to go to the registry web UI to register your namespace first
  * If you go to the origin's web UI, the same error message will appear at the status section, under the `Registry` component
* Before you close your origin (you don't have to though). Go to the public key endpoint and download your origin's public key at `/.well-known/issuer.jwks`
* Go to your registry web UI, should be at `https://localhost:8444`, and register a new namespace with your origin's federation prefix and the public key
* Switch to your origin terminal, wait for a minute and it should not complain about the error anymore
* Go to your origin web UI, the error in the Registry component should be gone, and the `Director` component should report OK (green color)

If you want to take one step further, you can test if a file can be transferred through this origin and the federation (I tested locally and it works for me)

### DIFF
--- a/registry/registry_ui.go
+++ b/registry/registry_ui.go
@@ -423,7 +423,6 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
 	}
 
 	ns := Namespace{}
-	ns.Topology = false
 	if ctx.ShouldBindJSON(&ns) != nil {
 		ctx.JSON(http.StatusBadRequest, gin.H{"error": "Invalid create or update namespace request"})
 		return
@@ -443,7 +442,6 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
 		return
 	}
 	ns.Prefix = updated_prefix
-	ns.Topology = false
 
 	if !isUpdate {
 		// Check if prefix exists before doing anything else. Skip check if it's update operation
--- a/registry/registry_ui.go
+++ b/registry/registry_ui.go
@@ -495,6 +495,9 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
 		ns.AdminMetadata.UserID = user
 		// Overwrite status to Pending to filter malicious request
 		ns.AdminMetadata.Status = Pending
+		if inTopo {
+			ns.AdminMetadata.Description = "[ Attention: Prefix exists in OSDF topology ] " + ns.AdminMetadata.Description
+		}
 		if err := AddNamespace(&ns); err != nil {
 			log.Errorf("Failed to insert namespace with id %d. %v", ns.ID, err)
 			ctx.JSON(http.StatusInternalServerError, gin.H{"error": "Fail to insert namespace"})
@@ -502,7 +505,7 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
 		}
 		msg := "success"
 		if inTopo {
-			msg = "Prefix is successfully registered. Note that there is an existing namespace prefix in the OSDF topology. The register admin will review your request and approve your namespace if this is expected."
+			msg = "Prefix is successfully registered. Note that there is an existing namespace prefix in the OSDF topology. The registry admin will review your request and approve your namespace if this is expected."
 		}
 		ctx.JSON(http.StatusOK, gin.H{"msg": msg})
 	} else { // Update
--- a/registry/registry_ui.go
+++ b/registry/registry_ui.go
@@ -464,7 +464,7 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
 	}
 
 	// Check if the parent or child path along the prefix has been registered
-	inTopo, valErr, sysErr := validateKeyChaining(ns.Prefix, pubkey)
+	inTopo, topoNss, valErr, sysErr := validateKeyChaining(ns.Prefix, pubkey)
 	if valErr != nil {
 		log.Errorln("Bad prefix when validating key chaining", valErr)
 		ctx.JSON(http.StatusBadRequest, gin.H{"error": valErr.Error()})
@@ -496,7 +496,7 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
 		// Overwrite status to Pending to filter malicious request
 		ns.AdminMetadata.Status = Pending
 		if inTopo {
-			ns.AdminMetadata.Description = "[ Attention: Prefix exists in OSDF topology ] " + ns.AdminMetadata.Description
+			ns.AdminMetadata.Description = fmt.Sprintf("[ Attention: A superspace or subspace of this prefix exists in OSDF topology: %s ] ", GetTopoPrefixString(topoNss))
 		}
 		if err := AddNamespace(&ns); err != nil {
 			log.Errorf("Failed to insert namespace with id %d. %v", ns.ID, err)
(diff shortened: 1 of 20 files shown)


haoming29:
@jhiemstrawisc this one should be good to go. See PR description for how to test e2e locally.

jhiemstrawisc:
One comment about the error I get from serving the origin:
```
The namespace /chtc/itb/jhiemstra already exists in the OSDF topology
```
I think this is somewhat misleading, because that full namespace prefix _isn't_ registered in topology. Is there a way to indicate specifically that in order to register a sub namespace of `/chtc/itb` (or maybe it's just `/chtc`, whichever is actually registered) the user has to follow the provided steps?

haoming29:
> One comment about the error I get from serving the origin:
> 
> ```
> The namespace /chtc/itb/jhiemstra already exists in the OSDF topology
> ```
> 
> I think this is somewhat misleading, because that full namespace prefix _isn't_ registered in topology. Is there a way to indicate specifically that in order to register a sub namespace of `/chtc/itb` (or maybe it's just `/chtc`, whichever is actually registered) the user has to follow the provided steps?

I agree the wording is kind of misleading, I can work on feeding the exact topology namespace that's taken but I think a quicker fix would be rewording the error message and say "part of the namespace ...." instead? I can do either way.

registry/registry_ui.go
```diff
@@ -497,12 +495,19 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
 		ns.AdminMetadata.UserID = user
 		// Overwrite status to Pending to filter malicious request
 		ns.AdminMetadata.Status = Pending
+		if inTopo {
+			ns.AdminMetadata.Description = "[ Attention: Prefix exists in OSDF topology ] " + ns.AdminMetadata.Description
```
jhiemstrawisc, line 499: And here as well -- we should be explicit about whether the namespace exists or whether some super-space of the attempted registration exists

registry/registry_validation.go
```diff
@@ -69,25 +69,24 @@ func validatePrefix(nspath string) (string, error) {
...
-	if !param.Registry_RequireKeyChaining.GetBool() {
-		return
-	}
+func validateKeyChaining(prefix string, pubkey jwk.Key) (inTopo bool, validationError error, serverError error) {
 	// We don't check keyChaining for caches
 	if strings.HasPrefix(prefix, "/caches/") {
 		return
 	}
+	// Here, we do the check anyway but only returns error (if any)
+	// we the flag is on. This is to make sure the topology check is independent
```
jhiemstrawisc, line 78: we --> when

jhiemstrawisc (changes requested):
A few cleanup items, but I'm convinced that this should solve our issue by giving federation admins a way to bypass the previous restrictions we had in place.

jhiemstrawisc (approved):
LGTM!

jhiemstrawisc:
@turetske, I believe we need to patch this into 7.6 before EoD

PR #1040, opened 2024-04-03
More error tolerant web ui check

### LINKED ISSUE #1033: v7.6.2 Registry repeatedly fails to start, until it doesn't

haoming29:
Should fix #1033 

The suspicion is that there's a racing condition at the registry start somehow, and the web engine is not "ready-ready" to serve requests, although it could be other issues. However, this PR will make the web ui check more robust in case of the error code returned

### DIFF
--- a/server_utils/server_utils.go
+++ b/server_utils/server_utils.go
@@ -41,7 +41,7 @@ import (
 
 // Wait until given `reqUrl` returns a HTTP 200.
 // Logging messages emitted will refer to `server` (e.g., origin, cache, director)
-// Pass true to statusMismatch to allow a mismatch of expected status code and what's returned not fail immediatelys
+// Pass true to statusMismatch to allow a mismatch of expected status code and what's returned not fail immediately
 func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expectedStatus int, statusMismatch bool) error {
 	expiry := time.Now().Add(10 * time.Second)
 	ctx, cancel := context.WithDeadline(ctx, expiry)
@@ -82,7 +82,7 @@ func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expect
 					statusError = errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Can't read response body with error %v.", reqUrl, resp.StatusCode, expectedStatus, err)
 					if statusMismatch {
 						if !statusErrLogged {
-							log.Info(statusError, "Will retry until timeedout")
+							log.Info(statusError, "Will retry until timeout")
 							statusErrLogged = true
 						}
 					} else {
@@ -94,7 +94,7 @@ func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expect
 						statusError = errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Response body: %s", reqUrl, resp.StatusCode, expectedStatus, string(bytes))
 						if statusMismatch {
 							if !statusErrLogged {
-								log.Info(statusError, "Will retry until timeedout")
+								log.Info(statusError, "Will retry until timeout")
 								statusErrLogged = true
 							}
 						} else {
@@ -105,7 +105,7 @@ func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expect
 						statusError = errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Response body is empty.", reqUrl, resp.StatusCode, expectedStatus)
 						if statusMismatch {
 							if !statusErrLogged {
-								log.Info(statusError, "Will retry until timeedout")
+								log.Info(statusError, "Will retry until timeout")
 								statusErrLogged = true
 							}
 						} else {
--- a/server_utils/server_utils.go
+++ b/server_utils/server_utils.go
@@ -41,13 +41,17 @@ import (
 
 // Wait until given `reqUrl` returns a HTTP 200.
 // Logging messages emitted will refer to `server` (e.g., origin, cache, director)
-func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expectedStatus int) error {
+// Pass true to statusMismatch to allow a mismatch of expected status code and what's returned not fail immediatelys
+func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expectedStatus int, statusMismatch bool) error {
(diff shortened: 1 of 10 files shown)


bbockelm (changes requested):
I don't understand what this code is trying to do.

`WaitUntilWorking` tries a URL every 50ms until it succeeds or 10 seconds has passed.  It seems the code adds another set of loops to retry on top of retries (which appears strange?) ... no?

haoming29:
> I don't understand what this code is trying to do.
> 
> `WaitUntilWorking` tries a URL every 50ms until it succeeds or 10 seconds has passed. It seems the code adds another set of loops to retry on top of retries (which appears strange?) ... no?

My guessing of William hitting that 404 error is that the gin server wasn't fully started before we hit the health check endpoint. Since 404 instead of 200 was returned, `WaitUntilWorking` returns immediately and we never get a chance to check it a bit later. I don't want to chance how `WaitUntilWorking` works because in most situations it makes sense to fail immediately if the status code is different from expected, but for the registry's case, it would probably worth waiting and checking again if the status code doesn't match

bbockelm:
Gotcha.  I would rather add another option to WaitUntilWorking to make the wrong response code non-fatal.  Layers of retries are a bit scary.

haoming29:
> Gotcha. I would rather add another option to WaitUntilWorking to make the wrong response code non-fatal. Layers of retries are a bit scary.

Sounds good. I'll refactor the code

jhiemstrawisc:
@haoming29 is this ready for review or should I wait a bit longer?

haoming29:
> @haoming29 is this ready for review or should I wait a bit longer?

Sorry I haven't get the chance to work on the refactoring. I will let you know when its' review for another round of review

haoming29:
@jhiemstrawisc Should be good for the review. The Ubuntu test failure should be a random one and re-running it will likely fix the issue

bbockelm (approved):
LGTM

server_utils/server_utils.go
```diff
@@ -75,24 +79,55 @@ func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expect
 				}
 				bytes, err := io.ReadAll(resp.Body)
 				if err != nil {
-					// We didn't get the expected status
-					return errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Can't read response body with error %v.", reqUrl, resp.StatusCode, expectedStatus, err)
+					statusError = errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Can't read response body with error %v.", reqUrl, resp.StatusCode, expectedStatus, err)
+					if statusMismatch {
+						if !statusErrLogged {
+							log.Info(statusError, "Will retry until timeedout")
```
jhiemstrawisc, line 85: timeedout --> timeout

server_utils/server_utils.go
```diff
@@ -41,13 +41,17 @@ import (
 
 // Wait until given `reqUrl` returns a HTTP 200.
 // Logging messages emitted will refer to `server` (e.g., origin, cache, director)
-func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expectedStatus int) error {
+// Pass true to statusMismatch to allow a mismatch of expected status code and what's returned not fail immediatelys
```
jhiemstrawisc, line 44: immediatelys --> immediately

server_utils/server_utils.go
```diff
@@ -75,24 +79,55 @@ func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expect
...
+						return statusError
+					}
 				} else {
 					if len(bytes) != 0 {
-						// We didn't get the expected status
-						return errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Response body: %s", reqUrl, resp.StatusCode, expectedStatus, string(bytes))
+						statusError = errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Response body: %s", reqUrl, resp.StatusCode, expectedStatus, string(bytes))
+						if statusMismatch {
+							if !statusErrLogged {
+								log.Info(statusError, "Will retry until timeedout")
```
jhiemstrawisc, line 97: Same here

server_utils/server_utils.go
```diff
@@ -75,24 +79,55 @@ func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expect
...
+						} else {
+							// We didn't get the expected status
+							return statusError
+						}
 					} else {
-						return errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Response body is empty.", reqUrl, resp.StatusCode, expectedStatus)
+						statusError = errors.Errorf("Received bad status code in reply to server ping at %s: %d. Expected %d. Response body is empty.", reqUrl, resp.StatusCode, expectedStatus)
+						if statusMismatch {
+							if !statusErrLogged {
+								log.Info(statusError, "Will retry until timeedout")
```
jhiemstrawisc, line 108: Here too

jhiemstrawisc (changes requested):
A few small typos, otherwise LGTM

haoming29:
> A few small typos, otherwise LGTM

Thanks for catching them!

jhiemstrawisc (approved):
LGTM. Thanks!

PR #1041, opened 2024-04-03
Fix error responses for local cache

bbockelm:
This PR does two primary things:

1. Allows the local cache to respond with a 404 when upstream does.
2. Allows the local cache to timeout before the plugin does, giving the client a 504 (Gateway timeout).

Along the way, two improvements:
1. The `TransferJob` object now takes a context and can be canceled.
2. You can retrieve the PID of the running xrootd process(es).  This is used to be able to SIGSTOP the xrootd daemon to make it unresponsive in unit tests.


joereuss12 (approved):
Any conflicts should be resolved now and everything looks good to me!

PR #1042, opened 2024-04-03
Update Origin UI to use API for the new `Origin.Exports` scheme

### LINKED ISSUE #1035: Update Data Export table in origin UI
In #927, we did a large refactoring of origin export configurations with changes to the parameters name, this results in a `NULL` value for the existing Data Export table:

Screenshot 2024-04-03 at 8 47 53 AM

We should update the UI to fix the null value.

Update: given the complexity of how we set up the `origin.exports` and `origin.enable*`/ `origin.exportvolume`, I'd suggest we have a dedicated API call to return the exported namespaces. I'll be working the API part first and see if I can tackle on the UI as well to off load Cannon's packed list of todos

haoming29:
Closes #1035 

Since @jhiemstrawisc has a refactoring PR to support multiple S3 exports, which may add/change the scheme of `Origin.Exports`. Mark this PR as draft until Justin's PR gets merged in.


CannonLock (approved):
LGTM. Ignored the resolved comment below, I patched that and forgot to delete the request.

PR #1043, opened 2024-04-03
Removed the setting of an AuthURL for Pelican serverAds

turetske:
Because osdf origins and caches still have separate AuthURLs, this PR doesn't remove them fully from ServerAds. Rather, if it is a Pelican ServerAd, the AuthURL isn't set.

The director has been changed so that if an AuthURL isn't set, it will return the regular url instead when an AuthURL is needed and an AuthURL otherwise.

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -42,18 +42,20 @@ func parseServerAd(server utils.Server, serverType server_structs.ServerType) se
 	// but endpoints do not have a scheme. As such, we need to add one for the.
 	// correct parsing. Luckily, we don't use this anywhere else (it's just to
 	// make the url.Parse function behave as expected)
-	if !strings.HasPrefix(server.AuthEndpoint, "http") { // just in case there's already an http(s) tacked in front
-		server.AuthEndpoint = "https://" + server.AuthEndpoint
-	}
 	if !strings.HasPrefix(server.Endpoint, "http") { // just in case there's already an http(s) tacked in front
 		server.Endpoint = "http://" + server.Endpoint
 	}
-	serverAuthUrl, err := url.Parse(server.AuthEndpoint)
-	if err != nil {
-		log.Warningf("Namespace JSON returned server %s with invalid authenticated URL %s",
-			server.Resource, server.AuthEndpoint)
+	if server.AuthEndpoint != "" {
+		if !strings.HasPrefix(server.AuthEndpoint, "http") { // just in case there's already an http(s) tacked in front
+			server.AuthEndpoint = "https://" + server.AuthEndpoint
+		}
+		serverAuthUrl, err := url.Parse(server.AuthEndpoint)
+		if err != nil {
+			log.Warningf("Namespace JSON returned server %s with invalid authenticated URL %s",
+				server.Resource, server.AuthEndpoint)
+		}
+		serverAd.AuthURL = *serverAuthUrl
 	}
-	serverAd.AuthURL = *serverAuthUrl
 
 	serverUrl, err := url.Parse(server.Endpoint)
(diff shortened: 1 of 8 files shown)


director/advertise.go
```diff
@@ -42,18 +42,20 @@ func parseServerAd(server utils.Server, serverType server_structs.ServerType) se
...
-	}
 	if !strings.HasPrefix(server.Endpoint, "http") { // just in case there's already an http(s) tacked in front
 		server.Endpoint = "http://" + server.Endpoint
 	}
-	serverAuthUrl, err := url.Parse(server.AuthEndpoint)
-	if err != nil {
-		log.Warningf("Namespace JSON returned server %s with invalid authenticated URL %s",
-			server.Resource, server.AuthEndpoint)
+	if server.AuthEndpoint != "" {
+		if !strings.HasPrefix(server.AuthEndpoint, "http") { // just in case there's already an http(s) tacked in front
```
haoming29, line 49: While we are fixing this, we should parse the url first, then set the `Scheme` field to `https` directly

director/director.go
```diff
@@ -91,6 +91,9 @@ func getRedirectURL(reqPath string, ad server_structs.ServerAd, requiresAuth boo
 	var serverURL url.URL
 	if requiresAuth {
```
haoming29, line 92:
```suggestion
	if requiresAuth && ad.AuthURL.String() != "" {
```

haoming29 (approved):
One comment on the style, but more a personal preference, feel free ignore it or fix it

director/advertise.go
```diff
@@ -42,27 +41,25 @@ func parseServerAd(server utils.Server, serverType server_structs.ServerType) se
...
 	// correct parsing. Luckily, we don't use this anywhere else (it's just to
 	// make the url.Parse function behave as expected)
-	if !strings.HasPrefix(server.Endpoint, "http") { // just in case there's already an http(s) tacked in front
-		server.Endpoint = "http://" + server.Endpoint
+	serverUrl, err := url.Parse(server.Endpoint)
+	if err != nil {
+		log.Warningf("Namespace JSON returned server %s with invalid unauthenticated URL %s",
+			server.Resource, server.Endpoint)
 	}
+	serverUrl.Scheme = "http"
```
haoming29, line 49: Looking at the comment around 40-43, the logic here is to make sure `url.Parse` work properly. However, my code snippet suggests that `url.Parse` works without the source string having a scheme: https://goplay.tools/snippet/0JAxY3I5hnt

Could you confirm this and remove the comment if so? Also I think we should set `https` instead of `http` (not 100% sure, so we probably want to check with Justin to confirm if the topology servers support https)

haoming29 (changes requested):
One more minor change

director/advertise.go
jhiemstrawisc: Unfortunately, in general the non-auth URL for origins/caches from topology are only set up to work with `http`. For example, the origin listed in topology with this endpoint `unl-origin.nationalresearchplatform.org:1094` gets a `403` when you curl with `http`, but gets a `refused connection` when you try to curl with `https`. 

This is all due to something I don't fully understand from the way things work in the legacy OSDF deployments. In general, though, all Pelican servers should be on HTTPS. Do we have a way to determine whether a particular origin/cache is Pelican or legacy? That might be really useful in the serverAd if not

haoming29: Since setting `http` for the `serverUrl` is what we had before, let's simply remove the old comment in line 40-43 and instead note what Justin reported here and we should be good for now

haoming29 (approved):
LGTM!

PR #1044, opened 2024-04-03
Sets Origin.Port to fix "port already in use" error that occurs occas…

turetske:
…ionally in S3 Origin tests


bbockelm:
Did the wrong branch get force pushed here?  It doesn't seem like the code changes match the PR title.

turetske:
> Did the wrong branch get force pushed here? It doesn't seem like the code changes match the PR title.

This was two issues. One, after realizing it was a more complex issue than the port change and that broke everything more, we decided to remove the virtual mode test (since it wasn't really serving a purpose right now) and I didn't update the description AND it looks like a rebase off of main went funky. I'll repair it today.

PR #1057, opened 2024-04-05
Config fixups

CannonLock:
Moved over the multilines to literal formatting rather then folding as Markdown expects literal. 

https://yaml-multiline.info/


haoming29 (approved):
LGTM once the linter error was fixed. Thanks for tidying this up! It would be useful for the record if you can briefly explain in the PR why we switch from `>-` to `|+`

CannonLock:
@haoming29 Good call, check new description.

PR #1058, opened 2024-04-05
Add topology timeout

turetske:


haoming29 (approved):
LGTM! Although we don't have a test for this improvement yet, we should use the `GetTransport` anyway and additional tests can be a separate ticket

PR #1063, opened 2024-04-06
Correctly return the transfer attempt information

bbockelm:
v7.6.x has a regression where the `downloadObject` function returns an error whenever an error occurs; rather, it's only supposed to return an error if no transfers are attempted.  That change causes the multiple attempts to not be returned to the transfer plugin -- instead, only the last one is.

This change:
- Returns all the failed attempts
- Fixes the "15s since start" in error messages by recording when the error occurred.
- Stubs out any paths in the plugin error message (since HTCSS also reports the path).
- Switches the client to golang time objects where appropriate, converting to integers (and floats) only at the plugin level.

### DIFF
--- a/client/errorAccum.go
+++ b/client/errorAccum.go
@@ -59,11 +59,15 @@ func NewTransferErrors() *TransferErrors {
 }
 
 func (te *TransferErrors) AddError(err error) {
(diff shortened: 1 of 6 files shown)


joereuss12 (approved):
LGTM!

PR #1065, opened 2024-04-08
Bump the revision of xrdcl-pelican for libcurl testing

bbockelm:
In unit tests, we've been seeing timeouts in xrdcl-pelican due to what appears overly-long delays for DNS lookups.  I believe this is due to the DNS lookup not correctly being waited on for the curl multi-wait.  This bumps the dev image to the curl testing code; if successful, we may do a similar thing for prod.


bbockelm:
Going to self-merge this given there is little risk in updating the ITB container code.

PR #1070, opened 2024-04-08
Move default S3 url style from defaults.yaml to config.go

jhiemstrawisc:
Because we have no way to tell whether defaults in defaults.yaml are user-provided or not, the default there of `Origin.S3UrlStyle` was throwing an error meant to notify users that they have incompatible config variables set.

This moves setting of that default into config.go where we now use `viper.SetDefault`, which later on allows us to `viper.IsSet("Origin.S3UrlStyle")` without catching our own default.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -983,6 +983,10 @@ func InitServer(ctx context.Context, currentServers ServerType) error {
 	viper.SetDefault("Cache.ExportLocation", "/")
 	viper.SetDefault("Registry.RequireKeyChaining", true)
 
+	// Set up the default S3 URL style to be path-style here as opposed to in the defaults.yaml becase
+	// we want to be able to check if this is user-provided (which we can't do for defaults.yaml)
+	viper.SetDefault("S3UrlStyle", "path")
+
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -983,6 +983,10 @@ func InitServer(ctx context.Context, currentServers ServerType) error {
 	viper.SetDefault("Cache.ExportLocation", "/")
 	viper.SetDefault("Registry.RequireKeyChaining", true)
 
+	// Set up the default S3 URL style to be path-style here as opposed to in the defaults.yaml becase
+	// we want to be able to check if this is user-provided (which we can't do for defaults.yaml)
+	viper.SetDefault("S3UrlStyle", "path")
```
haoming29, line 988:
```suggestion
	viper.SetDefault("Origin.S3UrlStyle", "path")
```

haoming29 (changes requested):
One naming typo otherwise LGTM

jhiemstrawisc:
Sorry for the typo, should be good to go now.

haoming29 (approved):
LGTM! Thanks for the fix!

PR #1071, opened 2024-04-08
Fix broken URLs of namespaces from the topology at the director

haoming29:
The URLs for the topology namespace in director are broken. The current code gives `http:8000` as `url` or `authUrl` for namespaces from the topology (try `/api/v1.0/director_ui/severs`). The original code https://github.com/PelicanPlatform/pelican/commit/10fd09f8b803019098f5bdbaf10c91b39957f866 was correct in that `url.Parse` requires the scheme to be present, which is contrary to what I tested in the playground: https://github.com/PelicanPlatform/pelican/pull/1043#discussion_r1553705373

But since `url.Parse` documented that parsing a url without scheme is invalid, let's keep doing what we were doing to ensure nothing breaks.

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -21,6 +21,7 @@ package director
 import (
 	"context"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


turetske (approved):
LGTM!

Issue #1075, opened 2024-04-08
Overhaul production docker container push and tagging


haoming29:
Currently, we push the production docker container with `:latest` tag if there's a new change in the main branch. This causes trouble as we don't want people using container to pull from our "dev" branch. Instead, we should only push when there's a new release cut, and we should mark the latest released version as `latest` for the container.

PR #1077, opened 2024-04-08
Only push the production container when the release was cut

haoming29:
This should fix the critical bug described in #1075 but leave the part untouched where "automatically tagging the latest release with the largest semantic version as the latest"

### DIFF
--- a/.github/workflows/build-prod-container.yml
+++ b/.github/workflows/build-prod-container.yml
@@ -11,10 +11,10 @@ on:
   workflow_dispatch:
   workflow_call:
     inputs:
-      from_workflow_call:
+      release_workflow:
         required: false
         type: boolean
-        default: true
+        default: false
       tag:
           description: 'Tag to use for the Docker image'
           required: false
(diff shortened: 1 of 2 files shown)


brianhlin (commented):
Is there a way to be even more specific here? I imagine that /any/ workflow call (even those that aren't for release) would trigger this `if`.

Also, is there a world in which you'd want `workflow_dispatch` to trigger a push?

haoming29:
> Is there a way to be even more specific here? I imagine that /any/ workflow call (even those that aren't for release) would trigger this `if`.
> 
> Also, is there a world in which you'd want `workflow_dispatch` to trigger a push?

Good call. I pass a variable from the `prelease` action to the build yaml to be sure it's called from the release

.github/workflows/build-prod-container.yml
```diff
@@ -192,7 +194,7 @@ jobs:
         with:
           context: .
           file: ./images/Dockerfile
-          push: ${{ github.repository == 'PelicanPlatform/pelican' && github.event_name != 'pull_request' }}
+          push: ${{ github.repository == 'PelicanPlatform/pelican' && (inputs.release_workflow || false) }}
```
brianhlin, line 197: What's with the `|| false` here? Kinda seems superfluous?

brianhlin (approved):
One question but otherwise LGTM. Pre-approving.

.github/workflows/build-prod-container.yml
haoming29: it does look superfluous but otherwise the yaml check will fail when the workflow was not triggered by the workflow call (in which case `release_workflow` is undefined) and gives the error:

```
Error: Input does not meet YAML 1.2 "Core Schema" specification: push
Support boolean input list: `true | True | TRUE | false | False | FALSE`
```

PR #1085, opened 2024-04-09
Set the largest semantic version tag to the latest image

### LINKED ISSUE #1079: Mark the production docker image with the latest semantic version tag as the `latest` image
As a follow up on #1077, we want to mark the image with the latest semantic version tag as the `latest` image instead of marking whatever was pushed most recent as the latest one.

haoming29:
Closes #1079 

This one should be merged in before subsequent releases after `7.7.0` to make sure our `latest` tag works as expected

### DIFF
--- a/.github/workflows/build-prod-container.yml
+++ b/.github/workflows/build-prod-container.yml
@@ -133,9 +133,9 @@ jobs:
           highest_tag=$(echo "${tags}" | tail -n1)
           echo "Highest version tag is ${highest_tag}"
           if [[ "${GITHUB_REF##*/}" == "${highest_tag}" ]]; then
-            echo "::set-output name=IS_LATEST::true"
+            echo "IS_LATEST=true" >> $GITHUB_OUTPUT
           else
-            echo "::set-output name=IS_LATEST::false"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-prod-container.yml
```diff
@@ -141,7 +155,7 @@ jobs:
             then
               GITHUB_TAG=${GITHUB_REF##*/}
             else
-              GITHUB_TAG="latest"
+              GITHUB_TAG="latest-dev"
```
haoming29, line 158: If there's no tags on the current commit, we set it to `latest-dev` instead of `latest` to avoid confusion; however, since we only push images if they are part of a release, I don't expect this to be present in the registry at all.

CannonLock:
@haoming29 This is not a blocker for 7.7 correct?

haoming29:
Nope, the only downside of not having this is that our `latest` tag won't be automatically updated. It will stay at `v7.6.5`

CannonLock:
@haoming29 Okay, I won't have time to review till next week. Until then we will have to manually tag latest.

brianhlin (approved):
LGTM

PR #1087, opened 2024-04-10
Turn off LotMan while we investigate dlopen issues

jhiemstrawisc:
This should prevent LotMan from being compiled into Pelican, which excises our import of purego.


web_ui/frontend/package-lock.json
haoming29, line 1: The changes to the `package-lock.json` looks a bit suspicious since this PR has nothing to do with the frontend stuff. @CannonLock could you confirm this is expected?

haoming29 (approved):
Build a production origin container and run locally with `--platform linux/amd64` works and I get Pelican binary running. Pre-approval with one question regarding the extra npm file changes

web_ui/frontend/package-lock.json
CannonLock: This is fine, the package-lock.json had a package committed that was removed from the package.json.

CannonLock (approved):

PR #1090, opened 2024-04-10
Generate public/private keys and server web ui htpasswd files

### LINKED ISSUE #1019: Pelican should generate the public key given the location of a private key, or generate the private key as well
We don't have a way for admins to look up the public key of the Pelican origin before running Pelican origin. Thus, it's very difficult for general admins to register their origin/namespace ahead of time. To mitigate the issue, Pelican should have a command to generate a private/public key-pair which admin can use as the private key of their production Origin to register, and the public to show up at the registry. 

A command like `pelican keygen -p/--private <private key location> -o/--output <key output location>` should work where pelican _generates_ a private key (if `-p` is empty), stores it to what `-o` reads, and output the corresponding public JWT to stdout.

haoming29:
Closes #1019 

This PR didn't include the function to generate keys/password file as a k8s secret object, due to the fact that existing `kubectl` command already includes the function to convert an existing file to a k8s secret file. After talking to @brianhlin , we agreed that it is sufficient to generate the keys, and we should document how to leverage `kubectl` for admins to generate their k8s secrets if desired.


cmd/keygen.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/tg123/go-htpasswd"
+)
+
+var (
+	keygenCmd = &cobra.Command{
+		Use:   "keygen",
+		Short: "Generate a public-private key-pair for Pelican issuer",
```
bbockelm, line 39: We combine "key pair generation" and "admin password file generation" into one command named "keygen".  This is likely to be confusing.  Either have two top-level commands or one generic command ("generate"?) with two subcommands ("keygen", "admin-password").

cmd/keygen.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	passwordFile := filepath.Join(outPath, "server-web-passwd")
+	viper.Set(param.Server_UIPasswordFile.GetName(), passwordFile)
+	file, err := os.OpenFile(passwordFile, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0600)
+	if err != nil {
+		return err
+	}
+	file.Close()
+
+	_, err = htpasswd.New(passwordFile, []htpasswd.PasswdParser{htpasswd.AcceptBcrypt}, nil)
```
bbockelm, line 70: How does this handle trailing newlines (or lack thereof)?

cmd/keygen.go
```diff
@@ -0,0 +1,126 @@
...
+	// and parse the private key and generate the corresponding public key for us
+	pubkey, err := config.GetIssuerPublicJWKS()
+	if err != nil {
+		return err
+	}
+	bytes, err := json.MarshalIndent(pubkey, "", "	")
+	if err != nil {
+		return errors.Wrap(err, "failed to generate json from jwks")
+	}
+	output, err := os.OpenFile(pubKeyPath, os.O_CREATE|os.O_WRONLY, 0444)
```
bbockelm, line 97:
```suggestion
	output, err := os.OpenFile(pubKeyPath, os.O_CREATE|os.O_WRONLY, 0644)
```

bbockelm: Actually, would it be better to just invoke `os.WriteFile` here?  Would save a few lines of boilerplate.

bbockelm (changes requested):
Three conceptual items:
1.  Improve the handling of the password so it's not visible to others on the system.
2. Break the two pieces of functionality into two commands instead of one command with two quite different behaviors.
3. Add unit test coverage for the new commands.

bbockelm (approved):
Revised changes look good.

PR #1098, opened 2024-04-11
Fix work deadlock

bbockelm:
This PR fixes an observed deadlock when the job-to-file routine is writing to the runMux routine while the runMux routine is blocked writing to the job-to-file routine.  The fix is to only accept more work (which is what triggers the runMux writing) when there's nothing buffered for job-to-file.

This was observed in the wild when running in "STASH" mode which we likely don't need anymore; the PR eliminates the STASH mode as well.

For good measure, this also:
- Fixes up a few error messages observed.
- Adds small buffering to both of the queues that were blocked.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -542,6 +542,7 @@ func NewTransferEngine(ctx context.Context) *TransferEngine {
 	)
 
 	// Start our cache for url metadata
+	// This is stopped in the `Shutdown` method
 	go pelicanURLCache.Start()
 
 	te := &TransferEngine{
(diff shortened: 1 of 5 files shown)


matyasselmeci:
What are the ramifications of removing the STASH mode? What will stop working? I think that should be a separate PR from the deadlock fix, since the deadlock fix is critical for getting 7.7 working in the OSPool, and removal of STASH mode is not.

bbockelm:
> What are the ramifications of removing the STASH mode?

The STASH mode was effectively introduced in 7.7.0.  It's always "existed" but, until 7.7.0, it was identical to OSDF.  In 7.7.0, it became distinct from OSDF due to code like this:

https://github.com/PelicanPlatform/pelican/blob/main/client/handle_http.go#L483-L501

(which is "if OSDF" / "else if Pelican" -- but not checking for the Stash case).

This is causing topology to be used unexpectedly which then triggered the deadlock.  Removing STASH mode has the ramification of restoring the 7.6.0 behavior.

matyasselmeci:
OK, sounds good

bbockelm:
@haoming29 - please backport to the 7.7.x series once merged.

joereuss12 (approved):
LGTM

joereuss12:
@haoming29 this is ready for backport

haoming29:
> @haoming29 this is ready for backport

Sounds good!

PR #1101, opened 2024-04-11
Made additional retryable errors

joereuss12:
Made StoppedTransferError and Cannot allocate memory a retryable error

### DIFF
--- a/client/errorAccum.go
+++ b/client/errorAccum.go
@@ -71,6 +71,11 @@ func (te *TransferErrors) AddPastError(err error, timestamp time.Time) {
 	}
 }
 
+// This resets the TransferErrors object for testing purposes
+func (te *TransferErrors) ResetErrors() {
+	te.errors = make([]error, 0)
+}
+
 func (te *TransferErrors) Unwrap() []error {
 	return te.errors
 }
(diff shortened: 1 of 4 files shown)


haoming29 (approved):
LGTM!

client/errorAccum.go
```diff
@@ -71,6 +71,11 @@ func (te *TransferErrors) AddPastError(err error, timestamp time.Time) {
 	}
 }
 
+// This resets the TransferErrors object for testing purposes
+func (te *TransferErrors) ResetErrors() {
```
bbockelm, line 75: It seems this is only used by unit tests - please change to not be exported outside the module.

client/handle_http.go
```diff
@@ -112,6 +112,10 @@ type (
 		Err error
 	}
 
+	AllocateMemoryError struct {
```
bbockelm, line 115: It doesn't seem this needs to be visible outside the module - please don't export.

bbockelm (changes requested):
Basic idea is good, just nits about the visibility.  Since we want others to eventually use the client module, we need to be extremely careful about exported types and methods.

bbockelm (approved):

Issue #1097, opened 2024-04-10
Cache UI didn't redirect to login page with unauthed user


haoming29:
Spin up the cache and head to the cache UI at `/view/cache/index.html`

The cache dashboard page renders with a bunch of 401/403 errors.

The expected behavior is to redirect unauthed user to the login page at `/view/login`

PR #1103, opened 2024-04-11
Fix cache dashboard page skips auth check and wrong health metric name

haoming29:
Closes https://github.com/PelicanPlatform/pelican/issues/1097 

Also fixed the issue where the component health metric has the wrong name `Xrootd.cache` instead of `xrootd`


turetske (approved):
LGTM!

PR #1111, opened 2024-04-12
Directory listing changes in Pelican client

joereuss12:
This is a different approach to the change of directory listings. Now, for recursive downloads, we will query the director for a PROPFIND to get the location of our "dirlisthost" (or origin). Then procede with the normal recursive download code. If we receive a 405 from the director doing this PROPFIND, we will revert back to the old way of doing things

### DIFF
--- a/client/fed_linux_test.go
+++ b/client/fed_linux_test.go
@@ -22,6 +22,7 @@ package client_test
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -391,4 +392,30 @@ func TestRecursiveUploadsAndDownloads(t *testing.T) {
 		// Throw in a viper.Reset for good measure. Keeps our env squeaky clean!
 		viper.Reset()
 	})
+}
+
+// This tests that is origins disable listings, we should fail the download
+// Note: origins disabling listings override the existance of dirlisthost, causing a failure
+func TestFailureOnOriginDisablingListings(t *testing.T) {
+	viper.Reset()
+	server_utils.ResetOriginExports()
+
+	viper.Set("Logging.Level", "debug")
+	viper.Set("Origin.StorageType", "posix")
+	viper.Set("Origin.ExportVolumes", "/test")
+	viper.Set("Origin.EnablePublicReads", true)
+	viper.Set("Origin.EnableListings", false)
+	fed := fed_test_utils.NewFedTest(t, "")
+
+	destDir := filepath.Join(fed.Exports[0].StoragePrefix, "test")
+	require.NoError(t, os.MkdirAll(destDir, os.FileMode(0755)))
+	log.Debugln("Will create origin file at", destDir)
+	err := os.WriteFile(filepath.Join(destDir, "test.txt"), []byte("test file content"), fs.FileMode(0644))
+	require.NoError(t, err)
+	uploadURL := fmt.Sprintf("pelican://%s:%s%s/%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()),
+		fed.Exports[0].FederationPrefix, "test")
+
+	_, err = client.DoGet(fed.Ctx, uploadURL, t.TempDir(), true)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "Origin and/or namespace does not support directory listings")
 }
--- a/client/fed_linux_test.go
+++ b/client/fed_linux_test.go
@@ -395,7 +395,7 @@ func TestRecursiveUploadsAndDownloads(t *testing.T) {
 }
 
 // This tests that is origins disable listings, we should fail the download
-// Note: origins disabling listings override the existance of dirlisthost, causing a failure
+// Note: origins disabling listings override the existence of dirlisthost, causing a failure
 func TestFailureOnOriginDisablingListings(t *testing.T) {
 	viper.Reset()
 	server_utils.ResetOriginExports()
--- a/client/fed_linux_test.go
+++ b/client/fed_linux_test.go
@@ -412,10 +412,10 @@ func TestFailureOnOriginDisablingListings(t *testing.T) {
 	log.Debugln("Will create origin file at", destDir)
 	err := os.WriteFile(filepath.Join(destDir, "test.txt"), []byte("test file content"), fs.FileMode(0644))
 	require.NoError(t, err)
-	uploadURL := fmt.Sprintf("pelican://%s:%s%s/%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()),
+	downloadURL := fmt.Sprintf("pelican://%s:%s%s/%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()),
 		fed.Exports[0].FederationPrefix, "test")
 
-	_, err = client.DoGet(fed.Ctx, uploadURL, t.TempDir(), true)
+	_, err = client.DoGet(fed.Ctx, downloadURL, t.TempDir(), true)
 	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "Origin and/or namespace does not support directory listings")
+	assert.Contains(t, err.Error(), "origin and/or namespace does not support directory listings")
 }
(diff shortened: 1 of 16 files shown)


client/handle_http.go
```diff
@@ -2122,31 +2126,52 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
 // Walk a remote directory in a WebDAV server, emitting the files discovered
-func (te *TransferEngine) walkDirDownload(job *clientTransferJob, transfers []transferAttemptDetails, files chan *clientTransferFile, url *url.URL) error {
-	// Create the client to walk the filesystem
-	rootUrl := *url
-	if job.job.namespace.DirListHost != "" {
-		// Parse the dir list host
-		dirListURL, err := url.Parse(job.job.namespace.DirListHost)
+func (te *TransferEngine) walkDirDownload(job *clientTransferJob, transfers []transferAttemptDetails, files chan *clientTransferFile, remoteURL *url.URL) error {
+	// Query the director a PROPFIND to see if we can get our directory listing
+	resp, err := queryDirector("PROPFIND", remoteURL.Path, job.job.directorUrl+"/api/v1.0/director/origin")
```
jhiemstrawisc, line 2131: Can you construct the `job.job.directorUrl` with `url.JoinPath` to make sure it's valid? Or are we otherwise sure this construction won't result in a double `/`?

client/handle_http.go
```diff
@@ -2122,31 +2126,52 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+	if err != nil {
+		return err
+	}
+
+	collectionsUrl := *remoteURL
+	// If we get a "method not allowed", we go back to the old method
```
jhiemstrawisc, line 2137: For someone who doesn't already know what this code does, this comment won't be very elucidating. How about something like "If the director responds with 405 (method not allowed), we're working with an old Director. In that event, we try to fallback and use the deprecated dirlisthost"

client/handle_http.go
```diff
@@ -2122,31 +2126,52 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+	if resp.StatusCode == http.StatusMethodNotAllowed {
+		log.Debugln("PROPFIND against director failed, trying to get collections url from namespace")
```
jhiemstrawisc, line 2139: Since this is at the client, our error messages will be much more user-facing than they would be in a Pelican server like the Director. You should make the final decision, but I think we could come up with an error message that's a little more user friendly -- I don't think the average person knows what PROPFIND is or what a collections URL means.

client/handle_http.go
```diff
@@ -2122,31 +2126,52 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+			dirListURL, err := url.Parse(job.job.namespace.DirListHost)
+			if err != nil {
+				log.Errorln("Failed to parse dirlisthost from namespaces into URL:", err)
+				return err
+			}
+			collectionsUrl = *dirListURL
+		} else {
+			// Both methods to get directory listings failed
+			log.Errorln("Host for directory listings is unknown")
+			return errors.New("Host for directory listings is unknown")
```
jhiemstrawisc, line 2151: Does this result in duplicate error messages if we both log and return the same message?

client/handle_http.go
```diff
@@ -2160,7 +2185,7 @@ func (te *TransferEngine) walkDirDownloadHelper(job *clientTransferJob, transfer
 	}
 	infos, err := client.ReadDir(remotePath)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "Failed to PROPFIND remote directory")
```
jhiemstrawisc, line 2188: Same with this one, let's find a way to make this both something a user can interpret, and a developer can debug with.

client/handle_http.go
```diff
@@ -2122,31 +2126,52 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+				return err
+			}
+			collectionsUrl = *dirListURL
+		} else {
+			// Both methods to get directory listings failed
+			log.Errorln("Host for directory listings is unknown")
+			return errors.New("Host for directory listings is unknown")
+		}
+	} else if resp.StatusCode == http.StatusTemporaryRedirect {
+		collectionsStr := resp.Header.Get("Location")
```
jhiemstrawisc, line 2154: Should we also take into consideration any of the metalink header urls? For example, if the first origin from the `Location` header doesn't respond, is it worth checking any other origins that the Director indicates might serve the object?

jhiemstrawisc (changes requested):

joereuss12:
Note: this still needs a unit test to test multiple origin support so hold off on a review for now.

joereuss12:
@bbockelm see if you like these changes better, I think I hit all of the suggestions you gave. Also, we want listings to be true by default correct?

server_structs/director.go
```diff
@@ -76,6 +76,7 @@ type (
 		Latitude     float64    `json:"latitude"`
 		Longitude    float64    `json:"longitude"`
 		Writes       bool       `json:"enable_write"`
+		Listings     bool       `json:"Listing"`              // True if the origin allows directory listings
```
bbockelm, line 79: Match the style of the other flags
```suggestion
		Listings     bool       `json:"enable_listing"`              // True if the origin allows directory listings
```

director/director_ui.go
```diff
@@ -48,6 +48,7 @@ type (
 		Longitude    float64                   `json:"longitude"`
 		Writes       bool                      `json:"enableWrite"`
 		DirectReads  bool                      `json:"enableFallbackRead"`
+		Listings     bool                      `json:"enableListings"`
```
bbockelm, line 51: Match the fact the other flags use the singular.

```suggestion
		Listings     bool                      `json:"enableListing"`
```

client/fed_linux_test.go
```diff
@@ -392,3 +393,29 @@ func TestRecursiveUploadsAndDownloads(t *testing.T) {
...
+	viper.Set("Origin.EnablePublicReads", true)
+	viper.Set("Origin.EnableListings", false)
+	fed := fed_test_utils.NewFedTest(t, "")
+
+	destDir := filepath.Join(fed.Exports[0].StoragePrefix, "test")
+	require.NoError(t, os.MkdirAll(destDir, os.FileMode(0755)))
+	log.Debugln("Will create origin file at", destDir)
+	err := os.WriteFile(filepath.Join(destDir, "test.txt"), []byte("test file content"), fs.FileMode(0644))
+	require.NoError(t, err)
+	uploadURL := fmt.Sprintf("pelican://%s:%s%s/%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()),
```
bbockelm, line 415: Nitpick -- this is called uploadURL but is for downloading.  Please rename.

client/handle_http.go
```diff
@@ -1069,6 +1073,45 @@ func (tc *TransferClient) NewTransferJob(ctx context.Context, remoteUrl *url.URL
...
+			return nil, err
+		}
+		// If the director responds with 405 (method not allowed), we're working with an old Director.
+		// In that event, we try to fallback and use the deprecated dirlisthost
+		if resp.StatusCode == http.StatusMethodNotAllowed {
+			log.Debugln("Failed to find collections url from director, attempting to find directory listings host in namespace")
+			// Check for a dir list host in namespace
+			if tj.namespace.DirListHost == "" {
+				// Both methods to get directory listings failed
+				return nil, errors.New("Origin and/or namespace does not support directory listings")
```
bbockelm, line 1091: Nitpick: make this a non-exported symbol at the module level (put it in the `var` block at the top of the file).  This way, we can do an `errors.Is` against it in the future.

Also, please follow go style and make error message not capitalized.

client/handle_http.go
```diff
@@ -1304,8 +1347,6 @@ func (te *TransferEngine) createTransferFiles(job *clientTransferJob) (err error
 			log.Errorln("Failed to get namespaced caches (treated as non-fatal):", err)
 		}
 
-		log.Debugln("Matched caches:", closestNamespaceCaches)
```
bbockelm, line 1307: What's this diff from?

bbockelm (approved):
Some minor nitpicks.  Please resolve and, once done, I'm pre-approving this to go in.

jhiemstrawisc (approved):
👍

Issue #1112, opened 2024-04-12
Populate `StartdAttrs` in -classad output of Pelican plugin


joereuss12:
Pelican Service:
- [X] Plugin

The condor team recently enabled the ability to show the Pelican version in the starter ad. However, on the Pelican end of things, we need to populate the attribute `StartdAttrs` within the Pelican -classad output. We need to include PluginVersion/PelicanVersion in that attribute so it shows in the starter ad.

Ex:
```
MultipleFileSupport = true
PluginVersion = "0.2"
PluginType = "FileTransfer"
ProtocolVersion = 2
SupportedMethods = "http,https,ftp,file,dav,davs"
StartdAttrs = "CurlPluginProtocol, CurlPluginMethods"
CurlPluginProtocol = join({"Multi", ProtocolVersion})
CurlPluginMethods = SupportedMethods
```
Then the Starter, and consequently the STARTD will have these attributes
```
CurlPluginMethods = "http,https,ftp,file,dav,davs"
CurlPluginProtocol = "Multi2"
```

PR #1113, opened 2024-04-12
Populate StartdAttrs in plugin -classad output

joereuss12:
Added the PluginVersion to the StartdAttrs so that it can show up as a starter ad.

Jira ticket: https://opensciencegrid.atlassian.net/browse/HTCONDOR-1051
Pelican ticket: https://github.com/PelicanPlatform/pelican/issues/1112

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -133,6 +133,7 @@ func stashPluginMain(args []string) {
 			fmt.Println("PluginType = \"FileTransfer\"")
 			fmt.Println("ProtocolVersion = 2")
 			fmt.Println("SupportedMethods = \"stash, osdf\"")
+			fmt.Println("StartdAttrs = \"PluginVersion\"")
 			os.Exit(0)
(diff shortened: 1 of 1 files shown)


cmd/plugin.go
```diff
@@ -133,6 +133,7 @@ func stashPluginMain(args []string) {
 			fmt.Println("PluginType = \"FileTransfer\"")
 			fmt.Println("ProtocolVersion = 2")
 			fmt.Println("SupportedMethods = \"stash, osdf\"")
+			fmt.Println("StartdAttrs = \"PluginVersion\"")
```
bbockelm, line 136: Does this get copied literally into the startd ad?  Or does it get a prefix based on the plugin name somehow?

Given there are multiple plugins at play, I think we'll want some prefix (`PelicanPluginVersion`).

bbockelm (changes requested):

bbockelm (approved):
LGTM.

PR #1115, opened 2024-04-12
Plugin returns last attempted source & filters \n

joereuss12:
This fixes a bug where condor was not returning a HoldReasonCode/Subcode and it was caused by the Pelican plugin returning a classad that contained a newline character. In addition, we weren't getting a URL returned either, because the TransferUrl classad was not being populated on error. The fixes here make it so that this classad will be populated in all cases it is needed.


joereuss12:
Also, I am not sure if we want to throw this in a bugfix release or anything

classads/classads.go
```diff
@@ -62,6 +62,7 @@ func (c *ClassAd) String() string {
 		buffer.WriteString(" = ")
 		switch v := value.(type) {
 		case string:
+			v = strings.ReplaceAll(v, "\n", "")
```
bbockelm, line 65: This doesn't make sense.

The next line removes all newlines from the string.  What does removing them here do?

cmd/plugin.go
```diff
@@ -411,17 +415,18 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
 	for {
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
+			return lastAttemptedSource, ctx.Err()
```
bbockelm, line 418:
```suggestion
			err = ctx.Err()
			return
```

cmd/plugin.go
```diff
@@ -411,17 +415,18 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
...
 		case transfer, ok := <-workChan:
 			if !ok {
 				tc.Close()
 				workChan = nil
 				break
 			}
+			lastAttemptedSource = transfer.url.String()
```
bbockelm, line 425: What's returned here?  Is it the URL that was used for the transfer or the URL that was requested by HTCondor?

cmd/plugin.go
```diff
@@ -411,17 +415,18 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
...
 			// Check we have valid query parameters
 			err := checkValidQuery(transfer.url)
 			if err != nil {
-				return err
+				return lastAttemptedSource, err
```
bbockelm, line 429:
```suggestion
				return
```

cmd/plugin.go
```diff
@@ -448,12 +453,12 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
 			project := client.GetProjectName()
 			tj, err = tc.NewTransferJob(context.Background(), &urlCopy, transfer.localFile, upload, recursive, project, client.WithAcquireToken(false), client.WithCaches(caches...))
 			if err != nil {
-				return errors.Wrap(err, "Failed to create new transfer job")
+				return lastAttemptedSource, errors.Wrap(err, "Failed to create new transfer job")
```
bbockelm, line 456:
```suggestion
				err = errors.Wrap(err, "Failed to create new transfer job")
				return
```

cmd/plugin.go
```diff
@@ -448,12 +453,12 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
...
 			}
 			jobMap[tj.ID()] = transfer
 
 			if err = tc.Submit(tj); err != nil {
-				return err
+				return lastAttemptedSource, err
```
bbockelm, line 461:
```suggestion
				return
```

bbockelm (changes requested):
Two things I don't understand here:

1. What does removing the newline characters do?  It's followed by a line of code that removes all the newline characters (`strconv.QuoteToASCII`) ... so this occurs twice?  What's problematic, exactly?  If there's some issue this is triggering then that sounds quite subtle and we should include a unit test to make sure we avoid regressions in the future.
2. What does recording the last attempted source solve?  In what cases does that make a difference?  For example, if you get an error after a successful ClassAd was created, you now have two ClassAds for a single URL, one successful and one failed.  What's the semantics of that?  If we need something like this, I'd like to ensure there's some structural mechanism to avoid bugs.
   - Note, for example, you still end up with an empty `TransferUrl` if a panic occurs during a transfer

classads/classads.go
Todd-L-Miller: The new line converts all newlines into the new ClassAd syntax's escape for a newline, which _actually gets handled properly_ all the way through until it gets to the schedd, which is rejects a string containing a newline as invalid; see https://opensciencegrid.atlassian.net/browse/HTCONDOR-2379 for details.

bbockelm: @Todd-L-Miller: To be clear - 

The contents of the message currently should contain two bytes, `\` and `n`, not the single byte `\n`.  So you're saying that these bytes are re-interpreted as the newline character and then the interpreted string is printed into the Hold message?

We could squash newlines specifically - but are there other forbidden characters?  What about \r or general escape sequences?  Should we be escaping `\` to `\\`?

bbockelm: Talked a bit with Jaime about this over Slack.  He believes that `\n` is the only special character in old ClassAds (Todd: note that this disagrees with your comment in the JIRA ticket -- please let me know if you guys decide on a different conclusion; we're flexible!).

So here's what I propose:
1. Add a comment explaining why `\n` is special and needs to be escaped.  Just squashing it is too much "magic".
2. Replace `\n` with `\\n` before quoting the string.  This preserves what was originally there as best we can.
3. Find the error message generating this newline and remove the newline.  It obviously shouldn't be there in the first place!

joereuss12: I removed the `\n` in the error thrown from the director so it should not show up anymore. However, I'm confused as to why to remove `\\n` instead of just `\n`. Because with the old director error message (with the `\n` at the end) if I replace `\n` it works and successfully removes it but if I replace `\\n` it does not work and the `\n` is still there in the error message still causing problems.

joereuss12:
>     2. What does recording the last attempted source solve?  In what cases does that make a difference?  For example, if you get an error after a successful ClassAd was created, you now have two ClassAds for a single URL, one successful and one failed.  What's the semantics of that?  If we need something like this, I'd like to ensure there's some structural mechanism to avoid bugs.
>        
>        * Note, for example, you still end up with an empty `TransferUrl` if a panic occurs during a transfer

The issue that I found was that Pelican fails and returns before any `TransferUrl` classad is set. This is problematic because to set the `URL file = ...` in the returned error message from condor, it relies on `TransferUrl` being set. My thought process led to recording the last attempted transfer url so that this can be recorded in any case. However, it is tricky to find the right approach to this as the current url being used is in runPluginWorker and we would need a way to ensure the classad is being updated somehow.

classads/classads.go
bbockelm: I think the current discussion is leaning toward:

```suggestion
			v = strings.ReplaceAll(v, "\n", "\\n").ReplaceAll(v, "\r", "\\r")
```

for step (2).

bbockelm (approved):
LGTM after revisions.

PR #1116, opened 2024-04-14
"Airplane mode" for unit tests

bbockelm:
This removes unintended usage of topology (due to unit tests not reseting their original state) and allows nearly every unit test run without network connectivity (exception is the S3 unit test).

One aspect of this causing the most code churn is deferring the lookup of federation metadata until it is needed.  This allows various simple client invocations to succeed (think: `pelican help` and similar -- anything where the client config is initialized but not needed) and avoiding failures in the corresponding unit tests.

### DIFF
--- a/broker/broker_test.go
+++ b/broker/broker_test.go
@@ -129,7 +129,9 @@ func doRetrieveRequest(t *testing.T, ctx context.Context, dur time.Duration) (*h
 
 	reqReader := bytes.NewReader(reqBytes)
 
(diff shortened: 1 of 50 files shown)


bbockelm:
@joereuss12 - you test on Windows, right?  Can you take a look at the failing unit test?  I suspect it's something simple (basically, it looks like the unit test isn't fully shut down before cleanup) but it's really tough to debug without Windows access.

jhiemstrawisc (approved):
LGTM

PR #1119, opened 2024-04-15
Replace `jwks.cache` by TTL cache for namespace jwks at the director

### LINKED ISSUE #1108: Director JWKS cache becomes invalid when the registry is down/back online
For production director, if the registry is down for a while (say 10 min or longer) and brought back online, the director jwks cache becomes invalid, and the following origin/cache advertisement will be rejected due to failure to authenticate the token (grabbing the jwks from the jwks cache)

An example error message:

```
time="2024-04-11T21:00:16Z" level=warning msg="Failed to verify token:  url \"https://osdf-registry.osg-htc.org:443/api/v1.0/registry/sage/.well-known/issuer.jwks\" is not registered (did you make sure to call Register() first?)"
```

Currently the solution is to restart the director after the registry was restarted, but their restart should be independent.

haoming29:
Fixes #1108 

The issue #1108 describes in the production server is likely triggered with the following conditions:
* The registry server pod is down (for whatever reason), but visiting registry URLs gives 404 (probably due to traefik)
* When the director attempts to verify a token origin/cache server sent to advertise, it calls `server_utils.GetJWKSURLFromIssuerURL(issuerUrl)`. This function first attempts to request `issuerUrl` and get the `jwks` endpoint from `issuerUrl` response. The `issuerUrl` is `https://<registry>/path/to/issuer`, which returns 404. The function recognize the response 404 as a compatibility issue and thought it hit the legacy OSDF topology, so it instead construct the jwks endpoint by concatenating  `Federation.IssuerURL` with predefined path `/api/v1.0/registry/<namespace>/.well-known/issuer.jwks`
* Director then try `namespaceKeys` cache with the key `<namespace>`
* If the key returns a jwks cache item (ar), we call `ar.Get()` with the jwks endpoint (keyLoc) as the cache key for getting the jwks
* However, if the registry-returned jwks endpoint is different from director-constructed jwks endpoint, we will hit the issue where the jwks cache does not have one item or the other recorded.

The root cause is that `server_utils.GetJWKSURLFromIssuerURL` is not atomic in returning jwks endpoint given an issuer url. 

This PR fixes the issue by stripping off the layer of indirect and instead uses TTL cache for storing jwks, which prevents us from noting being able to get the jwks due to inconsistent jwks urls

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -782,8 +782,6 @@ func getPrefixByPath(ctx *gin.Context) {
 		ctx.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "Bad request. Path is empty or '/' "})
 		return
 	}
-	namespaceKeysMutex.Lock()
-	defer namespaceKeysMutex.Unlock()
 
 	originNs, _, _ := getAdsForPath(pathParam)
 
--- a/director/director_api.go
+++ b/director/director_api.go
@@ -135,8 +135,6 @@ func ConfigTTLCache(ctx context.Context, egrp *errgroup.Group) {
 		log.Info("Gracefully stopping director TTL cache eviction...")
 		serverAdMutex.Lock()
 		defer serverAdMutex.Unlock()
-		namespaceKeysMutex.Lock()
-		defer namespaceKeysMutex.Unlock()
 		serverAds.DeleteAll()
 		serverAds.Stop()
 		namespaceKeys.DeleteAll()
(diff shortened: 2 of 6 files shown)


haoming29:
Let's backport this to 7.7.x once it's reviewed and merged

turetske (approved):

turetske:
Rebase to fix the macOS failures and then LGTM

PR #1126, opened 2024-04-16
Move Swagger to Top Level

### LINKED ISSUE #509: Move Swagger spec from `web_ui/frontend/...` to top level folder
For making `make web-build` work in container, we intentionally moved Swagger spec yaml file from the top-level folder to the frontend folder, assuming we will use auto-generated specs in the future.  However, this will not happen soon, but we instead want to do it manually via editing the spec yaml file. To make it easier to be discovered, let move the yaml file back to top level folder. We also need to take care of the make file in container build to copy Swagger spec at early stage.

CannonLock:
Closes #509


haoming29 (approved):
LGTM!

Issue #1105, opened 2024-04-12
Create origin UI element for registration


bbockelm:
After watching folks fumble through the CLI-based registration with containers, it's apparent that identity-based registration is too awkward outside the "developer" field.

Let's make an option so the registration workflow is completely web-based.

An origin without a valid registration for a namespace prefix should:

1.  Have a clear UI element on origin homepage (maybe the exports table) showing a big red X if the registration isn't complete.  This should only show up for admins.
2. When the user clicks on the broken registration, it should initiate the linking workflow between the CILogon identity and the registry.  I assume that the origin has already started the prefix registration process.
3. The user should land on the registry and be able to edit / update desired details such as institution and description.  Even once identity linking is done, for as long as there are required fields missing the "X" icon should show up in the origin.

An analogous workflow should exist for caches as well.

haoming29:
By "broken registration", what's our criteria besides the identity? The only other "required" field is the institution list. Do we want to include other fields as well?

bbockelm:
Ah, I chose my wording poorly.  By "broken", I really meant "incomplete".  This could be due to lacking the identity association, a lack of institution, or any future required field we decide to add.

The point is, we should have the ability to "click through" from the origin host to the registry to complete the process.

PR #1129, opened 2024-04-17
Add support for origin registration completeness status

### LINKED ISSUE #1130: Web UI didn't redirect to `redirectURL` once user logged in
Pelican Version:

Pelican Service:
- [x] Registry
- [x] Director
- [x] Origin
- [x] Cache

**Describe the bug**

If a user hit s page where requires login, the user will be taken to the login page. If the login is successful, the user should be redirect back to the original page. Current this doesn't work. The user was only redirected to the root page at `/view/`

**To Reproduce**

**Expected behavior**

**Screenshots**

 - OS: [e.g. Linux, MacOS]

**Additional context**

haoming29:
Related issue: #1105. This PR implements the API for enabling the feature mentioned in the ticket, but the UI components is not in the scope of this PR.
Also fixes #1130 

Some noticeable changes this PR made:
* Allow user with a valid access token to view and edit a namespace registration in the registry that has an empty `UserID` field. A valid access token is a JWT signed by the origin with the private key matches public key from the the namespace registration to access
* Origin UI API `/exports` now has three additional fields: `edit_url`, `status`, `status_description` as documented in the updated Swagger spec. `status` shows the registration status of the namespace, and `status_description` provides a detailed description of the status. `edit_url` will take the user from the origin to the corresponding registry page to edit the registration with a valid access token embedded.
* Streamlined web UI redirects with the following updated policies:
  * For an unauthenticated request hit an auth required page, the redirect to `/login` page is handled by the **frontend**, and frontend will "save" the URL to the current page as a query parameter `redirectURL` and add it to the `/login` URL.
  * When a user is logged in, either via admin password (`/login` page) or oauth callback (`/auth/oauth/callback`), the redirect to the original page was handled by the **backend**.


registry/registry_ui.go
```diff
@@ -480,22 +500,29 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
...
+			if !belongsTo {
+				// If the user doen's own the namespace, they can't edit it
+				// but if they have an access token generated by the private key
+				// that corresponds to the public key of the namespace they are trying to access,
+				//  we allow them to do it so that admin can fill out the rest of the registration if the user_id is empty
+				if accessToken == "" {
+					log.Errorf("Access denied from user %s for namespace with id=%d", user, id)
+					ctx.JSON(http.StatusNotFound, server_structs.SimpleApiResp{
+						Status: server_structs.RespFailed,
+						Msg:    "Namespace not found. Check the id or if you own the namespace"})
```
jhiemstrawisc, line 523: Is this actually a "namespace not found" error, or is it a "permission denied" error? The message seems to suggest both.

registry/registry_ui.go
```diff
@@ -512,7 +539,73 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
...
+						server_structs.SimpleApiResp{
+							Status: server_structs.RespFailed,
+							Msg:    fmt.Sprint("Invalid access token: ", err),
+						})
+					return
+				}
+
+				scopeValidator := token_scopes.CreateScopeValidator([]token_scopes.TokenScope{token_scopes.Registry_EditRegistration}, false)
+				if err = jwt.Validate(accessJWT, jwt.WithValidator(scopeValidator)); err != nil {
+					log.Errorf("Failed to verify the scope of the token: %v", err)
```
jhiemstrawisc, line 571: Should these logged errors also enforce a lowercase? Or do we only care about then when passing errors back up the error stack?

swagger/pelican-swagger.yaml
```diff
@@ -418,6 +418,20 @@ definitions:
   OriginExport:
     type: object
     properties:
+      status:
+        type: string
+        description:
+          The registration status of the federation prefix.
+          Can be "Not Supported"|"Completed"|"Incomplete"|"Registration Error"
```
jhiemstrawisc, line 425: How is "Not Supported" different from "Registration Error"?

web_ui/authentication_test.go
```diff
@@ -297,7 +297,9 @@ func TestPasswordBasedLoginAPI(t *testing.T) {
 	defer func() { require.NoError(t, egrp.Wait()) }()
 	defer cancel()
 
+	dirName := t.TempDir()
 	viper.Reset()
+	viper.Set("ConfigDir", dirName)
```
jhiemstrawisc, line 302: Good catch :)

web_ui/ui.go
```diff
@@ -181,15 +186,15 @@ func handleWebUIRedirect(ctx *gin.Context) {
 	// If requestPath doesn't have extension, is not a directory, and has a index file, redirect to index file
 	if !strings.Contains(requestPath, ".") && !strings.HasSuffix(requestPath, "/") {
 		if _, err := webAssets.ReadFile("frontend/out" + requestPath + "/index.html"); err == nil {
-			ctx.Redirect(http.StatusMovedPermanently, "/view/"+requestPath+"/")
+			ctx.Redirect(http.StatusMovedPermanently, "/view/"+requestPath+"/"+query)
```
jhiemstrawisc, line 189: Is the trailing `/` before the query necessary? Usually I see queries take the form `/some/path?query` and not `/some/path/?query`

web_ui/ui.go
```diff
@@ -181,15 +186,15 @@ func handleWebUIRedirect(ctx *gin.Context) {
...
 			ctx.Abort()
 			return
 		}
 	}
 
 	// If only one server is enabled, and user is requesting "/", redirect to that server
 	if len(config.GetEnabledServerString(true)) == 1 && requestPath == "/index.html" {
-		ctx.Redirect(http.StatusFound, "/view/"+config.GetEnabledServerString(true)[0]+"/")
+		ctx.Redirect(http.StatusFound, "/view/"+config.GetEnabledServerString(true)[0]+"/"+query)
```
jhiemstrawisc, line 197: See previous comment

jhiemstrawisc (changes requested):
A few minor changes, and a note that I skimmed any changes under web_ui (not something I know enough about to review).

registry/registry_ui.go
haoming29: This is the part I'm not very confident about. I saw people returning 404 for objects that does not belong to a user (although it exists), (to confuse attackers?), but this will also somehow confuse our users a bit.

registry/registry_ui.go
haoming29: The logged error are not chained together (unless the real Go error), so I don't think we should put them in lower case if they are meant to be a complete sentence.

swagger/pelican-swagger.yaml
haoming29: "Not Support" is when the registry is below 7.8.0 (or whichever version this PR is included in)
"Registration Error" means that there's an issue with the registration itself, like not having identity when registering a namespace overlapped with topology, or something else that leads to a failure of registering the namespace. In this case, the "completeness status" doesn't make any sense because the namespace is not registered at all.

haoming29: I will update the Swagger doc to detail them

web_ui/ui.go
haoming29: Yeah that's our set convention, we enforce the trailing slash for frontend routes

registry/registry_ui.go
jhiemstrawisc: I think we're probably safe to expose its under-construction existence here, especially if there are any other queries an attacker could perform to learn about it. Presumably, namespaces are public once they're registered anyway, so existence doesn't seem like it needs to be tightly protected.

jhiemstrawisc (approved):
LGTM! I'll approve and let you merge when you're ready -- I think we're still discussing one of the comments, but you get final say.

haoming29:
@jhiemstrawisc I agree with you, and for the sake of consistency, I updated the status code to use 403 and error messages for the `getNamespace` handler as well.

PR #1131, opened 2024-04-17
Allow plugin to work w/ non-url/localfile classads

joereuss12:
The plugin now works if it receives classads in the infile that are not the url or LocalFileName as long as the infile contains them at some point. This is useful for HTCSS devs as they will eventually pass things to the plugin through the infile and for now we don't fail when that happens. Updated unit test and fixed bug in writing the error message to check that we have a url (string) before parsing it (in cases of panic we don't pass a url).

This was a request from TJ and other HTCSS devs


jhiemstrawisc (approved):
LGTM

Issue #1091, opened 2024-04-10
Pelican origin token create does not respect issuer flag


jhiemstrawisc:
**Describe the bug**
`pelican origin token create --issuer "foobar.com"` should use `foobar.com` as its issuer, but I believe it ends up calling `config/config.go/GetServerIssuerURL()` instead. If I don't set `Server.IssuerUrl`, it always grabs my local hostname

**To Reproduce**
`pelican origin token create --issuer "foobar.com"`

**Expected behavior**
The generated token should have its issuer set to `foobar.com`

jhiemstrawisc:
Closed by #1135

PR #1135, opened 2024-04-18
Fix token creation CLI flags and add new key creation warning

### LINKED ISSUE #1092: `pelican origin token create` should use key from `--private-key` and not create new keys
A user I was helping recently reported that in Pelican v7.6.5, running:

```
pelican origin token create \
    --audience "https://wlcg.cern.ch/jwt/v1/any" \
    --scope "storage.read:/" \
    --subject "bar" \
    --lifetime "300" \
    --private-key "$PWD/issuer-key.pem"
```

Produces:
```
ERROR[2024-04-10T13:25:51-05:00] Fatal error occurred at the start of the program. Cleanup started: Cannot create token, failed to initialize configuration: Failed to load issuer private JWK: Failed to generate new private key: Private key file, </path/to/issuer-key.pem>, contains unsupported key type
```

against an EC256 key generated with the `scitokens-admin-create-key` tool.

Two things are wrong with this:
1. It would be helpful if the error message explained why it thinks the key type is unsupported.
2. It's concerning that the tool is trying to create a new key in the first place. This gets more tricky because I happen to know the token creation tool calls `InitServer()` under the hood. Perhaps we can take the input from `--private-key` and `viper.Set("Server.IssuerKey", "/path/to/key")` or something that points the tool to the correct key.

jhiemstrawisc:
Closes #1092 and #1091


cmd/origin_token.go
```diff
@@ -129,6 +130,13 @@ func cliTokenCreate(cmd *cobra.Command, args []string) error {
 		args = append(args, audSlice...)
 	}
 
+	// If a custom issuer is provided, we need to add it to the claims. If there is none, we'll end up falling back to
+	// the issuer generated by config.go/GetServerIssuerURL()
```
haoming29, line 134: Where do we implement the falling back? I didn't find it in the `origin_token` file

haoming29: After some digging I think I figured out the logic, and it looks good to me

haoming29 (approved):
LGTM!

cmd/origin_token.go
jhiemstrawisc: For the sake of leaving thorough breadcrumbs in case anyone else winds up asking this (or things break in the future), the fallback happens in `token/token_create.go` on the call to `tokenConfig.CreateTokenWithKey()`. If the token config is instantiated without an issuer, the token defaults to `config.GetServerIssuerURL()`.

PR #1138, opened 2024-04-19
Plugin include TransferUrl & TransferFileName

joereuss12:
Fixes up a very long error message from condor with the Plugin. This ensures that the plugin will return TransferUrl and TransferFileName even when empty so condor will be happy. Also fixes bug with error message for 404 no namespace found for path not returning correct error.

### DIFF
--- a/classads/classads.go
+++ b/classads/classads.go
@@ -21,7 +21,6 @@ package classads
 import (
 	"bufio"
 	"bytes"
-	"errors"
 	"fmt"
 	"io"
 	"strconv"
@@ -42,11 +41,11 @@ func NewClassAd() *ClassAd {
 // Get returns the value of the attribute with the given name.
 func (c *ClassAd) Get(name string) (interface{}, error) {
 	if c.attributes == nil {
-		return nil, errors.New("There are no attributes in the classad")
+		return nil, nil
 	} else if value, ok := c.attributes[name]; ok {
 		return value, nil
 	} else {
-		return nil, errors.New("Classad attribute not found")
+		return nil, nil
 	}
 }
 
--- a/classads/classads.go
+++ b/classads/classads.go
@@ -21,6 +21,7 @@ package classads
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"strconv"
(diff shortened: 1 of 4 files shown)


joereuss12:
@bbockelm see what you think of the changes from the last commit. I worked in some functionality to try to populate TransferUrl and TransferFileName in the case of failure and before we get results back. It might still need a tad of cleanup but I want to see if the functionality looks sound first. Here is an example of a hold message in the case the change covers:
> Transfer input files failure at execution point slot1_60@e2598.chtc.wisc.edu using protocol osdf. Details: Error from slot1_60@e2598.chtc.wisc.edu: Failed to transfer files: FILETRANSFER:1:non-zero exit (1) from stash_plugin. |Error: failed to get namespace information for remote URL osdf:///ospool/uc-shared/PUBLIC/OSG-Staff/validation/test.txt: 404: No namespace found for path. Either it doesn't exist, or the Director is experiencing problems ( URL file = osdf:///ospool/uc-shared/PUBLIC/OSG-Staff/validation/test.txt )|
        Code 13 Subcode 256
...

I know we have the URL twice at the moment but I think it is important to populate the (URL file = ...) field rather than filling it with `Not a URL-specific error` like in: https://opensciencegrid.atlassian.net/browse/HTCONDOR-2395 which could be confusing since this error does have to do with the URL provided.

Also, let me know if any more changes to the hold message should be made at this time in this PR as well.

bbockelm (changes requested):
The bones of the approach are good, just some nitpicks and requests for more unit tests.

One thing I spotted is the fact that `writeOutfile` calls `os.Exit` directly.  This prevents it from being easily unit-tested.  If you don't remove these calls as part of the addition of the requested unit tests, please make a follow-up ticket to clean that out.

jhiemstrawisc:
@joereuss12, it looks like you're chasing the pre-release rebase from all of the merges going on 😂 I'll start my review, but can you get this caught up in the meantime?

joereuss12:
Yes, will do lol. It's always my curse right up on release day

joereuss12:
@jhiemstrawisc luckily the merge conflicts were very easy to clean up. Should be all good now

jhiemstrawisc (approved):
LGTM

Issue #899, opened 2024-03-06
Casing restrictions


CannonLock:
Restrict parameters.yaml to PascalCase and API data to camelCase.

PR #1142, opened 2024-04-19
Purge topology origin servers when there is a duplicated server from Pelican

### LINKED ISSUE #1104: Remove topology serverAd if there's a Pelican advertisement with the same server name

haoming29:
Closes #1104 

The core idea of this PR is as follows:
* For all `serverAd`-s and their namespaces, we store as as-is, no matter they are from Topology or Pelican. We mark the ones from Topology with a new field `FromTopology`
* When direct received an object request, it will go over the `serverAds` and when doing match-making (related code is in `getAdsForPath` in `director/cache_ads.go`:
  * if origin candidate list is empty, set the matched origin server to the list, no matter from topology or not
  * if origin list is not empty, and the LAST item and the next matched origin ad are from the same source, simple add the next matched origin to the list
  * If they are different, then
    * If the next matched one is from topology, then ignore it (meaning existing candidates are all from Pelican)
    * If the next matched one is from Pelican, then remove the existing candidates and put the Pelican one into the list
  * For cache servers, as long as they matched the namespace, set them to the cache candidate list

What this PR hasn't covered yet, is the list of namespaces and their servers the director sends to the _cache_ for generating scitokens/authfile. We should do the same check ands filter out the duplicated topology server. Not sure if I should include this aspect in this PR or make it a separate one.

This PR also make sure we stable-sort the `serverAd` first in `getAdsForPath` and always generate the stable result (the current code gives random result depending on serverAds ordering of its `Items()` which is random ordering).

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -236,9 +236,6 @@ func redirectToCache(ginCtx *gin.Context) {
 	authzBearerEscaped := getAuthzEscaped(ginCtx.Request)
 
 	namespaceAd, originAds, cacheAds := getAdsForPath(reqPath)
-	log.Error("Origin Ads: ", server_structs.ServerAdsToServerNameURL(originAds))
-	log.Error("Cache Ads: ", server_structs.ServerAdsToServerNameURL(cacheAds))
-	log.Error("NamespaceAds: ", namespaceAd)
 	// if GetAdsForPath doesn't find any ads because the prefix doesn't exist, we should
 	// report the lack of path first -- this is most important for the user because it tells them
 	// they're trying to get an object that simply doesn't exist
--- a/director/director.go
+++ b/director/director.go
@@ -236,6 +236,9 @@ func redirectToCache(ginCtx *gin.Context) {
 	authzBearerEscaped := getAuthzEscaped(ginCtx.Request)
 
 	namespaceAd, originAds, cacheAds := getAdsForPath(reqPath)
+	log.Error("Origin Ads: ", server_structs.ServerAdsToServerNameURL(originAds))
+	log.Error("Cache Ads: ", server_structs.ServerAdsToServerNameURL(cacheAds))
+	log.Error("NamespaceAds: ", namespaceAd)
 	// if GetAdsForPath doesn't find any ads because the prefix doesn't exist, we should
 	// report the lack of path first -- this is most important for the user because it tells them
 	// they're trying to get an object that simply doesn't exist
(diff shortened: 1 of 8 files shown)


director/director.go
```diff
@@ -262,7 +262,7 @@ func redirectToCache(ginCtx *gin.Context) {
 			return
 		}
 	} else {
-		cacheAds, err = sortServers(ipAddr, cacheAds)
+		cacheAds, err = sortServerAdsOnIP(ipAddr, cacheAds)
```
jhiemstrawisc, line 265: This should probably be "ByIP"

server_structs/director.go
```diff
@@ -66,16 +67,17 @@ type (
...
+		AuthURL      url.URL    `json:"auth_url"`
+		BrokerURL    url.URL    `json:"broker_url"` // The URL of the broker service to use for this host.
+		URL          url.URL    `json:"url"`        // This is server's XRootD URL for file transfer
+		WebURL       url.URL    `json:"web_url"`    // This is server's Web interface and API
+		Type         ServerType `json:"type"`
+		Latitude     float64    `json:"latitude"`
+		Longitude    float64    `json:"longitude"`
+		Writes       bool       `json:"enable_write"`
+		DirectReads  bool       `json:"enable_fallback_read"` // True if reads from the origin are permitted when no cache is available
+		FromTopology bool       `json:"from_topology"`
```
jhiemstrawisc, line 80: It pains me that we have both `from_topology` and `from-topology`, but I understand why it has to be like that 😭

jhiemstrawisc (changes requested):
I think I already know the answer to this question, but I just want to confirm -- if an old origin/namespace advertises to this new director with an old server/namespace ad that doesn't have `FromTopology`, will the director automatically unmarshal the response and set that value to false? If that's the case, I think we're okay to modify the struct without worrying about hurting backwards compatibility.

server_structs/director.go
haoming29: and Cannon actually prefers to `fromTopology` #899 😂

haoming29:
> I think I already know the answer to this question, but I just want to confirm -- if an old origin/namespace advertises to this new director with an old server/namespace ad that doesn't have `FromTopology`, will the director automatically unmarshal the response and set that value to false? If that's the case, I think we're okay to modify the struct without worrying about hurting backwards compatibility.

The server advertisement should be fine, as we manually set the value (or let it be default value for Pelican server: `director/director.go line 612`), for namespace ad, I'm pretty sure if the incoming namespaceAd does not have `FromTopology`, the field is initialized to false by `encoding/json`, so there shouldn't be a compatibility issue.

jhiemstrawisc (approved):
LGTM. Do we need/want to patch this one into 7.7?

haoming29:
> LGTM. Do we need/want to patch this one into 7.7?

What would be a good question for @bbockelm

PR #1143, opened 2024-04-22
Set XRDHOST env var to server's DNS name for OSDF origins

### LINKED ISSUE #1110: OSDF mode should set `XRDHOST`
In the pure "Pelican" setup, the director will redirect Pelican caches to the `Server.ExternalWebUrl`.

By default, however, the legacy OSDF caches will use the xrootd protocol to contact the origin.  The xrootd protocol will redirect the cache to a hostname computed from the reverse DNS lookup of the public IP.

So, we can have the legacy caches use a different hostname from the Pelican caches, resulting in surprising failures.

The xrootd logic can be overridden by the `XRDHOST` environment variable for xrootd.  When OSDF mode is enabled, we should set that environment variable to the host name (removing the port) in the `ExternalWebUrl`.

jhiemstrawisc:
Legacy OSDF caches resolve origin hostnames via reverse DNS lookup, which can result in these caches finding a different hostname for origins than what would be used by Pelican caches. This sets the XRDHOST environment variable in OSDF origins to override that behavior.

Closes #1110

### DIFF
--- a/xrootd/xrootd_config.go
+++ b/xrootd/xrootd_config.go
@@ -660,6 +660,23 @@ func ConfigXrootd(ctx context.Context, origin bool) (string, error) {
 				return "", errors.New("Origin.Multiuser is set to `true` but the command was run without sufficient privilege; was it launched as root?")
 			}
 		}
+
+		// Legacy caches may attempt to reach out to the origin using the xroot protocol, which
+		// determines the origin's host via reverse DNS. Since we don't want that, we use the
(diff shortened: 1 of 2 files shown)


turetske (approved):
LGTM!

bbockelm:
@jhiemstrawisc -- one important change needed!

This should be the hostname _only_ -- http://foo.com isn't going to work.  You want foo.com.

Issue #712, opened 2024-01-29
Some tests `InitConfig()` without first isolating themselves to a testing dir.


jhiemstrawisc:
Tests should run in an environment isolated from the system they run on. However, some of our tests read values from `/etc/pelican/pelican.yaml` and use those values instead of the values we want the test to use.

For example, `TestDeprecateLogMessage()` in `config/config_test.go` has this snippet:
```
	t.Run("expect-deprecated-message-if-namespace-url-is-set", func(t *testing.T) {
		hook := test.NewGlobal()
		viper.Reset()
		// The default value is set to Error, but this is a warning message
		viper.Set("Logging.Level", "Warning")
		viper.Set("Federation.NamespaceUrl", "https://dont-use.com")
		InitConfig()
```

Without first setting `viper.Set("ConfigDir", <tmp test dir>)`, the call to `InitConfig()` also loads the values from `/etc/pelican/pelican.yaml`. When I have `Federation.RegistryUrl` set there, the test fails.

PR #1144, opened 2024-04-22
Isolate bevy of tests so they don't inherit from system pelican.yaml

jhiemstrawisc:
It has been frustrating me for a long time that a number of our unit tests were picking up values in my `pelican.yaml`, which was causing failures. Unit tests should really never look at the system pelican config, because the presumption is that each test is setting up its own configuration.

That being said, since the config directory is configured with viper, I couldn't immediately come up with a good way to uniformly enforce this across tests, or even with tests that toggle `viper.Reset()` between small test runs. For now, it's on us developers to remember that isolating our tests is a good practice.

Finally, there may be tests I missed. I found the tests covered in this PR by running `go test ./...` from the repo's root with an entire test federation configured in `/etc/pelican/pelican.yaml`. These were the tests that failed because of my system config.

Closees #712

### DIFF
--- a/client/handle_http_test.go
+++ b/client/handle_http_test.go
@@ -555,6 +555,7 @@ func TestNewPelicanURL(t *testing.T) {
 	t.Run("TestOsdfOrStashSchemeWithOSDFPrefixNoError", func(t *testing.T) {
 		viper.Reset()
 		_, err := config.SetPreferredPrefix(config.OsdfPrefix)
+		viper.Set("ConfigDir", t.TempDir())
 		assert.NoError(t, err)
 		// Init config to get proper timeouts
 		config.InitConfig()
@@ -583,6 +584,7 @@ func TestNewPelicanURL(t *testing.T) {
 	t.Run("TestOsdfOrStashSchemeWithOSDFPrefixWithError", func(t *testing.T) {
 		viper.Reset()
 		_, err := config.SetPreferredPrefix(config.OsdfPrefix)
+		viper.Set("ConfigDir", t.TempDir())
 		require.NoError(t, err)
 		config.InitConfig()
 		require.NoError(t, config.InitClient())
@@ -608,7 +610,7 @@ func TestNewPelicanURL(t *testing.T) {
 
 	t.Run("TestOsdfOrStashSchemeWithPelicanPrefixNoError", func(t *testing.T) {
 		viper.Reset()
-
+		viper.Set("ConfigDir", t.TempDir())
 		config.InitConfig()
 		require.NoError(t, config.InitClient())
 		te := NewTransferEngine(ctx)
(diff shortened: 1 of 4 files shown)


joereuss12 (changes requested):
Found a more instances of where we do an "InitConfig()" without setting a config dir:
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/director/discovery_test.go#L140-L144
A good amount of references in this one (I only included 1):
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/xrootd/xrootd_config_test.go#L53-L61
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/server_utils/registry_test.go#L40-L47
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/registry/registry_db_test.go#L744-L750
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/web_ui/ui_test.go#L92-L97
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/registry/client_commands_test.go#L185
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/registry/client_commands_test.go#L256
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/director/director_test.go#L669
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/director/stat_test.go#L392
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/token/token_create_test.go#L211
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/broker/token_utils_test.go#L34
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/director/origin_api_test.go#L84
3 cases in this file:
https://github.com/PelicanPlatform/pelican/blob/5442ed83b69ba85e7879cea87a405a78952fef25/web_ui/prometheus_test.go#L66

I just searched "InitConfig()" on github and went thru all the test files that call the function to find these. I don't think it'll be too hard to include all of them, just tedious.

web_ui/prometheus_test.go
```diff
@@ -169,7 +171,7 @@ func TestPrometheusProtectionOriginHeaderScope(t *testing.T) {
 
 	//Setup a private key and a token
 	viper.Set("IssuerKey", kfile)
-
+	viper.Set("ConfigDir", t.TempDir())
```
joereuss12, line 174: For some reason, occasionally t.TempDir() is cleaned up before a process is done using the directory. This is an issue in windows as windows does not allow the removal of files/directories if a process is using it. To fix this, simply replace the t.TempDir() with an os.MkDirTemp() and add a defer os.RemoveAll(<temp dir path>). The defer happens in a place windows typically likes.

joereuss12 (changes requested):

joereuss12 (approved):
LGTM

PR #1152, opened 2024-04-23
Populate registry institution options by institutionsUrl in field list

haoming29:
This PR is to populate the `options` field for the `admin_metadata.institution` field in the returned list of available registration fields in `/api/v1.0/registry_ui/namespaces`.

This PR also removes the existing `Institution` struct and TTL cache but instead to use `optionsCache` for caching the options given an optionsUrl, while maintaining backward-compatibility with the existing Institution parameters.

Finally, this PR enabled setting `description` for the hard-coded fields via the tag.

The PR will yield the following response (note that the institution options is truncated fo display)

`OPTION /api/v1.0/registry_ui/namespaces`

```json
[
    {
        "name": "prefix",
        "displayed_name": "Prefix",
        "type": "string",
        "required": true,
        "options": null,
        "description": ""
    },
    {
        "name": "pubkey",
        "displayed_name": "Pubkey",
        "type": "string",
        "required": true,
        "options": null,
        "description": "Pubkey is your origin's public key in JWKS form"
    },
    {
        "name": "admin_metadata.description",
        "displayed_name": "Description",
        "type": "string",
        "required": false,
        "options": null,
        "description": ""
    },
    {
        "name": "admin_metadata.site_name",
        "displayed_name": "Site Name",
        "type": "string",
        "required": false,
        "options": null,
        "description": ""
    },
    {
        "name": "admin_metadata.institution",
        "displayed_name": "Institution",
        "type": "string",
        "required": true,
        "options": [
            {
                "name": "Academia Sinica",
                "id": "https://osg-htc.org/iid/05bxb3784"
            },
            {
                "name": "Agricultural Research Service",
                "id": "https://osg-htc.org/iid/02d2m2044"
            },
            {
                "name": "Albert Einstein College of Medicine",
                "id": "https://osg-htc.org/iid/05cf8a891"
            },
            {
                "name": "American Museum of Natural History",
                "id": "https://osg-htc.org/iid/03thb3e06"
            }
        ],
        "description": ""
    },
    {
        "name": "admin_metadata.security_contact_user_id",
        "displayed_name": "Security Contact User Id",
        "type": "string",
        "required": false,
        "options": null,
        "description": "User Identifier of the user responsible for the security of the service"
    }
]
```


haoming29:
@CannonLock the PR is ready

launchers/registry_serve.go
```diff
@@ -41,11 +41,17 @@ func RegistryServe(ctx context.Context, engine *gin.Engine, egrp *errgroup.Group
...
-	if err != nil {
-		return err
+		err = registry.InitCustomRegistrationFields()
+		if err != nil {
+			return err
+		}
+
+		if err := registry.InitInstConfig(ctx, egrp); err != nil {
+			return err
+		}
```
jhiemstrawisc, line 54: For consistency, can you make the err handling in lines 47-50 the same as 52-54?
```
	if err := registry.InitCustomRegistrationFields(); err != nil {
			return err
	}
```

registry/custom_reg_fields.go
```diff
@@ -196,138 +190,48 @@ func checkUniqueOptions(options []registrationFieldOption) bool {
...
 // Initialize institutions list
 func InitInstConfig(ctx context.Context, egrp *errgroup.Group) error {
-	institutions := []Institution{}
+	institutions := []registrationFieldOption{}
 	if err := param.Registry_Institutions.Unmarshal(&institutions); err != nil {
 		log.Error("Fail to read Registry.Institutions. Make sure you had the correct format", err)
 		return errors.Wrap(err, "Fail to read Registry.Institutions. Make sure you had the correct format")
 	}
 
+	instRegIdx := -1
```
jhiemstrawisc, line 201: Can you add a comment explaining how you use/modify this index? What happens in line 231 if `registrationFields` is empty and `instRegIdx` isn't updated from -1?

registry/registry_db.go
```diff
@@ -64,8 +64,8 @@ type AdminMetadata struct {
 	UserID                string             `json:"user_id" post:"exclude"` // "sub" claim of user JWT who requested registration
 	Description           string             `json:"description"`
 	SiteName              string             `json:"site_name"`
-	Institution           string             `json:"institution" validate:"required"` // the unique identifier of the institution
-	SecurityContactUserID string             `json:"security_contact_user_id"`        // "sub" claim of user who is responsible for taking security concern
+	Institution           string             `json:"institution" validate:"required"`                                                                                // the unique identifier of the institution
```
jhiemstrawisc, line 67: It would be useful if you gave an example of what the unique identifier looks like in this comment. Is it a string name? Is it a random number? Is it a hash?

jhiemstrawisc (changes requested):
Couple of minor tweaks, but otherwise LGTM

registry/registry_db.go
haoming29: The type of the value really depends on whatever provided by `Registry.Institutions` or `Registry.InstitutionsUrl`, so I updated the comment to reflect this.

jhiemstrawisc (approved):
LGTM

haoming29:
> @haoming29 I assume I was added as reviewer just so I could track when this got merged? Would you like me to review as well?

Yes! But also feel free reviewing it and see if this is what you expect

haoming29:
I'm going to merge this in once the ubuntu test is passed, the current main branch had a test failure and I added the fix in this PR.

haoming29:
> I'm going to merge this in once the ubuntu test is passed, the current main branch had a test failure and I added the fix in this PR.

The current ubuntu test failed at the `test base code` which is the current main branch and that's what's expected to fail, so I'll go ahead and merge

PR #1154, opened 2024-04-24
Don't include scheme when setting XRDHOST for osdf origins

jhiemstrawisc:
PR #1143 had a mistake in that it set a scheme for the XRDHOST env var. This PR removes the scheme and updates the tests to check the correct thing.


turetske (approved):
LGTM

PR #1156, opened 2024-04-24
Plugin recursive download errors returned properly

joereuss12:
There was a bug with the plugin when doing recursive downloads. If some error happened such as a 404 on the directory we are requesting or a 500 which is doing the ?recursive query on a file instead of a directory, PROPFIND would fail and the error was not making it back to the plugin. Now, when we discover this error (eventually makes its way to RunMux() in client/handle_http.go) we populate the results channel with the failed result before closing it. This way, the plugin can see the results and record the failure.

In addition, I added some unit tests toward these failures as well as for recusive downloads for plugin in general.

To test: run `stash_plugin pelican://path/to/some/file?recursive=true .` and see that we get classads returned that record a TranferError as well as other classads. Feel free to run the unit tests as well as any other hand tests that you feel might break this.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -944,10 +944,6 @@ func (te *TransferEngine) runMux() error {
 			// If no transfers were created and we have an error, the job is no
 			// longer active
 			if job.job.lookupErr != nil && job.job.totalXfer == 0 {
-				// Make sure we update results of this error
-				log.Errorln(job.job.lookupErr)
-				results := &TransferResults{jobId: job.job.uuid, job: job.job, Error: job.job.lookupErr, Scheme: job.job.remoteURL.Scheme}
-				te.resultsMap[job.uuid] <- results
 				// Remove this job from the list of active jobs for the client.
 				activeJobs[job.uuid] = slices.DeleteFunc(activeJobs[job.uuid], func(oldJob *TransferJob) bool {
 					return oldJob.uuid == job.job.uuid
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -944,6 +944,10 @@ func (te *TransferEngine) runMux() error {
 			// If no transfers were created and we have an error, the job is no
 			// longer active
 			if job.job.lookupErr != nil && job.job.totalXfer == 0 {
+				// Make sure we update results of this error
+				log.Errorln(job.job.lookupErr)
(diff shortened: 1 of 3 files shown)


cmd/plugin_test.go
```diff
@@ -362,6 +362,175 @@ func TestPluginMulti(t *testing.T) {
...
+// Test recursive downloads from the plugin
+func TestPluginRecursiveDownload(t *testing.T) {
+	viper.Reset()
+	server_utils.ResetOriginExports()
+
+	dirName := t.TempDir()
+
+	viper.Set("Logging.Level", "debug")
+	viper.Set("Origin.StorageType", "posix")
+	viper.Set("Origin.ExportVolumes", "/test")
```
jhiemstrawisc, line 374: This really isn't how the `ExportVolumes` is meant to be used, and it is malformed because export volumes are meant to contain both a storage and a federation prefix in posix. If you're trying to set up a federation prefix, you should use `Origin.FederationPrefix`, and if this is meant to be a storage prefix, you should use `Origin.StoragePrefix`. The export volumes variable is really only meant to be configured via env vars or bound to the command line `-v` when serving origins.

jhiemstrawisc (changes requested):

client/handle_http.go
```diff
@@ -926,6 +926,10 @@ func (te *TransferEngine) runMux() error {
 			// If no transfers were created and we have an error, the job is no
 			// longer active
 			if job.job.lookupErr != nil && job.job.totalXfer == 0 {
+				// Make sure we update results of this error
+				log.Errorln(job.job.lookupErr)
+				results := &TransferResults{jobId: job.job.uuid, job: job.job, Error: job.job.lookupErr, Scheme: job.job.remoteURL.Scheme}
+				te.resultsMap[job.uuid] <- results
```
bbockelm, line 932: We can't do blocking operations in the runMux loop.  That will lead to deadlocks, no?

I'd need to page the specific knowledge back in but I recall some issues in having non-transfer results sent in the transfer result channel.

Can the plugin check for the error condition when the channel closes perhaps?

bbockelm (changes requested):
See question inline.

client/handle_http.go
joereuss12: Do you have any suggestions for how to have the plugin check for this error condition? Just not sure how I can check for error conditions if the channel is already closed and the context has no errors attached to it. Should I maybe have something for the client and check for a tc.Error or something?

jhiemstrawisc:
Merging, as it looks like the requested changes have all been made.

PR #1157, opened 2024-04-24
Add support for checking a sentinel file for the cache

### LINKED ISSUE #1056: Add a sentinel file for the cache xrootd
In #1003 we added the option to provide a sentinel file, and it would be nice if we can do the same thing for the cache server as well.

haoming29:
Closes #1056

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -461,7 +461,7 @@ description: |+
   - SentinelLocation: A filename under `StoragePrefix` path for Pelican to check the storage directory exists and is correctly mounted.
       The value must be a file and contain no directory. Leave it empty to skip the check.
 
-      You should always choose a distinguished name for `SentinelLocation`. It should not be reused for other servers.
+      You should always choose a distinct name for `SentinelLocation`. It should not be reused for other servers.
       If running in a containerized environment it should not be the name of the underlying physical host as that may change and lead to confusion.
       You need to manually create a file under path to `StoragePrefix` with the same name as `SentinelLocation`. Note that this parameter will be ignored
       if the origin StorageType is S3.
@@ -946,7 +946,7 @@ description: |+
   A filename under `Cache.DataLocation` path for Pelican to check the storage directory exists and is correctly mounted.
   The value must be a file and contain no directory. Leave it empty to skip the check.
 
-  You should always choose a distinguished name for `Cache.SentinelLocation`. It should not be reused for other servers.
+  You should always choose a distinct name for `Cache.SentinelLocation`. It should not be reused for other servers.
   If running in a containerized environment it should not be the name of the underlying physical host as that may change and lead to confusion.
   You need to manually create a file under path to `Cache.DataLocation` with the same name as `Cache.SentinelLocation`.
 type: filename
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -459,7 +459,9 @@ description: |+
       where each of these has the same effect as the corresponding "Origin.Enable*" configuration, except scoped to the
       given export. If "PublicReads" is included, "Reads" is inferred.
   - SentinelLocation: A filename under `StoragePrefix` path for Pelican to check the storage directory exists and is correctly mounted.
-      Leave it empty to skip the check. You should always choose a distinguished name for `SentinelLocation`. It should not be reused for other servers.
+      The value must be a file and contain no directory. Leave it empty to skip the check.
+
+      You should always choose a distinguished name for `SentinelLocation`. It should not be reused for other servers.
       If running in a containerized environment it should not be the name of the underlying physical host as that may change and lead to confusion.
       You need to manually create a file under path to `StoragePrefix` with the same name as `SentinelLocation`. Note that this parameter will be ignored
       if the origin StorageType is S3.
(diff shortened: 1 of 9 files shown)


docs/parameters.yaml
```diff
@@ -459,7 +459,9 @@ description: |+
       where each of these has the same effect as the corresponding "Origin.Enable*" configuration, except scoped to the
       given export. If "PublicReads" is included, "Reads" is inferred.
   - SentinelLocation: A filename under `StoragePrefix` path for Pelican to check the storage directory exists and is correctly mounted.
-      Leave it empty to skip the check. You should always choose a distinguished name for `SentinelLocation`. It should not be reused for other servers.
+      The value must be a file and contain no directory. Leave it empty to skip the check.
+
+      You should always choose a distinguished name for `SentinelLocation`. It should not be reused for other servers.
```
turetske, line 464:
```suggestion
      You should always choose a distinct name for `SentinelLocation`. It should not be reused for other servers.
```

docs/parameters.yaml
```diff
@@ -939,6 +941,18 @@ root_default: /run/pelican/xrootd/cache
 default: $XDG_RUNTIME_DIR/pelican/cache
 components: ["cache"]
 ---
+name: Cache.SentinelLocation
+description: |+
+  A filename under `Cache.DataLocation` path for Pelican to check the storage directory exists and is correctly mounted.
+  The value must be a file and contain no directory. Leave it empty to skip the check.
+
+  You should always choose a distinguished name for `Cache.SentinelLocation`. It should not be reused for other servers.
```
turetske, line 949:
```suggestion
  You should always choose a distinct name for `Cache.SentinelLocation`. It should not be reused for other servers.
```

server_utils/origin.go
```diff
@@ -492,7 +492,8 @@ from S3 service URL. In this configuration, objects can be accessed at /federati
 	return originExports, nil
 }
 
-func CheckSentinelLocation(exports []OriginExport) (ok bool, err error) {
+// Check the sentinel files from Origin.Exports
+func CheckOriginSentinelLocations(exports []OriginExport) (ok bool, err error) {
```
turetske, line 496:
```suggestion
func CheckOriginSentinelLocation(exports []OriginExport) (ok bool, err error) {
```

launchers/launcher.go
```diff
@@ -162,7 +162,7 @@ func LaunchModules(ctx context.Context, modules config.ServerType) (servers []se
 			return
 		}
 
-		ok, err = server_utils.CheckSentinelLocation(originExports)
+		ok, err = server_utils.CheckOriginSentinelLocations(originExports)
```
turetske, line 165:
```suggestion
		ok, err = server_utils.CheckOriginSentinelLocation(originExports)
```

server_utils/origin_test.go
```diff
@@ -437,7 +437,7 @@ func TestCheckSentinelLocation(t *testing.T) {
 		exports = append(exports, mockExportNoStn)
 		exports = append(exports, mockExportNoStn)
 
-		ok, err := CheckSentinelLocation(exports)
+		ok, err := CheckOriginSentinelLocations(exports)
```
turetske, line 440:
```suggestion
		ok, err := CheckOriginSentinelLocation(exports)
```

server_utils/origin_test.go
```diff
@@ -447,7 +447,7 @@ func TestCheckSentinelLocation(t *testing.T) {
 		exports = append(exports, mockExportNoStn)
 		exports = append(exports, mockExportValidStn)
 
-		ok, err := CheckSentinelLocation(exports)
+		ok, err := CheckOriginSentinelLocations(exports)
```
turetske, line 450:
```suggestion
		ok, err := CheckOriginSentinelLocation(exports)
```

server_utils/origin_test.go
```diff
@@ -458,7 +458,7 @@ func TestCheckSentinelLocation(t *testing.T) {
 		exports = append(exports, mockExportValidStn)
 		exports = append(exports, mockExportInvalidStn)
 
-		ok, err := CheckSentinelLocation(exports)
+		ok, err := CheckOriginSentinelLocations(exports)
```
turetske, line 461:
```suggestion
		ok, err := CheckOriginSentinelLocation(exports)
```

launchers/cache_serve.go
```diff
@@ -51,6 +54,21 @@ func CacheServe(ctx context.Context, engine *gin.Engine, egrp *errgroup.Group, m
 		return nil, err
 	}
 
+	// Check for the sentinel file
+	if param.Cache_SentinelLocation.IsSet() {
+		sentinelPath := param.Cache_SentinelLocation.GetString()
```
turetske, line 59: This seems redundant. Can the server_utils function be made more general to check both the cache and origin sentinel locations?

turetske (changes requested):
Can you add testing to check the cache sentinel location as well? Something similar to what's in origin_test.go?

server_utils/origin.go
haoming29: The `s` was intentional as we are checking multiple locations for different prefixes

launchers/cache_serve.go
haoming29: The first draft (not published) was to take advantage of the `server_utils` function and make them share the same core logic, but I figured the real "core" logic is just a `os.Stat` call. Given that the error message has different format for cache and origin, and that origin and cache reads from different parameters, I figured I'd be ok to have a separate function to handle them

turetske (approved):

Issue #1123, opened 2024-04-15
Deprecate OSDF_ environment variables


brianhlin:
Pelican Service:
- [X] Client
- [X] Plugin
- [X] Registry
- [X] Director
- [X] Origin
- [X] Cache
- [ ] Other (please give the detail)

**Is your feature request related to a problem? Please describe.**
It would be nice to have a single env var configuration interface for operators. Right now, we have some [heinous looking](https://github.com/opensciencegrid/tiger-osg-config/blob/master/manifests/base/osdf-pelican-origin/osdf-env-vars.yaml) env var references that only work for a subset of configuration values so that operators only have to bother setting `PELICAN_*` variables.

**Describe the solution you'd like**
Allow the OSDF binary to respect the `PELICAN_*` env vars and deprecate the `OSDF_*` env vars.

PR #1160, opened 2024-04-24
Deprecate OSDF/Stash prefix and use Pelican ones

### LINKED ISSUE #954: osdf binary should respect 'PELICAN_*' env vars
Pelican Service:
- [ ] Client
- [ ] Plugin
- [X] Registry
- [X] Director
- [X] Origin
- [X] Cache

In Kubernetes, we constructing different layers of generic Pelican vs OSDF services where the latter is based on the former. To provide a consistent config interface to the operator, we're having them just use `PELICAN_*` env vars. To do this, we have to do a fair amount of hackery with kustomize patching that only really works for the variables that we explicitly list in our hack:

For example, in the `osdf-origin` base, we have

```
patches:
  - path: deployment-patch.yaml
  # Vars need to be defined ahead of any references so we use a
  # JSON6902 style patch to make sure they're appended to the list
  - path: osdf-env-vars.yaml
    target:
      group: apps
      version: v1
      kind: Deployment
      name: pelican-origin
      labelSelector: "app=pelican-origin,federation=osdf"
```

And in `osdf-env-vars.yaml`:

```
---
- op: add
  path: /spec/template/spec/containers/0/env/-
  value:
    name: "OSDF_FEDERATION_DISCOVERYURL"
    value: "$(PELICAN_FEDERATION_DISCOVERYURL)"
- op: add
  path: /spec/template/spec/containers/0/env/-
  value:
    name: "OSDF_SERVER_WEBPORT"
    value: "$(PELICAN_SERVER_WEBPORT)"
<snip>
```

We could eliminate a lot of the above complexity if the `osdf` binary respected `PELICAN_*` style environment variables.

haoming29:
Closes #954 and #1123 

Since Viper only seems to support single env prefix, this PR has a workaround that set env with OSDF/Stash prefix to PELICAN prefix (based on the binary alias name) and uses `PELICAN` as the prefix internally.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -844,7 +844,9 @@ func InitConfig() {
 		viper.SetConfigName("pelican")
 	}
 
-	viper.SetEnvPrefix(string(prefix))
+	osdfEnvToPelican() // Deprecate OSDF env prefix but be compatible for now
+
+	viper.SetEnvPrefix("pelican")
 	viper.AutomaticEnv()
 	// This line allows viper to use an env var like ORIGIN_VALUE to override the viper string "Origin.Value"
 	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
--- a/config/env.go
+++ b/config/env.go
@@ -1,0 +1,53 @@
+/***************************************************************
+ *
+ * Copyright (C) 2024, Pelican Project, Morgridge Institute for Research
+ *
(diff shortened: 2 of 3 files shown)


haoming29:
After contemplating on the implementation, I don't like the approach I was taking so I will re-implement this with a separate PR.

PR #1162, opened 2024-04-25
Deprecate OSDF/STASH environment prefix and use PELICAN

### LINKED ISSUE #1123: Deprecate OSDF_ environment variables

haoming29:
Closes #1123 #954 

This PR makes it possible to take both OSDF/STASH prefix and PELICAN prefix for `osdf` / `stash` binary.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -844,7 +844,9 @@ func InitConfig() {
 		viper.SetConfigName("pelican")
 	}
 
-	viper.SetEnvPrefix(string(prefix))
+	osdfEnvToPelican() // Deprecate OSDF env prefix but be compatible for now
+
+	viper.SetEnvPrefix("pelican")
 	viper.AutomaticEnv()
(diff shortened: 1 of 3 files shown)


jhiemstrawisc (approved):
This one looks like it was fun to write :) I like it!

Issue #1165, opened 2024-04-25
Remove the need for =true for ?recursive query


joereuss12:
Pelican Service:
- [X] Client
- [X] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

When implementing ?directread query param, we discovered that we do not actually need a value set for the query and can just check if the query exists. This should be implemented for ?recursive for the Pelican plugin so the need for =true is not needed.

PR #1166, opened 2024-04-25
Implement client direct reads

### LINKED ISSUE #516: Allow cache bypass based on query parameters
For the NRAO run, we hit an issue where the CHTC cache was busy so it was fairly slow in delivering some one-time-use files to the 

The client should have an extension (much like we added the `pack=auto` query parameters) which provides it with the ability to download straight from the origin if allowed.

The work should interact with @jhiemstrawisc to determine the policy language with the origins on, exactly, how they indicate they are willing to serve direct clients in addition to caches.  As a starting point, I'm OK with assuming that an origin is OK with a direct client for non-public cases (as that gives the origin the ability to revoke access as needed).

joereuss12:
This implements direct read functionality within the client. Adds a query parameter to add to the transfer url ?directread. When this is set, the client will avoid using a cache and go directly to the origin for its downloads. This PR also adds some unit tests for this functionality with a few bug fixes along the way. This should work with both the client and the plugin as well as with the ?pack and ?recursive query.

Closes #516


cmd/plugin_test.go
```diff
@@ -475,6 +539,15 @@ func TestQuery(t *testing.T) {
 		assert.NoError(t, err)
 	})
 
+	t.Run("TestValidDirectRead", func(t *testing.T) {
+		transferStr := "pelican://something/here?directread"
```
jhiemstrawisc, line 543: Similar to your other tests, the comments above this test says more than one query should fail. After you consolidate the two `checkValidQuery()`, you can delete the duplicate tests and move what's left into utils.

client/fed_test.go
```diff
@@ -505,3 +505,78 @@ func TestStatHttp(t *testing.T) {
...
+			// Set path for object to upload/download
+			tempPath := tempFile.Name()
+			fileName := filepath.Base(tempPath)
+			uploadURL := fmt.Sprintf("pelican://%s:%s%s/%s?directread", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()),
+				export.FederationPrefix, fileName)
+
+			// Download the file with GET. Shouldn't need a token to succeed
+			transferResults, err := client.DoGet(fed.Ctx, uploadURL, t.TempDir(), false)
+			assert.NoError(t, err)
+			if err == nil {
```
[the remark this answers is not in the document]
joereuss12: Here we do because even if assert fails for assert.NoError, we still move on to the assert.Equal after. This is bad because if we fail, then transferResults is nil and we get a nil pointer dereference. That is why we want to do the if err == nil here

client/director.go
```diff
@@ -112,7 +112,19 @@ func CreateNsFromDirectorResp(dirResp *http.Response) (namespace namespaces.Name
...
+	sourceUrl, err := url.Parse(sourceUrlStr)
+	if err != nil {
+		return
+	}
+	if sourceUrl.RawQuery != "" {
+		source = sourceUrl.Path + "?" + sourceUrl.RawQuery
+	} else {
+		source = sourceUrl.Path
+	}
+
```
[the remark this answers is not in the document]
joereuss12: The issue with that I found is doing sourceUrl.String() includes the scheme (`i.e. pelican://aaa4a774c20c:8444/test/foo.txt?directread`). We are only interested in the path and the query to add to the url we query the director with. If you have any other suggestions for better ways to do it feel free to let me know but this is the way I found to remove the scheme from the url.

client/fed_test.go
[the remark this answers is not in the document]
jhiemstrawisc: Maybe do a `require.NoError` instead?

client/director.go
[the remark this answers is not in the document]
jhiemstrawisc: I think the solution is to only paths in `sourceUrlStr`. If you do that, parsing it will result in an empty scheme and host.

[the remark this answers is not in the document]
jhiemstrawisc: I think the solution is to only paths in sourceUrlStr (and maybe change the variable name to indicate that these should only be paths). If you do that, parsing it as a URL will result in an empty scheme and host.

jhiemstrawisc (changes requested):

utils/utils.go
```diff
@@ -51,3 +53,28 @@ func SnakeCaseToHumanReadable(input string) string {
...
+
+// This function checks if we have a valid query (or no query) for the transfer URL
+func CheckValidQuery(transferUrl *url.URL, isPlugin bool) (err error) {
+	query := transferUrl.Query()
+	_, hasRecursive := query["recursive"]
+	_, hasPack := query["pack"]
+	_, hasDirectRead := query["directread"]
+
+	// If we are not the plugin, we should not use ?recursive (we should pass a -r flag)
+	if !isPlugin && hasRecursive {
```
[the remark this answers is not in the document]
joereuss12: Will be added in #1165, for now, since we're kinda close to release and this PR is already pretty big, I think it is fine if we throw an error.

utils/utils.go
```diff
@@ -77,22 +76,27 @@ func CheckValidQuery(transferUrl *url.URL, isPlugin bool) (err error) {
 	query := transferUrl.Query()
 	_, hasRecursive := query["recursive"]
 	_, hasPack := query["pack"]
-	_, hasDirectRead := query["directread"]
+	directRead, hasDirectRead := query["directread"]
 
-	// If we are not the plugin, we should not use ?recursive (we should pass a -r flag)
+	// // If we are not the plugin, we should not use ?recursive (we should pass a -r flag)
```
jhiemstrawisc, line 81: Can you remove double comment?

utils/utils.go
```diff
@@ -77,22 +76,27 @@ func CheckValidQuery(transferUrl *url.URL, isPlugin bool) (err error) {
...
 	if !isPlugin && hasRecursive {
-		return fmt.Errorf("Cannot use the recursive query parameter when not utilizing the pelican plugin")
+		return errors.New("cannot use the recursive query parameter when not utilizing the pelican plugin")
```
jhiemstrawisc, line 83: On second thought, let's say "recursive query parameter not yet supported in URLs outside of plugin" or something to that effect. That tells the user it's coming, but not yet ready.

My second comment is that we need an issue to keep track of this. Can you open an issue and link it to this PR?

utils/utils.go
```diff
@@ -77,22 +76,27 @@ func CheckValidQuery(transferUrl *url.URL, isPlugin bool) (err error) {
...
 
 	// If we have both recursive and pack, we should return a failure
 	if hasRecursive && hasPack {
-		return fmt.Errorf("Cannot have both recursive and pack query parameters")
+		return errors.New("cannot have both recursive and pack query parameters")
+	}
+
+	// If we have both recursive and pack, we should return a failure
+	if hasDirectRead && directRead[0] != "" {
+		return errors.New("directread query parameter should not have any values assigned to it")
```
jhiemstrawisc, line 93: If we're changing this behavior, we should make sure things are backwards compatible for at least the next release or two, but with a clear indication that the old usage is deprecated and will be removed. For now, rather than throwing an error like this, we want a message that says something like "Arguments (true/false) for the `directread` query parameter have been deprecated and will be disallowed in a future release." Then in 7.9 we can flip the switch and make this a true error.

jhiemstrawisc (changes requested):

utils/utils.go
joereuss12: This is the issue, I added it to it https://github.com/PelicanPlatform/pelican/issues/1165

The issue includes this PR and I have notes to modify this function as well as the tests. Anything else I need to do?

utils/utils.go
joereuss12: Should we mention in the message that the argument will be ignored as well?

utils/utils.go
jhiemstrawisc: Nope, thanks for linking it :)

jhiemstrawisc (approved):

Issue #1107, opened 2024-04-12
Origin registration


bbockelm:
Now that we've released multi-exports -- and are starting to get more serious about multiple origins per prefix -- it's about time to separate the concept of a "namespace registration" and "origin registration".

I think we should use the same "trick" for origins as caches -- register them by the hostname inside `/origins` (can't recall if we settled on `/cache` or `/caches` as a prefix when registering caches; we should keep both singular or both plural).

- When starting up, the origin should register all its namespaces and itself.
- There should be tooling or a web UI element for the origin registration so the identity-based workflow can be all web-based if needed.  This can be done as part of this issue or #1105.
- Adjustments should be made to the registration UI so there are three clearly separate sections -- prefixes, origins, and caches.

haoming29:
Closes #623 as it's a duplicated issue

haoming29:
@CannonLock  This is not finished yet. Reopening it

PR #1176, opened 2024-04-26
Improve site name handling

### LINKED ISSUE #1106: Improve "site name" handling
The "site name" is currently a fairly nebulous concept.  Let's start fixing up semantics.

0. The site name should not be required for a prefix registration.  It should be required for the cache registration - and we'll have an origin registration in the future - but there's not much point for the prefix.
1. If the "site name" is not set in the configuration - but it is filled in in the cache or (in the future) origin registration - use that.
2.  In "pelican mode", the "site name" should default to the hostname in the external web URL, not the result of `os.Hostname`.  This should be a more sane default in container-based systems.
3. In "OSDF mode", the registration page should provide a drop-down populated with existing OSDF site names.

haoming29:
Closes #1106 

Some points to note (copied from the ticket above):

* We need to have https://github.com/PelicanPlatform/pelican/issues/1107 ASAP; otherwise, with multi-export support at the origin, we can't wisely determine which namespace to fetch the site name from the registry. Currently, this PR only grab the first exported namespace as the prefix to query the site name at the registry, which is a bit sketchy.
* ~~We need to add an extra serverAd field Hostname to the director (and during advertisement), with this change.~~ When implementing this PR for caches, the director fails to verify the advertisement token from the cache because it relies on the `Name` field from the `serverAd` to find the namespace for the cache and grab the public jwks from the registry. If the Name is different from the hostname of the cache (which is used to register the namespace for the cache at the registry), the registry won't be able to find the namespace registration for the cache.  This PR took the approach that instead of adding a new field, the director now concatenates the namespace prefix for the cache with a new value, by extracting the hostname from `DataURL` or `WebURL` field in `OriginServerAd`

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -566,12 +566,12 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 	ctx.Set("serverName", adV2.Name)
 	ctx.Set("serverWebUrl", adV2.WebURL)
 
-	ad_url, err := url.Parse(adV2.DataURL)
+	adUrl, err := url.Parse(adV2.DataURL)
 	if err != nil {
 		log.Warningf("Failed to parse %s URL %v: %v\n", sType, adV2.DataURL, err)
 		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
 			Status: server_structs.RespFailed,
-			Msg:    fmt.Sprintf("Invalid %s registration", sType),
+			Msg:    fmt.Sprintf("Invalid %s registration. DataURL %s is not a valid URL", sType, adV2.DataURL),
 		})
 		return
 	}
@@ -581,7 +581,7 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 		log.Warningf("Failed to parse server Web URL %v: %v\n", adV2.WebURL, err)
 		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
 			Status: server_structs.RespFailed,
-			Msg:    "Invalid server Web URL",
+			Msg:    fmt.Sprintf("Invalid %s registration. WebURL %s is not a valid URL", sType, adV2.WebURL),
 		})
 		return
 	}
@@ -591,7 +591,7 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 		log.Warningf("Failed to parse broker URL %s: %s", adV2.BrokerURL, err)
 		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
 			Status: server_structs.RespFailed,
-			Msg:    "Invalid broker URL",
+			Msg:    fmt.Sprintf("Invalid %s registration. BrokerURL %s is not a valid URL", sType, adV2.BrokerURL),
 		})
 	}
 
@@ -627,7 +627,7 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 	} else {
 		token := strings.TrimPrefix(tokens[0], "Bearer ")
 		// Use hostname from adv2.DataUrl instead of adv2.Name as Name is the site name
-		hostname := ad_url.Hostname()
+		hostname := adUrl.Hostname()
 
 		prefix := path.Join("/caches", hostname)
 		ok, err := verifyAdvertiseToken(engineCtx, token, prefix)
@@ -657,7 +657,7 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 
 	sAd := server_structs.ServerAd{
 		Name:        adV2.Name,
-		URL:         *ad_url,
+		URL:         *adUrl,
 		WebURL:      *adWebUrl,
 		BrokerURL:   *brokerUrl,
 		Type:        sType,
--- a/director/director.go
+++ b/director/director.go
@@ -566,6 +566,35 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 	ctx.Set("serverName", adV2.Name)
 	ctx.Set("serverWebUrl", adV2.WebURL)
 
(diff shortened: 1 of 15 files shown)


director/director.go
```diff
@@ -566,6 +566,35 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
...
+		})
+		return
+	}
+
+	brokerUrl, err := url.Parse(adV2.BrokerURL)
+	if err != nil {
+		log.Warningf("Failed to parse broker URL %s: %s", adV2.BrokerURL, err)
+		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    "Invalid broker URL",
```
jhiemstrawisc, line 594: Here as well

launcher_utils/advertise.go
```diff
@@ -103,14 +105,73 @@ func Advertise(ctx context.Context, servers []server_structs.XRootDServer) error
...
 }
 
+// Get the site name from the registry given a namespace prefix
+func getSitenameFromReg(ctx context.Context, prefix string) (sitename string, err error) {
+	fed, err := config.GetFederation(ctx)
+	if err != nil {
+		return
+	}
+	if fed.NamespaceRegistrationEndpoint == "" {
+		err = fmt.Errorf("unable to fetch site name from the registry. Federation.RegistryUrl is unset")
```
jhiemstrawisc, line 115: This error message could be confusing to someone who uses `Federation.DiscoveryUrl` instead of directly setting the registry/director URLs. Can you modify it to take that case into account?

launcher_utils/advertise_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/pelicanplatform/pelican/config"
+	"github.com/pelicanplatform/pelican/server_structs"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
jhiemstrawisc, line 33: Can you split the pelican/non pelican imports here?

launcher_utils/advertise_test.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+func TestGetSitenameFromReg(t *testing.T) {
+	t.Run("no-registry-url", func(t *testing.T) {
```
jhiemstrawisc, line 37: Can you add deferred resets to the beginning of the test for anything that has the potential to affect a later test that forgets to reset at the start?

registry/registry_test.go
```diff
@@ -21,12 +21,14 @@ package registry
...
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
 	"github.com/gin-gonic/gin"
 	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/pelicanplatform/pelican/server_structs"
```
jhiemstrawisc, line 31: Same here -- can you split the pelican/non-pelican imports?

registry/registry_ui_test.go
```diff
@@ -40,6 +40,7 @@ import (
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/pelicanplatform/pelican/config"
 	"github.com/pelicanplatform/pelican/param"
+	"github.com/pelicanplatform/pelican/server_structs"
```
jhiemstrawisc, line 43: Here too

registry/registry_validation_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	"github.com/jellydator/ttlcache/v3"
+	"github.com/pelicanplatform/pelican/server_structs"
```
jhiemstrawisc, line 25: And here

jhiemstrawisc (changes requested):
All of the requested changes should be fairly minor. Otherwise LGTM

director/director.go
```diff
@@ -566,12 +566,12 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
...
 	ctx.Set("serverWebUrl", adV2.WebURL)
 
-	ad_url, err := url.Parse(adV2.DataURL)
+	adUrl, err := url.Parse(adV2.DataURL)
 	if err != nil {
 		log.Warningf("Failed to parse %s URL %v: %v\n", sType, adV2.DataURL, err)
 		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
 			Status: server_structs.RespFailed,
-			Msg:    fmt.Sprintf("Invalid %s registration", sType),
+			Msg:    fmt.Sprintf("Invalid %s registration. DataURL %s is not a valid URL", sType, adV2.DataURL),
```
jhiemstrawisc, line 574: I don't think `DataURL` is something we expose to caches and origins, is it? I forget exactly how we bootstrap this URL, but I think we can use the error message to point the user to whatever configuration options they need to change to get this working. I know we have `Origin.Url/Cache.Url`, but we probably also set those to some defaults when they're not directly configured.

director/director.go
```diff
@@ -581,7 +581,7 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 		log.Warningf("Failed to parse server Web URL %v: %v\n", adV2.WebURL, err)
 		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
 			Status: server_structs.RespFailed,
-			Msg:    "Invalid server Web URL",
+			Msg:    fmt.Sprintf("Invalid %s registration. WebURL %s is not a valid URL", sType, adV2.WebURL),
```
jhiemstrawisc, line 584: Same here, if we can point the user to the actual configuration knobs at their server they need to adjust to fix the invalid registration, we'll empower them to fix it themselves.

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM

PR #1179, opened 2024-04-26
Fix wrong type for integer values in requests with customized registration fields at the registry

haoming29:
As @CannonLock found out, the current validation for customized registration fields fails on `integer` type values, since Go will by default marshal JSON numbers to `float64` type. This PR converts any customized fields with `float64` or `float32` type to `int` type, which should be a relatively safe assumption as we currently don't support float-time customized fields. 

This PR also improves test coverage for the registration to cover this case.


jhiemstrawisc (approved):
LGTM

PR #1182, opened 2024-04-26
Add Custom Registration UI Elements

CannonLock:


CannonLock:
Waiting for the API fixes to be merged so we can test efficiently.

haoming29:
@CannonLock  Just so you know, the related revisions to the APIs are merged to main

CannonLock:
@haoming29 

Good array of inputs to test with. 

```
registry:
    customregistrationfields:
        - description: Cool Boolean Description Required
          name: Boolean Field
          options: []
          optionurl: ""
          required: true
          type: bool
          validationurl: ""
        - description: String Not required
          name: String
          options: []
          optionurl: ""
          required: false
          type: string
          validationurl: ""
        - description: Datetime Required
          name: Datetime
          options: []
          optionurl: ""
          required: true
          type: datetime
          validationurl: ""
        - description: InstitutionEnum
          name: Enum Institution
          options: []
          optionsurl: https://topology-institutions.osg-htc.org/api/institution_ids
          required: true
          type: enum
          validationurl: ""
        - description: asdf
          name: Test.Period
          options: []
          optionsurl: ""
          required: false
          type: bool
          validationurl: ""
```

haoming29:
@CannonLock  The Ubuntu test is failing at TypeScript linter: https://github.com/PelicanPlatform/pelican/actions/runs/8900875291/job/24443440075?pr=1182

CannonLock:
@haoming29 Fixed linters and rebased.

web_ui/frontend/app/registry/components/CustomRegistrationField/EnumerationField.tsx
```diff
@@ -0,0 +1,37 @@
...
+            return <TextField
+                {...params}
+                required={required}
+                helperText={description}
+                label={displayed_name}
+                variant={"outlined"}
+            />
+        }}
+        onChange={(e, value) => {
+            const enumerateId = options?.find(option => option.name === value?.name)?.id
```
haoming29, line 24: The names of the options are not unique AFAIK (at least we don't check them), is there a way to get use ID as the value instead?

haoming29 (approved):
LGTM with one minor change requested. Pre-approved. Thanks for making this happen!

web_ui/frontend/app/registry/components/CustomRegistrationField/EnumerationField.tsx
CannonLock: @haoming29 I think we should make the names unique then. If the names aren't unique how is a user supposed to distinguish between 'x' and 'x'?

CannonLock: I'd offer adding the id to the ui but the ID is likely just Primary Key hash.

CannonLock: Since this is a bit of a edge case, that won't break anything if it hits ( first one would just get picked ) I say we move forward on this PR.

CannonLock: @haoming29 Just tested and the web ui prevents duplicate names. So this is even more of an edge case.

haoming29: Yeah I think it doesn't make sense to have duplicated item names and it sounds even better if the UI can handle this case

CannonLock: @haoming29 Sounds good, I will make a ticket to have the config verifier check for unique names and merge this.

PR #1186, opened 2024-04-29
Change director TTL cache key from `ServerAd` to server URL

### LINKED ISSUE #870: Make Director more robust against duplicate server ads
In https://github.com/PelicanPlatform/pelican/pull/852, @haoming29 fixed a subtle issue with the director where an empty URL pointer was causing duplicate server ads.

It appears we're doing a hash somewhere (where?) against the entire server ad itself when deciding whether we should create new internal structure (which ones?).

Instead, we should make a unique key from the ad, composed of critical updates, and consider two ads identical if the key is identical.

This is important as I'd like to have performance metrics become part of the server ad and incorporated into the cache selection process.

haoming29:
Closes #870 

This PR also removed the mutex locks we used for TTL cache as TTL cache is thread-safe.

### DIFF
--- a/director/advertise_test.go
+++ b/director/advertise_test.go
@@ -155,17 +155,16 @@ func TestAdvertiseOSDF(t *testing.T) {
 	err := AdvertiseOSDF()
 	require.NoError(t, err)
 
-	var foundServer server_structs.ServerAd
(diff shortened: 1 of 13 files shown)


jhiemstrawisc (approved):
Nice one! LGTM

Issue #1155, opened 2024-04-24
NEAREST_CACHE should be able to prepend in front of "normal" list


bbockelm:
Some sites are using the `NEAREST_CACHE` env var to point the client at an "institutional cache" that no one else should be using (director shouldn't send clients to it).

That works well.  However, what sites likely want is more "put this cache to the front of the list" not "replace the list from the director with a single entry of this cache".

Let's add something along the lines of:

```
NEAREST_CACHE=https://blah.com,+
```

(or however we specify multiple caches).  Let's have `+` signify "put the rest of the director list here".

PR #1187, opened 2024-04-29
NEAREST_CACHE preprend cache list when specified

joereuss12:
To specify the local cache, users need to set the NEAREST_CACHE env variable. In the old functionality, setting this means we do not look for any other caches and just try to use this "preferred cache". Now, if the specified NEAREST_CACHE has a '+' character at the end of it, it will now append the "normal" list of caches behind the preferred cache in the list of caches to try.

### DIFF
--- a/cmd/object_copy.go
+++ b/cmd/object_copy.go
@@ -181,12 +181,15 @@ func copyMain(cmd *cobra.Command, args []string) {
 	}
 	caches := make([]*url.URL, 0, 1)
 	if preferredCache != "" {
-		if preferredCacheURL, err := url.Parse(preferredCache); err != nil {
-			log.Errorf("Unable to parse preferred cache (%s) as URL: %s", preferredCache, err.Error())
-			os.Exit(1)
-		} else {
-			caches = append(caches, preferredCacheURL)
-			log.Debugln("Preferred cache for transfer:", preferredCacheURL)
+		cacheList := strings.Split(preferredCache, ",")
+		for _, cache := range cacheList {
+			if preferredCacheURL, err := url.Parse(cache); err != nil {
+				log.Errorf("Unable to parse preferred cache (%s) as URL: %s", cache, err.Error())
+				os.Exit(1)
+			} else {
+				caches = append(caches, preferredCacheURL)
+				log.Debugln("Preferred cache for transfer:", preferredCacheURL)
+			}
 		}
 	}
 
--- a/cmd/object_copy.go
+++ b/cmd/object_copy.go
@@ -29,6 +29,7 @@ import (
 	"github.com/pelicanplatform/pelican/config"
 	"github.com/pelicanplatform/pelican/namespaces"
 	"github.com/pelicanplatform/pelican/param"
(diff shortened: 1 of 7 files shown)


bbockelm:
@joereuss12 -
- Can you link this to the corresponding issue?  At least from the mobile UI it appears it's not linked.
- I think the original idea was to treat NEAREST_CACHE as a list and to make the + a special list entry.  Can we still do that approach?  That'll allow multiple site-local caches.

joereuss12:
> @joereuss12 -
> 
>     * Can you link this to the corresponding issue?  At least from the mobile UI it appears it's not linked.
> 

It says it is linked under development, but here: #1155

joereuss12:
>     * I think the original idea was to treat NEAREST_CACHE as a list and to make the + a special list entry.  Can we still do that approach?  That'll allow multiple site-local caches.

Ah, I see what you mean. Yes! I can work that into the implementation!

cmd/object_copy.go
```diff
@@ -181,12 +181,15 @@ func copyMain(cmd *cobra.Command, args []string) {
 	}
 	caches := make([]*url.URL, 0, 1)
 	if preferredCache != "" {
-		if preferredCacheURL, err := url.Parse(preferredCache); err != nil {
-			log.Errorf("Unable to parse preferred cache (%s) as URL: %s", preferredCache, err.Error())
-			os.Exit(1)
-		} else {
-			caches = append(caches, preferredCacheURL)
-			log.Debugln("Preferred cache for transfer:", preferredCacheURL)
+		cacheList := strings.Split(preferredCache, ",")
```
haoming29, line 184: Can we pull this logic into a function? I saw this was repeated a couple of times in different places so I'd be nice if the logic can be shared.

cmd/object_get.go
```diff
@@ -99,12 +100,15 @@ func getMain(cmd *cobra.Command, args []string) {
 	}
 	caches := make([]*url.URL, 0, 1)
 	if preferredCache != "" {
-		if preferredCacheURL, err := url.Parse(preferredCache); err != nil {
-			log.Errorf("Unable to parse preferred cache (%s) as URL: %s", preferredCache, err.Error())
-			os.Exit(1)
-		} else {
-			caches = append(caches, preferredCacheURL)
-			log.Debugln("Preferred cache for transfer:", preferredCacheURL)
+		cacheList := strings.Split(preferredCache, ",")
```
haoming29, line 103: same here

cmd/plugin.go
```diff
@@ -384,14 +384,17 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
 	}()
 
 	caches := make([]*url.URL, 0, 1)
+	var nearestCacheURL *url.URL
```
haoming29, line 387: here as well

haoming29 (changes requested):
One relatively minor change requested otherwise LGTM

haoming29:
You may also want to merge/rebase against main to let the macOS test pass

haoming29 (approved):
Revised code LGTM

PR #1191, opened 2024-04-30
Added the ability to specify a list of data/meta folders for the cache

turetske:
Allows a used to specify multiple cache/meta data file locations

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1034,13 +1034,13 @@ func InitServer(ctx context.Context, currentServers ServerType) error {
 		if currentServers.IsEnabled(CacheType) {
 			viper.SetDefault("Cache.RunLocation", filepath.Join("/run", "pelican", "xrootd", "cache"))
 		}
-		viper.SetDefault("Cache.LocalRoot", "/run/pelican/xcache")
+		viper.SetDefault("Cache.LocalRoot", "/run/pelican/cache")
 		if viper.IsSet("Cache.DataLocation") {
 			viper.SetDefault("Cache.DataLocations", []string{filepath.Join(param.Cache_DataLocation.GetString(), "data")})
 			viper.SetDefault("Cache.MetaLocations", []string{filepath.Join(param.Cache_DataLocation.GetString(), "meta")})
 		} else {
-			viper.SetDefault("Cache.DataLocations", []string{"/run/pelican/xcache/data"})
-			viper.SetDefault("Cache.MetaLocations", []string{"/run/pelican/xcache/meta"})
+			viper.SetDefault("Cache.DataLocations", []string{"/run/pelican/cache/data"})
+			viper.SetDefault("Cache.MetaLocations", []string{"/run/pelican/cache/meta"})
 		}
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -1034,7 +1034,14 @@ func InitServer(ctx context.Context, currentServers ServerType) error {
...
 			viper.SetDefault("Cache.RunLocation", filepath.Join("/run", "pelican", "xrootd", "cache"))
 		}
-		viper.SetDefault("Cache.DataLocation", "/run/pelican/xcache")
+		viper.SetDefault("Cache.LocalRoot", "/run/pelican/xcache")
+		if viper.IsSet("Cache.DataLocation") {
+			viper.SetDefault("Cache.DataLocations", []string{filepath.Join(param.Cache_DataLocation.GetString(), "data")})
+			viper.SetDefault("Cache.MetaLocations", []string{filepath.Join(param.Cache_DataLocation.GetString(), "meta")})
+		} else {
+			viper.SetDefault("Cache.DataLocations", []string{"/run/pelican/xcache/data"})
+			viper.SetDefault("Cache.MetaLocations", []string{"/run/pelican/xcache/meta"})
```
jhiemstrawisc, line 1043: I think the name `xcache` is taboo and one of the "swear words." Is it too late to replace it with something else?

xrootd/resources/xrootd-cache.cfg
```diff
@@ -62,10 +62,14 @@ throttle.throttle concurrency {{.Cache.Concurrency}}
 pss.origin {{.Cache.PSSOrigin}}
 # FIXME: the oss.space meta / data only works if the meta and data directories are different physical devices.
 # Otherwise, no data space is setup and the cache simply doesn't write out data.
-oss.localroot {{.Cache.DataLocation}}
-#pfc.spaces data meta
-#oss.space meta {{.Cache.DataLocation}}/meta*
-#oss.space data {{.Cache.DataLocation}}/data*
+oss.localroot {{.Cache.LocalRoot}}
+pfc.spaces data meta
```
jhiemstrawisc, line 66: Based on the comment above explaining that this only works if `meta/data` are different devices, I'm worried about someone setting the deprecated `Cache.DataLocation` that now enables this previously-disabled feature. Perhaps we can keep the behavior of `Cache.DataLocation` the same as it was and not have any defaults for these values?

jhiemstrawisc (changes requested):

xrootd/resources/xrootd-cache.cfg
turetske: It seems like that above comment isn't correct anymore. I was able to get this to work on the same device. I'm unsure why I put that there originally.

config/config.go
turetske: I can change it, no worries.

xrootd/resources/xrootd-cache.cfg
turetske: Talked with BrianB about this. We've decided it's probably an erroneous comment.

jhiemstrawisc (approved):
LGTM

PR #1192, opened 2024-04-30
Allow net timeouts to be retryable errors

joereuss12:
Adds net timeouts to the list of retryable errors. Also, in the plugin, where we check if we already failed we used to only return TransferError and TransferSuccess, now we return TransferRetryable as well.


jhiemstrawisc (commented):
LGTM

jhiemstrawisc (approved):

Issue #1184, opened 2024-04-28
"401 : undefined"


bbockelm:
When creating a new origin with no password set, I get the following if I try to login to the web interface:

<img width="366" alt="image" src="https://github.com/PelicanPlatform/pelican/assets/1093447/c011fb6f-3e6c-4e84-b2cc-ddca1f3ca9cd">

"401 : undefined" -- we need to provide a better error message.

Additionally, in this case, we shouldn't even be asking for an admin password if it hasn't been initialized.

CannonLock:
@bbockelm Any chance I have access to this origin?

CannonLock:
@haoming29 I think this is a bit of a bug with the new redirection rules. If the db is not initialized the login page should redirect to the code page.

haoming29:
> @haoming29 I think this is a bit of a bug with the new redirection rules. If the db is not initialized the login page should redirect to the code page.

If that's the case, #1193 should fix this issue as well.

CannonLock:
Closed by #1233

PR #1193, opened 2024-04-30
Fix infinite redirect of UI init page

haoming29:
As recently found out in the Denver cache deployment and NSDF S3 origin deployment, the web UI is trapped in an infinite redirect at the server start (not yet initialized with admin password). This PR is to fix this issue.

To test the fix, follow the instructions below:
* Remove your admin password file (backup first): `rm /etc/pelican/server-web-passwd`
* Build and run an origin `./pelican origin serve`
* Go to the admin website at https://localhost:8444/
* You should be redirect to the initialization code page at `/view/initialization/code/`

### DIFF
--- a/web_ui/ui.go
+++ b/web_ui/ui.go
@@ -219,19 +219,22 @@ func handleWebUIAuth(ctx *gin.Context) {
 	user, err := GetUser(ctx)
 
 	// Skip auth check for static files other than html pages
-	if path.Ext(requestPath) != ".html" {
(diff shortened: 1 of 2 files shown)


CannonLock (approved):
LGTM!

haoming29:
> LGTM! @haoming29 Note that this won't close #1184 as I also make the mistake of expecting the error response to take the form { error: "" } which is a mistake.

Sounds good. I'll remove the link to close #1184

PR #1195, opened 2024-04-30
Fix wrong type for Institution registration field in the registry 

### LINKED ISSUE #1194: Institution field has wrong type `string`
As reported by @CannonLock , the institution field has the wrong type `string` when getting the list of available registration options:

```
"name": "admin_metadata.institution",
"displayed_name": "Institution",
"type": "string",
"required": true
```

It should be `enum` instead

haoming29:
Fixes #1194 

Also added back the description for the fields (that are unintentionally removed during previous merge)

### DIFF
--- a/server_structs/registry.go
+++ b/server_structs/registry.go
@@ -44,8 +44,8 @@ type AdminMetadata struct {
 	UserID                string             `json:"user_id" post:"exclude"` // "sub" claim of user JWT who requested registration
 	Description           string             `json:"description"`
 	SiteName              string             `json:"site_name"`
-	Institution           string             `json:"institution" validate:"required"` // the unique identifier of the institution
-	SecurityContactUserID string             `json:"security_contact_user_id"`        // "sub" claim of user who is responsible for taking security concern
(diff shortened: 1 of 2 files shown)


server_structs/registry.go
```diff
@@ -44,8 +44,8 @@ type AdminMetadata struct {
 	UserID                string             `json:"user_id" post:"exclude"` // "sub" claim of user JWT who requested registration
 	Description           string             `json:"description"`
 	SiteName              string             `json:"site_name"`
-	Institution           string             `json:"institution" validate:"required"` // the unique identifier of the institution
-	SecurityContactUserID string             `json:"security_contact_user_id"`        // "sub" claim of user who is responsible for taking security concern
+	Institution           string             `json:"institution" validate:"required"`                                                                                // the unique identifier of the institution
```
jhiemstrawisc, line 47: There's a strangely large amount of space before the comment here. Is that some hack to deal with linting or can we reduce it?

jhiemstrawisc (approved):
One minor formatting question, feel free to fix or merge as is.

PR #1196, opened 2024-04-30
Shim in ability for the director to use load-aware cache redirects

### LINKED ISSUE #1173: Restructure cache redirects to take a "load" variable
Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ X] Cache

Right now the director redirects clients to caches based solely on the georesolution of the incoming client IP. This doesn't give us a mechanism to pass information about cache load to the redirect function

For now, restructure the cache redirect stuff to take in a shim variable that's always set to 1. Eventually, when we have a way to generate cache load information that can be passed to the Director, that variable will inform how the Director redirects to caches.

jhiemstrawisc:
Currently, the director blindly redirects clients to the cache nearest to them without any consideration of whether that cache may be overloaded. While we work to make cache load information available in the Director, this adds a "load-aware" load balancing function -- that currently assumes the same load for every cache.

When we have load information for each cache, this new feature should be able to grab that information from the cache's serverAd and do something intelligent with it. For now, I'm assuming load will be from [0,1] where 0 is no load and 1 is the highest.

This commit also makes it a bit easier to develop more load-balancing algorithms by hooking up algorithm selection to a configurable parameter. For testing, we can also randomly sort things.

Closes #1173

### DIFF
--- a/director/sort.go
+++ b/director/sort.go
@@ -209,7 +209,8 @@ func sortServerAdsByIP(addr netip.Addr, ads []server_structs.ServerAd) ([]server
 		case "random":
 			weights[idx] = SwapMap{rand.Float64(), idx}
 		default:
-			return nil, errors.Errorf("Invalid sort method '%s' set in Director.CacheSortMethod", param.Director_CacheSortMethod.GetString())
+			return nil, errors.Errorf("Invalid sort method '%s' set in Director.CacheSortMethod. Valid methods are 'distance',"+
+				"'distanceAndLoad', and 'random.'", param.Director_CacheSortMethod.GetString())
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


haoming29:
@jhiemstrawisc the linter test failed. Could you take a look?

jhiemstrawisc:
> @jhiemstrawisc the linter test failed. Could you take a look?
Should be fixed.

director/sort.go
```diff
@@ -165,29 +165,59 @@ func getLatLong(addr netip.Addr) (lat float64, long float64, err error) {
...
+				weights[idx] = SwapMap{0 - rand.Float64(), idx}
+			} else {
+				// Each server ad will have a load value that we can use for sorting
+				weights[idx] = SwapMap{distanceAndLoadWeight(clientCoord, ad),
+					idx}
+			}
+		case "random":
+			weights[idx] = SwapMap{rand.Float64(), idx}
+		default:
+			return nil, errors.Errorf("Invalid sort method '%s' set in Director.CacheSortMethod", param.Director_CacheSortMethod.GetString())
```
haoming29, line 212:
```suggestion
			return nil, errors.Errorf("invalid sort method '%s' set in Director.CacheSortMethod", param.Director_CacheSortMethod.GetString())
```
Do we want to mention available options in the error message as well?

haoming29 (approved):
LGMT! With one minor comment that's non-blocking. Feel free fix it and merge.

PR #1205, opened 2024-05-01
Use HTTPS when the director is sending clients to Pelican servers

### LINKED ISSUE #1204: Pelican redirects requests for all public objects to HTTP and never HTTPS
**Describe the bug**
Legacy OSDF caches/origins split public and authorized reads across HTTP and HTTPS, respectively, whereas all Pelican servers should be HTTPS all of the time. However, when the director redirects a client to a cache/origin to fetch public objects, it still uses HTTP, regardless of whether the server is Pelicanized or not. This means that people who fetch files through a browser get scary messages about insecure connections.

**To Reproduce**
Use curl to see how the director handles redirects for public objects:
```
curl -v https://osdf-director.osg-htc.org/ospool/uc-shared/public/OSG-Staff/validation/test.txt

...
< link: <http://osdf-uw-cache.svc.osg-htc.org:8443/ospool/uc-shared/public/OSG-Staff/validation/test.txt>;
rel="duplicate"; pri=1; depth=3, <http://osdf-uw-cache.svc.osg-htc.org:8443/ospool/uc-shared/public/OSG-Staff/validation/test.txt>;
rel="duplicate"; pri=2; depth=3, <http://sc-cache.chtc.wisc.edu:8443/ospool/uc-shared/public/OSG-Staff/validation/test.txt>;
rel="duplicate"; pri=3; depth=3, <http://stash-cache.osg.chtc.io:8000/ospool/uc-shared/public/OSG-Staff/validation/test.txt>;
rel="duplicate"; pri=4; depth=3, <http://osg-chicago-stashcache.nrp.internet2.edu:8000/ospool/uc-shared/public/OSG-Staff/validation/test.txt>;
rel="duplicate"; pri=5; depth=3, <http://dtn-pas.cinc.nrp.internet2.edu:8000/ospool/uc-shared/public/OSG-Staff/validation/test.txt>; ...
```

While we're bound to `http` for legacy OSDF caches/origins, any pelican caches/origins should use HTTPS. In this example, osdf-uw-cache and sc-cache are both Pelicanized.

**Expected behavior**
We only redirect to http endpoints for public objects that are coming through a legacy cache or from a legacy origin.

jhiemstrawisc:
Previously, we determined the scheme based on whether the requested object required an authorization token, but all Pelican servers should be HTTPS all of the time. The only time we need to redirect to HTTP is when when the client is fetching a public object from a legacy cache/origin.

Closes #1204

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -103,7 +103,10 @@ func getRedirectURL(reqPath string, ad server_structs.ServerAd, requiresAuth boo
 	if requiresAuth {
 		redirectURL.Scheme = "https"
 	} else {
-		redirectURL.Scheme = "http"
+		redirectURL.Scheme = "https"
+		if ad.FromTopology {
+			redirectURL.Scheme = "http"
+		}
 	}
 	redirectURL.Host = serverURL.Host
 	redirectURL.Path = reqPath
--- a/director/director_test.go
+++ b/director/director_test.go
@@ -1258,3 +1258,60 @@ func TestHandleAllowServer(t *testing.T) {
 		assert.Contains(t, string(resB), "'name' is a required path parameter")
 	})
 }
+
+func TestGetRedirectUrl(t *testing.T) {
+	adFromTopo := server_structs.ServerAd{
+		URL: url.URL{
+			Host: "fake-topology-ad.org:8443",
+		},
+		AuthURL: url.URL{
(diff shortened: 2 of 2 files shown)


turetske (approved):
LGTM!

PR #1207, opened 2024-05-01
Update S3 plugin in containers to fix segfault

jhiemstrawisc:
The `v0.1.1` plugin had a snippet of code that wasn't thread safe, but was only triggered in the case of PUTs. That has been fixed/tagged upstream in [this issue](https://github.com/PelicanPlatform/xrootd-s3-http/pull/33).

This PR updates the plugin version used in our containers.

@haoming29, when this is merged we should also patch it into 7.7


haoming29:
@jhiemstrawisc For the patch to 7.7, how do we want to do it? There's no source code change and it will only affect containers. Do we want to only release a container build with a release note (but without the actual git tag)?

turetske (approved):

jhiemstrawisc:
I think your approach is reasonable. Can we rebuild containers based on this PR for the previous v7.7.X releases without creating a new tag?

PR #1212, opened 2024-05-02
Bump s3 plugin version for OSX CI test

jhiemstrawisc:
My previous PR missed one place where the S3 version was declared. In a future release cycle, I hope we can centralize this so I don't have to remember the growing list of places that need these updates.


haoming29 (approved):
LGTM!

Issue #1010, opened 2024-03-28
Director web UI should display registry URL


haoming29:
In the discussion here https://github.com/PelicanPlatform/pelican/pull/966#discussion_r1543764228 we figured that there's no good way for a client user to know the registry URL (given the federation URL). And we are intended to allow general users to see the registry Web UI to browse through available namespaces. Given that director now hosts the federation URL, we should add the registry URL at the director web UI, as we did in the origin web UI, to allow users to quickly look up and go to the registry.

Another good idea would be to have an endpoint at the director to redirect user to the registry URL (could be overengineering here)

PR #1213, opened 2024-05-02
Add Registry URL

CannonLock:
- Refactor to make future additions more straight forward`


haoming29 (changes requested):
The layout is a bit off and the new panel got overflows to the edge:

Also, the panel disappears for the anonymous users, but they should be able to see it

Screenshot 2024-05-03 at 10 11 36 AM

Screenshot 2024-05-03 at 10 11 51 AM

web_ui/frontend/components/FederationOverview.tsx
```diff
@@ -23,22 +24,38 @@ const LinkBox = ({href, text} : {href: string, text: string}) => {
...
+    {key: ["Federation", "JwkUrl", "Value"], text: "JWK"}
+]
+
 const FederationOverview = () => {
 
-    const [config, setConfig] = useState<{ [key: string] : string | undefined} | undefined>(undefined)
+    const [config, setConfig] = useState<{text: string, url: string | undefined}[]>([])
 
     let getConfig = async () => {
         let response = await fetch("/api/v1.0/config")
```
haoming29, line 41: This call is not public, and it causes the director UI to jump to the login page
A better endpoint should be `/.well-known/pelican-configuration`, which will give you a json like the following:
```json
{
  "director_endpoint": "https://6ee28c5df997:8888",
  "namespace_registration_endpoint": "https://osdf-itb-registry.osg-htc.org",
  "jwks_uri": "https://6ee28c5df997:8888/.well-known/issuer.jwks",
  "broker_endpoint": ""
}

```

web_ui/frontend/components/FederationOverview.tsx
```diff
@@ -23,22 +24,38 @@ const LinkBox = ({href, text} : {href: string, text: string}) => {
...
-            setConfig({
-                JwkUrl: (responseData?.Federation as Config)?.NamespaceUrl?.Value as undefined | string,
-                NamespaceUrl: (responseData?.Federation as Config)?.NamespaceUrl?.Value as undefined | string,
-                DirectorUrl: (responseData?.Federation as Config)?.DirectorUrl?.Value as undefined | string,
-                TopologyNamespaceUrl: (responseData?.Federation as Config)?.TopologyNamespaceUrl?.Value as undefined | string,
-                DiscoveryUrl: (responseData?.Federation as Config)?.DiscoveryUrl?.Value as undefined | string,
+            const federationUrls = UrlData.map(({key, text}) => {
+                let url = getObjectValue<string>(responseData, key)
+                if(url && !url?.startsWith("http://")) {
+                    url = "http://" + url
```
haoming29, line 48: Also, for some reason, the links I got from the panel are all broken:
(note the duplicated, yet wrong protocols)

```
http://https//topology.opensciencegrid.org/stashcache/namespaces.json
```

But this is what I saw in the config page:
```
Federation.TopologyNamespaceUrl: https://topology.opensciencegrid.org/stashcache/namespaces.json
```

which is correct

haoming29 (changes requested):
See comments above

web_ui/frontend/components/FederationOverview.tsx
CannonLock: @haoming29 Where can I find what the keys will be for the TopologyNamespaceURL and the DiscoveryURL?

haoming29: Good question. I don't think it's available at that endpoint, and that link is usually what `DiscoveryURL` will be pointed to. Do you want to me to prepare a new API for the UI or add the two keys to the `pelican-configuration` endpoint?

haoming29: That sounds good to me

jhiemstrawisc:
Closes #1010

PR #1215, opened 2024-05-02
Overwrite topology server if there is a Pelican server with the same url

haoming29:
With the new TTL cache key in place #1186, there can be a case where a server from the topology shares the exact same url as a Pelican server, so there's a "key collision" and the two servers will fight against each other in the cache. This PR sets the rule that, in case of a key collision:
* If the existing server in the cache is a _Pelican_ server, and the incoming server is from the topology, then we ignore the incoming server/
* If the existing server in the cache is a topology server, and the incoming server is from Pelican, then we **overwrite** the existing server by the incoming server

### DIFF
--- a/director/cache_ads_test.go
+++ b/director/cache_ads_test.go
@@ -373,3 +373,59 @@ func TestServerAdsCacheEviction(t *testing.T) {
 		}
 	})
 }
+
+func TestRecordAd(t *testing.T) {
+	serverAds.DeleteAll()
+	t.Cleanup(func() {
+		serverAds.DeleteAll()
+	})
+
+	topologyServerUrl := url.URL{Scheme: "https", Host: "origin.chtc.wisc.edu"}
+	pelicanServerUrl := url.URL{Scheme: "https", Host: "pelican.chtc.wisc.edu"}
+	mockTopology := &server_structs.Advertisement{
+		ServerAd: server_structs.ServerAd{
+			URL:          topologyServerUrl,
+			FromTopology: true,
+		},
+		NamespaceAds: []server_structs.NamespaceAdV2{},
+	}
+	mockPelican := &server_structs.Advertisement{
+		ServerAd: server_structs.ServerAd{
+			URL:          pelicanServerUrl,
+			FromTopology: false,
+		},
+		NamespaceAds: []server_structs.NamespaceAdV2{},
+	}
+
+	t.Run("topology-server-added-if-no-duplicate", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


director/cache_ads_test.go
```diff
@@ -373,3 +373,59 @@ func TestServerAdsCacheEviction(t *testing.T) {
...
 }
+
+func TestRecordAd(t *testing.T) {
+	serverAds.DeleteAll()
+	t.Cleanup(func() {
+		serverAds.DeleteAll()
+	})
+
+	topologyServerUrl := url.URL{Scheme: "https", Host: "origin.chtc.wisc.edu"}
+	pelicanServerUrl := url.URL{Scheme: "https", Host: "pelican.chtc.wisc.edu"}
```
jhiemstrawisc, line 384: Do the URLs generated for advertisements that come from topology have a scheme? I know we don't have schemes in the actual topology JSON, but it's not clear to me whether we construct a scheme in the director for those ads, or if we don't store the scheme but [add it back on redirects](https://github.com/PelicanPlatform/pelican/blob/208b6e3019fa3bb79cdff62307e93c61497443e0/director/director.go#L92). The reason I ask is because a `ServerAd` has both a `URL` and an `AuthURL`. In the case of a server ad from topology, we need to use HTTP for the ad's `URL` and HTTPs for the `AuthURL`, so if we already have the scheme there, I would only expect the AuthURLs to match.

I think the point I'm trying to get to is that a Pelican ad with `ad.URL = "https://pelican.com"` should probably still overwrite a topology ad with `ad.URL = "http://pelican.com"`. The problem with doing that here is that it would require the TTL cache to use only the hostname as a key. Thoughts?

jhiemstrawisc (changes requested):

director/cache_ads_test.go
haoming29: Ooof. First, thanks for catching this! Besides what you mentioned, the two URLs should be the same. The current setup gives two different servers which is not intended.

I think you are right, if servers are from topology, we set their `ad.URL` with `http` scheme, which will mess things up as you demonstrated. However, hostname sounds like a scope too broad to be used as a key as this prevent users to serve multiple servers under different ports (although I don't think this is a good idea at all)

One less optimal but easier to implement solution I could think of is to strip `https` then `http` from the url when doing the comparison and use that to determine if there's an existing server in the ttl cache.

jhiemstrawisc (approved):
LGTM

PR #1218, opened 2024-05-03
Add a denied namespaces page

CannonLock:
- From here you can delete or approve previously denied namespace if you are an admin


haoming29:
@CannonLock There's a linter test failed. Could you check that one?

web_ui/frontend/components/Namespace/DeniedCard.tsx
```diff
@@ -0,0 +1,160 @@
...
+                                </>
+                            }
+                        </Box>
+                    </Box>
+                </Box>
+                <Box ref={ref}>
+                    <InformationDropdown adminMetadata={namespace.admin_metadata} transition={transition} parentRef={ref}/>
+                </Box>
+            </Box>
+            <AlertPortal alert={alert} onClose={() => setAlert(undefined)} />
```
haoming29, line 154: I didn't see any alert when I successfully deleted a namespace, but we do if we approve a namespace in the dashboard view. Can we give user a successful message alert here?

haoming29 (changes requested):
One small change requested. LGTM if the linter failure is fixed too

web_ui/frontend/components/Namespace/DeniedCard.tsx
CannonLock: @haoming29 Going to start transitioning to making success states less intrusive. On success the card now turns green before it disappears rather then also popping up a alert.

PR #1221, opened 2024-05-03
Collect origin/cache server load metric from xrootd g-stream and use it for calculating redirect priority

### LINKED ISSUE #873: Prometheus monitoring for the g-stream
Pelican Service: Cache, Origin

In https://github.com/xrootd/xrootd/pull/2206, there are new monitoring statistics added for measuring the overall I/O load on an origin or cache.  Once/if that work gets merged, let's add a corresponding Prometheus metric.

Goal is to have the I/O load eventually be part of the cache and origin selection logic alongside GeoIP.

haoming29:
Closes #873 

This PR is built on top of osg-testing build of xrootd-5.6.9-1.3. Without the patch, origin/cache server **will fail to start**. To test this locally, you need to rebuild dev container: 
`docker build -t pelican-dev -f images/dev.Dockerfile .`

The production containers are updated to use the patched version and are tested to be able to run locally.

The server IO metrics can be found via Prometheus query: `/api/v1.0/prometheus/query?query=xrootd_server_io`, and parsed/scaled IO load number can be found in director's API `/api/v1.0/director_ui/server`, the `io_load` field. This value is calculated by taking the first derivative of Prometheus `xrootd_server_io` metric over 5 minutes, divided by 100 (a magic number for now) to reduce sensitivity, finally normalized to [0,1] by a `sigmoid` activation function.

If you spin up a fed-in-a-box, create your temp file, and download it through Pelican client, you should be able to see the `io_load` number fluctuates.


director/director_api.go
```diff
@@ -154,3 +158,73 @@ func ConfigFilterdServers() {
 		filteredServers[sn] = permFiltered
 	}
 }
+
+// Start a goroutine to query director's Prometheus endpoint for origin/cache server I/O stats
+// and save the value to the corresponding serverAd
+func ConfigServerIOQuery(ctx context.Context, egrp *errgroup.Group) {
```
bbockelm, line 164: I have been trying to use `Launch` in the function name to indicate a goroutine will be created by a given routine.
```suggestion
func LaunchServerIOQuery(ctx context.Context, egrp *errgroup.Group) {
```

In general, this function is hard to read due to the level of indentation.  Instead of having an anonymous function inside `egrp.Go`, can you define the function in the body of `LaunchServerIOQuery` and then later call `egrp.Go`?

director/director_api.go
```diff
@@ -154,3 +158,73 @@ func ConfigFilterdServers() {
...
+					}
+					serverUrl := item.Key()
+					serverAd := item.Value()
+					if serverAd.FromTopology {
+						// Topology servers have no Prometheus metrics
+						wg.Done()
+						continue
+					}
+					// Run the queries in parallel, in hope that Prometheus is strong enough to handle ~100 concurrent requests
+					go func(ctx context.Context, wg *sync.WaitGroup, sUrl string, sAd *server_structs.Advertisement) {
```
bbockelm, line 190: Why parallel queries?

Why not just query for *all* series in prometheus then filter out the ones not corresponding to a server we're interested in?

director/director_api.go
```diff
@@ -154,3 +158,73 @@ func ConfigFilterdServers() {
...
+							ioDeriv, err := strconv.ParseFloat(ioDerivStr, 64)
+							if err != nil {
+								log.Debugf("Failed to update IO stat for server %s: failed to convert Prometheus response to a float number: %s", serverUrl, ioDerivStr)
+								return
+							}
+							// We scale down the derivative to avoid sensitive metric
+							// but this is a magic number and we need to figure out a heuristic for getting a reliable number
+							ioDeriv /= 100
+							// Here we use a sigmoid activate function to bound the value between [0,1]
+							sigmoidIO := utils.Sigmoid(ioDeriv)
```
bbockelm, line 219: For now, I'd like to avoid bounding it to 1 -- see https://github.com/PelicanPlatform/pelican/discussions/1198.

bbockelm (changes requested):
Just a few big-picture items.  Didn't get a chance to go in-depth (but I assume someone else can check the details).

director/prom_query.go
```diff
@@ -0,0 +1,267 @@
...
+								return
+							}
+							unixTime, ok := valueArr[0].(float64)
+							if !ok {
+								err = fmt.Errorf("parse error for result type %s: malformed time for values field for element at %d: %#v", data.ResultType, rIdx, elem)
+								return
+							}
+							vecValue, ok := valueArr[1].(string)
+							if !ok {
+								err = fmt.Errorf("parse error for result type %s: malformed metric value for values field for element at %d: %#v", data.ResultType, rIdx, elem)
```
jhiemstrawisc, line 252: Nice, I had seen `%+v` but `%#v` is a new format specifier for me

jhiemstrawisc (changes requested):

haoming29:
@bbockelm Drawing the metrics out shows that the `io_total` value isn't always positive, and it went wild at the start of the server to rocket high (max at 16000) then fall down to a reasonable value. The `io_active` is 0 for all the time. Is this expected?

Screenshot 2024-05-07 at 10 09 29 AM

director/director_api.go
```diff
@@ -154,3 +158,73 @@ func ConfigFilterdServers() {
...
+						if err != nil {
+							log.Debugf("Failed to update IO stat for server %s: %v", serverUrl, err)
+							return
+						}
+						if queryResult.ResultType != "vector" {
+							log.Debugf("Failed to update IO stat for server %s: Prometheus response returns type %s not vector", serverUrl, queryResult.ResultType)
+							return
+						}
+						if len(queryResult.Result) != 1 {
+							log.Debugf("Failed to update IO stat for server %s: Prometheus response contains more or less than 1 result: %d", serverUrl, len(queryResult.Result))
```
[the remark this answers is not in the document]
haoming29: This log was removed due to the revision of Brian's comment

haoming29:
@bbockelm The tests are failing. The Ubuntu one is expected because it uses the dev container that does not have the patch to xrootd in place, but the macOS failure is unexpected. I patched in your code with a git diff but the error message suggests xrootd can't start due to mismatched configuration: https://github.com/PelicanPlatform/pelican/actions/runs/9005682697/job/24742038342?pr=1221 Did I miss something?

bbockelm:
@haoming29 - for the Mac OS X test failure, I see the following:

> time="2024-05-08T17:10:38Z" level=info msg="240508 17:10:38 259 XrootdConfig: monitor dest endpoint is invalid; nodename nor servname provided, or not known" daemon=xrootd.origin

However, the xrootd configuration doesn't seem to be logged.  Do you know why this is occurring?

metrics/xrootd_metrics.go
```diff
@@ -308,6 +316,11 @@ var (
 		Help: "Number of bytes the data requested is in the cache or not",
 	}, []string{"path", "type"}) // type: hit/miss/bypass
 
+	CacheIO = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
bbockelm, line 319: I don't think the `Gauge` type is correct for all the data here.
- active I/O is the current number of I/O operations ongoing.
- total I/O is an operation counter: the total number of I/O operations seen by the throttle plugin.  This should be monotonically increasing.
- I/O wait time is the total time that the server was waiting on operations to complete.  This should be monotonically increasing.

So, a gauge type is appropriate for the first two but not the latter two.

haoming29:
> @haoming29 - for the Mac OS X test failure, I see the following:
> 
> > time="2024-05-08T17:10:38Z" level=info msg="240508 17:10:38 259 XrootdConfig: monitor dest endpoint is invalid; nodename nor servname provided, or not known" daemon=xrootd.origin
> 
> However, the xrootd configuration doesn't seem to be logged. Do you know why this is occurring?

I did see both xrootd and pelican configs are logged: https://github.com/PelicanPlatform/pelican/actions/runs/9005682697/job/24742038342?pr=1221#step:7:756

bbockelm (changes requested):
Getting there.
- Please backport the XRootD fixes for the load metrics as needed  (https://github.com/xrootd/xrootd/pull/2262)
- Let's use the derivative of the I/O wait as our load metrics (meaning it's not between 0 and 1) and drop the sigmoid-based approach for now.
- Small change requested in the use of the Prometheus metrics.

haoming29:
Jira ticket is created: https://opensciencegrid.atlassian.net/browse/SOFTWARE-5873

haoming29:
> @haoming29 -- I found a few bugs in the throttle code and sent a PR upstream: [xrootd/xrootd#2262](https://github.com/xrootd/xrootd/pull/2262)
> 
> Let's merge it into our build -- that should give more realistic numbers for the metrics.

Jira ticket is created: https://opensciencegrid.atlassian.net/browse/SOFTWARE-5873

haoming29:
Update: with @bbockelm 's fix to xrootd, the IO wait time seems to work fine now. However, the active IO is still not accurate and the total IO is not monotonically increasing.

Grafana:
Screenshot 2024-05-30 at 1 08 33 PM

As you can tell, for most of the time, the total IO is monotonically increasing, but it decreases around server start. After the restart of the process, the value seems to be monotonic

For active IO, the Prometheus read is always 0. However, when I printed the array of json strings from xrootd g-stream, I got a pattern similar to this:
 
```json
[{"event":"throttle_update","io_wait":2.5060,"io_active":1,"io_total":1990} {"event":"throttle_update","io_wait":2.5610,"io_active":0,"io_total":2019} {"event":"throttle_update","io_wait":2.6180,"io_active":0,"io_total":2048} 
....
{"event":"throttle_update","io_wait":3.3790,"io_active":0,"io_total":2600} {"event":"throttle_update","io_wait":3.3790,"io_active":0,"io_total":2600} {"event":"throttle_update","io_wait":3.3790,"io_active":0,"io_total":2600}] 
```

Notice that the `"io_active":1` in the first entry. After looking at a couple of such arrays, it seems that `io_active` appears and only appears to be 1 for the first couple of entries, and then it goes down to 0 for the rest. (There's a file transfer going on while I inspect the log). Since I iterate over the array and update Prometheus metric one by one, the 1 value is quickly overridden by the 0 values. However, this makes me wonder if there's still a bug somewhere in xrootd and flushes/resets the active IO.


The io_wait seems to be fine, although I'm not sure the unit of the time. The file I used to test is a 5GiB file that took ~1-2min to finish transfer, but the wait time is not in this range either if it's in second or minute

haoming29:
Also, since we do have `io_active` as a gauge value, and I use one metric with three different labels for the three data point, I can't change it from gauge type to counter type unless we want to have three metrics instead

bbockelm:
@haoming29 - I think this is actually potentially good news:

- On server restart, counters are supposed to go to zero.  There's special logic within Prometheus that will account for this in restarts and will ensure the derivative handles the discontinuity correctly.
- `io_active` doesn't capture the number of transfers ongoing but rather the number of storage operations ongoing.  If the network is the bottleneck and you're doing a single file transfer at a time, then via OS buffering and readahead, you'll rarely see a non-zero `io_active` value.
   - I've found the best way to capture this is purposely slow down the storage operations.  This could be done with a special compilation of XRootD (inserting a "sleep 1"-type statement into the read) or dropping the page cache and turning off readahead to mimic a slow IO device.

For the first item, I think we should just split this into three metrics, two counters and a gauge.  For `io_active`, I suspect we should keep an eye out on existing servers.  Since it's not a critical-path metric, I'm OK with letting the mystery remain for a release or two.

haoming29:
@bbockelm Sounds good to me. I'll change to three metrics then

haoming29:
Changed `io_total` and `io_wait` to counter and the plot looks good to me (the drops are at server restarts which is expected):
Screenshot 2024-06-05 at 2 22 35 PM

haoming29:
Looks like the test are passing now, OSX dependency issue should be fixed.

bbockelm (approved):
Getting close on this one.  Very minor items - only spotted two concerns:

1.  Since IOLoad is no longer in the interval [0, 1), the current sorting metrics are not useful.  We should either fix it here (though the PR is already large enough - maybe not the greatest idea) or do an immediate follow-up to make the sorting algorithms closer to what's described in https://github.com/PelicanPlatform/pelican/discussions/1198.
2. The `IOLoad` should probably be made atomic.

Once you handle the above two items, I think we are good to go!

director/sort_algorithms.go
```diff
@@ -76,7 +76,10 @@ func distanceAndLoadWeight(coord Coordinate, sAd server_structs.ServerAd) float6
 	// assume a server's load has more impact on its performance than its distance does. As long as load is
 	// hard coded, this function should act exactly like distanceWeight.
 	// TODO: Come up with a better function once we have an actual load value and know how it works
-	load := 0.5
+	load := sAd.IOLoad
```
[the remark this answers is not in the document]
haoming29: I'll do a follow-up ticket to fix the range issue instead

haoming29:
@bbockelm re-requested your review in case you want to take a look at the changes.

bbockelm:
@haoming29 - I just noticed the Y-axis here:

image

I'm puzzled why the derivative is going negative around restart.  Are you sure you're using the correct prometheus query?  Is it marked as a Counter in that database?

haoming29:
> @haoming29 - I just noticed the Y-axis here:
> I'm puzzled why the derivative is going negative around restart. Are you sure you're using the correct prometheus query? Is it marked as a Counter in that database?

This is the query I used for that plot: `deriv(xrootd_server_io_total{job="origin_cache_servers", server_type="Origin"}[5m])`

I think negative derivative is expected at the server restart because it's the value of the metric that's reset to 0, but the derivative still takes the values from the past 5min and do the math, and in our case there's a large drop of the value, and the derivative goes to negative (and slowly going back to positive once the server is up and running for a while)

bbockelm:
> This is the query I used for that plot: `deriv(xrootd_server_io_total{job="origin_cache_servers", server_type="Origin"}[5m])`

Ah, that's the problem.  From the Prometheus documentation,

> `deriv` should only be used with gauges.

The `deriv` function does not take into account server restarts.  I think you want the `rate` function instead.  From the Prometheus docs:

> `rate(v range-vector)` calculates the per-second average rate of increase of the time series in the range vector. Breaks in monotonicity (such as counter resets due to target restarts) are automatically adjusted for.

It's a big problem if the server load becomes negative!

haoming29:
@bbockelm Changed to use `rate` instead, which looks much more correct. The only issue is it takes a while (5min?) for the rate metric to dissipate to 0 at the server restart, but I'd assume this won't be a big problem.
Screenshot 2024-06-12 at 8 51 18 AM

haoming29:
@jhiemstrawisc I've included the two additional patches from xrootd for OSX dependecy, and it looks like the macOS test is passing. This PR should be good for a final review

director/prom_query.go
```diff
@@ -0,0 +1,266 @@
...
+	if strings.HasPrefix(query, "?query=") {
+		err = errors.Errorf("query argument should not contain \"?query=\"")
+		return
+	}
+	extWebUrl := param.Server_ExternalWebUrl.GetString()
+	baseUrl, err := url.JoinPath(extWebUrl, "/api/v1.0/prometheus/query")
+	if err != nil {
+		return
+	}
+	queryUrl := baseUrl + "?query=" + url.QueryEscape(query)
```
jhiemstrawisc, line 87: Why not construct this so that if the incoming query is already prefixed with `?query=` we simply pass it along, but if there is no prefix, we prepend it?

jhiemstrawisc (commented):
Have the endpoints for observing these values changed? I was unable to observe anything by firing this up and hitting `/api/v1.0/prometheus/query?query=xrootd_server_io` or `/api/v1.0/director_ui/server` locally.

director/prom_query.go
haoming29: Oh that's because we can't tell if someone is mistakenly passing ?query=xxx where xxx is not URLEncoded, so we require that they only pass the query body instead of with the prefix

haoming29:
> Have the endpoints for observing these values changed? I was unable to observe anything by firing this up and hitting `/api/v1.0/prometheus/query?query=xrootd_server_io` or `/api/v1.0/director_ui/server` locally.

Sorry things changed since I wrote the description. The new metric names are: `xrootd_server_io_total`, `xrootd_server_io_active`, and `xrootd_server_io_wait_time `, respectively

jhiemstrawisc (approved):
I've been able to reproduce, and it seems that many of the concerns @bbockelm pointed out have been addressed. LGTM

PR #1230, opened 2024-05-06
Fix origin registration status not updated

### LINKED ISSUE #1229: Origin registration status API doesn't update
Pelican Version:

Pelican Service:
- Origin

**Describe the bug**

The `/origin_ui/exports` API contains origin registration status. The data should reflect if the origin registration is complete. However, even after the admin fixed the issue at the registry, the API still returns the existing status and didn't pull the latest status.

haoming29:
Fixes #1229 

@CannonLock requested your review since you have the frontend set up to check this. Even with the fix, it takes ~1 min for the value to be updated unfortunately.

### DIFF
--- a/origin/reg_status.go
+++ b/origin/reg_status.go
@@ -59,7 +59,7 @@ type (
 // A TTL cache to save namespace registration status
 // The TTL depends on the Status: For Status == StatusRegistrationError, the cache item won't expire,
 // meaning a failure registration stays in the cache, until the registration is successful.
-// For other Status, TTL is set to 5min
+// For other Status, TTL is set to 1 min
 var registrationsStatus = ttlcache.New(ttlcache.WithTTL[string, RegistrationStatus](1 * time.Minute))
 
 var RegistryNotImplErr = errors.New("the running version of the registry didn't implmenet this function")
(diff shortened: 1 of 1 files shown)


origin/reg_status.go
```diff
@@ -60,7 +60,7 @@ type (
 // The TTL depends on the Status: For Status == StatusRegistrationError, the cache item won't expire,
 // meaning a failure registration stays in the cache, until the registration is successful.
 // For other Status, TTL is set to 5min
-var registrationsStatus = ttlcache.New(ttlcache.WithTTL[string, RegistrationStatus](5 * time.Minute))
+var registrationsStatus = ttlcache.New(ttlcache.WithTTL[string, RegistrationStatus](1 * time.Minute))
```
jhiemstrawisc, line 63: Does switching this to a 1-minute TTL negate the above comment?

jhiemstrawisc (approved):
One comment that might warrant a change, but otherwise LGTM. I'll approve and you can decide whether to merge as is.

origin/reg_status.go
haoming29: Thanks for the catch! Fixed

PR #1234, opened 2024-05-06
Plugin now also accepts PELICAN_NEAREST_CACHE

joereuss12:
As requsted from the condor devs, Pelican now accepts PELICAN_NEAREST_CACHE. This is so this set environment varible looks more important and not confusing for what it is for when set in condor configuration


haoming29 (approved):
LGTM

PR #1242, opened 2024-05-07
Fix director bug in looking up prefix for cache

### LINKED ISSUE #1241: Use `Server.Hostname` as the prefix for origin/cache namespace registration
As @williamnswanson reported, the ITB cache failed to start with 7.8.0 director with the following message: 
```
 Error: error during director registration: Authorization token verification failed failed to check namespace approval status: Registry returns error when checkNamespaceStatus 400 with body {"status":"error","msg":"The namespace /caches/2607:f388:2200:99:b984:4e7d:7cf7:1568 does not exist in the registry"}
```

The ITB cache has `Xrootd.Sitename` set to `CHTC-ITB-OSDF-PELICAN-CACHE` but the director was using the server IP to look up the prefix of the cache at the registry, causing a 400 error.

haoming29:
Fixes #1241 

The bug was introduced in 7.8.0 where the director uses `hostname` of `DataURL` of the cache to concatenate with `/caches/{hostname}` for looking up the namespace prefix for the cache in the registry. This is based on the premise that the default value of `Xrootd.Sitename` is `Server.Hostname`, which is what the cache used to register itself at the registry.

However, in most of our ITB/production servers, we changed `Xrootd.Sitename` to something more human-readable, which causes the discrepancy between director look-up and what register records. 

This PR fixes the issue by adding a new field in origin/cache advertisement, `RegistryPrefix`, which is the full namespace prefix the cache used to register itself at the registry, i.e. `/caches/{Xrootd.Sitename}`, so that the director can use this value for registry look up.

This extra field will also help use build #1107 in the future.

This PR also fixed a bug in the UI where editing cache registration will prepend `/caches/` to the existing namespace prefix, causing duplicated prefix `/caches/caches/blah/blah`

Come with tests.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -369,10 +369,6 @@ func keySignChallengeCommit(ctx *gin.Context, data *registrationData) (bool, map
 			}
 		}
 
-		// Since CLI/auto-registered namespace did not have site name as part of their registration
-		// Use their IP for an educated guess
-		ns.AdminMetadata.SiteName = ctx.ClientIP()
-
 		// Overwrite status to Pending to filter malicious request
 		ns.AdminMetadata.Status = server_structs.RegPending
 
--- a/cache/advertise.go
+++ b/cache/advertise.go
@@ -39,12 +39,14 @@ type (
 	}
 )
 
-func (server *CacheServer) CreateAdvertisement(name string, originUrl string, originWebUrl string) (*server_structs.OriginAdvertiseV2, error) {
+func (server *CacheServer) CreateAdvertisement(name, originUrl, originWebUrl string) (*server_structs.OriginAdvertiseV2, error) {
+	registryPrefix := "/caches/" + param.Xrootd_Sitename.GetString()
 	ad := server_structs.OriginAdvertiseV2{
-		Name:       name,
-		DataURL:    originUrl,
-		WebURL:     originWebUrl,
-		Namespaces: server.GetNamespaceAds(),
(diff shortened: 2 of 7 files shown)


haoming29:
@turetske since the bug was introduced in 7.8.0, we should be good to only patch this in for 7.8

registry/registry.go
```diff
@@ -369,10 +369,6 @@ func keySignChallengeCommit(ctx *gin.Context, data *registrationData) (bool, map
 			}
 		}
 
-		// Since CLI/auto-registered namespace did not have site name as part of their registration
-		// Use their IP for an educated guess
-		ns.AdminMetadata.SiteName = ctx.ClientIP()
```
turetske, line 374: Is this just being removed altogether?

haoming29: Yes, I realized this is not necessary because we want the user to explicitly set `Sitename`, and we use that value for the director to show it on the director web UI, and defaulting to an IP address doesn't help much

turetske (approved):
LGTM!

haoming29:
@turetske Since Brian suggested a change to the approach, let's hold the patch release for this PR and wait for a revision. I'll try to get this done today so that we can unblock @williamnswanson

haoming29:
@turetske Just talked to Brian and due to backward compatibility issues, we won't do another revision for 7.8, so you are good to release this patch. The revision will come in 7.9.

PR #1244, opened 2024-05-07
Fixed Cache shutting down if advertisement fails

### LINKED ISSUE #1189: Do not fail cache startup for a failed advertisement
Pelican Version: 7.7.2

When the cache starts, it invokes `Advertise` to the director as part of `CacheServeFinish`.

If the advertisement fails, then the whole cache shuts down.  This is surprising as the advertisement will fail if the cache registration is still pending.

For the origin, if the advertisement fails, then the periodic go routine still continues to run in the background.  It's not a hard requirement for startup.

We should provide a kinder user experience -- let's make it so the standalone cache has behavior analogous to the origin.  The admin should be able to bring up the service and, from the web UI, notice that the registration is still pending.

turetske:
Removed extra advertise in CacheServeFinish, now only advertises in the launcher's PeriodicAdvertise

closes #1189

### DIFF
--- a/launchers/cache_serve.go
+++ b/launchers/cache_serve.go
@@ -138,8 +138,5 @@ func CacheServeFinish(ctx context.Context, egrp *errgroup.Group, cacheServer ser
 		return err
 	}
 
-	log.Debug("Advertise Cache")
-	servers := make([]server_structs.XRootDServer, 1)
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (approved):
LGTM :)

PR #1252, opened 2024-05-08
Plugin advertises PluginVersion

joereuss12:
Ran into an issue on ospool since the pelican plugin stopped advertising PluginVersion with -classad, this change reimplements that advertisement in addition with PelicanPluginVersion.


joereuss12:
@haoming29 this is the issue Brian mentioned during standup. Should be fixed with this 1 line change as it reverts the change in: #1113

haoming29 (approved):
LGTM

joereuss12:
@turetske are you able to make a patch release for this for @matyasselmeci?

turetske:
@joereuss12 I can, I want to wait on another fix before I do so.

PR #1253, opened 2024-05-08
Add link to OpenAPI page on the UI

### LINKED ISSUE #1147: Add link to Swagger docs on server admin website
We deployed Swagger docs on each server website at `/api/v1,0.docs` but we didn't advertise this anywhere. It would be nice if we add the link to the dialer panel in the website menu (along with the various links and version number)

haoming29:
Closes #1147

### DIFF
--- a/web_ui/frontend/components/layout/Sidebar/SpeedDial.tsx
+++ b/web_ui/frontend/components/layout/Sidebar/SpeedDial.tsx
@@ -1,8 +1,8 @@
 "use client"
 
-import React, {ReactNode, useEffect, useState} from "react";
+import React, {ReactNode, useState} from "react";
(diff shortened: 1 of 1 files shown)


CannonLock (changes requested):
It is the `import Link from "next/link";` that is prepending /view/. Lets just use the Link component from material-ui and add this new value directly to actions to prevent the jumpiness of adding another element post mount.

haoming29:
> It is the `import Link from "next/link";` that is prepending /view/. Lets just use the Link component from material-ui and add this new value directly to actions to prevent the jumpiness of adding another element post mount.
> 
> Notably this swap only makes sense because all these pages link to the outside and won't benefit from client side rendering.

Sounds good to me

CannonLock (approved):
LGTM! Code reviewed and tested.

PR #1254, opened 2024-05-08
Fix bug when director returns 404 for PROPFIND

joereuss12:
Found a bug when we try to PROPFIND against an outdated director, the director returns a 404 not found (because it redirects to a cache and that cache cannot find the object to PROPFIND so it returns a 404). When this happens the client returns just the error and not the response. Therefore, the fix is to check if we fail querying the director with a PROPFIND, we check if the namespace has a dirlisthost. If dirlisthost is present, we continue and otherwise, we fail. We'll need to eventually create a way to test against outdated directors but for now, we'll stick to this band-aid.

### DIFF
--- a/client/director.go
+++ b/client/director.go
@@ -152,7 +152,11 @@ func queryDirector(ctx context.Context, verb, sourcePath, directorUrl string) (r
 	body, _ := io.ReadAll(resp.Body)
 
 	// If we get a 404, the director will hopefully tell us why. It might be that the namespace doesn't exist
-	if resp.StatusCode == 404 {
+	if resp.StatusCode == 404 && verb == "PROPFIND" {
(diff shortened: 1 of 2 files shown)


joereuss12:
made issue #1255 so we can remember to test this. Since it is so close to release (and this is a significant bug) and because we currently don't have the framework to test against old directors, we should put this fix out in a patch release and revisit making tests for this in the near future.

bbockelm:
Instead of blanket covering _all_ errors, can we just target the error we know old directors to do (404?)?

joereuss12:
> Instead of blanket covering _all_ errors, can we just target the error we know old directors to do (404?)?

Implemented just now

jhiemstrawisc (approved):

joereuss12:
@turetske are you able to throw this in the patch release as well?

PR #1257, opened 2024-05-08
Pelican object ls command implementation

joereuss12:
New command added `pelican object ls` that lists the objects of the specified dir in a namespace (or on the namespace itself). This PR also comes with tests for this new command as well as tests for a helper function to obtain 'dirlisthost'. The new command with a few sets of flags that I felt were the most important from the standard 'ls' command.

This still needs a little more polishing, but I thought to go ahead and make the PR if anyone wants to take a look and to push my current progress on this new command

### DIFF
--- a/client/fed_test.go
+++ b/client/fed_test.go
@@ -700,6 +700,17 @@ func TestNewTransferJob(t *testing.T) {
 	})
 }
 
+// A helper function to set up the TestObjectLs, gives us a buffer to capture stdout as well as our path to ls
+func setupObjectListTest(export server_utils.OriginExport) (listURL string, r, w, stdout *os.File, buf *bytes.Buffer) {
+	listURL = fmt.Sprintf("pelican://%s:%s%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()), export.FederationPrefix)
+	// Create a pipe for output
+	r, w, _ = os.Pipe()
+	stdout = os.Stdout
+	os.Stdout = w
+	buf = new(bytes.Buffer)
+	return listURL, r, w, stdout, buf
+}
+
 // A test that spins up a federation, and tests object list
 func TestObjectList(t *testing.T) {
 	viper.Reset()
@@ -755,28 +766,21 @@ func TestObjectList(t *testing.T) {
 
 	// This tests object ls with no flags set
 	t.Run("testPelicanObjectLsNoFlags", func(t *testing.T) {
-		// Set path for object to upload/download
+		// Set path for ls
 		for _, export := range fed.Exports {
-			uploadURL := fmt.Sprintf("pelican://%s:%s%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()), export.FederationPrefix)
-
-			// Create a pipe for output
-			r, w, _ := os.Pipe()
-			originalStdout := os.Stdout
-			os.Stdout = w
-
+			listURL, r, w, originalStdout, buf := setupObjectListTest(export)
 			go func() {
 				if export.Capabilities.PublicReads {
-					err := client.DoList(fed.Ctx, uploadURL, client.WithTokenLocation(""))
+					err := client.DoList(fed.Ctx, listURL, client.WithTokenLocation(""))
 					require.NoError(t, err)
 					w.Close()
 				} else {
-					err := client.DoList(fed.Ctx, uploadURL, client.WithTokenLocation(tempToken.Name()))
+					err := client.DoList(fed.Ctx, listURL, client.WithTokenLocation(tempToken.Name()))
 					require.NoError(t, err)
 					w.Close()
 				}
 			}()
-			var buf bytes.Buffer
-			_, err = io.Copy(&buf, r)
+			_, err = io.Copy(buf, r)
 			require.NoError(t, err)
 			os.Stdout = originalStdout
 
@@ -787,28 +791,21 @@ func TestObjectList(t *testing.T) {
 
 	// This tests object ls with json flag set
 	t.Run("testPelicanObjectLsJson", func(t *testing.T) {
-		// Set path for object to upload/download
+		// Set path for ls
 		for _, export := range fed.Exports {
-			uploadURL := fmt.Sprintf("pelican://%s:%s%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()), export.FederationPrefix)
-
-			// Create a pipe for output
-			r, w, _ := os.Pipe()
-			originalStdout := os.Stdout
-			os.Stdout = w
-
+			listURL, r, w, originalStdout, buf := setupObjectListTest(export)
 			go func() {
 				if export.Capabilities.PublicReads {
-					err := client.DoList(fed.Ctx, uploadURL, client.WithTokenLocation(""), client.WithJsonOption(true))
+					err := client.DoList(fed.Ctx, listURL, client.WithTokenLocation(""), client.WithJsonOption(true))
 					require.NoError(t, err)
 					w.Close()
 				} else {
-					err := client.DoList(fed.Ctx, uploadURL, client.WithTokenLocation(tempToken.Name()), client.WithJsonOption(true))
+					err := client.DoList(fed.Ctx, listURL, client.WithTokenLocation(tempToken.Name()), client.WithJsonOption(true))
 					require.NoError(t, err)
 					w.Close()
 				}
 			}()
-			var buf bytes.Buffer
-			_, err = io.Copy(&buf, r)
+			_, err = io.Copy(buf, r)
 			require.NoError(t, err)
 			os.Stdout = originalStdout
 
@@ -819,28 +816,21 @@ func TestObjectList(t *testing.T) {
 
 	// Test that dironly works and does not list files
 	t.Run("testPelicanObjectLsDirOnly", func(t *testing.T) {
-		// Set path for object to upload/download
+		// Set path for ls
 		for _, export := range fed.Exports {
-			uploadURL := fmt.Sprintf("pelican://%s:%s%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()), export.FederationPrefix)
-
-			// Create a pipe for output
-			r, w, _ := os.Pipe()
-			originalStdout := os.Stdout
-			os.Stdout = w
-
+			listURL, r, w, originalStdout, buf := setupObjectListTest(export)
 			go func() {
 				if export.Capabilities.PublicReads {
-					err := client.DoList(fed.Ctx, uploadURL, client.WithDirOnlyOption(true))
+					err := client.DoList(fed.Ctx, listURL, client.WithDirOnlyOption(true))
 					require.NoError(t, err)
 					w.Close()
 				} else {
-					err := client.DoList(fed.Ctx, uploadURL, client.WithDirOnlyOption(true), client.WithTokenLocation(tempToken.Name()))
+					err := client.DoList(fed.Ctx, listURL, client.WithDirOnlyOption(true), client.WithTokenLocation(tempToken.Name()))
 					require.NoError(t, err)
 					w.Close()
 				}
 			}()
-			var buf bytes.Buffer
-			_, err = io.Copy(&buf, r)
+			_, err = io.Copy(buf, r)
 			require.NoError(t, err)
 			os.Stdout = originalStdout
 
@@ -849,30 +839,23 @@ func TestObjectList(t *testing.T) {
 		}
 	})
 
-	// Test that fileonly works and only lists files
-	t.Run("testPelicanObjectLsFileOnly", func(t *testing.T) {
-		// Set path for object to upload/download
+	// Test that objectonly works and only lists files
+	t.Run("testPelicanObjectLsObjectOnly", func(t *testing.T) {
+		// Set path for ls
 		for _, export := range fed.Exports {
-			uploadURL := fmt.Sprintf("pelican://%s:%s%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()), export.FederationPrefix)
-
-			// Create a pipe for output
-			r, w, _ := os.Pipe()
-			originalStdout := os.Stdout
-			os.Stdout = w
-
+			listURL, r, w, originalStdout, buf := setupObjectListTest(export)
 			go func() {
 				if export.Capabilities.PublicReads {
-					err := client.DoList(fed.Ctx, uploadURL, client.WithFileOnlyOption(true))
+					err := client.DoList(fed.Ctx, listURL, client.WithObjectOnlyOption(true))
 					require.NoError(t, err)
 					w.Close()
 				} else {
-					err := client.DoList(fed.Ctx, uploadURL, client.WithFileOnlyOption(true), client.WithTokenLocation(tempToken.Name()))
+					err := client.DoList(fed.Ctx, listURL, client.WithObjectOnlyOption(true), client.WithTokenLocation(tempToken.Name()))
 					require.NoError(t, err)
 					w.Close()
 				}
 			}()
-			var buf bytes.Buffer
-			_, err = io.Copy(&buf, r)
+			_, err = io.Copy(buf, r)
 			require.NoError(t, err)
 			os.Stdout = originalStdout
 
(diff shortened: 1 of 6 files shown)


joereuss12:
This also comes with a rough design doc I used for my thought process: https://docs.google.com/document/d/1bCaFpQxbmySAnv2PT4jB6SP_45I2DuMfjstV2iE0HaM/edit?usp=sharing

client/handle_http.go
```diff
@@ -2381,6 +2421,94 @@ func (te *TransferEngine) walkDirUpload(job *clientTransferJob, transfers []tran
 	return err
 }
 
+// This function performs the ls command by walking through the specified directory and printing the contents of the files
```
bbockelm, line 2424: Don't we also have a function that does nearly the same thing -- this is basically the walkDavDir without recursing.  Let's avoid the duplication.

client/handle_http.go
```diff
@@ -2381,6 +2421,94 @@ func (te *TransferEngine) walkDirUpload(job *clientTransferJob, transfers []tran
...
+				return nil
+			}
+		}
+		// Otherwise, a different error occurred and we should return it
+		return errors.Wrap(err, "failed to read remote directory")
+	}
+
+	// if the -L flag was set, we print more information
+	if flagOptions["long"] {
+		w := tabwriter.NewWriter(os.Stdout, 1, 2, 10, ' ', tabwriter.TabIndent|tabwriter.DiscardEmptyColumns)
```
bbockelm, line 2465: The output of the API function should be a data structure - or a channel of data structures - describing the listed object.

Writing to stdout should be done in the client, not the library.

client/main.go
```diff
@@ -465,6 +465,77 @@ func schemeUnderstood(scheme string) error {
 	return nil
 }
 
+// TODO function header
+func DoList(ctx context.Context, remoteObject string, flagOptions map[string]bool, options ...TransferOption) (err error) { //TODO change from TransferOption to something with ls
```
bbockelm, line 469: Don't use two separate option mechanisms -- use the `TransferOption` approach from elsewhere in the library.

cmd/object_ls.go
```diff
@@ -0,0 +1,106 @@
...
+	fileOnly, _ := cmd.Flags().GetBool("fileonly")
+
+	options := map[string]bool{
+		"long":     long,
+		"dironly":  dirOnly,
+		"fileonly": fileOnly,
+	}
+
+	//isRecursive, _ := cmd.Flags().GetBool("recursive")
+	result := client.DoList(ctx, object, options, client.WithTokenLocation(tokenLocation))
```
bbockelm, line 89: Don't force all usage to utilize tokens.

bbockelm (changes requested):
I think the split between library and client is wrong here.

The library should be returning data structures; the client should print to stdout.

client/handle_http.go
```diff
@@ -2381,6 +2421,94 @@ func (te *TransferEngine) walkDirUpload(job *clientTransferJob, transfers []tran
...
+	dirListHost, err := getDirListHost(ctx, remoteObjectUrl, namespace, directorUrl)
+	if err != nil {
+		return
+	}
+	log.Debugln("Dir list host: ", dirListHost.String())
+
+	auth := &bearerAuth{token: token}
+	client := gowebdav.NewAuthClient(dirListHost.String(), auth)
+
+	// XRootD does not like keep alives and kills things, so turn them off.
```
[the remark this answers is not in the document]
joereuss12: I think this might just be an old comment that got copy/pasted over and over. I'll work to remove it because I'm also unsure what it means and must have missed it

cmd/object_ls.go
joereuss12: This shouldn't force token usage because if `tokenLocation` is "", the code assumes there is no token and should work fine without it. I'll update the tests to show this better and that functionality works without a token on public read origins. I wrote this portion looking at the other client commands, here is `object copy`:
https://github.com/PelicanPlatform/pelican/blob/44779fc26009a451c170a786729240969ae92b31/cmd/object_copy.go#L115
https://github.com/PelicanPlatform/pelican/blob/44779fc26009a451c170a786729240969ae92b31/cmd/object_copy.go#L201-L206

joereuss12:
Addressed most of the PR comments with this latest commit. Again, still needs a little polishing but wanted to keep track of current changes.

client/handle_http.go
joereuss12: I think the current changes made the functions a decent amount different from each other that this is not needed anymore. However, let me know if you still think otherwise.

client/fed_test.go
```diff
@@ -663,3 +665,296 @@ func TestNewTransferJob(t *testing.T) {
...
+				}
+			}()
+			var buf bytes.Buffer
+			_, err = io.Copy(&buf, r)
+			require.NoError(t, err)
+			os.Stdout = originalStdout
+
+			assert.Contains(t, buf.String(), "test")
+			// permissions for test
+			assert.Contains(t, buf.String(), "drwxrwxr-x")
```
jhiemstrawisc, line 879: @bbockelm if we're working on `ls` and `stat` in a way that starts reporting directory and permissions information like this, are we moving away from the object framework and toward something closer to a filesystem? Some of this doesn't make much sense to me unless we stop saying Pelican federates objects.

client/handle_http.go
```diff
@@ -2256,6 +2243,82 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
 
 }
 
+// This function queries the director with a PROPFIND to attempt to get the 'dirListHost'. If a propfind is not allowed on the director
+// We fall back to the deprecated dirlisthost in the namespace
+func getDirListHost(ctx context.Context, remoteObjectUrl *url.URL, namespace namespaces.Namespace, directorUrl string) (dirListHost *url.URL, err error) {
```
jhiemstrawisc, line 2248: I think we're trying to get rid of references to `dirListHost`. Wherever you can, let's swap it out for `collectionsUrl`.

client/handle_http.go
```diff
@@ -2256,6 +2243,82 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
 
 }
 
+// This function queries the director with a PROPFIND to attempt to get the 'dirListHost'. If a propfind is not allowed on the director
```
jhiemstrawisc, line 2246: I don't think we need to do a `PROPFIND` to get this information. Can you replace it with a `HEAD`?

client/handle_http.go
```diff
@@ -2256,6 +2243,82 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+				dirListHost, err = url.Parse(namespace.DirListHost)
+				if err != nil {
+					return nil, err
+				}
+				return dirListHost, nil
+			}
+		} else if resp.StatusCode == http.StatusTemporaryRedirect {
+			// If the director responds with a 307 (temporary redirect), we're working with a new Director.
+			// In that event, we can get the collections URL from our redirect
+			collections := resp.Header.Get("Location")
```
jhiemstrawisc, line 2282: Why get this from the `Location` header instead of the `X-Pelican-Namespace`'s built-in `collections-url`?

client/handle_http.go
```diff
@@ -2256,6 +2243,82 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+			// The resp redirect includes the full path to the object from the origin, we are only interested in the scheme and host
+			collectionsURL, err := url.Parse(collections)
+			if err != nil {
+				return nil, err
+			}
+			dirListHost = &url.URL{
+				Scheme: collectionsURL.Scheme,
+				Host:   collectionsURL.Host,
+			}
+			namespace.DirListHost = dirListHost.String()
```
jhiemstrawisc, line 2295: When you switch over to get the `X-Pelican-Namespace`'s `collection-url`, you won't have a URL path to worry about, and you can avoid recreating the URL like this. Otherwise, I'd say you could simplify this to:
```
collectionsUrl, err := url.Parse(collections)
if err != nil {
    return nil, err
}
collectionsUrl.Path = ""
namespace.DirListHost = collectionsUrl.String()
```

client/handle_http.go
```diff
@@ -2256,6 +2243,82 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+		err = &dirListingNotSupportedError{Err: errors.New("origin and/or namespace does not support directory listings")}
+		return nil, err
+	}
+	return
+}
+
+// This helper function creates a web dav client to walkDavDir's. Used for recursive downloads and lists
+func createWebDavClient(dirListHost *url.URL, token string) (client *gowebdav.Client) {
+	auth := &bearerAuth{token: token}
+	client = gowebdav.NewAuthClient(dirListHost.String(), auth)
```
jhiemstrawisc, line 2316: Same here. Let's get rid of references to `dirListHost` wherever we're not stuck with a field in an existing struct.

client/handle_http.go
```diff
@@ -2381,6 +2439,78 @@ func (te *TransferEngine) walkDirUpload(job *clientTransferJob, transfers []tran
...
+	log.Debugln("Dir list host: ", dirListHost.String())
+
+	client := createWebDavClient(dirListHost, token)
+	remotePath := remoteObjectUrl.Path
+
+	infos, err := client.ReadDir(remotePath)
+	if err != nil {
+		// Check if we got a 404:
+		if gowebdav.IsErrNotFound(err) {
+			err = errors.Wrap(err, "file not found")
```
jhiemstrawisc, line 2458: This might actually be a case where we don't want to wrap the error. If we get a 404, we know that it's a "not found" issue, so we can tell the user that instead of whatever ugly message pops out of the web dav client.

client/handle_http.go
```diff
@@ -2381,6 +2439,78 @@ func (te *TransferEngine) walkDirUpload(job *clientTransferJob, transfers []tran
...
+		} else if gowebdav.IsErrCode(err, 500) {
```
jhiemstrawisc, line 2459: Can you replace the 500 with an `http.StatusInternalServerError`?

client/handle_http_test.go
```diff
@@ -851,3 +851,142 @@ func TestSearchJobAd(t *testing.T) {
 		assert.Equal(t, "12345", jobId)
 	})
 }
+
+func TestGetDirListHost(t *testing.T) {
+	viper.Reset()
+	defer viper.Reset()
+	config.InitConfig()
```
jhiemstrawisc, line 858: This should be isolated so the tests don't pick up user configs.

client/main.go
```diff
@@ -465,6 +478,168 @@ func schemeUnderstood(scheme string) error {
 	return nil
 }
 
+// Function for the object ls command, we get target information for our remote object and eventually print out the contents of the specified object
+func DoList(ctx context.Context, remoteObject string, options ...TransferOption) (err error) {
+	// First, create a handler for any panics that occur
+	defer func() {
+		if r := recover(); r != nil {
+			log.Debugln("Panic captured while attempting to perform transfer (DoList):", r)
```
jhiemstrawisc, line 486: This should probably be a `log.Errorln`

client/main.go
```diff
@@ -465,6 +478,168 @@ func schemeUnderstood(scheme string) error {
...
+	te := NewTransferEngine(ctx)
+	defer func() {
+		if err := te.Shutdown(); err != nil {
+			log.Errorln("Failure when shutting down transfer engine:", err)
+		}
+	}()
+
+	pelicanURL, err := te.newPelicanURL(remoteObjectUrl)
+	if err != nil {
+		return errors.Wrap(err, "Failed to generate pelicanURL object")
```
jhiemstrawisc, line 514: Can you lowercase "Failed"?

client/main.go
```diff
@@ -465,6 +478,168 @@ func schemeUnderstood(scheme string) error {
...
+	}
+
+	if ns.UseTokenOnRead && token == "" {
+		token, err = getToken(remoteObjectUrl, ns, true, "", tokenLocation, acquire)
+		if err != nil {
+			return errors.Errorf("failed to get token for transfer: %v", err)
+		}
+	}
+
+	fileInfos, err := listHttp(ctx, remoteObjectUrl, pelicanURL.directorUrl, ns, token, WithLongOption(long), WithDirOnlyOption(dironly), WithFileOnlyOption(fileonly))
```
jhiemstrawisc, line 556: It seems like `dironly` and `fileonly` are mutually exclusive. Is it worth checking somewhere that the user can't configure both? Or is it fine to include them together?

client/main.go
```diff
@@ -465,6 +478,168 @@ func schemeUnderstood(scheme string) error {
...
+			} else if !info.IsDir && dironly {
+				// If the object is a file and we have dironly (-D) set, ignore it
+				continue
+			}
+			jsonInfo = append(jsonInfo, info.Name)
+		}
+		// Convert the FileInfo to JSON and print it
+		jsonData, err := json.Marshal(jsonInfo)
+		if err != nil {
+			log.Fatalf("Error converting to JSON: %v", err)
```
jhiemstrawisc, line 603: This probably needs a return so we don't continue to print an empty buffer.

cmd/object_ls.go
```diff
@@ -0,0 +1,101 @@
...
+package main
+
+import (
+	"os"
+
+	"github.com/pelicanplatform/pelican/client"
+	"github.com/pelicanplatform/pelican/config"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
```
jhiemstrawisc, line 28: Can you split pelican imports from non pelican imports? We've been doing this elsewhere to make pelican-specific imports a bit easier to see at a glance.
```
	"github.com/pkg/errors"
	log "github.com/sirupsen/logrus"
	"github.com/spf13/cobra"
	
	"github.com/pelicanplatform/pelican/client"
	"github.com/pelicanplatform/pelican/config"
```

cmd/object_ls.go
```diff
@@ -0,0 +1,101 @@
...
+		Run:   listMain,
+	}
+)
+
+func init() {
+	flagSet := lsCmd.Flags()
+	flagSet.StringP("token", "t", "", "Token file to use for transfer")
+	flagSet.BoolP("long", "L", false, "Include extended information")
+	flagSet.BoolP("dironly", "D", false, "List directories only")
+	flagSet.BoolP("fileonly", "F", false, "List files only")
```
jhiemstrawisc, line 44: I'll point out that I worry about referring to things as "files" when the second sentence of our own client documentation states:
```
One thing to note is that Pelican should be thought of as a tool that works with federated objects as opposed to files.
```
Can we change whichever is incorrect?

cmd/object_ls.go
```diff
@@ -0,0 +1,101 @@
...
+	object := args[len(args)-1]
+
+	log.Debugln("Object:", object)
+
+	long, _ := cmd.Flags().GetBool("long")
+	dirOnly, _ := cmd.Flags().GetBool("dironly")
+	fileOnly, _ := cmd.Flags().GetBool("fileonly")
+	json, _ := cmd.Flags().GetBool("json")
+
+	result := client.DoList(ctx, object, client.WithTokenLocation(tokenLocation), client.WithLongOption(long),
```
jhiemstrawisc, line 83: Can you change this to `err`? The underlying `DoList` returns an error, and calling this result makes me think it _should_ have something on success.

jhiemstrawisc (changes requested):
One other comment I had was that this fails unexpectedly for origins that enable listings but whose object stores don't support PROPFIND. For example, if I set up a local S3 origin and use `pelican object ls` against it, I get this confusing error:
```
ERROR[2024-05-21T18:32:22Z] Failure getting /prefix/different-path: failed to read remote directory: ReadDir /prefix/different-path/: 405
```
I think we should handle the 405 internally and report to the user that object listings is not supported instead of spitting out the raw HTML status code.

client/fed_test.go
joereuss12: That is a good question, the permissions might not make too much sense in some cases but might be useful in others. Let me know what you guys think and if these permissions make sense as something to be printed out. I can easily remove this if we do not want this feature

client/handle_http.go
joereuss12: I'm not sure if a HEAD request would work, we do the PROPFIND to tell the director to redirect us straight to the origin, if we do a regular HEAD request, we might get sent to a cache. I could be wrong about this though. This helper function includes all the changes I recently made to directory listings, just made to be a helper function so other functions can access it.

client/handle_http.go
joereuss12: Good point, can do!

client/handle_http.go
joereuss12: I need to look through more to remember why, I think it is because the `collections-url` is set from the old school `dirlisthost` being set in topology. Since we want to move away from that, we just look directly at the `Location` header to get the origin (if listings are enabled) rather than relying on dirlisthost being set

client/main.go
```diff
@@ -465,6 +478,168 @@ func schemeUnderstood(scheme string) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Take our fileInfos and print them in a nice way
+	// if the -L flag was set, we print more information
+	if long {
+		w := tabwriter.NewWriter(os.Stdout, 1, 2, 10, ' ', tabwriter.TabIndent|tabwriter.DiscardEmptyColumns)
+		// If we want JSON format, we append the file info to a slice of fileInfo structs so that we can marshal it
+		if JSON {
```
joereuss12, line 566: Making a note for me here, I realized that JSON as a variable name is not good since it will be exported. Silly mistake on my end and needs to be changed

client/handle_http.go
joereuss12: The webdav client only returns this: `PROPFIND /test/fo/: 404`. I just thought things might look more user-friendly if we include just a little `object not found` or something or even rewrite the error all-together since a user might not know/care about `PROPFIND`

How about we just have this for the error (I like it better without it saying PROPFIND since it is a simple and common error): `Failure getting pelican://90ba8eb4ebfb:8444/test/thisdoesnotexist: 404: object not found`

client/main.go
joereuss12: All of the other recover functions only have `Debugln`, is this something we should change everywhere as well? I think either `Debugln` or `Errorln` could make sense but if for some reason we panic when a user is using our software and the debugs are errors, they will get a super scary error message that might not make sense to them. So I'm unsure what route we want to take?

client/main.go
joereuss12: Good point, I'll throw in a check somewhere

client/fed_test.go
```diff
@@ -851,112 +834,125 @@ func TestObjectList(t *testing.T) {
...
+	})
+
+	// Test that when both dironly and objectonly are specified, we work as normal
+	t.Run("testPelicanObjectLsDirOnlyAndObjectOnly", func(t *testing.T) {
+		// Set path for ls
+		for _, export := range fed.Exports {
+			listURL, r, w, originalStdout, buf := setupObjectListTest(export)
+			go func() {
+				if export.Capabilities.PublicReads {
+					err := client.DoList(fed.Ctx, listURL, client.WithObjectOnlyOption(true), client.WithDirOnlyOption(true))
```
jhiemstrawisc, line 871: Does this do what it should? I'd think that passing both `withObjectOnlyOption(true)` and `withDirOnlyOption(true)` should causes an error because the two are mutually exclusive.

client/main.go
```diff
@@ -529,8 +531,8 @@ func DoList(ctx context.Context, remoteObject string, options ...TransferOption)
 	token := ""
 	long := false
 	dironly := false
-	fileonly := false
-	JSON := false
+	objectonly := false
+	jsn := false
```
jhiemstrawisc, line 535: What about using something like `asJSON` for the variable?

client/main.go
```diff
@@ -543,21 +545,27 @@ func DoList(ctx context.Context, remoteObject string, options ...TransferOption)
...
-			fileonly = option.Value().(bool)
+		case identTransferOptionObjectOnly{}:
+			objectonly = option.Value().(bool)
 		case identTransferOptionJson{}:
-			JSON = option.Value().(bool)
+			jsn = option.Value().(bool)
 		}
 	}
 
+	// If a user specifies dirOnly and fileOnly, this means basic functionality (list both files and directories) so just remove the flags
```
jhiemstrawisc: I think this should be an error instead of toggling both back to false.

jhiemstrawisc (changes requested):

PR #1260, opened 2024-05-09
Fetch topology downtime for cache servers

### LINKED ISSUE #1248: Topology Downtimes not picked up by the director
When a Pelican cache needs to put in a downtime, operators still create the downtime in topology (for example, that goes into the daily monitoring reports).

However, it seems that if there are Pelican server ads, then it overwrites the downtime from topology and we continue to use the cache.  For example, check out sc-cache.chtc.wisc.edu that @matyasselmeci operates: it should be in a months-long downtime but it is actively receiving redirects.

Unlike other attributes where we always prefer the Pelican version, we should have the director enforce a downtime if either source (topology or the pelican origin/cache) has a downtime listed.

Marking as critical for 7.9.0 but I think it should be backported to 7.8.x as well.

haoming29:
Fixes #1248

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -70,6 +70,11 @@ func parseServerAd(server utils.Server, serverType server_structs.ServerType) se
 	return serverAd
 }
 
+// Do a subtraction of excludeDowned set from the includeDowned set to find cache servers
+// that are in downtime
+//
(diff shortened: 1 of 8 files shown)


jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
👍

PR #1263, opened 2024-05-10
Throw clear error when InitClient() not called

joereuss12:
The transfer engine in the client now throws a clear error if config.InitClient() has not been called. Before this fix, if InitClient() was not called we would get an error from the TransferEngine about not having worker count set or other unexpected/undesired behavior. Now, InitClient() checks if the workercount has been set (sets a default value if not) and NewTransferEngine() now throws an error if the client has not been initialized.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -186,6 +186,8 @@ var (
 		StashPrefix:   true,
 		"":            true,
 	}
+
+	clientInitialized = false
 )
 
 // This function creates a new MetadataError by wrapping the previous error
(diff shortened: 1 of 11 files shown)


config/config.go
```diff
@@ -1477,6 +1489,13 @@ func InitClient() error {
 		viper.Set("Client.MinimumDownloadSpeed", downloadLimit)
 	}
 
+	// Ensure our worker count is set
+	if viper.IsSet("Client.WorkerCount") {
+		viper.SetDefault("Client.WorkerCount", param.Client_WorkerCount.GetInt())
```
jhiemstrawisc, line 1494: I don't think this check really makes sense. If you check and find the viper variable is already set, then setting a default has no effect because it's already set and always overridden. It looks like this value is also already set in `config/resources/defaults.yaml`, so setting a viper default shouldn't have an effect anyway.

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM

Issue #1200, opened 2024-05-01
Overhaull Installation page and how other pages refer to install instructions


jhiemstrawisc:
Some goals for this page:

- Reduce some of the redundancy in our tables
- Make it easier to figure out which package to install, and when to use Docker instead
- We should also be more clear about the fact that some of our packages don't come with the entire kit and caboodle (eg .deb doesn't set up xrootd)
- Clearly reference the requirements to work around #916 in our Docker instructions.

haoming29:
The download-card related issues are resolved with https://github.com/PelicanPlatform/docs/pull/23 merged in

PR #1273, opened 2024-05-10
Update installation pages

haoming29:
Finished part of the issue #1200 

Some of the points are converted to issues in the docs repo: https://github.com/PelicanPlatform/docs/issues/17 https://github.com/PelicanPlatform/docs/issues/18


docs/pages/install/alpine.mdx
```diff
@@ -6,13 +6,15 @@ This document explains how to install Pelican on a Alpine Linux operating system
 
 You can install Pelican by downloading an `.apk` package or by downloading a binary `.tar.gz` file. You only need to follow one section below to install.
 
+> **Note**: Pelican Alpine Linux package does not contain dependencies such as `XRootD` for Pelican servers. Use [Pelican Docker image](./docker.mdx) instead if you want to serve a Pelican server.
+
 ### Install the Pelican using an APK pacakge
```
joereuss12, line 11: Typo here: pacakge instead of package

docs/pages/install/docker.mdx
```diff
@@ -49,7 +49,7 @@ docker run -it -p 8444:8444 -p 8442:8442 --name=pelican-cache hub.opensciencegri
 Where most of the command overlaps with the one to run the Pelican origin server, with the following differences:
 * `-p 8444:8444 -p 8442:8442` publishes port `8444` and `8442` from the container to the same port on the host machine. Note that the admin website of the Pelican server runs on port `8444` by default, and the objects transfer endpoint of the Pelican cache server runs on port `8442` by default.
 
-> Note that Pelican docker image currently does not support binding a directory on your host machine as the directory for Pelican cache to store cache files, i.e. using `-v /host/machine:/run/pelican/cache/location`. We will have an update once we officialy support this flow.
+> **Note**: Pelican Docker image currently does not support binding a directory on your host machine as the directory for Pelican cache to store cache files, i.e. using `-v /host/machine:/run/pelican/cache/location`. We will have an update once we officialy support this flow.
```
joereuss12, line 52:
```suggestion
> **Note**: the Pelican Docker image currently does not support binding a directory on your host machine as the directory for a Pelican cache to store cached objects, i.e. using `-v /host/machine:/run/pelican/cache/location`. We will have an update once we officially support this flow.
```

And instead of "We will have an update once we officially support this flow." Maybe say something like this:
"This feature is a work in progress and will be available in a future Pelican version"

This sounds good to me but maybe others have another suggestion as well.

docs/pages/install/docker.mdx
```diff
@@ -140,3 +140,38 @@ It is recommended that admins bind a Pelican configuration file on the host mach
...
+
+Pelican server has a built-in web server to handle various API requests. By default, it uses server hostname and a port number as the server address (e.g. `https://127.0.0.1:8444`). In a container environment, the hostname is an alphanumeric value, e.g. `6ee28c5df997`, and the corresponding Pelican server address is `https://6ee28c5df997:8444`. This value is for container internal communications and is not accessible from outside of the container. In local deployment, to access the container ports from the host machine, one can publish a port from inside the container to the host machine.
+
+```bash
+docker run -p 8444:8444 origin
+```
+
+However, Pelican server still recognize `6ee28c5df997` as its hostname, causing a mismatch.
+
+In practice, there is usually a reverse-proxy service (such as Nginx, Traefic, etc.), to direct traffic from a human-readble domain name (e.g. https://example-origin.org) to your container (e.g. `6ee28c5df997:8444`). If Pelican server doesn't catch this information and uses it as its server address, there will be a mismatch when the server advertise itself to other Pelican services.
```
joereuss12, line 156:
```suggestion
In practice, there is usually a reverse-proxy service (such as Nginx, Traefic, etc.), to direct traffic from a human-readable domain name (e.g. https://example-origin.org) to your container (e.g. `6ee28c5df997:8444`). If Pelican server doesn't catch this information and uses it as its server address, there will be a mismatch when the server advertise itself to other Pelican services.
```

docs/pages/install/docker.mdx
```diff
@@ -140,3 +140,38 @@ It is recommended that admins bind a Pelican configuration file on the host mach
...
+
+Therefore, to keep things consistent, Pelican internally should use the same human-readble domain name as its web server's _external address_. To do so, configure `Server.ExternalWebURL` to the domain name the outside world uses to access the Pelican web service, e.g. https://example-origin.org:8444 for a production service or https://localhost:8444 for a local deployment.
```
joereuss12, line 158:
```suggestion
Therefore, to keep things consistent, Pelican internally should use the same human-readable domain name as its web server's _external address_. To do so, configure `Server.ExternalWebURL` to the domain name the outside world uses to access the Pelican web service, e.g. https://example-origin.org:8444 for a production service or https://localhost:8444 for a local deployment.
```

docs/pages/install/docker.mdx
```diff
@@ -140,3 +140,38 @@ It is recommended that admins bind a Pelican configuration file on the host mach
...
+
+Notice that there is a port number, `8444`, following the hostname in the URLs above. This is the default port for Pelican web server. When publishing ports from the container, you may choose a different port on the host machine to publish to. Two convential options are `443` or `8080`, which are default ports for a web service. If you publish `8444` port from the container to `443`, or `8080` on the host machine, i.e. `docker run -p 8444:443`, you may remove the port number from `Server.ExternalWebURL`, i.e. `https://example-origin.org`. For other ports, you need to leave the port number in the URL.
```
joereuss12, line 160:
```suggestion
Notice that there is a port number, `8444`, following the hostname in the URLs above. This is the default port for Pelican web servers. When publishing ports from the container, you may choose a different port on the host machine to publish to. Two conventional options are `443` or `8080`, which are default ports for a web service. If you publish `8444` port from the container to `443`, or `8080` on the host machine, i.e. `docker run -p 8444:443`, you may remove the port number from `Server.ExternalWebURL`, i.e. `https://example-origin.org`. For other ports, you need to leave the port number in the URL.
```

docs/pages/install/macos.mdx
```diff
@@ -2,36 +2,74 @@
 
 This document explains how to install Pelican on macOS.
 
-### Install Pelican as a standalone binary
+## Install Pelican as a standalone binary
 
 1. Navigate to [Pelican download page](../install.mdx) and select the Pelican macOS binary you want to install.
 
-2. In **Operating System** section, select **macOS**. In **Architectures** section, select **X84_64** (Intel) or **ARM64** (ARM) depending on the architecture of your machine.
+2. In **Operating System** section, select **macOS**. In **Architectures** section, select **AMD64** or **ARM64** depending on the architecture of your machine.
```
joereuss12, line 9:
```suggestion
2. In the **Operating System** section, select **macOS**. In **Architectures** section, select **AMD64** or **ARM64** depending on the architecture of your machine.
```

docs/pages/install/rhel.mdx
```diff
@@ -10,19 +10,17 @@ You can install Pelican from the standalone RPM, or with the binary `.tar.gz` fi
 
 1. Navigate to [Pelican download page](../install.mdx) and select the Pelican RPM you want to install.
 
-2. In **Operating System** section, select **Linux**. In **Architechtures** section, select **X84_64** (Intel/AMD) or **AARCH64** (ARM) depending on the architecture of your machine.
+2. In **Operating System** section, select **Linux**. In **Architechtures** section, select **AMD64** or **ARM64** depending on the architecture of your machine.
```
joereuss12, line 13:
```suggestion
2. In the **Operating System** section, select **Linux**. In **Architechtures** section, select **AMD64** or **ARM64** depending on the architecture of your machine.
```

docs/pages/install/rhel.mdx
```diff
@@ -32,9 +30,9 @@ You can install Pelican from the standalone RPM, or with the binary `.tar.gz` fi
 
 1. Navigate to [Pelican download page](../install.mdx) and select the Pelican RPM you want to install.
 
-2. In **Operating System** section, select **Linux**. In **Architechtures** section, select **X84_64** (Intel/AMD) or **AARCH64** (ARM) depending on the architecture of your machine.
+2. In **Operating System** section, select **Linux**. In **Architechtures** section, select **AMD64** or **AMR64** depending on the architecture of your machine.
```
joereuss12, line 33:
```suggestion
2. In the **Operating System** section, select **Linux**. In the **Architechtures** section, select **AMD64** or **AMR64** depending on the architecture of your machine.
```

joereuss12 (changes requested):
Here are my comments on an initial read through. Also, you seem to go back and forth between using i.e. and e.g., I think we should try to stay consistent and use one or the other. I'm not sure how important this is but it is at least a personal preference for me.

docs/pages/install.mdx
```diff
@@ -2,44 +2,35 @@ import DownloadsComponent from "/components/DownloadsComponent";
...
 # Install Pelican
 
-This document lists the operating system requirements to install Pelican and explains how a user can download Pelican client or Pelican server components.
+This document lists the operating system requirements for Pelican and explains how a user can download and install Pelican.
+
+## Before Starting
+
+Pelican executables can run as a **client** or a **server**.
+
+For client usage, such as accessing the data, you can download and install a Pelican package or Pelican binary [following the instructions below](#install-pelican-binary).
```
jhiemstrawisc, line 11: Maybe "To use the client for interacting with data, download and install the Pelican package or Pelican Binary"?

docs/pages/install.mdx
```diff
@@ -2,44 +2,35 @@ import DownloadsComponent from "/components/DownloadsComponent";
...
+
+For server usage, such as serving your data to a Pelican data federation or start a new Pelican data federation, you need to run Pelican using [Pelican Docker images](./install/docker.mdx). It is not recommended to run Pelican server with a binary. Although possible, additional software dependencies and configurations are required to allow the server to be functional.
```
jhiemstrawisc, line 13: Small typo -- "... to a Pelican data federation or _to_ start a new Pelican ..."

docs/pages/install.mdx
```diff
@@ -48,3 +39,44 @@ For example: 7.1.2
...
+1. Run the following command after you installed Pelican RPM
+
+    ```bash
+    which pelican
+    ```
+
+    It should output
+
+    ```console
+    /bin/pelican
```
jhiemstrawisc, line 68: I don't think this will be true if they install the client on Windows.

docs/pages/install.mdx
```diff
@@ -48,3 +39,44 @@ For example: 7.1.2
...
+    ```
+
+2. Test Pelican client function by running an **object copy** command that download a test file from [Open Science Data Federation (OSDF)](https://osg-htc.org/services/osdf) to your current directory
```
jhiemstrawisc, line 71: Small typo -- "**object copy** command that download***s***"

docs/pages/install.mdx
```diff
@@ -48,3 +39,44 @@ For example: 7.1.2
...
+
+    ```bash
+    $ pelican -f osg-htc.org object copy /ospool/uc-shared/public/OSG-Staff/validation/test.txt .
```
jhiemstrawisc, line 74: I think I'm in favor of promoting pelican-style URLs wherever we can. Can you switch the command to:
```
$ pelican object copy pelican://osg-htc.org/ospool/uc-shared/public/OSG-Staff/validation/test.txt .
```

docs/pages/install/docker.mdx
```diff
@@ -11,7 +11,7 @@ Pelican builds separate images for each Pelican server components, e.g. origin,
 * Pelican director server: `hub.opensciencegrid.org/pelican_platform/director:latest`
 * Pelican registry server: `hub.opensciencegrid.org/pelican_platform/registry:latest`
 
-> Note: the `latest` tag will pull the Pelican server image with the latest released version. You may pull an image with explicit Pelican version by passing the version as the tag (e.g. `v7.6.4`) instead. For a list of available tags, refer to [Pelican Harbor repository](https://hub.opensciencegrid.org/harbor/projects/883/repositories/origin/artifacts-tab?publicAndNotLogged=yes).
+> **Note**: the `latest` tag will pull the Pelican server image with the latest released version. You may pull an image with explicit Pelican version by passing the version as the tag (e.g. `v7.6.4`) instead. For a list of available tags, refer to [Pelican Harbor repository](https://hub.opensciencegrid.org/harbor/projects/883/repositories/origin/artifacts-tab?publicAndNotLogged=yes).
```
jhiemstrawisc, line 14: This Harbor URL brings users to a page with a big scary red banner that says "Unauthorized". Presumably we have an image for every tagged/released version, so maybe we can point them to the github Releases page instead?

docs/pages/install/docker.mdx
jhiemstrawisc: Following up on Joe's `e.g` versus `i.e.` comment, they mean two different things and aren't interchangeable so we can't pick one over the other. The way I remember which to use is that "i.e." is Latin for "id est" and means "that is" whereas "e.g." is Latin for "exempli gratia" and means "for example". The former is to clarify something when there's only one way it can be clarified, and the latter is for any time you can construct an arbitrary example. In this case I think "e.g." is correct, but things look correct in most of the other usage.

docs/pages/install/docker.mdx
```diff
@@ -140,3 +140,38 @@ It is recommended that admins bind a Pelican configuration file on the host mach
...
+### Error running Pelican Docker image on macOS with a M-series chip
+
+When running Pelican Docker image on macOS with a M-series chip, you may get the following error message:
+
+```console
+docker: no matching manifest for linux/arm64/v8 in the manifest list entries.
+See 'docker run --help'
+```
+
+This is because Pelican Docker images are built for Linux/AMD64 architecture only. To fix the issue, run the image with `--platform linux/amd64` instead:
```
jhiemstrawisc, line 173: Let's add a note that this isn't a great idea for production servers because the x86 emulation makes things much less efficient.

docs/pages/install/macos.mdx
```diff
@@ -2,36 +2,74 @@
...
+    Build Commit: d260a07d3b057d19b7fdd36125f91a8768531258
+    Built By: goreleaser
+    ```
+
+    * Add Pelican binary to your `PATH` permanently
+
+    ```bash
+    $ cd pelican_Darwin_arm64 # Go to the binary folder
+    $ echo 'export PATH=$PWD:$PATH' >> ~/.bashrc # Add the current folder to your .bashrc file
+    $ source ~/.bashrc # Apply the change
```
jhiemstrawisc, line 61: My mac says the default shell is now `zsh`. Are changes to `~/.bashrc` picked up by zsh, or do we need to include an additional set of instructions? (the only difference would be adding it to `~/.zshrc` instead)

jhiemstrawisc (changes requested):

docs/pages/install.mdx
aowen-uwmad, line 13:
```suggestion
For server usage, such as serving your data to a Pelican data federation or start a new Pelican data federation, you need to run Pelican using [Pelican Docker images](./install/docker.mdx). It is not recommended to run Pelican server with a binary. Although technically possible, additional software dependencies and configurations are required to make the binary functional for server usage. 
```

docs/pages/install.mdx
```diff
@@ -2,44 +2,35 @@ import DownloadsComponent from "/components/DownloadsComponent";
...
-Pelican supports the following operating systems:
+Pelican supports the following operating systems, when running as a **client**:
 
 * [Red Hat Enterprise Linux (CentOS, Fedora, Alma, Rocky, or openSUSE)](./install/rhel.mdx)
 * [Debian or Ubuntu](./install/debian.mdx)
 * [Alpine Linux](./install/alpine.mdx)
-* [macOS](./install/macos.mdx) [*]
-* Windows [*]
+* [macOS](./install/macos.mdx)
+* Windows
```
aowen-uwmad, line 24: Is Windows supported only via Docker image? If so, then maybe 

```suggestion
* Windows ([via Docker](./install/docker.mdx))
```

docs/pages/install.mdx
aowen-uwmad: I think this depends on how the client is "installed" on Windows. 
From the old version, it looks like using on Windows required using a Docker container.
So then the question is whether the Docker container being launched is still a Linux environment; if so, then these instructions are fine.

docs/pages/install.mdx
aowen-uwmad: I second this suggestion.

docs/pages/install.mdx
```diff
@@ -48,3 +39,44 @@ For example: 7.1.2
 - 2 represents a bug fix/patch release
 
 The download section above always shows the latest released version of Pelican and we recommend you to use the latest feature or major release version so it includes the latest bug fixes and features. If you would like to download a previous version, please refer to our [GitHub release page](https://github.com/PelicanPlatform/pelican/releases).
```
aowen-uwmad, line 41: Is this statement still true with the newer version of the Downloads table?

docs/pages/install/docker.mdx
```diff
@@ -1,6 +1,6 @@
 # Run Pelican Server with Docker Image
 
-This document explains how to run a Pelican server using Pelican docker image. If you are installing Pelican to use its client functionalities (download or upload an object), refer to [previous sections](../install.mdx) to download and install a Pelican binary instead.
+This document explains how to run a Pelican server using Pelican docker image. If you want to use Pelican client functionalities, such as download or upload an object, refer to [previous sections](../install.mdx) to download and install a Pelican binary instead.
```
aowen-uwmad, line 3: Currently this comment suggests that you can't use the Docker image for client functionality; I think this is incorrect.

```suggestion
This document explains how to run a Pelican server using Pelican docker image. For Pelican client functionalities, such as download or upload an object, you can use the docker image as well, or you can install the Pelican binary instead as described [elsewhere](../install.mdx).
```

docs/pages/install/docker.mdx
aowen-uwmad: I'm also not a fan of the scary red banner. That probably requires something to change on the Harbor side. 

Maybe link to a document (in Github?) that tables the versions and the Harbor tags / docker addresses? That way people can find the docker addresses without encountering the scary red banner.

docs/pages/install/docker.mdx
```diff
@@ -69,7 +69,7 @@ To run the latest pelican registry server, run the following command:
 docker run -it -p 8444:8444 --name=pelican-registry hub.opensciencegrid.org/pelican_platform/registry:latest serve -f <federation>
 ```
 
-> Note that to successfully run a Pelican registry server, additional configuration is required. Follow [Serve a Pelican Federation](../serving_a_federation.mdx) for instructions.
+To successfully run a Pelican registry server, additional configuration is required. Follow [Serve a Pelican Registry](../serving_a_federation.mdx#serve-a-registry) for instructions. For how to pass configurations to Docker container, refer to [the next section](#configure-pelican-server-in-a-container).
```
aowen-uwmad, line 72: Should this statement go before you show the example command for running the director container? 
My concern is that people will see and try to run the command before reading the note that it won't work as it is.

docs/pages/install/docker.mdx
```diff
@@ -140,3 +140,38 @@ It is recommended that admins bind a Pelican configuration file on the host mach
 
 
 4. You should note that your running Pelican origin server is logging debug level messages.
+
+## Additional Configurations for Running Pelican in a Container
+
+### Server hostname and port
+
+Pelican server has a built-in web server to handle various API requests. By default, it uses server hostname and a port number as the server address (e.g. `https://127.0.0.1:8444`). In a container environment, the hostname is an alphanumeric value, e.g. `6ee28c5df997`, and the corresponding Pelican server address is `https://6ee28c5df997:8444`. This value is for container internal communications and is not accessible from outside of the container. In local deployment, to access the container ports from the host machine, one can publish a port from inside the container to the host machine.
```
aowen-uwmad, line 148: Perhaps link out to the Docker network documentation (https://docs.docker.com/network/) for more information on how the Docker networking works?

docs/pages/install/docker.mdx
haoming29: Thanks Justin for the explanation!

docs/pages/install.mdx
haoming29: For Windows users, I'd like to have someone test the installation of the Windows binary then decide how I should revise the instruction. I've separate verification for Windows from Linux/macOS, and use `where` command for Windows, but I want to make sure that the Windows installer adds the binary to PATH, otherwise I'll update instruction for it.

docs/pages/install/docker.mdx
haoming29: I'd suggest change the GitHub release note to include links to various Docker images, or have a page on our docs/Pelican website listing available containers. @CannonLock What's your thought on this?

docs/pages/install/macos.mdx
haoming29: Thanks for reminding me! I'll do `~/.zshrc` instead. I copied that from Linux but I think macOS uses zsh years ago so I'll default to zsh

docs/pages/install.mdx
haoming29: It's a complicated story. We do produce Windows binary and user can download it, but most of our devs don't use Windows machine so I'm not confident instructing user to do it. I'll try find a Windows machine and fill out that page.

docs/pages/install.mdx
haoming29: I'll update this expression here

docs/pages/install/docker.mdx
haoming29: They can, but our Docker images aren't set up in a way to facilitate client usage, and we don't recommend this behavior.

haoming29:
@jhiemstrawisc and @aowen-uwmad I've updated the docs addressing all the feedback. I've also included new page for installing Pelican on Windows so you might be interested in taking a look at it.

docs/pages/install.mdx
jhiemstrawisc: IIRC, we at least try to make sure Pelican is installable on Windows because the client is supposed to work across platforms. All the servers are packed into the same binary, so anyone who wants to run more than `pelican object get/put`-type commands will need the container.

jhiemstrawisc (changes requested):
All minor comments at this point, I think we're close! I'm not knowledgeable enough about doing things via PowerShell to provide useful feedback for the new Windows installation section.

docs/pages/install/linux_binary.mdx
```diff
@@ -0,0 +1,70 @@
...
+Build Date: 2024-03-01T18:13:00Z
+Build Commit: d260a07d3b057d19b7fdd36125f91a8768531258
+Built By: goreleaser
+```
+
+### Add Pelican binary to your `PATH` for the current terminal
+
+```bash
+$ cd pelican_Linux_arm64 # Go to the binary folder
+$ export PATH=$PWD:$PATH # Add current folder to the PATH
```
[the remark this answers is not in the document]
haoming29: Did a quick search over docs, we prepend to a user's `$PATH`

jhiemstrawisc (approved):
@haoming29 I think all of the changes I asked for were made, tentatively approving and you can decide if you want to wait for other reviewers as well.

docs/pages/install/docker.mdx
aowen-uwmad: Ah, got it. 
Totally did not override the entrypoint of the container, add pelican to the PATH, and use that as a client to download a test file. That would be silly..

docs/pages/install/linux_binary.mdx
```diff
@@ -0,0 +1,70 @@
...
+Build Date: 2024-03-01T18:13:00Z
+Build Commit: d260a07d3b057d19b7fdd36125f91a8768531258
+Built By: goreleaser
+```
+
+### Add Pelican binary to your `PATH` permanently
+
+```bash
+$ cd pelican_Linux_arm64 # Go to the binary folder
+$ echo 'export PATH=$PWD:$PATH' >> ~/.bashrc # Add the current folder to your .bashrc file
```
aowen-uwmad, line 62: Have to be careful with this expression. Currently, because the string is wrapped with `'`, no substitution of the variables will occur, so this will add the literal line `export PATH=$PWD:$PATH` to the `.bashrc` file, which won't be helpful.

We can wrap the string in quotations `"` to make sure the value of $PWD is substituted, but we don't want the value of $PATH substituted, because the current value of $PATH may be considerably altered from the value of $PATH after sourcing the `.bashrc` file. (For example, if in the previous step the user ran `export PATH=$PWD` instead of copying/pasting..)
We can prevent the value of $PATH from being substituted by escaping the dollar sign with a backslash.

Altogether, I suggest this instead:

```suggestion
$ echo "export PATH=$PWD:\$PATH" >> ~/.bashrc # Add the current folder to your .bashrc file
```

Things get complicated if we also want that value wrapped in quotations. Could do 

```
echo "export PATH=\"$PWD:\$PATH\"" >> ~/.bashrc
```

but that's starting to get messy..

docs/pages/install/macos.mdx
```diff
@@ -53,12 +55,12 @@ This document explains how to install Pelican on macOS.
...
     ```
 
-    * Add Pelican binary to your `PATH` permanently
+  * Add Pelican binary to your `PATH` permanently
 
     ```bash
     $ cd pelican_Darwin_arm64 # Go to the binary folder
-    $ echo 'export PATH=$PWD:$PATH' >> ~/.bashrc # Add the current folder to your .bashrc file
-    $ source ~/.bashrc # Apply the change
+    $ echo 'export PATH=$PWD:$PATH' >> ~/.zshrc # Add the current folder to your .zshrc file
```
aowen-uwmad, line 62: See note in `linux_binary.mdx` for more information.

```suggestion
    $ echo "export PATH=$PWD:\$PATH" >> ~/.zshrc # Add the current folder to your .zshrc file
```

Might want someone to test that this works as expected.

docs/pages/install/windows.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+## Install Pelican as a standalone binary
+
+Pelican provides a binary file instead of a `MSI` installer for Windows. You need to download and extract the binary and add the binary to your `PATH` environment variable.
+
+1. Navigate to [Pelican download page](../install.mdx#download-pelican-binary) and select the Pelican Windows binary.
+
+2. In the **Operating System** section, select **Windows**. Copy the link to `pelican_Windows_x86_64.zip`.
+
+3. Download the file to your computer and extract the binary from the `.zip` file.
```
aowen-uwmad, line 13:
```suggestion
2. In the **Operating System** section, select **Windows**. Download the file `pelican_Windows_x86_64.zip` to your computer.

3. Using Windows Explorer, navigate to folder containing the downloaded `.zip` file and extract it.
```

docs/pages/install/windows.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+4. You may run the binary in the extracted folder, but it is recommended that you add Pelican binary to your `PATH` environment variable to allow `pelican` to be called directly from your PowerShell. To run Pelican binary from the extracted folder:
```
aowen-uwmad, line 15:
```suggestion
4. You may run the binary in the extracted folder, but it is recommended that you add Pelican binary to your `PATH` environment variable to allow `pelican` to be called directly from your PowerShell. To run Pelican binary from the extracted folder, do the following in PowerShell:
```

docs/pages/install/windows.mdx
```diff
@@ -0,0 +1,82 @@
...
+## Add Pelican binary to your `PATH`
+
+* Add Pelican binary to your `PATH` for the current PowerShell
+
+  ```bat
+  rem Go to the binary folder
+  $ cd pelican_Linux_arm64
+
+  rem Add current folder to the PATH
+  $ set PATH=%PATH%;%CD%
```
aowen-uwmad, line 39: This did not work for me. I do not use PowerShell, so I don't know what the correct command is. Could also be a version issue. I'm using

```
Edition	Windows 11 Pro
Version	23H2
Installed on	‎2/‎14/‎2023
OS build	22631.3593
Experience	Windows Feature Experience Pack 1000.22700.1003.0
```

and

```
PS C:\Users\a_owe\Downloads> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.22621.2506
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.22621.2506
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

docs/pages/install/windows.mdx
```diff
@@ -0,0 +1,82 @@
...
+      * In the Environment Variables window, under the `System variables` section, scroll down and find the `Path` variable, then select it and click `Edit`.
+      * In the Edit Environment Variable window, click `New` and add the path to the directory containing `pelican.exe` (e.g., `C:\path\to\directory`).
+      * Click `OK` to close all windows.
+
+   4. Verify Pelican is added to `PATH`
+      * In a new PowerShell window, run the following command:
+
+        ```bat
+        rem Run Pelican binary
+        $ pelican --version
```
aowen-uwmad, line 69: Should this be ?

```suggestion
        $ pelican.exe --version
```

aowen-uwmad (changes requested):

docs/pages/install/linux_binary.mdx
haoming29: Thanks for catching this!

docs/pages/install/windows.mdx
haoming29: I'm fairly sure you won't need to add `.exe` if the executable is already in the `PATH`, I'll double check once I get access to my Windows machine

haoming29: I verified that if I added pelican to `PATH`, I can evoke pelican by either `pelican` or `pelican.exe`. I'll leave this as-is here to be consistent with the previous instructions.

haoming29:
@aowen-uwmad Could you take another look at the changes? I've addressed the PATH issue and tested on my Windows machine

jhiemstrawisc:
That I can tell, all requested changes have been made. Merging so this doesn't get blocked.

PR #1276, opened 2024-05-13
Aggregate gin routes for Prometheus metrics

### LINKED ISSUE #1235: Aggregate gin routes for Prometheus metrics
We currently have very high cardinality for gin-related Prometheus metrics, especially at the director side, where it need to handle a object requests. This can slow down Prometheus and increase memory/disk usage of Prometheus. Also, the frontend loads many static resources which are spamming the metric and should be aggregated to `/_next` level.

haoming29:
Closes #1235 

To test, run your director server, login to web UI, then go to `/api/v1.0/prometheus/query?query=gin_requests_total` and you should be able to see the aggregated time series where the `path` label has values ended with `/:xxx`

### DIFF
--- a/web_ui/ui.go
+++ b/web_ui/ui.go
@@ -36,21 +36,21 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/pelicanplatform/pelican/config"
-	"github.com/pelicanplatform/pelican/server_structs"
-	"github.com/pelicanplatform/pelican/server_utils"
-	"github.com/spf13/viper"
-
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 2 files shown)


haoming29:
@turetske We should backport this to 7.8 once it's merged. When I was testing this, I noticed that there is a small cardinality explosion at the ITB director Prometheus instance, which bumps Prometheus memory usage a lot. I hope this fix can mitigate that memory increase

jhiemstrawisc (approved):
LGTM

jhiemstrawisc:
@turetske Now that this is merged, can you patch it when you get the chance?

PR #1284, opened 2024-05-14
Fix issuer for self-monitoring and scitokens issuer name

### LINKED ISSUE #1275: Pelican origin/cache uses external issuer for self-tests
Pelican Version:

Pelican Service:
- Origin
- Cache

**Describe the bug**

Pelican allows user to use an external issuer for authentication/authorization. However, when doing so, the issuer for self-testing is also changed to the external issuer, which invalidate token for self-tests because that's generating with the internal issuer of the server.

We should leave the internal issue as-is if the user set an external issuer.

haoming29:
Fixes #1275

### DIFF
--- a/origin/self_monitor.go
+++ b/origin/self_monitor.go
@@ -22,21 +22,18 @@ import (
 	"context"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+
 	"github.com/pelicanplatform/pelican/config"
 	"github.com/pelicanplatform/pelican/metrics"
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/server_utils"
-	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


turetske (changes requested):
Can you create a test to show that the issuers are generated correctly in xrootd/authorization_test.go?

haoming29:
> Can you create a test to show that the issuers are generated correctly in xrootd/authorization_test.go?

There's existing test coverage and I updated the test file to reflect the change: `xrootd/resources/test-scitokens-monitoring.cfg`

turetske:
> > Can you create a test to show that the issuers are generated correctly in xrootd/authorization_test.go?
> 
> There's existing test coverage and I updated the test file to reflect the change: `xrootd/resources/test-scitokens-monitoring.cfg`

Ah, okay. I missed that somehow.

turetske (approved):
LGTM! Ran it locally just to be sure.

PR #1288, opened 2024-05-15
Bump base os to el9, use osg 23 main repo, and fix maven vulnerability

### LINKED ISSUE #1199: Update published containers from CentOS 8 to CentOS 9
@aowen-uwmad reports in discussion #1185 that our containers are running on CentOS 8 stream, which has an upcoming EOL (May 31, 2024 if I'm not mistaken). We should move our published containers to CentOS 9, which should be available through the OSG repos.

haoming29:
Closes #1199 

For the rest of vulnerabilities, they are coming from the `scitoken-oauth-server` image we inherit, essentially the OA4MP software we depend on. I've opened an issue in the OA4MP repo for fixing those vulnerabilities: https://github.com/ncsa/oa4mp/issues/183

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -17,7 +17,8 @@
 # ***************************************************************
 
 ARG BASE_YUM_REPO=release
-ARG BASE_OSG_SERIES=3.6
+ARG BASE_OSG_SERIES=23
+ARG BASE_OS=el9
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (approved):
LGTM

PR #1291, opened 2024-05-15
Add http backend to origin data export API

### LINKED ISSUE #1180: Include https backend in the origin export API for the UI
We added support for origins to take https storage backend recently #958 , and we want to expose this type of backend to the origin export API and correctly display the related values

@CannonLock , since S3/POSIX/http or the future Globus backend each has a different combination of `Origin.Exports` entries and other configurations to set up (i.e. `Origin.S3Bucket` for s3 backend and `Origin.HttpServiceUrl` for `http` backend), do we want to render different panels based on different storage type?

haoming29:
Closes #1180 

This added extra fields to the `/origin_ui/exports`, additional UI work required to render the data.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -477,7 +477,7 @@ description: |+
     ```
 
   There are additional configuration fields if Origin.StorageType == "s3".
-  - S3Bucket: [REQUIRED] See `Origin.S3Bucket` for details
+  - S3Bucket: [OPTIONAL] See `Origin.S3Bucket` for details
   - S3AccessKeyfile: [OPTIONAL] See `Origin.S3AccessKeyfile` for details
   - S3SecretKeyfile: [OPTIONAL] See `Origin.S3SecretKeyfile` for details
 
@@ -776,7 +776,7 @@ components: ["origin"]
 name: Origin.S3ServiceName
 description: |+
   [Deprecated] Origin.S3ServiceName was previously used in part to determine an export's FederationPrefix, but
-  upstream changes no longer rely on this value. As of Pelican 7.7.0, setting this value no longer has any effect.
+  upstream changes no longer rely on this value. As of Pelican `7.7.0`, setting this value no longer has any effect.
   AWSv4 signatures used by S3 servers to handle authentication now hardcode "s3" as their service name.
 
   When constructing signed URLs for S3, this value is used as a part of the signature. It is almost always "s3". For more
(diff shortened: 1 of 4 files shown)


haoming29:
@jhiemstrawisc Let's hold off merging this PR until your #1285 PR gets merged in. I want to include `xroot` backend in this pass as well.

swagger/pelican-swagger.yaml
```diff
@@ -476,7 +476,19 @@ definitions:
       type:
         type: string
         example: "posix"
-        description: "The storage type of the origin server. Is either \"posix\" or \"s3\""
+        description: "The storage type of the origin server. Values can be posix, s3, and https"
+      s3Region:
+        type: string
+        description: "The server region for AWS S3 storage. Only available when type == s3"
```
jhiemstrawisc, line 482: Let's remove references to AWS, since this is a core part of the protocol (as opposed to the service offered by Amazon)

swagger/pelican-swagger.yaml
```diff
@@ -476,7 +476,19 @@ definitions:
...
+      s3ServiceUrl:
+        type: string
+        description: "The URL that provides API access the the S3 objects. Only available when type == s3"
+      s3UrlStyle:
+        type: string
+        description: "The style of S3 urls used by the service URL host. Can be either \"path\" or \"virtual\". Only available when type == s3"
```
jhiemstrawisc, line 488: Is it worth providing an example of the difference between the two here, or are you happy with what's in our docs elsewhere?

swagger/pelican-swagger.yaml
```diff
@@ -476,7 +476,19 @@ definitions:
       type:
         type: string
         example: "posix"
-        description: "The storage type of the origin server. Is either \"posix\" or \"s3\""
+        description: "The storage type of the origin server. Values can be posix, s3, and https"
```
jhiemstrawisc, line 479: Not a question we need to answer in this PR, but I wonder what our long-term strategy is for enumerating backends in places like this. Ideally we want to add many more backends, but it seems like we list available options manually in a variety of locations. It might be hard to keep things up to date if we don't come up with a solution.

docs/parameters.yaml
```diff
@@ -475,6 +475,12 @@ description: |+
         Capabilities: ["Reads", "PublicReads", "Writes", "Listings", "DirectReads"]
         SentinelLocation: demoproject_origin_A
     ```
+
+  There are additional configuration fields if Origin.StorageType == "s3".
+  - S3Bucket: [REQUIRED] See `Origin.S3Bucket` for details
```
jhiemstrawisc, line 480: Ahh -- S3Bucket isn't actually required. When we want to export an entire endpoint like amazon's open data registry, we can't manually enumerate all of the buckets. In that case we get around it by omitting the bucket (which then becomes a part of the object name)

jhiemstrawisc (changes requested):

swagger/pelican-swagger.yaml
haoming29: I think this is mostly a rough intro for people looking up the API for what they'll get out of that endpoint. If they are interested in learning what exactly the value is for, they can refer to the docs website. I'll link to it

swagger/pelican-swagger.yaml
haoming29: Yeah I had the same question. I don't know how we can sync Swagger docs with `parameters.yaml` and other places, but I'd be nice to document what to change for an enum type configuration

jhiemstrawisc (approved):
LGTM

PR #1294, opened 2024-05-16
Client de-duplicate cache entries from director

joereuss12:
The Pelican client now will look out for duplicate caches to try returned from the director. This is done with a new helper function that will get the caches to try and check for any duplicates. If we find a duplicate, it will not be added to the list of the caches to try. In addition, there are various cleanups of old debug logging as well as some new logging that seems cleaner without removing anything important.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -1265,7 +1265,6 @@ func newTransferDetails(cache namespaces.Cache, opts transferDetailsOptions) []t
 		cacheURL.Scheme = ""
 		cacheURL.Opaque = ""
 	}
-	log.Debugf("Parsed Cache: %s", cacheURL.String())
 	if opts.NeedsToken {
 		if cacheURL.Scheme != "unix" {
 			cacheURL.Scheme = "https"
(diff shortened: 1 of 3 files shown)


client/handle_http.go
```diff
@@ -1336,6 +1335,39 @@ func generateTransferDetailsUsingCache(cache CacheInterface, opts transferDetail
...
+			caches = append(caches, cache)
+			cacheList[cache] = true
+			td := transferDetailsOptions{
+				NeedsToken: job.namespace.ReadHTTPS || job.namespace.UseTokenOnRead,
+				PackOption: packOption,
+			}
+			transfers = append(transfers, generateTransferDetailsUsingCache(cache, td)...)
+		}
+	}
+	log.Debugln("Trying the caches:", caches)
```
joereuss12, line 1367: @jhiemstrawisc since you are going to review this PR, I know you said you wanted this debug statement gone but IMO I like this debug line. It's nice to see the specific caches we are trying outside of the massive list we get printed from the director. Then later on in the client code, we print out what specific cache we are tackling before we attempt a transfer.

jhiemstrawisc: As luck would have it, Brian is overhauling most of these error messages in #1301. We'll get your changes in first to make sure his changes are the final say.

jhiemstrawisc (approved):
LGTM

PR #1295, opened 2024-05-17
Display legacy namespace registrations in the registry web UI

### LINKED ISSUE #1220: Registry UI doesn't display legacy namespace registrations
In our production registry, we had a bunch of legacy namespace registrations pre `AdminMetadata` was introduced, and we didn't show them in the UI at all:

https://github.com/PelicanPlatform/pelican/blob/1827884b79e648bad72e2afbe595c56496e6685b/registry/registry_db.go#L356-L364

We should treat those registrations as `pending` registrations.

haoming29:
Fixes #1220 

This is built upon the existing `GET /registry_ui/namespaces` API with an additional query parameter `legacy`. By default the API ignores any legacy registrations, but with `?legacy`, it will only display the legacy registrations. Note that you may pass the parameter either in `legacy` or `legacy=true`.
 
@CannonLock  The idea for the UI is to render a list of legacy registrations as a separate table and allow registry admin to edit/delete those registrations.

### DIFF
--- a/registry/custom_reg_fields.go
+++ b/registry/custom_reg_fields.go
@@ -163,7 +163,7 @@ func convertCustomRegFields(configFields []customRegFieldsConfig) []registration
 }
 
 // Helper function to exclude pubkey field from marshaling into json
-func excludePubKey(nss []*server_structs.Namespace) (nssNew []NamespaceWOPubkey) {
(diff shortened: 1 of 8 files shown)


jhiemstrawisc (approved):
LGTM

PR #1298, opened 2024-05-17
Check uniqueness of option names for custom registration fields

### LINKED ISSUE #1231: Add name uniqueness as check for options enumerates
_A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]_

Add a check to make sure that options have unique names. 

The issue rises from the fact that from a user perspective it makes no sense to have to names be the same value when the ID value is likely a primary key int/hash.

haoming29:
Closes #1231


turetske (approved):
LGTM!

PR #1301, opened 2024-05-19
Report cache hits/misses

bbockelm:
This PR tracks whether there was a cache hit or miss based on the `Age` header in the server response.  The goal is to provide some indication to users in the hold message as to why their transfer failed -- was it a slow origin or a slow cache?  (The assumption is if it was a cache hit and a slow transfer then the cache is at fault.)

Also in this PR:
- Minor cleanup of the debug log messages when doing transfers.  Less repetition; do not print caches that we won't ever use.
- Fix the pre-transfer queries to selected cache hosts; particularly, put caches that respond with an error to the back of the list.
- Send correct header for PROPFIND request.
- Fix a small "airplane mode" issue causing tests to fail when there's no internet connectivity.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -106,6 +106,7 @@ type (
 		BytesPerSecond   int64
 		BytesTotal       int64
 		Duration         time.Duration
+		CacheAge         time.Duration
 	}
 
 	// ConnectionSetupError is an error that is returned when a connection to the remote server fails
@@ -155,6 +156,7 @@ type (
 		TimeToFirstByte   time.Duration // how long it took to download the first byte
 		TransferEndTime   time.Time     // when the transfer ends
 		TransferTime      time.Duration // amount of time we were transferring per attempt (in seconds)
+		CacheAge          time.Duration // age of the data reported by the cache
 		Endpoint          string        // which origin did it use
 		ServerVersion     string        // version of the server
 		Error             error         // what error the attempt returned (if any)
@@ -178,6 +180,12 @@ type (
 
 		// Specifies the pack option in the transfer URL
 		PackOption string
+
+		// Cache age, if known
+		CacheAge time.Duration
+
+		// Whether or not the cache has been queried
+		CacheQuery bool
 	}
 
 	// A structure representing a single file to transfer.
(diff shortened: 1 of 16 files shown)


client/handle_http.go
```diff
@@ -1796,7 +1883,7 @@ func downloadHTTP(ctx context.Context, te *TransferEngine, callback TransferCall
 	req = req.WithContext(ctx)
 
 	// Test the transfer speed every 5 seconds
```
jhiemstrawisc, line 1885: Can you update this comment

client/handle_http.go
```diff
@@ -2144,14 +2246,14 @@ func uploadObject(transfer *transferFile) (transferResult TransferResults, err e
...
 	var lastKnownWritten int64
-	t := time.NewTicker(20 * time.Second)
-	defer t.Stop()
 	uploadStart := time.Now()
 
 	go runPut(request, responseChan, errorChan)
 	var lastError error = nil
 
 	tickerDuration := 100 * time.Millisecond
+	stoppedTransferTimeout := time.Duration(param.Client_StoppedTransferTimeout.GetInt()) * time.Second
```
jhiemstrawisc, line 2255: Can this be simplified to use param's `GetDuration()`?

client/handle_http_test.go
```diff
@@ -851,3 +861,203 @@ func TestSearchJobAd(t *testing.T) {
 		assert.Equal(t, "12345", jobId)
 	})
 }
+
+// Test error messages when a 504 Gateway Timeout occurs
+func TestGatewayTimeout(t *testing.T) {
+	viper.Reset()
+	viper.Set("Logging.Level", "debug")
+	config.InitConfig()
```
jhiemstrawisc, line 869: And here.

jhiemstrawisc (changes requested):

config/config.go
```diff
@@ -1438,10 +1438,6 @@ func InitClient() error {
 
 	upper_prefix := GetPreferredPrefix()
 
-	viper.SetDefault("Client.StoppedTransferTimeout", 100)
-	viper.SetDefault("Client.SlowTransferRampupTime", 100)
-	viper.SetDefault("Client.SlowTransferWindow", 30)
```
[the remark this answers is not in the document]
bbockelm: The code above breaks in a different direction: these parameters need to be set for the local cache (which calls `InitServer`) but only have defaults in `InitClient`.

By moving this to `defaults.yaml`, it's available to both client and server.

The problem Ismael was hitting is he didn't invoke `InitClient` at all -- I believe the best approach there is to have the client API throw an error when it's used without `InitClient()` being called.

jhiemstrawisc (approved):
LGTM

PR #1304, opened 2024-05-20
Reduce number of caches in `Link` header on director cache redirects to 6

### LINKED ISSUE #1247: Only return the first 6 cache endpoints
The list of returned caches or origins is getting quite long -- I suspect we're starting to hit the supported header length limits in various clients.

Given we currently stop the clients at 3 attempts, I think there's no reason to send a list larger than 6.

jhiemstrawisc:
Closes #1247

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -281,7 +281,11 @@ func redirectToCache(ginCtx *gin.Context) {
 
 	linkHeader := ""
 	first := true
-	for idx, ad := range cacheAds {
+	cachesToSend := 6
+	if numCAds := len(cacheAds); numCAds < cachesToSend {
+		cachesToSend = numCAds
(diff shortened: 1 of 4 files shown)


director/director.go
```diff
@@ -281,7 +281,11 @@ func redirectToCache(ginCtx *gin.Context) {
 
 	linkHeader := ""
 	first := true
-	for idx, ad := range cacheAds {
+	cachesToSend := 6
```
haoming29, line 284: Do we want this to be a top-level constant? 
Also, can you add a brief comment explaining this magic number? i.e. what's our intent to limit caches to 6

haoming29 (changes requested):

PR #1305, opened 2024-05-20
Added filtering on osdf topology namespaces for production and its

### LINKED ISSUE #1250: Ability to filter on production topology resources
The topology importer that runs in the director appears to take in *all* caches and origins in topology.

However, topology sorts the resource groups into either "Production: true" (the production infrastructure) or "Production: false" (the testbed infrastructure).

@williamnswanson needs a configuration variable that allows him to filter on this attribute.  Otherwise, we end up importing all the test caches into production -- and that will only result in Bad Things (heads-up to @brianhlin).

Please consider this a priority to backport into the 7.8.x series.

turetske:
closes #1250 

Note that this passes a query to the topology namespace URL. Since we all the topology testing is faked/offline, there isn't an easy way to test it in unit tests as really the only true way to test it is to actually call out to the OSDF topology url and test that the query is filtering properly.

### DIFF
--- a/config/resources/osdf.yaml
+++ b/config/resources/osdf.yaml
@@ -23,6 +23,8 @@ Federation:
   TopologyURL: https://topology.opensciencegrid.org
   TopologyNamespaceURL: https://topology.opensciencegrid.org/osdf/namespaces
   TopologyReloadInterval: 4.5m
+  TopologyITB: false
+  TopologyProduction: true
 Registry:
   RequireCacheApproval: true
   RequireOriginApproval: true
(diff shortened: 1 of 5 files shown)


haoming29 (changes requested):
I thought with the way Matyas implemented the endpoint, aren't we be able to simply telling the ITB folks to manually append that query parameter to the URL set to `TopologyNamespaceUrl`? I think the real modification required is to take care of https://github.com/PelicanPlatform/pelican/blob/main/utils/web_utils.go#L118-L120 if the URL passed in has query parameter already. Since this feature is really meant for ITB only, I'm not sure if we even want to update the parameter docs description to introduce this "hidden" query

PR #1312, opened 2024-05-20
Added a function to deduplicate any duplicated base paths in the gene…

### LINKED ISSUE #1245: De-duplicate `base_path` entries in generated scitokens.cfg
We observed in the generated `scitokens.cfg` that the same directory in `base_path` may be listed multiple times.  Given this is unordered and the repeated ones add no permission, we should just de-duplicate before generating the file.

(Marked this as an origin and cache issue but I've only observed it on the origin -- I suspect it affects both)

turetske:
…rated scitokens files

close #1245


jhiemstrawisc (approved):
LGTM

PR #1313, opened 2024-05-20
Add "Continue" method for config file patching

### LINKED ISSUE #1122: Add multi-file configuration support
Pelican Service:
- [X] Client
- [ ] Plugin
- [X] Registry
- [X] Director
- [X] Origin
- [X] Cache

In Kustomize-based Kubernetes deployments, we make heavy use of environment variables because it's easier to layer configuration between various bases and specific instances (e.g., all OSDF origins should have the same federation discovery URL). However, env var config is harder to read and makes it harder to implement some of the more interesting features like multiple exports.

Ideally this would come in the form of files in `config.d` directories (yes, more than one!) that are parsed in lexicographic order. We make use of multiple `config.d` directories as packagers: we could provide a default federation discovery URL in `/usr/share/pelican/config.d` for the OSDF in the OSDF packaging, leaving `/etc/pelican/config.d` as the location for administrators to add their own config.

jhiemstrawisc:
This adds the ability to define "Continue" directories in your config, where doing so causes Pelican to merge in any additional files in that directory. For example, if my `/etc/pelican/pelican.yaml` has the line `Continue: /foo`, this addition causes Pelican to merge any extra yamls located in `/foo`. If `/foo/config.yaml` further defines `Continue: /bar`, Pelican will then look to `/bar` for extra config.

Config files in a "Continue" directory are parsed according to lexicographical order. This does not currently support adding entries to viper slices, so, for example, if `/etc/pelican/pelican.yaml` defines:
```
Origin:
  Exports:
    - < values for export 1 >
Continue: /foo
```

and `/foo/config.yaml` defines:
```
Origin:
  Exports:
    - < values for export 2 >
```

the exports slice will be overwritten with export 2, instead of appended to.

Closes #1122

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -853,13 +853,13 @@ func setupTranslation() error {
 	})
 }
 
-// If the user defines `Continue: /path/to/dir` in their configuration, this function will
+// If the user defines `ContinuedConfigLocation: /path/to/dir` in their configuration, this function will
 // crawl through that directory and parse all the yaml files in lexicographical order, overriding
 // viper keys that are redefined as it goes. This will NEVER override a key that's hard-coded in source
 func handleContinuedCfg() error {
 	merged := map[string]bool{}
-	for viper.IsSet("Continue") {
-		nextCfgDir := viper.GetString("Continue")
+	for viper.IsSet("ContinuedConfigLocation") {
+		nextCfgDir := viper.GetString("ContinuedConfigLocation")
 		if _, ok := merged[nextCfgDir]; ok {
 			break
 		}
@@ -868,7 +868,7 @@ func handleContinuedCfg() error {
 		// Check that the directory exists
 		if _, err := os.Stat(nextCfgDir); err != nil {
 			if os.IsNotExist(err) {
-				return errors.Errorf("directory %s specified in config 'Continue' does not exist", nextCfgDir)
+				return errors.Errorf("directory %s specified in config 'ContinuedConfigLocation' does not exist", nextCfgDir)
 			} else {
 				return errors.Wrapf(err, "failed to load extra configuration")
 			}
--- a/config/config.go
+++ b/config/config.go
@@ -860,7 +860,7 @@ func handleContinuedCfg() error {
 	if cfgDirs := viper.GetStringSlice("ConfigLocations"); len(cfgDirs) != 0 {
 		for _, cfgDir := range cfgDirs {
 
-		// Check that the directory exists
+			// Check that the directory exists
 			if _, err := os.Stat(cfgDir); err != nil {
 				if os.IsNotExist(err) {
 					return errors.Errorf("directory %s specified by the 'ConfigLocations' key does not exist", cfgDir)
--- a/config/config.go
+++ b/config/config.go
@@ -856,53 +856,57 @@ func setupTranslation() error {
 // If the config file defines a "ConfigLocations" key and a list of corresponding directories, we parse all the yaml
 // files in those directories according to directory-scoped lexicographical order. This allows users/admins to split
 // their configuration across multiple directories/files.
+//
+// Config merging is handled by viper. For more information, see https://pkg.go.dev/github.com/spf13/viper#MergeConfig
 func handleContinuedCfg() error {
-	if cfgDirs := viper.GetStringSlice("ConfigLocations"); len(cfgDirs) != 0 {
-		for _, cfgDir := range cfgDirs {
+	cfgDirs := viper.GetStringSlice("ConfigLocations")
+	if len(cfgDirs) == 0 {
+		return nil
+	}
 
-			// Check that the directory exists
-			if _, err := os.Stat(cfgDir); err != nil {
-				if os.IsNotExist(err) {
-					return errors.Errorf("directory %s specified by the 'ConfigLocations' key does not exist", cfgDir)
-				} else {
-					return errors.Wrapf(err, "failed to load extra configuration from %s", cfgDir)
-				}
-			}
-
-			// Get all files from the directory, sorted in lexicographical order (sorting handled by WalkDir)
-			configFiles := []string{}
-			fileSystem := os.DirFS(cfgDir)
-			err := fs.WalkDir(fileSystem, ".", func(path string, d fs.DirEntry, err error) error {
-				if err != nil {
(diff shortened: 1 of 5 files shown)


bbockelm:
I would like to see some discussion of the config name - I don't find "Continue" to be all that self-descriptive versus, say, "ConfigDir".

What are the semantics around recursion?  Can you have a config directory parameter discovered inside a config directory?

Should this be single-valued or a list?

jhiemstrawisc:
I went with "Continue" because that's the same key used by XRootD, and it seemed like @brianhlin was asking for the same behavior. For people who are already familiar with XRootD, this should feel natural.

For your second question, if a `Continue` directory includes a config file that has another `Continue`, this will follow that chain until we enter a directory with config files that don't define a `Continue`.

For your last question, making this a list makes things a bit more nuanced and less intuitive. In that case, we'd have to decide whether to perform a depth-first or breadth-first crawl -- if I configure:
```
Continue: ["/foo", "/bar"]
```
and both of these directories contain further continues, do I first parse everything in `/foo` and then in all of `/foo`'s continue directories? Or do I do `/foo` and then `/bar` and then move back to `/foo`'s specified continue directories, etc etc. I favor the simplicity of the approach as it stands.

jhiemstrawisc:
If we don't like `Continue`, I'll propose `ContinuedConfigLocation`. Everywhere else in our config, we use "Location" instead of "Dir", and I don't think `ConfigLocation` by itself captures that this is a type of "goto" statement.

brianhlin:
> I went with "Continue" because that's the same key used by XRootD, and it seemed like @brianhlin was asking for the same behavior. For people who are already familiar with XRootD, this should feel natural.

As discussed during standup, let's not base our model on XRootD's config engine. I'd strongly prefer a `ConfigLocations` approach rather than thinking with continues but that does lead to interesting questions like Brian's about recursion and general Pelican config behavior (can we append to lists?).

config/config.go
```diff
@@ -898,6 +966,12 @@ func InitConfig() {
 		}
 	}
 
+	// If the config file defines a "Continue" key, we parse all the yaml files in that directory
```
joereuss12, line 969: Update this comment to say "ContinuedConfigLocation" key or whatever the term is decided to be.

config/config_test.go
```diff
@@ -177,6 +182,128 @@ func TestInitConfig(t *testing.T) {
...
+	viper.AddConfigPath(rootCfgDir)
+	viper.SetConfigType("yaml")
+	viper.SetConfigName("pelican")
+
+	err := os.WriteFile(filepath.Join(rootCfgDir, "pelican.yaml"), []byte(fmt.Sprintf("ContinuedConfigLocation: %s\nOtherVal: bar", continueDir)), 0644)
+	require.NoError(t, err)
+	err = viper.MergeInConfig()
+	require.NoError(t, err)
+}
+
```
joereuss12, line 198: Can you add a comment header here describing the test? Just want to stay consistent with everything

config/config_test.go
```diff
@@ -177,6 +182,128 @@ func TestInitConfig(t *testing.T) {
...
+		setupContConfig(t, continueDir+"-dne")
+
+		continueFile := filepath.Join(continueDir, "continue.yaml")
+		err := os.WriteFile(continueFile, []byte("TestVal: foo"), 0644)
+		require.NoError(t, err)
+
+		err = handleContinuedCfg()
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "does not exist")
+	})
```
joereuss12, line 304: What if a user tries to continue a config dir that is empty or contains "config" files that are not `yaml`'s? I think we might want test cases for that.

joereuss12 (changes requested):
Just have some small items but otherwise it LGTM

jhiemstrawisc:
After feedback, the current mechanism works with the `ConfigLocations` config key and sets a list of extra dirs. For example:
```
ConfigLocations: ["/dir1", "/dir2"]
```
First we load all the configs according to lexicographical order in `/dir1` and then `/dir2`. There is no recursion into subdirectories. Config keys that are redefined are patched over, with the last config loaded winning.

docs/parameters.yaml
```diff
@@ -30,6 +30,23 @@ default: "~/.config/pelican"
 components: ["*"]
 type: filename
 ---
+name: ConfigLocations
+description: |+
+  `ConfigLocations` provides administrators a way to define a list of directories containing Pelican configuration files. Within a given directory,
+  files are read in lexicographical order, and any keys that are defined in multiple files will take the value from the last file read. Directories are
+  read in the order provided by the list. For example, specifying:
+
+  `ConfigLocations: ["/configs1", "/configs2"]`
```
haoming29, line 39: Are we still handling recursion here? Would it worth mentioned the behavior of nested `ConfigLocations`?

config/config.go
```diff
@@ -847,6 +848,60 @@ func setupTranslation() error {
 	})
 }
 
+// If the config file defines a "ConfigLocations" key and a list of corresponding directories, we parse all the yaml
+// files in those directories according to directory-scoped lexicographical order. This allows users/admins to split
+// their configuration across multiple directories/files.
+func handleContinuedCfg() error {
+	if cfgDirs := viper.GetStringSlice("ConfigLocations"); len(cfgDirs) != 0 {
```
haoming29, line 855: For better readability, can we pull the body of this if statement out by simply returning the function if the length of the slice is 0?

config/config.go
```diff
@@ -847,6 +848,60 @@ func setupTranslation() error {
...
+
+			for _, file := range configFiles {
+				fHandle, err := os.Open(filepath.Join(cfgDir, file))
+				if err != nil {
+					return errors.Wrapf(err, "failed to open extra configuration file %s", filepath.Join(cfgDir, file))
+				}
+				defer fHandle.Close()
+
+				reader := io.Reader(fHandle)
+				err = viper.MergeConfig(reader)
```
haoming29, line 891: Since we use `viper.MergeConfig` for handling the merging, let's document this at least in the function header (Or docs) and pointing to viper docs for the merging mechanism under the hood.

haoming29 (approved):
LGTM overall, conditionally approved if the comments are addressed.

PR #1318, opened 2024-05-21
Origin Registration

### LINKED ISSUE #1107: Origin registration

haoming29:
Closes #1107 

**Note** This PR changes the frontend routing for registrations. The original "namespace" registration path is changed from `/registry/origin/*` to `/registry/namespace/*`, and the `origin` path now becomes "origin" registration.

It is recommended to upgrade the registry service first with this version, then the origin/cache servers. For the origin server running `7.8.x`, there will be a **backward compatibility** issue, where the "complete namespace registration" feature in the origin web UI still directs the user to `/registry/origin` endpoint which now becomes the "origin registration" instead of "namespace registration". The user should still be able to do the update, but the texts on the UI would be wrong.

The workflow for completing registration is as follows:
* At server start, the server auto-registers Origin, but it's incomplete, missing Institutions for example.
* The top-right "Data Exports" panel shows that the origin registration is incomplete
* The admin clicks "Complete Registration" button and is redirected to the registry page to finish the registration
* The registry redirect the admin user back to the origin website
* The status becomes "Registration Compeleted"

For any reason the admin wants to know the public key of the origin, there's now a new button at the corner of the "Data Exports" table to copy the public key to the clipboard

Screenshot 2024-05-23 at 10 50 38 AM
Incomplete Origin Registration

Screenshot 2024-05-23 at 10 50 11 AM
Completed Origin Registration

Screenshot 2024-05-23 at 10 50 44 AM
Copy public key to clipboard


server_structs/xrootd_server.go
```diff
@@ -41,8 +44,49 @@ type (
 	NamespaceHolder struct {
 		namespaceAds []NamespaceAdV2
 	}
+
+	ServerPrefix string // The base namespace prefix for origin/cache server
+)
+
+const (
+	CachePrefix  ServerPrefix = "/caches"
```
jhiemstrawisc, line 52: These should probably contain a trailing `/` to be safe. Otherwise something like
```
prefix := "/cachesfoo"
isCache := strings.HasPrefix(prefix, server_structs.CachePrefix.String())
```
might tell us the prefix indicates a cache. You have functions that do a better job of checking this already, but it's very possible a future developer won't realize they exist and end up checking things like this.

server_utils/origin.go
```diff
@@ -38,18 +38,18 @@ var originExports []OriginExport
 
 type (
 	OriginExport struct {
-		StoragePrefix    string `json:"storage_prefix"`
-		FederationPrefix string `json:"federation_prefix"`
+		StoragePrefix    string `json:"storagePrefix"`
```
jhiemstrawisc, line 41: I'm guessing it's the registry that normally ingests these JSON values? Does changing the cases break any backwards compatibility with an old registry?

web_ui/frontend/app/registry/namespace/edit/page.tsx
```diff
@@ -0,0 +1,45 @@
+/***************************************************************
+ *
+ * Copyright (C) 2023, Pelican Project, Morgridge Institute for Research
```
jhiemstrawisc, line 3: Can you update the copyright year to 2024?

web_ui/frontend/app/registry/namespace/register/page.tsx
```diff
@@ -0,0 +1,45 @@
+/***************************************************************
+ *
+ * Copyright (C) 2023, Pelican Project, Morgridge Institute for Research
```
jhiemstrawisc, line 3: Same for this one.

jhiemstrawisc (changes requested):
One other question I had that didn't fit directly into inline comments. Do we have a way to prevent users from trying to register cache/origin-prefixed namespaces as a proper namespace? For example, do we prevent the user from setting something like `Origin.FederationPrefix = /caches` or `Origin.FederationPrefix = /origins/foo`? I can't think of _why_ a user would want to do that, but it seems like the kind of thing that may have security implications.

director/director.go
```diff
@@ -672,6 +672,49 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
...
+			registryPrefix = server_structs.GetCacheNS(adV2.Name)
+		}
+	}
+
+	if verifyServer {
+		ok, err := verifyAdvertiseToken(engineCtx, token, registryPrefix)
+		if err != nil {
+			if err == adminApprovalErr {
+				log.Warningf("Failed to verify token. %s %q was not approved", string(sType), adV2.Name)
+				ctx.JSON(http.StatusForbidden, gin.H{"approval_error": true, "error": fmt.Sprintf("%s %q was not approved by an administrator", string(sType), ad.Name)})
```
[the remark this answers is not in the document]
haoming29: Yes! Thanks for pointing out. This would be very useful for the users

launcher_utils/advertise.go
```diff
@@ -136,27 +135,19 @@ func advertiseInternal(ctx context.Context, server server_structs.XRootDServer)
 	name := ""
 	var err error
 	// Fetch site name from the registry, if not, fall back to Xrootd.Sitename.
-	// However, currently the registry only supports registering namespaces, not origins.
-	// If multiple namespaces fall under the same origin, we will use the first namespace to fetch the site name
 	if server.GetServerType().IsEnabled(config.OriginType) {
-		originExports, err := server_utils.GetOriginExports()
+		// Note we use Server_ExternalWebUrl as the origin prefix
+		// But caches still use Xrootd_Sitename, which will be changed to Server_ExternalWebUrl in the future
```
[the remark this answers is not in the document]
haoming29: Here's the issue: https://github.com/PelicanPlatform/pelican/issues/1351 I'll update the comment as well

server_utils/origin.go
haoming29: No, the JSON of this struct is only used for UI. We have different structs for registry/director to ingest

haoming29:
> One other question I had that didn't fit directly into inline comments. Do we have a way to prevent users from trying to register cache/origin-prefixed namespaces as a proper namespace? For example, do we prevent the user from setting something like `Origin.FederationPrefix = /caches` or `Origin.FederationPrefix = /origins/foo`? I can't think of _why_ a user would want to do that, but it seems like the kind of thing that may have security implications.

I added checks in `validateFederationPrefix` when doing export checks

jhiemstrawisc (approved):
LGTM

PR #1321, opened 2024-05-22
Implementation of pelican object stat

joereuss12:
The initial implentation of object stat command. This allows users to run the command `pelican object stat` on an object in a namespace and get the following information: Name, Size, ModTime, and IsDir. The only current issue is that if we stat a directory we are currently unable to give a size as the webdav client does not support it.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -2490,10 +2490,6 @@ func (te *TransferEngine) walkDirUpload(job *clientTransferJob, transfers []tran
 // is made without.
 func statHttp(ctx context.Context, dest *url.URL, namespace namespaces.Namespace, directorUrl, token string, jsn bool) (info fileInfo, err error) {
 	statHosts := make([]url.URL, 0, 3)
-	type statResults struct {
-		info fileInfo
-		err  error
-	}
 	var dirListNotSupported *dirListingNotSupportedError
 	collectionsUrl, err := getCollectionsUrl(ctx, dest, namespace, directorUrl)
 	// If we have a dirListingNotSupported error, we can attempt to stat the caches instead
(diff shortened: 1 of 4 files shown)


joereuss12:
Making a draft since it still needs unit testing. Just want to start documenting progress.

client/handle_http.go
```diff
@@ -2256,6 +2262,73 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
 
 }
 
+// This function queries the director with a PROPFIND to attempt to get the 'collections url'. If a propfind is not allowed on the director
+// We fall back to the deprecated dirlisthost in the namespace
+func getCollectionsUrl(ctx context.Context, remoteObjectUrl *url.URL, namespace namespaces.Namespace, directorUrl string) (collectionsUrl *url.URL, err error) {
```
joereuss12, line 2267: Note: this function is taken from this PR: https://github.com/PelicanPlatform/pelican/pull/1257

So that PR should most likely get approved and merged first but I don't think it should matter.

cmd/object_stat.go
```diff
@@ -0,0 +1,92 @@
...
+	flagSet.StringP("token", "t", "", "Token file to use for transfer")
+	flagSet.BoolP("json", "j", false, "Print results in JSON format")
+	objectCmd.AddCommand(statCmd)
+}
+
+func statMain(cmd *cobra.Command, args []string) {
+	ctx := cmd.Context()
+	err := config.InitClient()
+	if err != nil {
+		log.Errorln(err)
```
haoming29, line 50: Probably merge this error logging with the one on line 53

haoming29 (changes requested):
Small changes requested

PR #1322, opened 2024-05-22
Fix bug w/ client not working for WithToken option

### LINKED ISSUE #1320: Fix client upload code not working for `WithToken` option
Pelican Version:
found in v7.8.3

Pelican Service:
- [X] Client

Ismael discovered a bug in the client code that shows it does not work with the `WithToken` client transfer option. Here is the snippet from slack describing the bug and its fix:
I think I found a bug causing your issue: https://github.com/PelicanPlatform/pelican/blob/3edb26b657f0e3dff91ee437f60af105ddcc7bc7/client/handle_http.go#L1091-L1096
We're missing some parentheses in that if statement. It seems like it should be: if (upload || ns.UseTokenOnRead) && tj.token == "" . Basically, what is happening is tj.token should be set to whatever the client (tc)'s token is, but is overwritten in this block here meaning that the token you are sending is getting erased in our code. I'll work in a patch for you so this can be fixed.

joereuss12:
Fixes #1320

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -1088,7 +1088,7 @@ func (tc *TransferClient) NewTransferJob(ctx context.Context, remoteUrl *url.URL
 	}
 	tj.namespace = ns
 
-	if upload || ns.UseTokenOnRead && tj.token == "" {
+	if (upload || ns.UseTokenOnRead) && tj.token == "" {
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
👍

PR #1325, opened 2024-05-22
Warn users about unrecognized configuration

### LINKED ISSUE #998: Pelican should warn users/admin about unrecognized inputs
Pelican Service:
All

As a developer, I have lots of handy recipes baked into a "catchall" pelican.yaml. To toggle various features, I often comment or uncomment various blocks in my configuration, but sometimes I do things incorrectly or I forget to update something in my pelican.yaml with new config names that have changed since I last tested them. Pelican is totally silent in most cases when I pass config options that don't exist (eg I made a typo) or are nested under the wrong header (eg adding MaxMindKeyFile under Origin instead of Director).

At a minimum, Pelican should throw a warning when it doesn't recognize a config option it's passed via config file.

jhiemstrawisc:
It turns out Pelicans are picky birds, and they only like eating config options they recognize. Instead of feeding our beloved bird things it doesn't like and can't eat, this PR warns users when config options that are passed via config files or env vars that are unrecognized. Ewww!

This is accomplished by adding a mapstructure value to entries in our generated param.Config struct consisting of each field's lowercased name, and then using that to match against viper keys read in from config. When a viper key exists and can't be mapped to a field name in param.Config, we add the key to a list that is passed to the user as a warning.

To test, try passing some garbage values in your config file, or passing something like `PELICAN_BAD_KEY` through the environment.

Closes #998

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,8 +41,6 @@ jobs:
       if: runner.os == 'macOS'
       run: |
         make web-build
-        # Explicitly set XRootD to IPv4-only on Mac OS X to avoid a subtle config bug.
-        export PELICAN_XROOTD_IPV4ONLY=true
         go test -v -coverpkg=./... -coverprofile=coverage.out -covermode=count ./...
     - name: Test
       if: runner.os != 'macOS'
--- a/config/config.go
+++ b/config/config.go
@@ -31,7 +31,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"reflect"
 	"slices"
 	"sort"
 	"strconv"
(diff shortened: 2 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -41,8 +41,6 @@ jobs:
       if: runner.os == 'macOS'
       run: |
         make web-build
-        # Explicitly set XRootD to IPv4-only on Mac OS X to avoid a subtle config bug.
-        export PELICAN_XROOTD_IPV4ONLY=true
```
haoming29, line 45: I'd assume we remove this because there's no corresponding Pelican config for `XRootD.IPV4Only`, but let's confirm with whoever put this here to make sure it won't break anything. Looking at git blame, @bbockelm seems to add this line.

config/config.go
```diff
@@ -849,6 +850,79 @@ func setupTranslation() error {
 	})
 }
 
+// findFieldByTag searches for a field in a struct by the value of a tag. This is used to
```
haoming29, line 853: Can you pull the two functions out from `config.go`? I'm worried that we put too much things in this single file. Since this is about config validation, probably put them in `validation.go`

config/config_test.go
```diff
@@ -228,6 +228,59 @@ func TestDeprecateLogMessage(t *testing.T) {
 	})
 }
 
+func TestBadConfigKeys(t *testing.T) {
```
haoming29, line 231: If you decide to pull the validation logic out from `config.go`, the testing can also go to their dedicated file too

docs/parameters.yaml
haoming29, line 1: Thanks for filling out these missing definitions!

haoming29 (changes requested):
LGTM overall with small changes to organization

.github/workflows/test.yml
jhiemstrawisc: Yep, I confirmed with Brian that this is old and hasn't been doing anything for awhile. Should be safe to remove.

haoming29 (approved):
LGTM!

PR #1328, opened 2024-05-22
Increase `WaitUntilWorking` transfer timeout and adjust error message

### LINKED ISSUE #1324: `WaitUntilWorking` only useful for local services
The `WaitUntilWorking` function was originally designed as a quick "ping" for local services to enforce certain startup ordering -- e.g., make sure XRootD is fully functioning before proceeding to advertise it to the director.

However, this has been reused as a generic health check for remote services, such as in this code:
```go
		// Give five seconds for the origin to finish advertising to the director
		desiredURL := fedInfo.DirectorEndpoint + "/.well-known/openid-configuration"
		if err = server_utils.WaitUntilWorking(ctx, "GET", desiredURL, "director", 200, false); err != nil {
			log.Errorln("Director does not seem to be working:", err)
			return
		}
```

*However*, the original use case assumed that one could quickly connect and put in a 50ms timeout to quickly retry.  Outside of unit tests, it's not safe to assume that the director can service a request (3 round trips!) within 50ms.  So, in this new context, the check is always failing for physically-distant services.

jhiemstrawisc:
When `WaitUtilWorking` was pointed at the OSDF Director (located in Madison) from a distant server, it couldn't get a response in the previously-configured 50ms. After peeking at this function, it doesn't seem like there's a good reason not to have a more permissive timeout.

I also adjusted the error message that was output from the previous circumstance. Setting up a distant cache resulted in saying "Server director <url> did not startup after 10s of waiting," which makes it seem like the cache is trying to start up a director.

Closes #1324

### DIFF
--- a/server_utils/server_utils.go
+++ b/server_utils/server_utils.go
@@ -39,7 +39,7 @@ import (
 	"github.com/pelicanplatform/pelican/config"
 )
 
-// Wait until given `reqUrl` returns a HTTP 200.
+// Wait until given `reqUrl` returns the expected status.
 // Logging messages emitted will refer to `server` (e.g., origin, cache, director)
 // Pass true to statusMismatch to allow a mismatch of expected status code and what's returned not fail immediately
 func WaitUntilWorking(ctx context.Context, method, reqUrl, server string, expectedStatus int, statusMismatch bool) error {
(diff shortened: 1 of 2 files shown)


turetske (changes requested):
I think the ticker on line 49 if server_utils/server_utils.go should be changed to match the 1 second recurrence.

bbockelm:
Oh!  And the ticker is missing a `defer ticker.Stop()` -- the resource leaks, I think.

If I'm reading the documentation right, however:
> The ticker will adjust the time interval or drop ticks to make up for slow receivers
That means the for loop will sleep for up to 50ms before trying the download again (i.e., we won't busy-loop if failures are instantaneous).

So, I'm actually OK with keeping the short timeout.  If we increase the ticker duration to 1s, then we will try once every second on immediate failure -- a bit too long a time for my tastes.

turetske (approved):

PR #1331, opened 2024-05-24
Fix plugin bug for incorrect start time

joereuss12:
The plugin was reporting the start time of the transfer in the wrong place. Originally, it was recorded right when we get the results for the transfer (which is wrong). Now, we record the start time in the TransferResults object so each transfer will have its own start time and it will be properly recording in the plugin output ad.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -141,12 +141,13 @@ type (
 	// Represents the results of a single object transfer,
 	// potentially across multiple attempts / retries.
 	TransferResults struct {
-		jobId            uuid.UUID // The job ID this result corresponds to
-		job              *TransferJob
-		Error            error
-		TransferredBytes int64
-		Scheme           string
-		Attempts         []TransferResult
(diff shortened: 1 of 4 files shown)


client/handle_http.go
```diff
@@ -1698,6 +1700,7 @@ func downloadObject(transfer *transferFile) (transferResults TransferResults, er
 			break
 		}
 	}
+	transferResults.TransferStartTime = transferStartTime
```
haoming29, line 1703: It looks like the `TransferStartTime` is the start time of the last transfer attempt (successful or not). Could you leave a comment here or around line 1645 where you defined this variable? I'm was a bit confused when I first saw its definition and where it gets used.

haoming29: Also, I was wondering if this is actually the value we want to record? Why not recording the time at the start of the first attempt?

haoming29 (commented):
Some questions about definition of transfer start time

PR #1337, opened 2024-05-27
Fixed `object copy` directory checking for multiple sources

turetske:


joereuss12 (approved):
LGTM

joereuss12:
Actually after further investigation, it looks like 'object get' has the same issue. I can make a PR for that quick though

PR #1338, opened 2024-05-28
Fixed `object get` directory checking for multiple sources

joereuss12:
Object get had incomplete error checking for whether the destination was a directory when copying multiple sources. Found from #1337


turetske (approved):
LGTM!

PR #1343, opened 2024-05-28
Getting started documentation

turetske:
Beginning getting started documentation. Cleaned up some language and culled a lot of extraneous information. Those should be moved into more detailed Registry/Director/Origin pages. The culled sections have been put into their own tickets to rework, so the information is available outside of the GitHub history.


turetske:
Very stale

PR #1356, opened 2024-05-30
First draft of Pelican architectural description and overview updates

jhiemstrawisc:
I still need to tidy up presentation by making the architectural description section either a subpage or a separate page/tab under the "About Pelican" section.

This also reflects the first major change of "Cache" to "Staging Server" in our docs, which is something Miron would like to see. If the rest of the group isn't onboard with that, I can revert those references until we've discussed this more.


docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+
+This process serves two purposes -- first, whenever the Origin re-advertises with the federation's Director, the Director can verify the authenticity of those advertisements through public/private key asymmetric cryptography by looking at the Registry's stored public key for that Origin and namespace. Second, the Registry's persistent database prevents other Origins from registering namespaces under an already-registered namespace without first proving they're allowed to do so by the namespace owner (i.e. the entity that possesses the appropriate private key).
+
+## Making Bytes Accessible and Moving Them -- A First Look Under The Hood
```
haoming29, line 89: I forget the conclusion we draw during the RCF meeting. Do we want to split the glossary from the arch sections or just keep them on the same spot? Looking at the meeting notes, it seems that we want to have overview and core concepts as two separate pages:
https://docs.google.com/document/d/11knkvPU78eBvRZjcoYGzKV2F5y_Sp76qIeginR5NkIk/edit
```
About Pelican
– Core Concepts

```

docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+
+Pelican serves two sides of the same coin -- Data owners who want to federate their data from wherever it lives natively, and data consumers who want to access and compute on data wherever they need it.
+
+
+<ImageRow alt={"Pelican and OSDF"} src={"/pelican/arch-repo-and-consumer.png"}>
+	The federation's core goal is connecting data owners and data consumers.
+</ImageRow>
+
+
+As such, the primary prerequisite for data to be moved via a Pelican federation is for a data owner to make their data accessible to the federation. This happens when an Origin is placed in front of the repository and registered with the federation. While federations like the OSDF *may* wish to control or filter any Origin registrations to vet the data they make available, this example assumes the Origin's registration is automatically approved. The red arrow in the following graphic represents the vetting/approval step, should the federation require it.
```
haoming29, line 101: I didn't find any red arrows in the graphics in this PR. Should this be updated?

Also, I'm a bit hesitated to give out the details for approval in the overview doc, but it's a good topic for the federation overview page for sure.

docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+
+<ImageRow alt={"Pelican and OSDF"} src={"/pelican/arch-origin-registration.png"}>
+	When the Origin starts, it's told which federation it belongs to. It discovers the
```
haoming29, line 104:
```suggestion
	When the Origin starts, it's told which federation it belongs to. It discovers the
```
I'd say the origin owner pass the federation root to the origin config before it starts

docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+	hostnames for its Registry and Directory by using the federation root to construct
+	the URL "https://osg-htc.org/.well-known/pelican-configuration", which contains a JSON
```
haoming29, line 106: Do we want to define this URL so that it's easier to refer back to it in the following docs? Like Federation discovery endpoint?

docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+	detailing the federation's central services.
+
+	Next, the Origin registers its namespace and public key with the Registry, proving that it
+	owns the corresponding private key. Finally, the Origin begins advertising its namespace
+	information and hostname to the Director.
+</ImageRow>
+
+While somewhat simplified, this example illustrates the process origins must take to make themselves known within the federation. After completing these steps, the objects from the Data Repository are available via Pelican.
+
+The next step is for the data consumer to actually *move* the data. Pelican assumes the data consumer already knows the federation that provides the data they want, along with the name of the object within the federation. These two pieces of information are combined and provided to the Client as a `pelican://`-schemed URL
```
haoming29, line 116: With our ongoing work for client list functionality, would we expect this to be changed soon since we do (and want to?) provide a way for user to discover data?

docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+
+<ImageRow alt={"Pelican and OSDF"} src={"/pelican/arch-orig-discovery.png"}>
+	The data consumer provides their Pelican client of choice the pelican:// URL that defines
+	the object they want to download, where osg-htc.org is the federation and
```
haoming29, line 120:
```suggestion
	the object they want to download, where `osg-htc.org` is the federation and
```

docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+	(/weather/cloud.jpg) is the object. Just as the origin discovered the Director's hostname
```
haoming29, line 121:
```suggestion
	(`/weather/cloud.jpg`) is the object. Just as the origin discovered the Director's hostname
```

docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+	by visiting "https://osg-htc.org/.well-known/pelican-configuration", so too does the client.
+
+	After the client has performed federation metadata discovery, it issues an HTTP GET
+	request to the Director, using the object name as a URL path. The Director responds
+	with an HTTP 307 Redirect, forwarding the client on to the a server that can provide
+	the object, in this example an Origin.
+
+	Finally, the Client follows the redirect and downloads the object by issuing an HTTP
```
haoming29, line 129: Is there a specific reason that we skip the staging components? I'm asking because I don't think by default origin (or director) allows direct access to the origin

docs/pages/index.mdx
```diff
@@ -2,39 +2,134 @@ import ImageRow from "@/components/ImageRow";
...
+	GET request to "https://my-origin.com/weather/cloud.jpg"
+
+	Notice that the Origin continues advertising with the Director throughout.
+</ImageRow>
 
-Central to Pelican is the concept of the origin service.  The origin is the intermediary between
-the existing storage and the federation.  The origin is responsible for serving data as well
-as issuing tokens (credentials) authorizing access to datasets based on the local policy.
+Once again, this example is simplified, mainly because the Director typically sends the client to a Staging device capable of fetching the object, not directly to the Origin. In any case, the object is delivered to the Client without passing through the federation's Central Services. When the object is fetched through a Staging device, the Staging device performs the same discovery step as the Client by asking the Director for an Origin that exports the object.
```
haoming29, line 135: I see...but I'm not sure if we want to skip caches completely, which really depends on who our target users are for this overview..

haoming29 (changes requested):

docs/pages/index.mdx
jhiemstrawisc: Thanks for catching the image, I'll update it.

As for the approval step, I think it's worth mentioning because the majority of people reading this for the foreseeable future will be registering with the OSDF.

docs/pages/index.mdx
jhiemstrawisc: My thinking is that the process the cache goes through to find an origin is the same as the client goes through to find an origin (bar a few gritty details that live in xrdcl-pelican). From the perspective of "how does a client discover a service that can provide data", there's essentially no difference between the cache and the origin, and throwing in the cache right away adds an extra layer of complexity. I'm personally in favor of leaving it out in this example, but if enough other people disagree, I'm happy to add it. @aowen-uwmad do you think adding the cache layer to the middle of this and getting into those details would help users or confuse them more?

docs/pages/index.mdx
jhiemstrawisc: I'd argue yes and know -- providing a way to `ls` for data is still pretty far from providing a way to discover data, because all `ls` tells you is the name of some objects, and nothing about what information they contain, how the user came to be looking for data at a given path prefix, etc. In order to say "I want to compute on weather data from the `/ncar`  namespace," the user still has to be told that they should look there for weather data, even if they can now discover the names of the objects it hosts.

docs/pages/index.mdx
jhiemstrawisc: Technically neither of those has an enforcement mechanism -- If I ask the director for an origin redirect, it gives it to me, and if some client shows up to the origin with a valid token, the origin doesn't ask whether the client is a cache or something else.

haoming29 (approved):
The revised doc LGTM! @aowen-uwmad I'm not sure if you want to take another look at this

docs/pages/index.mdx
aowen-uwmad: @jhiemstrawisc , I think the proposed change is fine, I think the reference to the "Staging device" (which was also defined earlier in the terminology section) is enough for this level of detail. If people want more info on how the Cache comes into play, they can read up the documentation about the Cache.

aowen-uwmad (approved):

PR #1359, opened 2024-05-31
Build deb pelican-server package and make sure it pulls in its required dependencies

### LINKED ISSUE #1272: Include xrootd as a dependency in other `pelican-server` packaging
We included xrootd as a dependency in our `rpm` packaging, but not in others. XRootD seems to support at least Debian: https://github.com/xrootd/xrootd?tab=readme-ov-file#supported-operating-systems, and we want to be consistent in server packaging, so let's include xrootd and related dependencies for Debian as well.

matyasselmeci:
Note that it is not installable on debian:latest or ubuntu:22.04 because they do not have the required version of xrootd.  It does install on ubuntu:24.04 (ubuntu:latest).

Closes #1272

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -221,19 +221,12 @@ nfpms:
     license: ASL 2.0
     meta: true
     formats:
-      # XXX Deb has some different conventions; planned for a later release
-      # - deb
+      - deb
       - rpm
     release: 1
     section: default
     priority: extra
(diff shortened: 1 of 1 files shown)


haoming29 (approved):
@matyasselmeci Can you create an issue for updating Debian/Ubuntu system requirement docs for installing `pelican-server` package? You can assign me for that ticket

PR #1366, opened 2024-06-03
Start phasing out NamespaceAdV2.PublicRead in favor of NamespaceAdV2.Caps.PublicReads

jhiemstrawisc:
NamespaceAdV2 has a weird duplication between `ad.PublicRead` and `ad.Caps.PublicReads`. Since we have the `Capabilities` struct, we should switch to populating that and to checking that. This will have to be done over several releases because this will break backwards compatibility if an old director (prior to whichever release this PR ends up in) tries to receive an ad from a new Origin.

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -218,7 +218,7 @@ func AdvertiseOSDF(ctx context.Context) error {
 		}
 		nsAd := server_structs.NamespaceAdV2{
 			Path:         ns.Path,
-			PublicRead:   !ns.UseTokenOnRead,
+			PublicRead:   caps.PublicReads,
 			Caps:         caps,
 			Generation:   []server_structs.TokenGen{tGen},
 			Issuer:       tokenIssuers,
--- a/director/cache_ads_test.go
+++ b/director/cache_ads_test.go
@@ -52,8 +52,8 @@ func TestGetAdsForPath(t *testing.T) {
 	*/
 	nsAd1 := server_structs.NamespaceAdV2{
 		// PublicRead: false,
-		Caps:       server_structs.Capabilities{PublicReads: false},
-		Path:       "/chtc",
+		Caps: server_structs.Capabilities{PublicReads: false},
+		Path: "/chtc",
 		Issuer: []server_structs.TokenIssuer{{
 			IssuerUrl: url.URL{
 				Scheme: "https",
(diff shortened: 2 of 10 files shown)


director/advertise.go
```diff
@@ -184,7 +184,7 @@ func AdvertiseOSDF() error {
 		}
 		nsAd := server_structs.NamespaceAdV2{
 			Path:         ns.Path,
-			PublicRead:   !ns.UseTokenOnRead,
```
haoming29, line 187: I think at the receiver side (director in this case), it's safe to leave `PublicRead` unset in this case, or copy the value of `PublicRead` to `Caps.PublicReads` and use/read `Caps.PublicReads` internally instead, so that internally we start using the new struct

director/cache_ads_test.go
```diff
@@ -51,9 +51,9 @@ func TestGetAdsForPath(t *testing.T) {
 			- Query for a few paths and make sure the correct ads are returned
 	*/
 	nsAd1 := server_structs.NamespaceAdV2{
-		PublicRead: false,
-		Caps:       server_structs.Capabilities{PublicReads: false},
-		Path:       "/chtc",
+		// PublicRead: false,
```
haoming29, line 54: Can you remove the commented fields or this is intentional?

director/director.go
```diff
@@ -334,14 +334,14 @@ func redirectToCache(ginCtx *gin.Context) {
 	// This is because the configuration of the origin/namespace should override the inclusion of "dirlisthost" for that origin.
 	// Listings is true by default so if it is ever set to false we should accept that config over the dirlisthost.
 	if namespaceAd.Caps.Listings && originAds[0].Listings {
-		if !namespaceAd.PublicRead && originAds[0].AuthURL != (url.URL{}) {
+		if !namespaceAd.Caps.PublicReads && originAds[0].AuthURL != (url.URL{}) {
```
haoming29, line 337: Looks like we already reading `Caps.PublicReads` here! So we should be abele to unset `PublicReads` for reading OSDF servers

haoming29 (changes requested):
Small cleanups requested

director/advertise.go
jhiemstrawisc: I changed the director to look at the new struct in this PR, but if we send this advertisement to an old director that checks `nsAd.PublicRead`, we'll have a backwards compatibility issue, won't we?

director/director.go
jhiemstrawisc: Same as for the previous comment, these changes happened in 7.9, so I think this would break a new advertisement (>v7.9.0) being used by an old director (<=v7.8.x).

director/advertise.go
haoming29: Yes we will, but there we are at the director side receiving the data, so leaving `PublicRead` unset and don't read this value in the director should be fine. This is not a big deal and we should be able to clean this up whenever we deprecate and remove the old capabilities. So no changes needed as this point!

haoming29 (approved):
Revised code LGTM!

PR #1371, opened 2024-06-03
Cleanup a few deprecation warnings from the old Federation.NamespaceUrl config key

jhiemstrawisc:
@haoming29, I came across these old deprecation warnings, and thought they were old enough that we could safely clean them up. If you disagree, you can close the PR as `wontfix` and we can open a separate issue to remove them in a later release.


haoming29 (approved):
LGTM!

PR #1376, opened 2024-06-05
Disable ZTN authentication on Pelican origins

### LINKED ISSUE #1375: Non-Pelican caches can't talk to Pelican origin over xrootd protocol
Non-Pelican caches will try to talk to the Pelican origin over the xrootd protocol.

To use the xrootd protocol, a session needs to be established.  There are two "authenticated" session types, "ZTN" (token-based) and GSI (X.509-based).  The latter is broken (see https://github.com/xrootd/xrootd/issues/2280) on Alma9 it appears.  The former doesn't work with caches (just was never implemented) and is also broken.

It turns out we don't need an authenticated session, really: we always authenticate individual requests with tokens to a Pelican-based origin (at least, the Pelican-based origins currently in use: one could at least theoretically setup X.509).

If we turn off all possible authentication methods, then no authentication will be required for the xrootd protocol.  This should be safe as long as TLS is required for all connections (so the token can safely be passed).

I believe this can be done by removing the following line from the origin configuration:

```
sec.protocol ztn
```

[Note this doesn't affect pure-Pelican setups as Pelican communicates over the HTTP, not xrootd, protocol.]

bbockelm:
Caches cannot talk ZTN so this never worked anyhow.  Further, if a non-Pelican cache tries to talk to a Pelican origin over the xrootd protocol, by having ZTN enabled anonymous sessions are forbidden.

This change will allow anonymous sessions for the xrootd origin, meaning that requests must be authorized by tokens.

Fixes #1375

### DIFF
--- a/xrootd/resources/xrootd-origin.cfg
+++ b/xrootd/resources/xrootd-origin.cfg
@@ -89,7 +89,6 @@ pss.origin {{.Origin.XRootServiceUrl}}
 ofs.osslib libXrdPss.so
 {{end}}
 xrootd.seclib libXrdSec.so
-sec.protocol ztn
 ofs.authorize 1
 acc.audit deny grant
 acc.authdb {{.Origin.RunLocation}}/authfile-origin-generated


bbockelm:
Note: this got tested by hand on ap40.  Verified both the positive case (non-Pelican caches began working with the origin) and the negative case (could not access the data without a token).

Once @jhiemstrawisc is happy, I think this can go in and be backported to the 7.8.x series.

jhiemstrawisc (approved):

PR #1377, opened 2024-06-05
Add mechanism to change authfile reload intervals

### LINKED ISSUE #1335: Refresh of authfile too infrequent
The defaults for the Authfile reloads are surprisingly long.  From the XRootD docs, it is:

```
acc.authrefresh 43200
```

Accordingly, it takes 12 hours for a namespace change to propagate out to the system.

I believe a 5 minute refresh interval is reasonable for both origin and cache.

jhiemstrawisc:
Caches take too long to pull permissions changes from upstream origins (12hours, from Xrootd's default `acc.authrefresh 43200`). This sets the default refresh time to 5m and adds a mechanism for admin to adjust as needed.

Closes #1335

### DIFF
--- a/xrootd/xrootd_config.go
+++ b/xrootd/xrootd_config.go
@@ -647,8 +647,8 @@ func authRefreshStrToSecondsHookFunc() mapstructure.DecodeHookFuncType {
 			return nil, err
 		}
 
-		if duration < time.Second {
-			log.Warningf("'Xrootd.AuthRefreshInterval' appears as less than a second. Using fallback of 5m")
+		if duration < 60*time.Second {
+			log.Warningf("'Xrootd.AuthRefreshInterval' of %s appears as less than 60s. Using fallback of 5m", durStr)
 			duration = time.Minute * 5
 		}
 
--- a/xrootd/xrootd_config.go
+++ b/xrootd/xrootd_config.go
@@ -629,7 +629,7 @@ func authRefreshStrToSecondsHookFunc() mapstructure.DecodeHookFuncType {
 		}
 
 		// Sanitize the input to guarantee we have a unit
-		suffixes := []string{"s", "m", "h", "d"}
+		suffixes := []string{"s", "m", "h"}
 		hasSuffix := false
 		for _, suffix := range suffixes {
 			if strings.HasSuffix(durStr, suffix) {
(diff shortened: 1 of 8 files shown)


xrootd/xrootd_config.go
```diff
@@ -605,6 +607,67 @@ func LaunchXrootdMaintenance(ctx context.Context, server server_structs.XRootDSe
 	)
 }
 
+// The default config has `Xrootd.AuthRefreshInterval: 5m`, which we need to convert
+// to an integer representation of seconds for our XRootD configuration. This hook
+// handles that conversion during unmarshalling, as well as some sanitization of user inputs.
+func authRefreshStrToSecondsHookFunc() mapstructure.DecodeHookFuncType {
```
haoming29, line 613: I've done a similar string to xrootd int value conversion for cache watermark parameter: https://github.com/PelicanPlatform/pelican/blob/c5ad188c8f1c43d2265843fb37dde3459933e1b0/config/config.go#L802-L839

I didn't do it at the marshaling level but in the config validation, where I read in the value, check the value, and return the xrootd value if everything goes well. At some point we should consolidate different approaches into a unified one.

One repetition I saw here is handling the duration conversion. Viper has the nice check against the config value and we should be able to reuse that and get the value in second instead. I'll let you decide

xrootd/xrootd_config.go
```diff
@@ -605,6 +607,67 @@ func LaunchXrootdMaintenance(ctx context.Context, server server_structs.XRootDSe
...
+			durStr = durStr + "s"
+		}
+
+		duration, err := time.ParseDuration(durStr)
+		if err != nil {
+			return nil, err
+		}
+
+		if duration < time.Second {
+			log.Warningf("'Xrootd.AuthRefreshInterval' appears as less than a second. Using fallback of 5m")
```
haoming29, line 651: I passed `Xrootd.AuthRefreshInterval: 10s` and xrootd refused to start with the following error:

```console
INFO[2024-06-10T20:32:42Z] 240610 20:32:42 27084 acc_a2x: authrefresh value 10 may not be less than 60  daemon=xrootd.origin
```

We should do this check at Pelican level to get more user friendly error message

haoming29 (changes requested):

xrootd/xrootd_config.go
jhiemstrawisc: I think I'd rather leave this in the unmarshal hook for now, since it handles type conversion of `time.Duration --> int`.

haoming29 (approved):
LGTM!

PR #1387, opened 2024-06-06
Add prettier to keep code style consistent

CannonLock:


haoming29:
@CannonLock can you fix the merge conflict here? I'll merge this in once the conflicts are resolved.

.github/workflows/pre-commit-linter.yml
```diff
@@ -44,3 +44,20 @@ jobs:
...
         uses: pre-commit/action@v3.0.0
+
+  prettier-linter:
+    name: prettier-linter
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
```
haoming29, line 55: Let's upgrade the action to `@v4`

.github/workflows/pre-commit-linter.yml
```diff
@@ -44,3 +44,20 @@ jobs:
...
+        with:
+          node-version: '22'
```
haoming29, line 57: Any specific reason we are using 22 instead of 20 here? I saw we use 20 in other places: https://github.com/PelicanPlatform/pelican/blob/1ec45a9618e4a64991230a5504650ed013e01fc1/.github/workflows/test.yml#L21

web_ui/frontend/package.json
```diff
@@ -8,6 +8,8 @@
     "start": "next start",
     "export": "next export",
     "lint": "next lint",
+    "format": "prettier --check --ignore-path .gitignore .",
```
haoming29, line 11: Do we want to include `next lint` here or the other way around? if both of them are running eslint

haoming29 (changes requested):

web_ui/frontend/package.json
CannonLock: I believe ( and just double checked ) that prettier just formats and doesn't do any linting. The linting is run during the build step.

haoming29 (approved):
LGTM

PR #1388, opened 2024-06-06
Update Director responses and client handling

### LINKED ISSUE #1369: Better Formatting for HTTP 500 Error Code on Pelican Client
Pelican Version:
Version: 0.0.1-next

Pelican Service:
- [x] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

**Describe the bug**
There is insufficient logging for when there is an unanticipated HTTP error code such as 500. It seems to print out a struct of some sort but does not format it in an interpretable or meaningful way.

**To Reproduce**
I ran into this issue when I was following the contributor guide found in the `CONTRIBUTE.md` file. I ran into the issue when I ran the following command:
`./pelican object copy -d -f https://osg-htc.org /ospool/uc-shared/public/OSG-Staff/validation/test.txt downloaded-testfile.txt`

**Expected behavior**
Ideally there would logging of the response body of the request and maybe some other info to help debug the problem.

**Screenshots**
Screenshot 2024-06-03 at 3 07 20 PM

**Desktop**
 - OS: MacOS running ARM64 

**Additional context**
- Additionally I was attached to a dev container running locally on my Mac

patrickbrophy:
This PR was made to fix #1369

### DIFF
--- a/client/director.go
+++ b/client/director.go
@@ -147,6 +147,11 @@ func queryDirector(ctx context.Context, verb, sourcePath, directorUrl string) (r
 	// Check HTTP response -- should be 307 (redirect), else something went wrong
 	body, _ := io.ReadAll(resp.Body)
 
+	contentType := req.Header.Get("Content-Type")
+	if contentType != "application/json" {
+		log.Debugln(string(body))
+	}
+
 	// If we get a 404, the director will hopefully tell us why. It might be that the namespace doesn't exist
 	if resp.StatusCode == 404 && verb == "PROPFIND" {
 		// If we get a 404 response from a PROPFIND, we are likely working with an old director so we should return a response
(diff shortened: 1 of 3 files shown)


haoming29:
A couple of things spotted and nitpicking here since this is your first PR:
* Capitalize PR title (first letter should be fine)
* No period in the title, and try use one sentence/phrase for your title
* Fix test failure. There seems to be two failure according to the test log. Let me know if you need help accessing/identifying the source of the failure
* Request at least one contributor to review when your PR is ready (me for this one)
* Tag the PR by an appropriate label. For this one, `client` and `bug` are good ones

client/director.go
```diff
@@ -162,11 +158,11 @@ func queryDirector(ctx context.Context, verb, sourcePath, directorUrl string) (r
 		// If we get a 405 with a PROPFIND, the client will handle it
 		return
 	} else if resp.StatusCode != 307 {
```
haoming29, line 160: Oops. One thing we missed here is to check the response content type. If the content type is not "application/json", let's simply log `string(body)`

haoming29 (changes requested):
LGTM with one small change requested. Once you get that one fixed and tested locally for backward compatibility, we should be good to go!

client/director.go
```diff
@@ -162,11 +158,15 @@ func queryDirector(ctx context.Context, verb, sourcePath, directorUrl string) (r
...
-		var respErr directorResponse
+		contentType := req.Header.Get("Content-Type")
+		if contentType != "application/json" {
+			log.Debugln(string(body))
```
haoming29, line 163: Ah my bad. I meant to return a new error with `string(body)` as the error message instead. Sorry for the confusion!

haoming29 (changes requested):
One last step!

haoming29 (approved):
LGTM! Congrats on your first PR to the Pelican repo! Feel free to merge this in.

Issue #1239, opened 2024-05-07
Create client command 'stat' on an object in a namespace


joereuss12:
Pelican Service:
- [X] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

Like adding the 'ls' command in Pelican, we should also add a 'stat' command, a lot of the framework for stat is there but we need to make this an actual command in Pelican

PR #1400, opened 2024-06-10
Client `stat` branch

haoming29:
This is a new PR for #1239 

FYI, there's [existing code review feedback](https://github.com/PelicanPlatform/pelican/pull/1321) that's addressed in this PR


haoming29:
@bbockelm Looking at the `stat` code, I didn't find anywhere in the code that the function directly prints out the result. We packaged the result and returned back to `main.go` for printing. Are you referring to the `ls` command instead?

client/handle_http.go
```diff
@@ -2426,6 +2432,73 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
 
 }
 
+// This function queries the director with a PROPFIND to attempt to get the 'collections url'. If a propfind is not allowed on the director
+// We fall back to the deprecated dirlisthost in the namespace
+func getCollectionsUrl(ctx context.Context, remoteObjectUrl *url.URL, namespace namespaces.Namespace, directorUrl string) (collectionsUrl *url.URL, err error) {
+	// If we are a recursive download and using the director, we want to attempt to get directory listings from
+	// PROPFINDing the director
+	if directorUrl != "" {
```
jhiemstrawisc, line 2440: Would it ever be the case this far along in Pelican that we could have an unpopulated director URL? One of the 7.9 PRs I bumped to 7.10 is removing most of the code paths that have this conditional, since I don't think we'll ever _not_ enter this section.

client/handle_http.go
```diff
@@ -2426,6 +2432,73 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+			}
+			// The resp redirect includes the full path to the object from the origin, we are only interested in the scheme and host
+			collectionsURL, err := url.Parse(collections)
+			if err != nil {
+				return nil, err
+			}
+			collectionsUrl = &url.URL{
+				Scheme: collectionsURL.Scheme,
+				Host:   collectionsURL.Host,
+			}
```
jhiemstrawisc, line 2483: What's the reason for re-instantiating this URL from the URL we just parsed?

client/handle_http.go
```diff
@@ -2426,6 +2432,73 @@ func runPut(request *http.Request, responseChan chan<- *http.Response, errorChan
...
+			namespace.DirListHost = collectionsUrl.String()
```
jhiemstrawisc, line 2484: I don't think this has an effect, does it? The input `namespace` is a struct and not a pointer.

client/handle_http.go
```diff
@@ -2607,31 +2679,36 @@ func statHttp(ctx context.Context, dest *url.URL, namespace namespaces.Namespace
...
 				}
+				client.SetTransport(transport)
 
-				resp, err = client.Do(req)
+				fsinfo, err := client.Stat(endpoint.Path)
 				if err == nil {
+					info = fileInfo{
+						Name:    path.Base(endpoint.Path),
+						Size:    fsinfo.Size(),
+						IsDir:   fsinfo.IsDir(),
```
jhiemstrawisc, line 2697: What does this look like outside of POSIX origins? If I were to use `client.Stat` against an S3 prefix, do we say "IsDir: true", or is that behavior undefined?

It's fine if we don't have an answer now, but something we should be asking.

cmd/object_stat.go
```diff
@@ -0,0 +1,92 @@
...
+* limitations under the License.
+*
+***************************************************************/
+
+package main
+
+import (
+	"os"
+
+	"github.com/pelicanplatform/pelican/client"
```
jhiemstrawisc, line 24: Can you split these imports?

cmd/object_stat.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if len(args) < 1 {
+		log.Errorln("no object provided")
+		err = cmd.Help()
+		if err != nil {
+			log.Errorln("failed to print out help:", err)
+		}
+		os.Exit(1)
+	}
+	object := args[len(args)-1]
```
jhiemstrawisc, line 71: Is it worth handling multiple objects in a single stat call?

jhiemstrawisc (changes requested):

client/handle_http.go
haoming29: Looks like we *do* have a check for empty director URL at the start of the cmd file, which means it's safe to assume non-empty director url in the downstream functions. I'll remove this check here then.
https://github.com/PelicanPlatform/pelican/blob/1bfb0bfe5f2f54c75ce0286747833039c1bfb6a3/client/handle_http.go#L550

https://github.com/PelicanPlatform/pelican/blob/1bfb0bfe5f2f54c75ce0286747833039c1bfb6a3/client/handle_http.go#L630-L632

haoming29: Actually, in `newPelicanUrl`, not all if branches check the returned directorURL and see if it's empty. I'll keep this as-is and let's do a one-off change for all such occurrences if we want to

client/handle_http.go
haoming29: It seems like we are trying to get the `DirListHost` around line 2484 out of the `collectionsURL`, which contains the path to the file, and we want to get rid of the path. I'll change naming to be less confusing

client/handle_http.go
haoming29: Honestly I don't know what's going on here. I'll remove this section

cmd/object_stat.go
haoming29: I'd say let's do one object for now just to get this through. This also acts like Linux stat

jhiemstrawisc (approved):
LGTM. There were a few failing tests, but I suspect the failures are transient. Let's make sure they pass before merging.

haoming29:
@jhiemstrawisc While I was revising #1393 which shares the `getColletionsUrl` function with this PR, I saw Brian's comments on how to make this function compatible with upcoming changes in the director, and I need to change this function to get there. I'm not sure how we want to handle that change here? We can merge this PR in and I can do a rebase against main, or let me know if you have different thoughts.

Also, I think Brian mentioned in the `ls` PR that the client library shouldn't handle printing in the client library and I need to move the logic to the CLI level. So I'd need to change that too.

I'll probably ask for another round of review after all these updates. I wasn't aware of that when I was revising this PR. Sorry about the last minute change!

Issue #1407, opened 2024-06-12
Include namespace prefix as a label in `pelican_director_total_advertisements_received`


haoming29:
* Include the namespace in the metric to allow more useful monitoring message
* Change the metric name to follow Prometheus metric/label naming best practice: https://prometheus.io/docs/practices/naming/, so `pelican_director_advertisements_received_total` is a better choice

PR #1410, opened 2024-06-12
Updating `pelican_director_total_advertisements_received` to a better name and to include the namespace prefix as a label

patrickbrophy:
This PR updates the `pelican_director_total_advertisements_received` metric to `pelican_director_advertisements_received_total`. This PR also adds the namespace prefix as a label to the metric. This PR directly addresses #1407.


bbockelm:
Before we merge, can we please clean up the git history a bit?  Also run the commit messages through spell check.

patrickbrophy:
I just cleaned up the commits using rebase. Sorry about the mess.

director/director.go
```diff
@@ -757,6 +757,7 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 	// Set to ctx for metrics handler downstream
 	ctx.Set("serverName", adV2.Name)
 	ctx.Set("serverWebUrl", adV2.WebURL)
+	ctx.Set("namespacePrefix", adV2.RegistryPrefix)
```
haoming29, line 760: Ah, our naming isn't very clear and confuse people. What we want is the federation namespace prefix (i.e. namespace prefix used for object access). Here, the `RegistryPrefix` in `adV2` is the *server* namespace, i.e. namespace for the origin/cache server. What you should look at instead is around line 845 where we iterate over the `adV2.Namespaces`, this is what we want to pass along. You want to pass specifically `NamespaceAdV2.Path` as the values. Also note that an origin can export multiple federation namespace prefixes, so this will be an array. In your following middleware, you want to iterate over the array and add those paths ad different label values, if that makes sense.

haoming29 (changes requested):
See comment below

director/director.go
```diff
@@ -758,6 +758,16 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
 	ctx.Set("serverName", adV2.Name)
 	ctx.Set("serverWebUrl", adV2.WebURL)
 
+	// Iterate over each advertised namespace and join the paths together
+	// into a string where each path is separated by a space
+	// i.e. "<path> <path> <path>"
+	var namespacePaths string
+	for _, namespace := range adV2.Namespaces {
+		path := namespace.Path
+		namespacePaths = fmt.Sprintf("%s %s", namespacePaths, path)
```
haoming29, line 767: This will introduce a leading whitespace in your result, can you try trim whitespaces of the final result? `strings.TrimSpace` should be good

director/director.go
```diff
@@ -758,6 +758,16 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
...
+	}
+	ctx.Set("namespacePaths", namespacePaths)
```
haoming29, line 769: You set the context to be `namespacePaths` but later you read `namespacePrefixes` instead. I'd suggest we use `namespacePrefixes` for both

director/director.go
```diff
@@ -903,13 +913,19 @@ func serverAdMetricMiddleware(ctx *gin.Context) {
...
-		prometheus.Labels{
-			"server_name":    serverName,
-			"server_web_url": serverWebUrl,
-			"server_type":    ctx.GetString("serverType"),
-			"status_code":    strconv.Itoa(statusCode),
-		}).Inc()
+	// Separate each path by the space separator
+	namespacePaths := strings.Split(ctx.GetString("namespacePrefixes"), " ")
+	// Update metrics for each path
+	for _, namespacePath := range namespacePaths {
```
haoming29, line 919: We should also skip empty `namespacePath` string here

haoming29 (changes requested):
We are getting close! A couple other things I spotted.

haoming29 (approved):
LGTM! Can you create a follow-up ticket to document this metric on the docs website?

haoming29:
Unfortunately this didn't get merged to our 7.9 release (we were so close!), but your other PRs did! And the good news is this is likely to be the first contribution to the 7.10 release.

PR #1412, opened 2024-06-12
Consolidate multi-export topology origins

### LINKED ISSUE #1406: Wrong key for aggregating namespaces under the same origin for topology servers
In https://github.com/PelicanPlatform/pelican/blob/main/director/advertise.go#L238 , we are aggregating the namespaces under the same origin/cache servers, but we used `serverAd` as the unique key for the server. This will cause wrong aggregation result as `serverAd` can be different for the same origin, if the namespaces it exports have different capabilities. Take ap20 origin as an example, it exports two namespaces, `/ospool/ap20/data` and `/ospool/ap20/.well-known`. Below is the topology json for the two namespaces:

* `/ospool/ap20/data`: 
```json
{
    "caches": [...], // entries skipped for readability
    "credential_generation": {
        "base_path": null,
        "issuer": "https://osg-htc.org/ospool",
        "max_scope_depth": 4,
        "strategy": "OAuth2",
        "vault_issuer": null,
        "vault_server": null
    },
    "dirlisthost": null,
    "origins": [
        {
            "auth_endpoint": "ap20.uc.osg-htc.org:1095",
            "endpoint": "ap20.uc.osg-htc.org:1094",
            "production": true,
            "resource": "UChicago_OSGConnect_ap20"
        }
    ],
    "path": "/ospool/ap20/data",
    "readhttps": true,
    "scitokens": [
        {
            "base_path": [
                "/ospool/ap20"
            ],
            "issuer": "https://ap20.uc.osg-htc.org:1094/ospool/ap20",
            "restricted_path": []
        }
    ],
    "usetokenonread": true,
    "writebackhost": "https://ap20.uc.osg-htc.org:1095"
}
```

* `/ospool/ap20/.well-known`:
```json
{
    "caches": [],
    "credential_generation": null,
    "dirlisthost": null,
    "origins": [
        {
            "auth_endpoint": "ap20.uc.osg-htc.org:1095",
            "endpoint": "ap20.uc.osg-htc.org:1094",
            "production": true,
            "resource": "UChicago_OSGConnect_ap20"
        }
    ],
    "path": "/ospool/ap20/.well-known",
    "readhttps": false,
    "scitokens": [],
    "usetokenonread": false,
    "writebackhost": null
}
```

Notice that although both namespaces have the exact same `origin` field, they differ in the capabilities and issuer, and we parse the two entries into `serverAd`, the result will be different too. Causing duplicated entries.

To fix the issue, we should 
1.Always sort the `namespace` filed from topology json by `path` field to make sure we are getting stable result
2.Come up with a mechanism for identifying same origin exporting multiple namespaces and "correctly" parse the origin `serverAd`. In this case, we shouldn't use the information from `/ospool/ap20/.well-known` for the ap20 origin but from `/ospool/ap20/data`, as it provides the correct issuer and auth information.

haoming29:
Fixes #1406 

A quick check can be running the director in `osdf` mode, then go to `/api/v1.0/director_ui/servers`. Look for `unl-origin.nationalresearchplatform.org:1094` which should export 4 namespaces, and `sdsc-origin.nationalresearchplatform.org:1094` which should export 11 namespaces.

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -32,6 +32,26 @@ import (
 	"github.com/pelicanplatform/pelican/utils"
 )
 
+// Consolite two ServerAds that share the same ServerAd.URL. For all but the capability fields,
+// the existing ServerAds takes precedence. For capability fields, an OR is made between two ads
+// to get a union of permissions.
+func consolidateDupServerAd(newAd, existingAd server_structs.ServerAd) server_structs.ServerAd {
+	consolidatedAd := existingAd
+
+	// Update new serverAd capabilities by taking the OR operation so that it's more permissive
+	consolidatedAd.Caps.DirectReads = existingAd.Caps.DirectReads || newAd.Caps.DirectReads
+	consolidatedAd.Caps.PublicReads = existingAd.Caps.PublicReads || newAd.Caps.PublicReads
+	consolidatedAd.Caps.Reads = existingAd.Caps.Reads || newAd.Caps.Reads
+	consolidatedAd.Caps.Writes = existingAd.Caps.Writes || newAd.Caps.Writes
+	consolidatedAd.Caps.Listings = existingAd.Caps.Listings || newAd.Caps.Listings
+
+	consolidatedAd.DirectReads = existingAd.DirectReads || newAd.DirectReads
+	consolidatedAd.Writes = existingAd.Writes || newAd.Writes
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (approved):
The tests you provided worked like a charm. LGTM

PR #1415, opened 2024-06-13
Fix only one namespace read from topology cache

haoming29:
* Corrected typo in map name and add unit test with real topology data

To test, run the director server and head to `/api/v1.0/director_ui/servers?server_type=cache`, and search for `/chtc/PUBLIC`, you should get 29 result, meaning there are 29 caches willing to serve this namespace. If you go to topology endpoint https://topology.opensciencegrid.org/stashcache/namespaces.json and find `/chtc/PUBLIC`, the `cache` array should also have 29 items


jhiemstrawisc:
The fix here looks simple enough, but is it possible to perform these tests by adding to or modifying the mock topology JSON? Otherwise we end up adding ~12k lines of code that become obsolete the next time the topology schema is updated.

haoming29:
> The fix here looks simple enough, but is it possible to perform these tests by adding to or modifying the mock topology JSON? Otherwise we end up adding ~12k lines of code that become obsolete the next time the topology schema is updated.

Good point. I was adding to the mock topology but thought it makes the mock too complicated to maintain. I'll instead shorten the real topology json for the current test if that makes thing cleaner

haoming29:
@jhiemstrawisc I've shortened the topology json to be minimum to test

jhiemstrawisc (approved):
LGTM.

PR #1416, opened 2024-06-13
Add bearer token config to HTTP backends for Globus compat

### LINKED ISSUE #1413: Pass Globus token location to xrootd

jhiemstrawisc:
This adds a generic config for adding an bearer token to the `Authorization` header on requests from the origin to an http backend. In particular, this is needed for Globus collections.

Closes #1413

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -907,6 +907,17 @@ type: string
 default: none
 components: ["origin"]
 ---
+name: Origin.HttpAuthTokenFile
+description: |+
+  When set, all requests from the origin to the http backend will include the contents of the file as a bearer token in the
+  Authorization header.
+
+  If the origin backend is configured with the `globus` storage type, any value set here will be overridden with the filepath to
+  the first file ending in `.tok` found in the $(Origin.GlobusConfigLocation)/tokens directory
+type: filename
+default: none
+components: ["origin"]
+---
 name: Origin.XRootServiceUrl
 description: |+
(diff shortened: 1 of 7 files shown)


jhiemstrawisc:
@haoming29, when I was deciding whether to export the `origin.GlobusActivated` field or the general `origin.isExportActivated` function, I settled on the former because the collections don't yet support multiple collections in `xrootd_config.go` where we only look at the first collection in the exports slice. If you think I should change that, let me know.

jhiemstrawisc:
Also, thoughts on the security warning? Outside of exporting the `origin.GlobusTokenFileExt` var, I haven't changed the way this directory is handled.

haoming29:
> Also, thoughts on the security warning? Outside of exporting the `origin.GlobusTokenFileExt` var, I haven't changed the way this directory is handled.

Yeah, but this should be from the previous code I wrote. We are reading the user-provided `cid` from query param, but this is the only way to match the collection. I've added basic path cleaning and security check to avoid path traversal. Let me know if this is still a concern.

haoming29 (approved):
LGTM. It makes sense to me to make the three variable public instead of exporting that `isExportActivated` function because that function relies on internal struct to work and publishing this may impose null-pointer risk

haoming29:
@jhiemstrawisc I just hand-tested and realized I need to change ownership of the token path to xrootd. To push things through, I've patched the fix to this PR. Feel free to take a look

PR #1432, opened 2024-06-18
Removed logging shoveler message at debug level and fix empty version

haoming29:
As @bbockelm reported, the current logging setup in the shoveler logs every message it sent to the MQ, which causes lots of noises at the debug level. This PR removed logging such messages.

Also, currently the `version` field in the message sent to MQ is empty, due to the fact that we shoveler is now part of Pelican code. This PR set the related shoveler build info with that from Pelican


patrickbrophy (approved):
Looks good! LGTM

PR #1436, opened 2024-06-20
Add two Prometheus metrics for director `stat` queries

### LINKED ISSUE #1431: Add metric to the `stat` call to record # of stat requests in the system per storage server

haoming29:
Closes #1431 

To test, first spin up your director. If you'd like to also test a local file, then you'll need to spin up your local federation (director, registry, origin)
* Get the test file through the origin: `curl -k -o /dev/null https://localhost:8444/api/v1.0/director/ospool/uc-shared/public/OSG-Staff/validation/test.txt`. Assuming that your local director runs at `8444`
* Check the Prometheus `/metric` endpoint to confirm the value. Login to your director web UI and go to `https://localhost:8444/metrics`. You should see two metrics named `pelican_director_stat_active` and `pelican_director_stat_total` where the value of `pelican_director_stat_total` should be 1 (for the test file you just queried)

### DIFF
--- a/director/monitor.go
+++ b/director/monitor.go
@@ -223,8 +223,8 @@ func LaunchPeriodicDirectorTest(ctx context.Context, serverAd server_structs.Ser
 									"server_name":    serverName,
 									"server_web_url": serverWebUrl,
 									"server_type":    string(serverAd.Type),
-									"status":         string(metrics.FTXTestSucceeded),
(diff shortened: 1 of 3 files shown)


haoming29:
@jhiemstrawisc This should be one of the items to backport to 7.9.1

patrickbrophy (approved):
Looks good! LGTM

PR #1439, opened 2024-06-20
Fix deadlock in `recordAd` and `LaunchTTLCache` and add support to turn off `stat`

### LINKED ISSUE #1438: Deadlock in the director between `statUtilsMutex` and `healthTestUtilsMutex`
Pelican Version:

Pelican Service:
- [ ] Director

**Describe the bug**

We flipped the order to locking when acquiring both `statUtilsMutex` and `healthTestUtilsMutex` in `recordAd` and `LaunchTTLCache`, causing the deadlock.

**To Reproduce**

It's difficult to reproduce the issue as multiple attempts in ITB and locally failed. But the code is relatively straightforward in suggesting the deadlock condition

**Expected behavior**

**Screenshots**

 - OS: [e.g. Linux, MacOS]

**Additional context**

### LINKED ISSUE #1440: Add config knob to turn off `stat` in the director

haoming29:
Fixes #1438 and closes #1440

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -397,7 +397,7 @@ func redirectToOrigin(ginCtx *gin.Context) {
 	reqPath := path.Clean("/" + ginCtx.Request.URL.Path)
 	reqPath = strings.TrimPrefix(reqPath, "/api/v1.0/director/origin")
 
-	disableStat := param.Director_DisableStat.GetBool()
+	disableStat := !param.Director_EnableStat.GetBool()
 
 	// Skip the stat check for object availability
 	// If either disableStat or skipstat is set, then skip the stat query
--- a/director/director.go
+++ b/director/director.go
@@ -397,8 +397,15 @@ func redirectToOrigin(ginCtx *gin.Context) {
 	reqPath := path.Clean("/" + ginCtx.Request.URL.Path)
 	reqPath = strings.TrimPrefix(reqPath, "/api/v1.0/director/origin")
 
+	disableStat := param.Director_DisableStat.GetBool()
+
 	// Skip the stat check for object availability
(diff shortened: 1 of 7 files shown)


director/director.go
```diff
@@ -397,8 +397,15 @@ func redirectToOrigin(ginCtx *gin.Context) {
 	reqPath := path.Clean("/" + ginCtx.Request.URL.Path)
 	reqPath = strings.TrimPrefix(reqPath, "/api/v1.0/director/origin")
 
+	disableStat := param.Director_DisableStat.GetBool()
```
jhiemstrawisc, line 400: Can we change the config name to match the style we use elsewhere, i.e. `Director.EnableStat` and then flip the default to be `true`?

jhiemstrawisc (changes requested):

director/director.go
haoming29: I thought the idea/guideline is to name the config in a way that the default value is `false`? No?

jhiemstrawisc: Hmm, I don't know that we've settled on a formal guideline, have we? Looking over `docs/parameters.yaml` shows that we're mostly using `.Enable*`-type parameters (some of which default to true, and some to false), although we do have one "Disable" param that doesn't come from a legacy StashCP configuration (`Logging.DisableProgressBars`). If you prefer the param as is, I'll approve and let you make the decision.

jhiemstrawisc (approved):

director/director.go
haoming29: I'll do `Enable*` then just to keep things consistent

haoming29:
@jhiemstrawisc I've updated to use `EnableStat` instead. Not sure if you want to take another look

Issue #1425, opened 2024-06-17
Record client IP address as a label to Prometheus metrics for xrootd access


haoming29:
* Xrootd already reports client IP address in the detailed monitoring packet: https://xrootd.slac.stanford.edu/doc/dev56/xrd_monitoring.htm#_Toc138968517 (see `host` field) and we have parsed this information into `XrdUserId`, which is used as the key for `userids` TTL cache.
* We creates `sessions` user sessions which tells us information about the user
* We associates `sessions` with `transfers` which contains user file accesses
* We listen to specific type of events for updating these three maps and when there's a file close event, we generate a new Prometheus record for file access
* In this ticket, we want to attach the `host` as a new label for access-related metrics, i.e. `xrootd_transfer_readv_segments_count`, `xrootd_transfer_operations_count`, and `xrootd_transfer_bytes`. 
---

* To do this, we first need to pass the `host` information from `XrdUserId` struct to `UserRecord`, which is the struct for values in `sessions`" . We want to add `host` field to the `UserRecord` struct. https://github.com/PelicanPlatform/pelican/blob/25df30c4bec61d0dea04b1db29615448867e7682/metrics/xrootd_metrics.go#L928-L929
* When we are generating a new Prometheus record when the file is closed, we read from the `sessions` map to get the `host`: https://github.com/PelicanPlatform/pelican/blob/25df30c4bec61d0dea04b1db29615448867e7682/metrics/xrootd_metrics.go#L660-L661
* You also want to keep an eye on wherever we update the value for `sessions` and `userids` map to make sure you aren't missing the `host` field.

To reduce the cardinality of this new label, let's do a `/16` subnet mask and record the masked value.

All the updates should be in `xrootd_metrics.go`, for the subnet function, we should put it in the `utils` package.

bbockelm:
I think a /16 is too much for IPv4; let's use /24 instead.

For IPv6, let's use an aggregation at /64.

PR #1445, opened 2024-06-21
Add IP address (host) label to XRootD access metrics

patrickbrophy:
This PR addresses #1425. This PR adds a IP address (host) label to XRD access metrics (xrootd_transfer_readv_segments_count, xrootd_transfer_operations_count, and xrootd_transfer_bytes). This PR also updated some tests to handle the new label.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -19,6 +19,7 @@
 package utils
 
 import (
+	"net"
 	"net/url"
 	"strings"
 	"unicode"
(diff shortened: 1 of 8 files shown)


utils/utils.go
```diff
@@ -102,3 +103,49 @@ func CheckValidQuery(transferUrl *url.URL) (err error) {
...
+	mask := net.CIDRMask(64, 128)
+	maskedIP := ip.Mask(mask)
+	return maskedIP.String()
+}
+
+// MaskIP applies the appropriate subnet mask to the input IP address.
+func MaskIP(ipStr string) string {
+	ip := net.ParseIP(ipStr)
+	if ip == nil {
+		// Invalid IP
```
bbockelm, line 135: Does this cause an explosion in cardinality if a serves is using hostnames and not an IP?

Good habits would have an indication to the caller that a special case occurred. Maybe also return a Boolean?

utils/utils.go
```diff
@@ -102,3 +103,49 @@ func CheckValidQuery(transferUrl *url.URL) (err error) {
...
+		return ipStr
+	}
+
+	if ip.To4() != nil {
+		// IPv4
+		return IPv4Mask(ipStr)
```
bbockelm, line 141: Can we avoid parsing the IP string twice?

bbockelm (changes requested):
Three requests:
- can we leave out the unrelated movement of structs to a separate module?  That causes code churn unrelated to the PR (but may be fine for a separate PR).
- there's a Xrootd setting, `xrd.network nodnr`, that should disable the use of reverse DNS to create a hostname in the monitoring records.  Can we try setting that here?  See https://xrootd.slac.stanford.edu/doc/dev56/xrd_config.htm#_Toc152596878
- given we are using networks and not hostnames, let's rename the labels to "network" perhaps?  Looking around on Google, it's not clear if there's a standard practice here.

patrickbrophy:
I'm closing this PR to separate out concerns. I'll open a PR for this metric and a separate one for moving around the structs from the utils folder.

PR #1451, opened 2024-06-24
Upgrade viper to handle unmarshaling env into nested struct

### LINKED ISSUE #1446: Mitigate viper bug in failing to bind env to nested config
Ref: https://github.com/spf13/viper/issues/761

As Fabio reported, when setting the `Server.TLSCACertificateDirectory` config via environment variable, i.e. `PELICAN_SERVER_TLSCACERTIFICATEDIRECTORY=/etc/grid-security/certificates`, the value isn't propagated to xrootd config. This is because we used `viper.Unmarshal` to unmarshal viper configuration to our defined xrootd configuration struct: https://github.com/PelicanPlatform/pelican/blob/b269e79872031b5d7cb02149960b7f3df12ea933/xrootd/xrootd_config.go#L680-L682

However, given the bug in viper, the values of nested fields are lost in unmarshaling. People in the the referred issue proposed a couple of work-around and we should pick one and replace all our instances where we used `viper.Unmarshal`.

haoming29:
Fixes #1446 

This PR upgraded viper to a pre-release version v1.20.0-alpha3 (came out this morning that fixes the issue), as they recently added this experimental support of handling unmarshaling env into nested struct via a viper option. They initially added support in v1.18.0 (https://github.com/spf13/viper/pull/1429) but with some reported rare use case failure, they then did a patch to hide this feature behind a feature flag. In 1.20, they will make this available by config option (https://github.com/spf13/viper/pull/1854). 

If all tests are passing, I don't think there's anything that blocks us from using this pre-released version.

To test locally, follow bug reproduction in the original issue.


jhiemstrawisc (approved):
Tested against the env var that had been causing problems and confirmed this works.

Issue #1309, opened 2024-05-20
Authenticated redirect call to the director


haoming29:
If a namespace is reported to be protected, instead of allowing the client to contact the namespace directly, we want to check if the token the client carries if valid first, if not, we should return the issuer url for the namespace back to the client with a 403 error.

@joereuss12 as a heads up, this should require client-side change to handle the 403 error

bbockelm:
Closing as not planned due to inactivity.

PR #1454, opened 2024-06-25
Add `stat` query to `redirectToCaches`

### LINKED ISSUE #1308: Add `stat` query to `redirectToCaches`
As we integrate `stat` query to `redirectToOrigins`, let's also do this to the cache, but for caches, instead of filtered out ones that does not have the object, we rank them to a lower priority, to allow them to be redirect to if none of the caches have the object

haoming29:
Closes #1308 

The idea is that for redirectToCaches, if `stat` is enabled (by default):
* Query **all** origins and caches that claim to serve the namespace
* If no caches claim to serve the namespace (not that they have the object), then iterate queried origins that claim to have the object and check if they enable "DirectRead". If none, then return 404. If there is, use the first one as the object server.
* If there's one or more caches claim to serve the namespace, we sort the caches based on existing GeoIP, but then sort again by availability. Caches that have the object have higher priority in the list
* Return the sorted caches

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -317,7 +317,12 @@ func redirectToCache(ginCtx *gin.Context) {
 		q := NewObjectStat()
 		st := config.NewServerType()
 		st.SetList([]config.ServerType{config.OriginType, config.CacheType})
-		qr := q.Query(context.Background(), reqPath, st, 1, 6,
+		// Set max response to all available origin and cache servers to ensure we stat against origins
+		// if no cache server has the file
+		// TODO: come back and re-evaluate if we need this many responses and potential origin/cache
+		// server performance issue out of this
+		maxRes := len(cacheAds) + len(originAds)
+		qr := q.Query(context.Background(), reqPath, st, 1, maxRes,
 			withOriginAds(originAds), withCacheAds(cacheAds), WithToken(reqParams.Get("authz")))
 		log.Debugf("Stat result for %s: %s", reqPath, qr.String())
 
@@ -375,7 +380,8 @@ func redirectToCache(ginCtx *gin.Context) {
 		if len(cacheAds) == 0 {
 			ginCtx.JSON(http.StatusNotFound, server_structs.SimpleApiResp{
 				Status: server_structs.RespFailed,
-				Msg:    "No cache found for path",
+				// TODO: make this message clearer by saying no fallback origin found
+				Msg: "No cache found for path",
 			})
 			return
 		}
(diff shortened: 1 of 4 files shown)


director/sort_test.go
```diff
@@ -280,3 +281,26 @@ func TestSortServerAdsByIP(t *testing.T) {
 		assert.NotEqualValues(t, notExpected, sorted)
 	})
 }
+
+func TestSortServerAdsByAvailability(t *testing.T) {
+	firstUrl := url.URL{Host: "first.org", Scheme: "https"}
+	secondUrl := url.URL{Host: "second.org", Scheme: "https"}
+	thirdUrl := url.URL{Host: "third.org", Scheme: "https"}
+	forthUrl := url.URL{Host: "forth.org", Scheme: "https"}
```
jhiemstrawisc, line 289: I think this should be fourthUrl

director/director.go
```diff
@@ -288,9 +299,79 @@ func redirectToCache(ginCtx *gin.Context) {
...
+				ginCtx.JSON(http.StatusInternalServerError, server_structs.SimpleApiResp{
+					Status: server_structs.RespFailed,
+					Msg:    fmt.Sprintf("Failed to query origins with error %s: %s", string(qr.ErrorType), qr.Msg),
+				})
+				return
+			}
+			// For denied servers, append them to availableOriginAds
+			// The qr.DeniedServers is a list of AuthURLs of servers that respond with 403
+			// Here, we need to match against the AuthURL field of originAds
+			for _, ds := range qr.DeniedServers {
```
jhiemstrawisc, line 356: I don't follow why we'd be getting denied from these servers. Don't we pass an auth token along with the stat in line 325?

jhiemstrawisc (changes requested):

director/director.go
haoming29: This is in case the user doesn't know if the origin is protected or not. The future plan is to return 403 with issuer URLs to the denied servers, if all servers denied the stat request, so that the client and automatically invoke OAuth flow

jhiemstrawisc: But shouldn't the Director know whether access for a particular resource requires authorization? If the user tries to stat `/protected/foo` and the Director knows `/protected` requires a token, why not just return a 403 to the user instead of performing the actual stat?

My gut says we should handle "the director knows an origin is protected and the user didn't provide a stat token" differently from "the director knows an origin is protected and the user provided an access token, but the token wasn't accepted by the origin".

haoming29: Yes! That'll be addressed in the issue https://github.com/PelicanPlatform/pelican/issues/1309 I think by then we can say 1. If accessing object from a protected namespace, the director returns 403 straight. 2. If token is given but the all responded stat gave 403, director should respond 403 as well

jhiemstrawisc (approved):

PR #1456, opened 2024-06-25
Take web config values as highest precedence

### LINKED ISSUE #1447: Web override should trump environment variables
The web override is a convenient way to update the Pelican configuration.

However, it appears that environment variables trump all other settings -- meaning you hit "save" on the web interface only to find that the server restarts with the prior value (if the configuration was set through an environment variable; there's no indication this happens in the web UI!).

Let's have the order of precedence go:

- File-based configuration.
- Environment-variable configuration.
- Web-override configuration.

Tagging @brianhlin @matyasselmeci @biozit as operators who might have an opinion on this.

haoming29:
Fixes #1447 

To test, run any of the pelican server with a `PELICAN_*` env variable, and head to the web config view and change the same config to something else, refresh the config page and you should be able to see the updated config on the page.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -911,10 +911,39 @@ func handleContinuedCfg() error {
 	return nil
 }
 
+// Read config file from web UI changes, and call viper.Set() to explicitly override the value
+// so that env wouldn't take precedence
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
LGTM

PR #1458, opened 2024-06-26
GZIP web responses

CannonLock:
8.2mb to 1.8mb


CannonLock:
Going to keep this as a draft as I test to make sure that this doesn't break any endpoints.

CannonLock:
@haoming29 This is ready, but it will cause issues on any endpoints that already gzip such as the prometheus endpoints as it will double zip them. 

Can you think of any endpoints other the prometheus that we don't produce?

haoming29:
@CannonLock looks like there's a unhandled merge conflict that causes tests to fail

web_ui/ui.go
```diff
@@ -557,6 +558,7 @@ func GetEngine() (*gin.Engine, error) {
 	gin.SetMode(gin.ReleaseMode)
 	engine := gin.New()
 	engine.Use(gin.Recovery())
+	engine.Use(gzip.Gzip(gzip.DefaultCompression, gzip.WithExcludedPaths([]string{"/api/v1.0/prometheus"})))
```
haoming29, line 561: When I was testing, the `/metrics` endpoint was zipped into a file, which is not expected. This endpoint should be added to the excluded list as well

haoming29 (changes requested):
One other endpoint that need to be excluded; otherwise LGTM

turetske (changes requested):
Seems good to me, but can you rebase/deal with the merge conflicts before I approve?

CannonLock:
@turetske What merge conflicts?

turetske:
> @turetske What merge conflicts?

Ah, my bad, I was misinterpreting the test failures. Hopefully they are just intermittent failures.

turetske:
@CannonLock The broker tests keep having a timeout. Could adding the gzip like this potentially slow down a callback to be beyond our threshold? (10 minutes)?

jhiemstrawisc:
I noticed this PR is creeping on 5 months since it was opened and 4 months since it was last updated. Is it something that we plan on picking back up?

CannonLock:
@jhiemstrawisc I can close and reopen in the future when complete? This is a small upgrade that looks like it is going to take more time then it is worth to fix so it isn't high on my priority list.

jhiemstrawisc:
This PR is creeping up on a year. If there's no plan to come back and finish it soon, can we close it?

bbockelm:
Cleaning out backlog -- if we want to take another swing, we can reopen.

PR #1462, opened 2024-06-26
Add storage type as part of serverAd and allow origin to turn off the director-based tests

### LINKED ISSUE #1386: Origin should report storage type to the director in its advertisement
For non-POSIX origins, we can't upload files to the origin, thus give us 403 errors when doing so at the director file tests. Let's report the storage type from the origin to the director and skip issuing director tests to that origin if it's not in POSIX, and document the reason why the test is off in various places (prometheus, server log, etc)

haoming29:
Closes #1386


haoming29:
Assigned @turetske as the reviewer and removed @jhiemstrawisc to balance the load

director/director.go
```diff
@@ -1062,17 +1063,24 @@ func registerServeAd(engineCtx context.Context, ctx *gin.Context, sType server_s
...
-		WebURL:      *adWebUrl,
-		BrokerURL:   *brokerUrl,
-		Type:        sType,
-		Caps:        adV2.Caps,
-		Writes:      adV2.Caps.Writes,
-		DirectReads: adV2.Caps.DirectReads,
-		Listings:    adV2.Caps.Listings,
-		IOLoad:      0.5, // Defaults to 0.5, as 0 means the server is "very free" which is not necessarily true
+		Name:                adV2.Name,
+		StorageType:         st,
```
turetske, line 1073: What happens when it's a V1 version of the ads and the storage type is checked and or DisableDirectorTest? Will the break backwards compatibility? Can we test for this?

generate/param_generator.go
```diff
@@ -483,6 +483,10 @@ func (slP StringSliceParam) GetStringSlice() []string {
 	return viper.GetStringSlice(slP.name)
 }
 
+func (slP StringSliceParam) GetName() string {
```
turetske, line 486: Can you create an issue for someone to go through and change all our viper calls to use this instead of the string names? It's busy work, but it would be nice to implement at some point.

director/director_test.go
```diff
@@ -677,6 +683,47 @@ func TestDirectorRegistration(t *testing.T) {
 		teardown()
 	})
 
+	t.Run("origin-storage-type-and-disable-test", func(t *testing.T) {
+		c, r, w := setupContext()
```
turetske, line 687: Can you also create a test that is a posix storage type but tests are disabled an ensure that tests aren't run?

turetske (changes requested):
A few small changes/questions.

generate/param_generator.go
haoming29: Issue created here: https://github.com/PelicanPlatform/pelican/issues/1526

turetske (approved):
LGTM

PR #1470, opened 2024-07-02
Implement intelligent cache/origin selection algorithm

### LINKED ISSUE #1461: Intelligent Director cache selection logic
See details in https://github.com/PelicanPlatform/pelican/discussions/1198

haoming29:
Closes #1461 

This is pending on #1454 to get required object availability data for the sorting algorithm

### DIFF
--- a/director/sort.go
+++ b/director/sort.go
@@ -196,8 +196,8 @@ func sortServerAds(clientAddr netip.Addr, ads []server_structs.ServerAd, availab
 
 	// For each ad, we apply the configured sort method to determine a priority weight.
 	for idx, ad := range ads {
-		switch sortMethod {
-		case "distance":
+		switch server_structs.SortType(sortMethod) {
+		case server_structs.DistanceType:
 			// If we can't get the client coordinates, then simply use random sort
 			if !clientCoordOk {
 				weights[idx] = SwapMap{rand.Float64(), idx}
@@ -212,7 +212,7 @@ func sortServerAds(clientAddr netip.Addr, ads []server_structs.ServerAd, availab
 				weights[idx] = SwapMap{distanceWeight(clientCoord.Lat, clientCoord.Long, ad.Latitude, ad.Longitude, false),
 					idx}
 			}
-		case "distanceAndLoad":
+		case server_structs.DistanceAndLoadType:
 			weight := 1.0
 			// Distance weight
 			if clientCoordOk {
@@ -226,7 +226,7 @@ func sortServerAds(clientAddr netip.Addr, ads []server_structs.ServerAd, availab
 			weight *= lWeighted
 
 			weights[idx] = SwapMap{weight, idx}
-		case "adaptive":
+		case server_structs.AdaptiveType:
 			weight := 1.0
 			// Distance weight
 			if clientCoordOk {
@@ -250,7 +250,7 @@ func sortServerAds(clientAddr netip.Addr, ads []server_structs.ServerAd, availab
 			weight *= lWeighted
 
 			weights[idx] = SwapMap{weight, idx}
-		case "random":
+		case server_structs.RandomType:
 			weights[idx] = SwapMap{rand.Float64(), idx}
 		default:
 			return nil, errors.Errorf("Invalid sort method '%s' set in Director.CacheSortMethod. Valid methods are 'distance',"+
@@ -258,7 +258,7 @@ func sortServerAds(clientAddr netip.Addr, ads []server_structs.ServerAd, availab
 		}
 	}
 
-	if sortMethod == "adaptive" {
+	if sortMethod == string(server_structs.AdaptiveType) {
 		candidates, _ := stochasticSort(weights, candidateLimit)
 		resultAds := []server_structs.ServerAd{}
 		for _, cidx := range candidates[:candidateLimit] {
--- a/director/sort.go
+++ b/director/sort.go
@@ -19,40 +19,19 @@
 package director
 
 import (
-	"archive/tar"
 	"cmp"
-	"compress/gzip"
-	"context"
-	"fmt"
-	"io"
 	"math/rand"
 	"net"
-	"net/http"
 	"net/netip"
-	"os"
-	"path"
-	"path/filepath"
 	"slices"
 	"sort"
 	"strings"
-	"sync/atomic"
-	"time"
 
-	"github.com/oschwald/geoip2-golang"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
-	"github.com/spf13/viper"
 
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/server_structs"
-)
-
-const (
-	maxMindURL string = "https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-City&license_key=%s&suffix=tar.gz"
-)
-
-var (
-	maxMindReader atomic.Pointer[geoip2.Reader]
 )
 
 type (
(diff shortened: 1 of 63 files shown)


haoming29:
@bbockelm one question came up when I'm implementing this. Should we do the same algorithm to `redirectToOrigins`? Right now they share the same sorting function, but we can separate them out if this is not desired.

director/sort.go
```diff
@@ -175,42 +169,87 @@ func getClientLatLong(addr netip.Addr) (coord Coordinate, ok bool) {
...
-				weights[idx] = SwapMap{0 - rand.Float64(), idx}
-			} else {
-				// Each server ad will have a load value that we can use for sorting
-				weights[idx] = SwapMap{distanceAndLoadWeight(clientCoord, ad),
-					idx}
+			weight := 1.0
+			// Distance weight
+			if clientCoordOk {
+				distance := distanceWeight(clientCoord.Lat, clientCoord.Long, ad.Latitude, ad.Longitude, true)
+				dWeighted := gatedHavlingMultiplier(distance, distanceHalvingThreshold, distanceHalvingFactor)
```
jhiemstrawisc, line 220: There's a typo in the function name -- it should be `gatedHalvingMultiplier` (switched v/l)

director/sort_algorithms_test.go
```diff
@@ -0,0 +1,111 @@
...
+	mockSwapMaps := SwapMaps{
+		{Weight: 0.2, Index: 2},
+		{Weight: 0.1, Index: 1},
+		{Weight: 0.4, Index: 6},
+		{Weight: 0.05, Index: 3},
+		{Weight: 10.5, Index: 4},
+		{Weight: 8, Index: 5},
+		{Weight: 2.1, Index: 0},
+	}
+	t.Run("0-maxOut-returns-all", func(t *testing.T) {
```
jhiemstrawisc, line 66: Is this safe if someone accidentally passes a negative value? Can we make this a `uint`?

director/sort_algorithms.go
```diff
@@ -57,29 +58,83 @@ func distanceOnSphere(lat1 float64, long1 float64, lat2 float64, long2 float64)
...
+//
+//  4. Repeat step 2-3 to select a the rest weights
+//
+// Returnss the sorted list of SwapMap.Index and the generated random weights for reference.
+// You may specify the maxOut argument to limit the output.
+func stochasticSort(sm SwapMaps, maxOut int) (candidates []int, randWeights []float64) {
+	if maxOut == 0 {
+		maxOut = len(sm)
+	} else if maxOut > len(sm) {
+		return
```
jhiemstrawisc, line 109: Why return here? If we tell the sort algorithm "don't give me more than 10 back" but there are only 5 candidates, shouldn't we still sort those 5 and return them?

docs/parameters.yaml
```diff
@@ -1268,6 +1268,8 @@ description: |+
   - "distanceAndLoad": Sorts caches according to both their distance and a calculated load. This is currently a placeholder,
     and returns the same ordering as "distance".
   - "random": Sorts caches randomly.
+  - "smart": Sorts caches by distance from the client, server IO load, object availability, and stochastically sort the result.
```
jhiemstrawisc, line 1271: How about something like "Sorts caches according to stochastically-generated weights that consider a combination of factors, including a cache's distance from the client, its IO load, and whether the cache already has the requested object."

On a side note, I don't love the term "smart" because I think this is the kind of thing Miron would have issues with. What about something like "balanced", "adaptive" or "dynamic"? Feel free to disagree to disagree with me on this, I'll defer to you.

jhiemstrawisc (changes requested):
Excellent, this looks like it was a fun one! A few minor comments inline, and a few bigger issues listed here.

When I tested locally, I served an OSDF director with `Director.CacheSortMethod: "smart"`. Curling the director with `curl -v -k https://<hostname>/chtc/PUBLIC/foo` resulted in a broken list of responses where many of the server hostnames have disappeared:
```
< link: <https:///chtc/PUBLIC/foo>; rel="duplicate"; pri=1; depth=1, <https:///chtc/PUBLIC/foo>; rel="duplicate"; pri=2; depth=1, <https:///chtc/PUBLIC/foo>; rel="duplicate"; pri=3; depth=1, <http://osg-kansas-city-stashcache.nrp.internet2.edu:8000/chtc/PUBLIC/foo>; rel="duplicate"; pri=4; depth=1, <http://stash.farm.particle.cz:8000/chtc/PUBLIC/foo>; rel="duplicate"; pri=5; depth=1, <http://osdf-cache.sprace.org.br:8000/chtc/PUBLIC/foo>; rel="duplicate"; pri=6; depth=1
< location: https:///chtc/PUBLIC/foo
< x-pelican-namespace: namespace=/chtc/PUBLIC, require-token=false
< x-pelican-token-generation: max-scope-depth=0
<a href="https:///chtc/PUBLIC/foo">Temporary Redirect</a>.
```
In all tests, it's the first three servers that are missing their hostnames. This is probably coming from the configured `candidateLimit` and some mishandling of the returned candidates, but I haven't gone that far to investigate.

Another issue I encountered is that when I set the sorting method to "distanceAndLoad" and performed the same curl, the Director gets a private IP address and correctly reports "GeoIP Resolution of the address 172.17.0.4 resulted in the null lat/long." When this happens, I get what appears to be a deterministic response with the same ordering:
```
< link: <http://amst-osdf-xcache01.es.net:8000/chtc/PUBLIC/foo>; rel="duplicate"; pri=1; depth=1, <http://dtn-pas.bois.nrp.internet2.edu:8000/chtc/PUBLIC/foo>; rel="duplicate"; pri=2; depth=1, <http://stash-cache.osg.chtc.io:8000/chtc/PUBLIC/foo>; rel="duplicate"; pri=3; depth=1, <http://dtn-pas.cinc.nrp.internet2.edu:8000/chtc/PUBLIC/foo>; rel="duplicate"; pri=4; depth=1, <http://osg-stash-sfu-computecanada-ca.nationalresearchplatform.org:8000/chtc/PUBLIC/foo>; rel="duplicate"; pri=5; depth=1, <http://dtn-pas.denv.nrp.internet2.edu:8443/chtc/PUBLIC/foo>; rel="duplicate"; pri=6; depth=1
< location: http://amst-osdf-xcache01.es.net:8000/chtc/PUBLIC/foo
< x-pelican-namespace: namespace=/chtc/PUBLIC, require-token=false
< x-pelican-token-generation: max-scope-depth=0
<a href="http://amst-osdf-xcache01.es.net:8000/chtc/PUBLIC/foo">Temporary Redirect</a>.
```
Shouldn't there still be something stochastic here as observed when using `distance`, even if we know the load values?

haoming29:
@jhiemstrawisc Regarding your top-level comments:
1. Yes! Thanks for catching that. There's a bug in the sorting code where I preemptively created a slice with size of `cadidateLimit` then did the `append`, which basically means the first three items are empty and not useful. This is fixed
2. This behavior is expected based on the current implementation, which is that if the client IP resolves to nothing, we basically ignores any "distance' related weights. In practice, this means that:
* For `distance`, we do random weight
* For `distanceAndLoad`, we ignore the distance part and only takes load into consideration and generate the weight based on load only.
* For `smart`, similar to  `distanceAndLoad`, but we still consider `availability` as another weight.

The reason you always receive the same ordering is due to a known bug https://github.com/PelicanPlatform/pelican/issues/1538 where the server IO load are not available among cache servers (bummer!)

director/sort.go
```diff
@@ -64,18 +43,33 @@ type (
...
+
+// Constants for the director sorting algorithm
+const (
+	souceServerAdsLimit      = 6    // Number of servers under consideration
+	distanceHalvingThreshold = 10   // Threshold where the distance havling factor kicks in, in miles
+	distanceHalvingFactor    = 200  // Halving distance for the GeoIP weight, in miles
+	objAvailabilityFactor    = 2    // Multiplier for knowing whether an object is present
+	loadHalvingThreshold     = 10.0 // Threshold where the load havling factor kicks in
+	loadHalvingFactor        = 4.0  // Halving interval for load
+	candidateLimit           = 3    // Number of servers return
```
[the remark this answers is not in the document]
haoming29: This is a good question for @bbockelm

director/sort_algorithms_test.go
haoming29: Alright, I will take `max(len(sm), maxOut)` as the final `maxOut` value to be safe.

director/sort_algorithms.go
haoming29: See reply above

docs/parameters.yaml
haoming29: Naming is always hard! I like "adaptive" and really open to any suggestions

turetske (commented):
Solely looking at the changes since your last review, it looks good. I'll approve once we figure out the issue with the build so that the tests start passing.

turetske (approved):
LGTM

jhiemstrawisc (approved):

PR #1477, opened 2024-07-11
Fix various issues in `stat`

### LINKED ISSUE #1474: Follow up `stat` issues

haoming29:
Fixes #1474 

This PR fixes two issues related to `stat`:
- If `stat` uses AuthURL instead of URL due to Topology server authentication, there is a mismatch when we compare the `stat` result and available serverAds and generate the final server response, thus a successful stat returns 0 matched servers and a 404
- Shadowed variable name in the `stat` goroutine that gave back wrong URL in both log and as the `issuer` url
- The default `300ms` timeout is too short for the Origins distant from the director, and this PR raised default value to 1000ms 

With this patch in, we should be able to fix most of the 404s the director returned in the production

### DIFF
--- a/director/stat_test.go
+++ b/director/stat_test.go
@@ -25,6 +25,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


director/stat_test.go
```diff
@@ -89,24 +91,32 @@ func TestQueryServersForObject(t *testing.T) {
 			Type:    server_structs.OriginType}
 		mockServerAdPrivateTopology := server_structs.ServerAd{
 			Name:         "originPrivateTopology",
-			URL:          url.URL{Host: "mock-origin-5.com", Scheme: "https"},
-			AuthURL:      url.URL{Host: "mock-origin-5-auth.com:8444", Scheme: "https"},
+			URL:          url.URL{Host: "mock-private-topo-origin.com", Scheme: "https"},
```
jhiemstrawisc, line 94: This is probably a minor-enough detail that it shouldn't effect anything, but the `URL` field from Topology is always `http`, while the `AuthURL` field should always be `https`.

jhiemstrawisc (changes requested):

haoming29:
@jhiemstrawisc  Things get a little tricker as I made too much changes around the director logic. For this PR, we want to release it as a patch fix to `7.9`, so I'll merge this in first, and I hope there's minimum merge conflicts

However, the issue fixed in this PR are not in #1308 and #1457, and #1308 is merged. So I'll address the issue in #1308 before I hit the merge button AND have a separate PR to fix the issue in #1457 (which does not need to go into 7.9 as we don't support `stat` for caches in 7.9)

Hopefully this all make sense and let me know if there's any questions!

haoming29:
@jhiemstrawisc This PR is good for another round of review and should be merged in first than the rest of my pending PRs. I've attached a brief description of the dependencies between my PRs and hopefully that makes sense.

jhiemstrawisc (approved):
The logic is tricky, but we're gradually finding/fixing the edge cases w.r.t. topology integration. LGTM!

PR #1487, opened 2024-07-16
Make server startup timeout configurable

### LINKED ISSUE #1486: Configurable startup timeout
Ok, this one pinches --

We use the `server_utils.WaitUntilWorking` method to stall the startup of the origin until OA4MP is fully ready.

*Unfortunately*, OA4MP is a big honking Java application and some of the Tiger hosts are too slow to startup a big honking Java application within the hardcoded 10s timeout.

We should make the timeout a configurable value -- I think the current default is probably fine.

bbockelm:
Fixes #1486


turetske (approved):
LGTM!

Issue #1483, opened 2024-07-15
Collect client version number metric


patrickbrophy:
We should collect Pelican client versions and track the amount of requests from different versions within the director.

haoming29:
@patrickbrophy Same thing for this issue. Could you also link this issue with the PR? For tracking purposes

PR #1489, opened 2024-07-17
Implemented client version metric

patrickbrophy:
This PR addresses issue #1483. This PR adds a new middleware to collect and parse the user-agent from client requests and increment a counter metric with the label being the client version.

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -1166,6 +1166,7 @@ func collectClientVersionMetric(c *gin.Context) {
 
 	reqVer, err := version.NewVersion(reqVerStr)
 	if err != nil {
+		log.Error("Failed to parse client version")
 		return
 	}
 	versionSegments := reqVer.Segments()
--- a/director/director.go
+++ b/director/director.go
@@ -214,15 +214,7 @@ func extractVersionAndService(ginCtx *gin.Context) (reqVer *version.Version, ser
 	return reqVer, service, nil
 }
 
-func versionCompatCheck(ginCtx *gin.Context) error {
-	reqVer, service, err := extractVersionAndService(ginCtx)
-	if err != nil {
-		return err
-	}
-	if reqVer == nil {
-		return nil
-	}
-
+func versionCompatCheck(reqVer *version.Version, service string) error {
 	var minCompatVer *version.Version
 	switch service {
 	case "client":
@@ -231,8 +223,14 @@ func versionCompatCheck(ginCtx *gin.Context) error {
 		minCompatVer = minOriginVersion
 	case "cache":
 		minCompatVer = minCacheVersion
+	case "": // service not provided, ie: using curl
+		return nil
 	default:
 		return errors.Errorf("Invalid version format. The director does not support your %s version (%s).", service, reqVer.String())
+	}
+
+	if reqVer == nil { // version not provided, ie: using curl
+		return nil
 	}
 
 	if reqVer.LessThan(minCompatVer) {
@@ -243,7 +241,8 @@ func versionCompatCheck(ginCtx *gin.Context) error {
 }
 
 func redirectToCache(ginCtx *gin.Context) {
-	err := versionCompatCheck(ginCtx)
+	reqVer, service, _ := extractVersionAndService(ginCtx)
+	err := versionCompatCheck(reqVer, service)
 	if err != nil {
 		log.Warningf("A version incompatibility was encountered while redirecting to a cache and no response was served: %v", err)
 		ginCtx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
@@ -253,7 +252,6 @@ func redirectToCache(ginCtx *gin.Context) {
 		return
 	}
 
-	reqVer, service, _ := extractVersionAndService(ginCtx)
 	collectClientVersionMetric(reqVer, service)
 
 	reqPath := path.Clean("/" + ginCtx.Request.URL.Path)
@@ -469,7 +467,8 @@ func redirectToCache(ginCtx *gin.Context) {
 }
 
 func redirectToOrigin(ginCtx *gin.Context) {
-	err := versionCompatCheck(ginCtx)
+	reqVer, service, _ := extractVersionAndService(ginCtx)
+	err := versionCompatCheck(reqVer, service)
 	if err != nil {
 		log.Warningf("A version incompatibility was encountered while redirecting to an origin and no response was served: %v", err)
 		ginCtx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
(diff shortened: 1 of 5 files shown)


director/director.go
```diff
@@ -1156,15 +1156,33 @@ func getHealthTestFile(ctx *gin.Context) {
 	}
 }
 
+func collectClientVersionMiddleware(c *gin.Context) {
```
jhiemstrawisc, line 1159: Why add this to a middleware instead of collecting the metric directly in `redirectToCache` and `redirectToOrigin`? My concern is that by the `ShortcutMiddleware` route we have set up in `launchers/director_serve.go` is what gets hit in the majority of client requests (but not cache requests), and it looks like that'll bypass this middleware.

director/director.go
```diff
@@ -1156,15 +1156,33 @@ func getHealthTestFile(ctx *gin.Context) {
...
+	userAgentSlc := c.Request.Header["User-Agent"]
+	if len(userAgentSlc) < 1 {
+		log.Error("Failed to parse User-Agent")
+	}
+
+	userAgent := userAgentSlc[0]
+
+	uaRegExp := regexp.MustCompile(`^pelican-[^\/]+\/\d+\.\d+\.\d+`)
+	if matches := uaRegExp.MatchString(userAgent); !matches {
+		log.Error("Failed to match User-Agent for Pelican client version")
```
jhiemstrawisc, line 1169: Since we use this same regex in other parts of the Director, let's pull it out and functionalize it to guarantee that we can't update one but forget the other.

A second note - I don't know if this really constitutes an "error" because we often use non-pelican clients to interact with the federation as well, such as `curl`. After reading the Cloudflare article you sent, I think we want to avoid creating any prom labels based on an arbitrary incoming UA header, but we should think about how to make this flexible enough to handle the case where the incoming UA isn't Pelican. Maybe it's enough to have a single label that tracks non-pelican clients?

Lastly, w.r.t. cardinality explosion and Pelican headers, do you have a sense of the "cost" of tracking the 56 versions we've already released, knowing that we release between 3-5 new versions every month?

metrics/director.go
```diff
@@ -81,4 +81,9 @@ var (
 		Name: "pelican_director_stat_total",
 		Help: "The total stat queries the director issues. The status can be Succeeded, Cancelled, Timeout, Forbidden, or UnknownErr",
 	}, []string{"server_name", "server_url", "server_type", "result"}) // status: see enums for DirectorStatResult
+
+	PelicanDirectorClientVersionTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "pelican_director_client_version_total",
+		Help: "The total amount of requests from client versions.",
```
jhiemstrawisc, line 87: Let's say "number" instead of "amount" -- just like "fewer" vs "less", amount describes continuous quantities.

utils/utils.go
```diff
@@ -143,3 +144,15 @@ func ExtractAndMaskIP(ipStr string) (maskedIP string, ok bool) {
 		return ApplyIPMask(ipStr)
 	}
 }
+
+func ExtractVersionAndServiceFromUserAgent(userAgent string) (reqVer, service string) {
```
jhiemstrawisc, line 148: Can you add a top-level comment to the function that describes its intended use along with the contents of the outputs (i.e. what reqVer and service look like when they're returned)? Also, can you add a unit test with any edge cases you can think of?

director/director.go
```diff
@@ -1156,6 +1156,33 @@ func getHealthTestFile(ctx *gin.Context) {
 	}
 }
 
+func collectClientVersionMetric(c *gin.Context) {
```
jhiemstrawisc, line 1159: Can you add a top-level comment describing how the function works and what it's supposed to do in the case of non-pelican clients?

utils/utils.go
```diff
@@ -143,3 +144,15 @@ func ExtractAndMaskIP(ipStr string) (maskedIP string, ok bool) {
...
+	uaRegExp := regexp.MustCompile(`^pelican-[^\/]+\/\d+\.\d+\.\d+`)
+	if matches := uaRegExp.MatchString(userAgent); !matches {
+		return "", ""
```
jhiemstrawisc, line 151: Maybe we can create a standard "non-pelican" client version for recording?

jhiemstrawisc (changes requested):

jhiemstrawisc (changes requested):
This is getting very close. My last cleanup requests are:

1. Malicious users can still pass mostly-arbitrary user agent headers to Prometheus, potentially leading to cardinality issues. For example, for my local director I can do `curl -v -k -A "pelican-cache/77.123123123.345345345" https://`hostname`:8445/some/object`, and the metric is recorded. Let's make the assumption that the Director will only record versions "other", `[7.0.0, director's version]` and maybe something like "pelican-future" to indicate that the request UA header has a version newer than the director (without the patch version, as you've already done). You can get the Director's version from `config.GetVersion()`, but you'll have to create `version.Version` object to perform comparisons. See `director.versionCompatCheck` for an example of version comparison. I actually think the `pelican-future` version metric is interesting, because it can quickly tell a federation administrator when the majority of client requests are coming from newer versions of pelican -- a good sign that it's time to update your central services! One thing to note is that the only reason I can't get through with a user agent like `pelican-foobar/7.0.0` is that the `versionCompatCheck` errors because foobar is an unrecognized service. If we ever collect this metric in a Director route that doesn't go through `redirectToCache/Origin`, however, this may also be recorded.

2. Right now, both `redirectToCache` and `redirectToOrigin` do a `versionCompatCheck` and then a `collectClientVersionMetric`, but a lot of the guts in these two functions are shared (and can be deduplicated). My proposal: instead of passing the gin context to both the version check and the metric collection, call `ExtractVersionAndServiceFromUserAgent` directly on the gin context and get the version/service. Then pass these values to the version check and metric collection functions. If you agree with point 3, you'll pass both the version and the service to the metric function, and you can validate the input there.

3. This is one is something to think about -- when we record the version metric, should we also record the service as a metric label? For example, what if I want to know "of the Director's traffic, how much is coming from user clients, and how much is coming from caches, and how much is coming from origins?" If we think this is something to record (I lean in favor of yes, even if it bumps the cardinality a bit), I'm in favor of recording this in a separate label in the version metric, so that we can get answers to those types of questions through direct promql queries, instead of having to post process the results.

haoming29:
I'll leave this to @jhiemstrawisc for the final round of review as he has more context to the PR than me at this point, but I hope the changes should be minor

jhiemstrawisc (approved):
LGTM. Thanks for all the work on this, @patrickbrophy!

PR #1490, opened 2024-07-17
Add X-Pelican-[Authorization/Token-Generation] headers to origin redirects

### LINKED ISSUE #1488: X-Pelican-Token-Generation header missing on PUT
When responding to a PUT request (writing an object to the federation), the director does not respond with the `X-Pelican-Token-Generation` header.

Accordingly, clients do not have the information needed to generate a new token and always return a failure instead of performing an OAuth2 flow.

The director should return the header on PUTs in the same way it does GETs.

Worth backporting to 7.9.x in order to simplify the PEARC demo.

jhiemstrawisc:
Brian noticed that PUTS using the Pelican client couldn't get the required information for token generation when an origin was configured with its own OA4MP issuer.

This PR adds the needed headers to origin redirects, and tests that both object/origin redirects contain the expected headers.

Closes #1488

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -184,6 +184,44 @@ func getRequestParameters(req *http.Request) (requestParams url.Values) {
 	return
 }
 
+// Generates the X-Pelican-Authorization header (when applicable) for responses that have
+// issued a request where token generation may be needed. This header informs the client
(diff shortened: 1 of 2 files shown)


turetske (approved):
LGTM!

PR #1504, opened 2024-07-23
Update S3 backend version to v0.1.4 to fix trailing / bug in URLs

jhiemstrawisc:
@turetske, I'll cut a `v7.9` release when this is passing tests and gets merged. It fixes a bug in the S3 plugin that mishandles S3 service URLs with a trailing slash. Contents of the bugfix can be found [here](https://github.com/PelicanPlatform/xrootd-s3-http/pull/37)


jhiemstrawisc:
Bypassing branch protections to merge this since tests are passing and this is an internal container version upgrade.

PR #1505, opened 2024-07-24
Ensure unpack directory name is correct

bbockelm:
The destination file name was being overwritten by `parseDestination` when `pack=auto` was being set.

Move the destination file name logic to the `parseDestination` helper function, correct it, and make sure there's unit test coverage for the name generation.


turetske (approved):
LGTM!

I'll abandon my own fix for this.

PR #1513, opened 2024-07-25
Rewrite TCP connection reset / broken pipe messages

bbockelm:
Goal is to improve the user understanding of the error messages:

```
the existing TCP connection was broken (potentially caused by server restart or NAT/firewall issue)
```

Also marked these are retryable.

No unit tests here -- it's not clear how to inject TCP-level problems into the testing.


turetske (approved):
LGTM!

PR #1514, opened 2024-07-25
Retry metadata lookup

bbockelm:
Given consistent (but infrequent) metadata lookup errors on NRP, this retries the metadata lookup after a few seconds.

This will still trigger a retry on metadata failures but hopefully will save a few stageouts.

Includes unit test.


turetske (approved):
LGTM!

Issue #1492, opened 2024-07-17
Collect the number of pelican and non-pelican servers in a Prometheus metric


patrickbrophy:
Director should collect the number of servers, delineated by pelican/non-pelican servers, origin/cache servers, and the existing server labels (url, web url, etc). This should be part of the periodically updated Go metric read from the `serverAds` TTL cache.

haoming29:
@patrickbrophy Can you link the issue with the PR?

PR #1515, opened 2024-07-25
Server count metric

patrickbrophy:
This PR address issue #1492. This PR implements a new metric in the director to count the number of pelican an non-pelican servers.

### DIFF
--- a/director/director_api.go
+++ b/director/director_api.go
@@ -21,6 +21,7 @@ package director
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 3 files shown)


jhiemstrawisc (approved):
LGTM

jhiemstrawisc:
FWIW, the failing test is one we've identified as being generally flaky. We could sit and re-run it until it passes, but I've seen this enough that I'm confident it's not an issue with the PR.

PR #1516, opened 2024-07-25
Changed user facing references to directories to collections on origins

### LINKED ISSUE #1499: Scrub references to "directory"
Before people start scripting against our machine readable outputs, I'd like to scrub out a few references to "directories" given our emphasis on "Pelican is not a filesystem".

Let's use the phrasing "collection" instead.

So,

```
$ pelican object stat pelican://osdf-itb.osg-htc.org/pearc24-tutorial/home/bockelman@wisc.edu/
Name: bockelman@wisc.edu
Size: 0
ModTime: 2024-07-22 13:37:09 +0000 GMT
IsDir: true
```

should be:

```
$ pelican object stat pelican://osdf-itb.osg-htc.org/pearc24-tutorial/home/bockelman@wisc.edu/
Name: bockelman@wisc.edu
Size: 0
ModTime: 2024-07-22 13:37:09 +0000 GMT
IsCollection: true
```

(please review the error strings, help strings in the client, and the outputs of both `stat` and `ls`)

turetske:
Removed references to "directories" on Pelican origins in user facing code.

I didn't remove anything named within the code because I was concerned about backwards compatibility.

closes #1499


jhiemstrawisc (changes requested):
I think there are still quite a few "directories" in here. For example, `docs/parameters.yaml:796` and `cmd/object_ls.go:89` are both cases where we mean "collections". The caveat is that not every occurrence of "directories" is bad, because in some places we're actually referring to directories on the POSIX filesystem hosting the origin.

If you don't want to clean these extra spots up in this PR, we can try to tackle them in another pass and leave the issue open for now.

jhiemstrawisc (approved):
Pre approving so you can merge when you think it's ready :)

PR #1517, opened 2024-07-25
Client cleanup

### LINKED ISSUE #1497: Client long listing command line flag issues
Noticed this during the tutorial run-through:

```
$ pelican object ls -l pelican://osdf-itb.osg-htc.org/pearc24-tutorial/home/bockelman@wisc.edu/100MB.dat
ERRO[0000] Failed to access/create specified directory. Error: mkdir pelican:/osdf-itb.osg-htc.org: permission denied 
$ pelican object ls -L pelican://osdf-itb.osg-htc.org/pearc24-tutorial/home/bockelman@wisc.edu/100MB.dat
100MB.dat          104857600          2024-07-22 13:37:12
```

Given we are using `ls` for the sub-command, using `-L` for the long listing is pretty horrible (as everyone has muscle memory for `ls -l`).  Let's change it before people start scripting this up.

### LINKED ISSUE #1498: Long output of client should be "absolute"
Given there's no such thing as a directory in Pelican, I find the current output of the `ls -L --json` confusing:

```
$ pelican object ls -L pelican://osdf-itb.osg-htc.org/pearc24-tutorial/home/bockelman@wisc.edu/100MB.dat --json
[{"Name":"100MB.dat","Size":104857600,"ModTime":"2024-07-22T13:37:12Z","IsDir":false}]
```

To me, we should be giving the full object name:

```
[{"Name":"/pearc24-tutorial/home/bockelman@wisc.edu/100MB.dat","Size":104857600,"ModTime":"2024-07-22T13:37:12Z","IsDir":false}]
```

For the "human readable" output, I'm OK with giving the shortened name:

```
$ pelican object ls -L pelican://osdf-itb.osg-htc.org/pearc24-tutorial/home/bockelman@wisc.edu/100MB.dat
100MB.dat          104857600          2024-07-22 13:37:12
```

Interested in hearing the thoughts of others on the topic.

turetske:
This PR addresses two issues, both involving cleaning up client interaction and responses.

The first changes "ls -L" to "ls -l" This also changed the log flag to "-L" to avoid conflicts.

The second changes the output of "ls -l" and "stat" to return the full object name with the namespace rather than just the top level name.

closes #1498 
closes #1497

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -29,7 +29,6 @@ import (
 	"net/url"
 	"os"
 	"path"
-	"path/filepath"
 	"reflect"
 	"regexp"
 	"slices"
@@ -2683,9 +2682,10 @@ func listHttp(ctx context.Context, remoteObjectUrl *url.URL, directorUrl string,
 	}
 
 	for _, info := range infos {
+		jPath, _ := url.JoinPath(remotePath, info.Name())
 		// Create a FileInfo for the file and append it to the slice
 		file := FileInfo{
-			Name:    filepath.Join(remotePath, info.Name()),
+			Name:    jPath,
 			Size:    info.Size(),
 			ModTime: info.ModTime(),
 			IsDir:   info.IsDir(),
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -2665,7 +2665,7 @@ func listHttp(ctx context.Context, remoteObjectUrl *url.URL, directorUrl string,
 			if !info.IsDir() {
 				// NOTE: we implement our own FileInfo here because the one we get back from stat() does not have a .name field for some reason
 				file := FileInfo{
-					Name:    path.Base(remotePath),
+					Name:    remotePath,
 					Size:    info.Size(),
 					ModTime: info.ModTime(),
 					IsDir:   false,
@@ -2762,7 +2762,7 @@ func statHttp(ctx context.Context, dest *url.URL, namespace namespaces.Namespace
 				fsinfo, err := client.Stat(endpoint.Path)
 				if err == nil {
 					info = FileInfo{
-						Name:    path.Base(endpoint.Path),
+						Name:    endpoint.Path,
 						Size:    fsinfo.Size(),
 						IsDir:   fsinfo.IsDir(),
 						ModTime: fsinfo.ModTime(),
@@ -2801,7 +2801,7 @@ func statHttp(ctx context.Context, dest *url.URL, namespace namespaces.Namespace
 			}
 
 			resultsChan <- statResults{FileInfo{
-				Name:    path.Base(endpoint.Path),
+				Name:    endpoint.Path,
 				Size:    info.Size,
 				IsDir:   info.IsDir,
 				ModTime: info.ModTime,
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -29,6 +29,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"slices"
@@ -2684,7 +2685,7 @@ func listHttp(ctx context.Context, remoteObjectUrl *url.URL, directorUrl string,
 	for _, info := range infos {
 		// Create a FileInfo for the file and append it to the slice
 		file := FileInfo{
-			Name:    info.Name(),
+			Name:    filepath.Join(remotePath, info.Name()),
 			Size:    info.Size(),
 			ModTime: info.ModTime(),
 			IsDir:   info.IsDir(),
--- a/client/fed_test.go
+++ b/client/fed_test.go
@@ -488,7 +488,7 @@ func TestObjectStat(t *testing.T) {
 				require.NoError(t, err)
 			}
 			assert.Equal(t, int64(13), got.Size)
-			assert.Equal(t, "hello_world.txt", got.Name)
+			assert.Equal(t, fmt.Sprintf("%s/hello_world.txt", export.FederationPrefix), got.Name)
 		}
 	})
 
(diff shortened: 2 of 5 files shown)


jhiemstrawisc (changes requested):
A few questions/requests for change:
1. Can you add some unit test coverage for the `ls -l` features to prevent a future regression?

2. This is a pretty hard backwards compat break. @bbockelm was it your intention to change the UI over a single release, or do you want to add something akin to a deprecation notice and do this over 2 releases like we've been trying to do elsewhere? If the latter, we can keep this PR open with a plan to cleanup/merge into 7.12, and then merge a new warning into 7.10 in the meantime. I don't know how many people currently use `object ls -L`, but I _have_ seen people using `-l` to set a log location.

3. I'm confused about the change from base names to full path names when I `ls` a full object name, but not when I `ls` an object prefix. For example, `pelican object ls -l pelican://osg-htc.org/chtc/PUBLIC/eturetsky` spits out:
```
ncar-subset          0             2024-05-22 19:05:48
titanic.csv             60301     2024-05-21 14:56:12
data                      0             2024-05-21 14:56:51
hrrrzarr                0               2024-03-20 15:50:39
```
But `pelican object ls -l pelican://osg-htc.org/chtc/PUBLIC/eturetsky/titanic.csv` spits out the full object name:
```
/chtc/PUBLIC/eturetsky/titanic.csv          60301          2024-05-21 14:56:12
```

I think we should produce the full object prefixes in the case that we're not calling ls on the full object name, such that the first query also produces:
```
/chtc/PUBLIC/eturetsky/ncar-subset          0             2024-05-22 19:05:48
/chtc/PUBLIC/eturetsky/titanic.csv             60301     2024-05-21 14:56:12
/chtc/PUBLIC/eturetsky/data                      0             2024-05-21 14:56:51
/chtc/PUBLIC/eturetsky/hrrrzarr                0               2024-03-20 15:50:39
```

turetske:
> A few questions/requests for change:
> 
> 1. Can you add some unit test coverage for the `ls -l` features to prevent a future regression?
> 2. This is a pretty hard backwards compat break. @bbockelm was it your intention to change the UI over a single release, or do you want to add something akin to a deprecation notice and do this over 2 releases like we've been trying to do elsewhere? If the latter, we can keep this PR open with a plan to cleanup/merge into 7.12, and then merge a new warning into 7.10 in the meantime. I don't know how many people currently use `object ls -L`, but I _have_ seen people using `-l` to set a log location.
> 3. I'm confused about the change from base names to full path names when I `ls` a full object name, but not when I `ls` an object prefix. For example, `pelican object ls -l pelican://osg-htc.org/chtc/PUBLIC/eturetsky` spits out:
> 
> ```
> ncar-subset          0             2024-05-22 19:05:48
> titanic.csv             60301     2024-05-21 14:56:12
> data                      0             2024-05-21 14:56:51
> hrrrzarr                0               2024-03-20 15:50:39
> ```
> 
> But `pelican object ls -l pelican://osg-htc.org/chtc/PUBLIC/eturetsky/titanic.csv` spits out the full object name:
> 
> ```
> /chtc/PUBLIC/eturetsky/titanic.csv          60301          2024-05-21 14:56:12
> ```
> 
> I think we should produce the full object prefixes in the case that we're not calling ls on the full object name, such that the first query also produces:
> 
> ```
> /chtc/PUBLIC/eturetsky/ncar-subset          0             2024-05-22 19:05:48
> /chtc/PUBLIC/eturetsky/titanic.csv             60301     2024-05-21 14:56:12
> /chtc/PUBLIC/eturetsky/data                      0             2024-05-21 14:56:51
> /chtc/PUBLIC/eturetsky/hrrrzarr                0               2024-03-20 15:50:39
> ```

The second one is on me. I think I'll fix that.

client/handle_http.go
```diff
@@ -2661,7 +2662,7 @@ func listHttp(ctx context.Context, remoteObjectUrl *url.URL, directorUrl string,
 	for _, info := range infos {
 		// Create a FileInfo for the file and append it to the slice
 		file := FileInfo{
-			Name:    info.Name(),
+			Name:    filepath.Join(remotePath, info.Name()),
```
jhiemstrawisc, line 2665: This is one of the few cases where I don't think we want to use `filepath.Join` -- Federation namespaces are OS independent because they always look like POSIX (i.e. they have the `/` delimiter), so this would produce an invalid path in a client run on Windows. If I missed this in my previous review, we should go back and fix it.

jhiemstrawisc (changes requested):
One change request, otherwise I think we're looking good here.

jhiemstrawisc:
Absent other responses, I think we should make this a warning for the next release cycle and then make the hard switch in `v7.11.0`. For the time being we can probably keep this PR open and merge it to main after we cut `v7.10.0`. Then in a separate PR for 7.10, make a big loud warning for these flags that informs users of the upcoming change in 7.11.

jhiemstrawisc (changes requested):
This doesn't appear to generate the intended warning in the terminal or the log file it produces:
```
[root@618916e9b865 pelican]# pelican object get -l log.txt pelican://osg-htc.org/chtc/PUBLIC/matyas/robots.txt .
Logging.LogLocation is set to log.txt. All logs are redirected to the log file.
[root@618916e9b865 pelican]# cat log.txt 
[root@618916e9b865 pelican]# 
```

I think I've run into a similar issue before, where this check is happening before the flags have actually been parsed. IIRC, the fix is to wrap the root command with a `PersistentPreRun` function, i.e. somewhere at the end of root.go's `init()`, you should be able to do:
```
    rootCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
        if cmd.Flags().Changed("log") {
            log.Warningln("The '-l' for logging will be changed to '-L' in the 7.11.0 pelican release")
        }
    }
```
The other change here is switching from `fmt.Println` to `log.Warningln` and removing the word "WARNING" from the message -- this standardizes the format with the other warnings we throw, and gives everything nice color coding:
```
time="2024-08-09T20:12:20Z" level=warning msg="The '-l' flag for logging will be changed to '-L' in the 7.11.0 pelican release"
```

Can you add a similar warning to the `-L` flag for `pelican object ls -L`?

jhiemstrawisc (approved):
LGTM. Thanks!

PR #1518, opened 2024-07-25
Change copy command into get in the example to avoid confusion

Kristy-an:


docs/pages/install.mdx
```diff
@@ -103,7 +103,7 @@ The server components require `xrootd-server >= 5.6.3` because of some of the pa
 2. Test Pelican client function by running an **object copy** command that downloads a test file from [Open Science Data Federation (OSDF)](https://osg-htc.org/services/osdf) to your current directory
```
jhiemstrawisc, line 103: Can you change this line to contain `**object get**` as well, and make sure any other references to `object copy` in this page are converted to their corresponding get/put?

jhiemstrawisc (changes requested):

docs/pages/install.mdx
Kristy-an: This part should be the only one using copy. I have changed it.

jhiemstrawisc (approved):
Looks good to me. Thanks for the updates!

PR #1522, opened 2024-07-30
Include Grafana template for Pelican director dashboard

haoming29:
This adds a Grafana template file for visualizing various Pelican Director and federation metrics


haoming29:
Close this PR as the said dashboard was mostly for internal use. Will move this to Pelican Google drive instead.

PR #1523, opened 2024-07-30
Fix server URL mismatch `redirectToCaches` if the object is from topology and is protected

haoming29:
This is a follow up PR to #1308 that fixes the mismatch in URL if we used AuthURL for `stat` for the protected object in the topology.


jhiemstrawisc (approved):
LGTM.

PR #1541, opened 2024-08-06
Fixed error where queries were being added to filename on get/copy

### LINKED ISSUE #1398: Implicit naming of fetched object includes query statement.
Pelican Version:
Version: 7.8.6
Build Date: 2024-06-06T21:27:21Z
Build Commit: f911e9940cc4ba00827475dbd4a33545bd927554
Built By: goreleaser

Pelican Service:
- [x] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

**Describe the bug**
When using `pelican object get` and providing a destination but not an explicit name, along with a query statement such as `?directread`, the name of the fetched object will include the query statement.

**To Reproduce**
1. Download the client and unpack it.
2. Use the client to get an object and pass a query, where the destination is a path but not a file name. 
3. Use `ls` to see the name of the file obtained.

For example,

```
$ wget https://github.com/PelicanPlatform/pelican/releases/download/v7.8.6/pelican_Linux_x86_64.tar.gz
$ tar -xzf pelican_Linux_x86_64.tar.gz
$ ls
 LICENSE   README.md   pelican   pelican_Linux_x86_64.tar.gz
$ ./pelican -f osg-htc.org object get /ospool/uc-shared/public/OSG-Staff/validation/test.txt?directread ./
$ ls
 LICENSE   README.md   pelican   pelican_Linux_x86_64.tar.gz  'test.txt?directread'
```

**Expected behavior**
Expect the file name to be the same as when a query statement is not used.
For example,

```
$ ./pelican -f osg-htc.org object get /ospool/uc-shared/public/OSG-Staff/validation/test.txt ./
$ ls
 LICENSE   README.md   pelican   pelican_Linux_x86_64.tar.gz  test.txt
```

**Screenshots**
N/A

 - OS: `PRETTY_NAME="CentOS Stream 9"`
 - Browser: N/A

**Additional context**
The query is (correctly) not included in the file name if the file name is explicitly passed in the destination argument.
For example, 

```
$ ./pelican -f osg-htc.org object get /ospool/uc-shared/public/OSG-Staff/validation/test.txt?directread ./test.txt
$ ls
 LICENSE   README.md   pelican   pelican_Linux_x86_64.tar.gz  test.txt
```

turetske:
closes #1398

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -149,6 +149,11 @@ func UrlWithFederation(remoteUrl string) (string, error) {
 			newErr := errors.New(fmt.Sprintf("error parsing discovery url: %s", err))
 			return remoteUrl, newErr
 		}
+		if parsedDiscUrl.Scheme == "" {
+			parsedDiscUrl.Scheme = "https"
+			updatedDiscString := parsedDiscUrl.String()
+			parsedDiscUrl, _ = url.Parse(updatedDiscString)
+		}
 		if parsedDiscUrl.Path != "" {
 			newErr := errors.New(fmt.Sprintf("provided federation url %s has a path component", param.Federation_DiscoveryUrl.GetString()))
 			return remoteUrl, newErr
--- a/utils/client_test.go
+++ b/utils/client_test.go
@@ -254,6 +254,14 @@ func TestUrlWithFederation(t *testing.T) {
 		assert.Equal(t, pelUrl, str)
 	})
 
+	t.Run("testFederationNoHost", func(t *testing.T) {
+		viper.Set(param.Federation_DiscoveryUrl.GetName(), "somefederation.org")
+		namespaceOnly := "/namespace/test.txt"
+		str, err := UrlWithFederation(namespaceOnly)
+		assert.NoError(t, err)
+		assert.Equal(t, pelUrl, str)
+	})
+
 	t.Run("testFederationWithFedHost", func(t *testing.T) {
 		viper.Set(param.Federation_DiscoveryUrl.GetName(), "https://somefederation.org")
 		namespaceOnly := "/namespace/test.txt"
@@ -264,6 +272,14 @@ func TestUrlWithFederation(t *testing.T) {
 
 	t.Run("testFederationWithPathComponent", func(t *testing.T) {
 		viper.Set(param.Federation_DiscoveryUrl.GetName(), "somefederation.org/path")
+		namespaceOnly := "/namespace/test.txt"
+		_, err := UrlWithFederation(namespaceOnly)
+		assert.Error(t, err)
+		assert.EqualError(t, err, fmt.Sprintf("provided federation url %s has a path component", param.Federation_DiscoveryUrl.GetString()))
+	})
+
+	t.Run("testFederationPathComponentWithHost", func(t *testing.T) {
+		viper.Set(param.Federation_DiscoveryUrl.GetName(), "https://somefederation.org/path")
 		namespaceOnly := "/namespace/test.txt"
 		_, err := UrlWithFederation(namespaceOnly)
 		assert.Error(t, err)


client/fed_test.go
```diff
@@ -155,6 +155,42 @@ func TestGetAndPutAuth(t *testing.T) {
...
+		for _, export := range fed.Exports {
+			tempPath := tempFile.Name()
+			fileName := filepath.Base(tempPath)
+			uploadURL := fmt.Sprintf("pelican://%s:%s%s/%s/%s", param.Server_Hostname.GetString(), strconv.Itoa(param.Server_WebPort.GetInt()),
+				export.FederationPrefix, "osdf_osdf", fileName)
+
+			// Upload the file with PUT
+			transferResultsUpload, err := client.DoPut(fed.Ctx, tempFile.Name(), uploadURL, false, client.WithTokenLocation(tempToken.Name()))
+			assert.NoError(t, err)
+			if err == nil {
```
jhiemstrawisc, line 176: What's the logic for doing the `assert.NoError` and then checking `if err == nil`? Doesn't the first imply the second?

turetske: Copy paste error, really. It's how it was in the original test, but you are correct.

jhiemstrawisc: While we're in this section of code, can you remove the redundancy and do the same with any of the other tests that do it?

jhiemstrawisc:
For posterity, we also discussed a change in approach here after noticing this issue does not exist with `pelican object get pelican:/// ... ?foo`-type queries because of the URL parsing we do. Rather than strip queries as part of the path, use `pelican -f <fed URL> </path/to/object>` to construct the appropriate Pelican URL. This guarantees we treat the two equivalent requests the same.

utils/utils.go
```diff
@@ -127,3 +130,22 @@ func ExtractVersionAndServiceFromUserAgent(userAgent string) (reqVer, service st
...
+
+func UrlWithFederation(remoteUrl string) (string, error) {
+	if param.Federation_DiscoveryUrl.IsSet() {
+		parsedUrl, err := url.Parse(remoteUrl)
+		if err != nil {
+			newErr := errors.New(fmt.Sprintf("error parsing source url: %s", err))
+			return "", newErr
+		}
+		if parsedUrl.Host != "" {
+			newErr := errors.New("Source URL should not have a host when the Federation_DiscoveryUrl is set")
```
jhiemstrawisc, line 142: If we've configured a `Federation.DirectorUrl` in `pelican.yaml` but we run `pelican object get pelican://<some different federation>`, I think this will cause a failure. We should treat the provided Pelican URL's federation as an override.

jhiemstrawisc (changes requested):
A few comments:
1. The following command should be valid, but results in bad URL parsing:
```
pelican object get -f https://osg-htc.org /chtc/PUBLIC/matyas/robots.txt?skipstat .
```
results in the error:
```
ERROR[2024-08-09T18:55:23Z] Failed to parse source URL: parse "pelican://https:%2F%2Fosg-htc.org/chtc/PUBLIC/matyas/robots.txt?skipstat": invalid port ":%2F%2Fosg-htc.org" after host 
ERROR[2024-08-09T18:55:23Z] Failure getting pelican://https:%2F%2Fosg-htc.org/chtc/PUBLIC/matyas/robots.txt?skipstat: parse "pelican://https:%2F%2Fosg-htc.org/chtc/PUBLIC/matyas/robots.txt?skipstat": invalid port ":%2F%2Fosg-htc.org" after host 
```

I think in your UrlWithFederation function, you can strip out the scheme to prevent it from being added back to the host of the constructed URL

3. Related to 1, we should throw an error and shutdown if the provided federation URL has a path component, e.g. `-f https://osg-htc.org/foo`. Currently this also gets parsed into the resultant URL

2. The following command runs, but nothing is downloaded:
```
[root@618916e9b865 director]# pelican object get pelican://osg-htc.org/chtc/PUBLIC/matyas/robots.txt?skipstat robots.txt
DEBUG[2024-08-09T18:57:55Z] Launch progress bars display                 
DEBUG[2024-08-09T18:57:55Z] Len of source: 2                             
DEBUG[2024-08-09T18:57:55Z] Sources: [pelican://osg-htc.org/chtc/PUBLIC/matyas/robots.txt?skipstat] 
DEBUG[2024-08-09T18:57:55Z] Destination: robots.txt  
```
I think this means something broke with standard Pelican URLs. I think the debug output for "sources" should still be the path component and not the entire Pelican URL.

utils/utils.go
turetske: Ah, that's a good point.

jhiemstrawisc (approved):
LGTM!

Issue #1512, opened 2024-07-25
Gracefully handle GeoIP failures


bbockelm:
When MaxMind doesn't know where an IP is, it returns `37.751, -97.822` and sets the "accuracy" to 1000KM.  Take as an example the IP 67.58.51.123.

However, we treat the lat/long in this case (a location in Kansas) as the gospel truth.

If the accuracy returned is ever greater than 100KM, then we should:
- Ensure a warning is displayed on both the director and service (cache/origin) web interface.
- Display an admin-only widget in the director's web interface that is the list of the 5 most recent GeoIP lookup failures
- Take accuracy into account in our sorting algorithms.  If either location (server or client) is at the Kansas location (or close enough -- something like `abs(lat - 37.751) + abs(long + 97.822) < 0.1`), then randomly select any cache within the US.
- If the location is *not* in Kansas (I don't know if MaxMind has a generic European location, for example), take the top 5 "nearest" caches and select one at random.

haoming29:
You might want to wait until #1470 is merged in because that one changed a lot for how we do the sorting the GeoIP lookup.

PR #1546, opened 2024-08-08
Modify sort algorithm to treat unresolvable IP addresses as random

jhiemstrawisc:
The underlying issue relates to the "Kansas" problem, where valid IP addresses that don't have a known GeoIP resolution are assigned the default Kansas lat/long.

This approach is _slightly_ different when compared to the GH issue, but I think it resolves the issue in a simpler fashion. Rather than incorporate logic that says "if Kansas AND large accuracy radius", this only looks at accuracy radius because a high radius, set to 900 km here, is as good as "I don't know where this IP comes from".

When we see such a radius, we set the lat/long to null (0,0) -- and this happens for both client IPs AND server registrations.

The sort logic was modified to detect any null lat/long and create a random, negative distance weight for these. The result is we sort these to the end of our list. If the client IP is null, everything gets sorted to the end of the list, so we either follow a load-aware sorting only (if sort method is distanceAndLoad/adaptive) or completely random (if sort method is distance). When this happens with a server IP but not the client IP, that server is sorted behind other servers with a valid distance.

Finally, the Director logs errors when either of these situations occur. On startup, a director run in OSDF mode spits out the origins/caches that currently have no GeoIP resolution as a WARNING that mentions the accuracy radius, and does the same for client requests. This is handled differently for IP addresses that are invalid (perhaps because they come from a private IP range), with an error that indicates a true null.

This PARTIALLY addresses issue #1512 -- I'm not the guy to make widgets for the web interface 🤓

### DIFF
--- a/director/cache_ads.go
+++ b/director/cache_ads.go
@@ -249,6 +249,8 @@ func updateLatLong(ad *server_structs.ServerAd) error {
 	if !ok {
 		return errors.New("Failed to create address object from IP")
 	}
+	// NOTE: If GeoIP resolution of this address fails, lat/long are set to 0.0 (the null lat/long)
+	// This causes the server to be sorted to the end of the list whenever the Director requires distance-aware sorting.
 	lat, long, err := getLatLong(addr)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


turetske (approved):

jhiemstrawisc:
I think I understand why the test is failing -- my described logic isn't totally correct because when distance is unknown, we randomly assign a weight and then use a load weight as a multiplier. This makes it more likely that a high-load server is placed behind a low-load server, but it doesn't guarantee a particular ordering.

I'll need to think a bit more about whether this is desired, and if so how to make the test reliable.

turetske (approved):

PR #1553, opened 2024-08-09
Switch container builds to use osg-testing instead of osg repos

### LINKED ISSUE #1552: Revert `osg` repo changes to `osg-testing` to inherit important S3 patch
There's an XRootD patch in `osg-testing` we desperately need for improved performance with S3 origins, and the patch that was breaking builds yesterday has since been fixed. We need to revert the changes I made previously when switching `osg-testing` to `osg`.

jhiemstrawisc:
This gets us a fixed patch to improve S3 origin performance by switching to the `osg-testing` repos for xrootd installation.

Closes #1552

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -48,7 +48,7 @@ RUN useradd -o -u 10940 -g 10940 -s /bin/sh xrootd
 
 # Install dependencies
 RUN yum -y update \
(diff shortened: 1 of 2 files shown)


matyasselmeci (approved):
LGTM

PR #1563, opened 2024-08-16
Selectively enable client X509 authentication

### LINKED ISSUE #1562: Add support for X.509-authenticated origins
In the legacy OSDF, some prefixes (particularly, LIGO-related ones) may utilize X.509 authentication for authorization at the origin.

When X.509-based clients contact the caches, the cache needs to use X.509 authentication to contact the origin -- otherwise, the origin cannot authorize the cache.

In https://github.com/PelicanPlatform/xrdcl-pelican/pull/9, I added some new configuration variables that allow X.509 authentication to be used if specified in a fixed file.  That should at least enable testing.  A follow-up commit will also enable this behavior based on director headers.

This issue is to create the Pelican-side of the infrastructure.  We should:
1.  Maintain a list of these prefixes at the director side.  This should be a "hidden" configuration knob.
2. When the cache requests an object starting with one of these prefixes, set the extra header `X-Osdf-X509: true`.  That will indicate to the cache X.509 auth may be necessary.
3. Set the appropriate environment variables when configuring xrootd for the cache.

bbockelm:
When a requested object matches a pre-configured set of paths, then indicate to the cache it should turn on client X509 authentication.

This fixes support for legacy OSDF namespaces which may require the cache to use client X509 authentication when requesting from the origin because a token is not otherwise available.

Fixes #1562

### DIFF
--- a/config/resources/osdf.yaml
+++ b/config/resources/osdf.yaml
@@ -28,6 +28,6 @@ Registry:
   RequireOriginApproval: true
 Director:
   X509ClientAuthenticationPrefixes:
-  - /xenon/PROTECTED
-  - /user/ligo
-  - /igwn
+  - /xenon/PROTECTED/
+  - /user/ligo/
+  - /igwn/
--- a/config/resources/osdf.yaml
+++ b/config/resources/osdf.yaml
@@ -26,3 +26,8 @@ Federation:
 Registry:
   RequireCacheApproval: true
   RequireOriginApproval: true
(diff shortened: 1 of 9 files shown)


jhiemstrawisc (approved):
I tested for the header and everything looks good. I changed the specified default prefixes for OSDF to include a trailing /, which I think is marginally safer given the way we do matching to construct the `X-Osdf-X509` header.

PR #1565, opened 2024-08-16
Improve director restart behavior

bbockelm:
When the director restarts,
- Detect whether a response is coming from Pelican or a SSL terminator application. In the latter case, retry the request a few times to allow the director to restart.
- If the director has recently restarted, instead of sending a 404, return a 429 (too many requests) indicating the client should retry again soon.

### DIFF
--- a/client/director.go
+++ b/client/director.go
@@ -22,19 +22,29 @@ import (
 	"context"
 	"encoding/json"
 	"io"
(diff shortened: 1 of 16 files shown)


bbockelm:
Closing given this is superseded by @jhiemstrawisc's work in #1890.

Issue #1566, opened 2024-08-20
Split config for enabling `stat` against caches & origins


jhiemstrawisc:
Currently, there's no way to disable stat requests against only origins _or_ caches. They're either both enabled or both disabled through the Director's `Director.EnableStat` config variable.

These two should be split so that an issue with only one doesn't require disabling both.

PR #1575, opened 2024-08-26
Enable caching of object existence queries

bbockelm:
This PR enables the director to keep a local cache of job existence query results.

That is, if it issues a HEAD to see if a remote cache already has an object, then keep around that result for a minute and elide future queries.

This also takes the opportunity to rename and split the stat option, fixing #1566.

### DIFF
--- a/director/stat.go
+++ b/director/stat.go
@@ -28,6 +28,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/jellydator/ttlcache/v3"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
@@ -263,6 +264,21 @@ func WithToken(tk string) queryOption {
 	}
 }
 
+func getStatUtils(ads []server_structs.ServerAd) map[string]*serverStatUtil {
+	statUtilsMutex.RLock()
+	defer statUtilsMutex.RUnlock()
+
+	result := make(map[string]*serverStatUtil, len(ads))
+	for _, ad := range ads {
+		url := ad.URL.String()
+		statUtil, ok := statUtils[url]
+		if ok {
+			result[url] = &statUtil
+		}
+	}
+	return result
+}
+
 // Implementation of querying origins/cache servers for their availability of an object.
 // It blocks until max successful requests has been received, all potential origins/caches responded (or timeout), or cancelContext was closed.
 //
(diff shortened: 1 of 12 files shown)


director/stat.go
```diff
@@ -345,15 +360,24 @@ func (stat *ObjectStat) queryServersForObject(ctx context.Context, objectName st
...
-				if serverAd.FromTopology && (!serverAd.Caps.PublicReads || cfg.protected || cfg.token != "") && serverAd.AuthURL.String() != "" {
-					baseUrl = serverAd.AuthURL
-				}
+
+			baseUrl := serverAd.URL
+			// For the topology server, if the server does not support public read,
+			// or the token is provided, or the object is protected, then it's safe to assume this request goes to authenticated endpoint
+			// For Pelican server, we don't populate authURL and only use server URL as the base URL
+			if serverAd.FromTopology && (!serverAd.Caps.PublicReads || cfg.protected || cfg.token != "") && serverAd.AuthURL.String() != "" {
+				baseUrl = serverAd.AuthURL
```
jhiemstrawisc, line 369: In my testing, I observed that a tokenless stat request for `/chtc/PUBLIC/matyas/robots.txt` wound up being denied at every topology cache because their representative server ads had "Caps.PublicReads = false". This doesn't make sense for caches, and it doesn't make sense to send a tokenless stat requests to a cache's auth URL. I think the issue is in the way we parse topology ads. Unfortunately, the `serverAd` struct doesn't have a way to encode whether the server is a cache or an origin, so we can't directly check that here. One temporary solution I identified is to modify `director/advertise.go::parseServerAdFromTopology:65` to set the PublicReads capability to true. A quick test of that locally bumped the number of caches found to have the object from 1 to 8.

jhiemstrawisc (changes requested):
In addition to the inline comments, I noticed while testing that the default stat timeout (1000ms) seems to be too low. When serving a local OSDF director, every stat request to a cache across the pond (London, Amsterdam, Scotland, Singapore, wherever https://stash.farm.particle.cz:8443 is located, etc) timed out, and the only timeouts were for these caches.

jhiemstrawisc:
@bbockelm, is this ready for a re-review?

bbockelm:
Yes - sorry, please do re-review.

jhiemstrawisc (approved):
LGTM.

jhiemstrawisc:
I'll clean up the merge conflict and merge shortly.

jhiemstrawisc:
@bbockelm, are you able to debug these failing tests any more? I haven't had much progress figuring out where in the stack the timeouts are coming from. While I was trying to figure it out, I reverted these changes to their state before my rebase and noticed that the timeouts seem tied to higher value for `Director.StatTimeout`. The error logs also indicate that failures are related to a timeout awaiting response headers from the Director, so my suspicion is that the increased Director stat timeout is causing a different context to hit its limit while the Director is waiting for stat results. I have been unable to find where that happens.

jhiemstrawisc:
One other thing that's not clear to me -- when the ttl cache presence capacity is reached, what happens on the call to `Set()` a result? The jellydator ttlcache docs didn't provide any insight. I think it's important to enforce a cache size limit because the strings being cached come from user queries and we'd otherwise expose the Director to an OOM-style attack that could be triggered by a malicious client asking for many bogus objects. But when that limit is hit, do we pop the oldest result to store the new, or do we just ignore the `Set()`?

jhiemstrawisc:
The current MacOS test failure is due to a flaky test and I don't believe it to be affected by the contents of this PR.

PR #1581, opened 2024-08-29
Fix topology downtime integration by using downtime URL and not ns JSON

### LINKED ISSUE #1258: Explicitly ignore Pelican endpoints in topology
As a follow-up to #1250:

Let's say a cache is in both topology and is a Pelican cache.  When everything is working, the director will see both the Pelican server ad and the topology entry and prefer the information from Pelican (the exception being the downtimes - see #1248).

However, if the service goes offline - say, the host crashes - then there is no more Pelican ad and the topology information takes over.  This is bad: it ensures clients are redirected to the cache even when it is down!

I see two options:
1.  Let's add either a `PelicanService: true` label in topology - causing the OSDF endpoints to ignore the service - or perhaps change the service type from `XRootD cache server` to `Pelican cache` (and analogously for `Pelican origin`).
2. Similarly add a label in topology but change the director code to filter out / ignore such entities.

I like (1) for its simplicity but I worry about the other downstream impacts on things that filter on caches (for example, @CannonLock's maps).

### LINKED ISSUE #1578: Hostnames in `Director.FilteredServers` list are still being used in redirects
Pelican Version:
Pelican v7.10

Pelican Service:
Director

**Describe the bug**
Hostnames listed in `Director.FilteredServers` are still used for redirects. We discovered this after `sc-cache.chtc.wisc.edu` continued showing up in the Director's cache response after it was listed there. We also tried adding the hostname:port combination, e.g. `sc-cache.chtc.wisc.edu:8443`, but the redirects continued.

The WebUI button to put a server in downtime does appear to function, so there's a disconnect happening somewhere here.

jhiemstrawisc:
This changes the overall approach for grabbing downtime information from Topology. Previously, we generated the list of downed servers by getting a diff of the servers returned in the namespaces JSON when querying that endpoint with `?include_downed=<0/1>`. The issue with this approach is that Topology services marked as Pelican services are explicitly excluded from the returned JSON.

However, Pelican services that are recorded in Topology are still included in the downtime URL, so this new approach grabs the list of resource names from that instead. That was probably the right approach all along, but topology endpoints remain inscrutable ¯\_(ツ)_/¯

This also corrects a documentation issue that claimed filtered servers could be added in the Director using their hostname. In fact, they need to be filtered using their site name!

Closes #1578 and partially closes #1258

### DIFF
--- a/director/advertise_test.go
+++ b/director/advertise_test.go
@@ -19,12 +19,15 @@
 package director
 
 import (
+	"bytes"
 	"context"
 	_ "embed"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"testing"
+	"text/template"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	logrustest "github.com/sirupsen/logrus/hooks/test"
@@ -33,7 +36,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/pelicanplatform/pelican/server_structs"
-	"github.com/pelicanplatform/pelican/server_utils"
 )
 
 var (
@@ -98,7 +100,7 @@ func TestConsolidateDupServerAd(t *testing.T) {
 
 func TestParseServerAdFromTopology(t *testing.T) {
 
-	server := server_utils.Server{
+	server := server_structs.TopoServer{
 		Endpoint:     "http://my-endpoint.com",
 		AuthEndpoint: "https://my-auth-endpoint.com",
 		Resource:     "MY_SERVER",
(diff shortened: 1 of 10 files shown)


director/advertise_test.go
```diff
@@ -301,149 +303,85 @@ func TestAdvertiseOSDF(t *testing.T) {
...
+			{
+				// start time is after current time. Should NOT be filtered
+				ResourceName: "DENVER_INTERNET2_OSDF_CACHE",
+				ResourceFQDN: "dtn-pas.denv.nrp.internet2.edu",
+				StartTime:    time.Now().Add(24 * time.Hour).Format("Jan 2, 2006 03:04 PM MST"),
+				EndTime:      time.Now().Add(25 * time.Hour).Format("Jan 2, 2006 03:04 PM MST"),
+			},
+			{
+				// end time is before current time. Should NOT be filtered
+				ResourceName: "HOW_MUCH_CASH_COULD_A_STASHCACHE_STASH",
```
turetske, line 325: 😆

turetske (approved):

PR #1586, opened 2024-09-07
Add "sync" functionality to the client

bbockelm:
This adds a "rsync-like" interface, `object sync`, to the command line client.  Example:

```
pelican object sync pelican://localhost:8444/tmp/test /tmp/download
```

This can be called multiple times; partially-downloaded files will be completed and completed files will be skipped (similar behavior on uploads).

The intent for this functionality is to allow someone to do a transfer of a large number of files into the OSDF from a host that might be rebooted or otherwise interrupted.

This includes multiple small bugfixes discovered in the client while testing, including:
- Fixed client hang for transfers of empty directories.
- Fixed client hang if the transfers finish before the directory walk finishes
- Do not consider a zero-sized file an error.
- If the director notices a namespace requires authorization -- but the client didn't provide a token -- then don't bother performing a stat (as it's just going to fail immediately).  In this case, the client will retry the director lookup once it's acquired a token.

### DIFF
--- a/client/director.go
+++ b/client/director.go
@@ -38,7 +38,7 @@ import (
 
 // Make a request to the director for a given verb/resource; return the
 // HTTP response object only if a 307 is returned.
-func queryDirector(ctx context.Context, verb, sourcePath, directorUrl string) (resp *http.Response, err error) {
+func queryDirector(ctx context.Context, verb, sourcePath, directorUrl string, token string) (resp *http.Response, err error) {
 	resourceUrl := directorUrl + sourcePath
 	// Here we use http.Transport to prevent the client from following the director's
 	// redirect. We use the Location url elsewhere (plus we still need to do the token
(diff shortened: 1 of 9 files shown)


turetske (changes requested):
@bbockelm Should this also work on a single large file rather than just a folder? Because currently it only works for folders and trying to do it on a single file fails poorly (tries to PROPFIND a folder with the file name which then fails). If it's the former, then that needs to be fixed. If it's the latter, then it should fail more gracefully. Maybe with something that says "source needs to be a folder"

bbockelm:
Good catch!  Indeed, it should to the correct thing when you pass it a file name.

That said, what's the current behavior without the patch?  If the "sync" behavior matches the current behavior, I'd consider the behavior a bug and can be fixed in a separate PR.

turetske:
@bbockelm What do you mean by current behavior? A `pelican object get` of a single file works fine.

turetske (approved):

bbockelm:
@turetske - can you redo that last commit with proper git author metadata?  GitHub renders the author as "your name" now.

turetske:
> @turetske - can you redo that last commit with proper git author metadata? GitHub renders the author as "your name" now.

@bbockelm Done

PR #1589, opened 2024-09-17
Add Pelican URL Parsing library to standardize how we handle pelican URLs

jhiemstrawisc:
This new pelican_url package and its overall integration should feel like our beloved, familiar `net/url` package, but with some added niceties for Pelican URL parsing. In particular, it handles
- Query param validation
- Metadata discovery
-  URL caching (to avoid frequent metadata lookup for URLs with the same discovery host)

A few caveats -- this grew more than I anticipated. The reviewer should pay particular attention to tests that I deleted -- in many cases, I moved the essence of the test to what I thought was a more appropriate location (such as within the `pelican_url` module), but there isn't a 1-1 relationship here. If I deleted something I shouldn't have, do let me know.


jhiemstrawisc:
One to-do that came from a discussion with Brian -- go back through this and restore any exported APIs from the client to have the same signature. There were a few places where I changed whether a URL or a string were passed.

turetske:
> One to-do that came from a discussion with Brian -- go back through this and restore any exported APIs from the client to have the same signature. There were a few places where I changed whether a URL or a string were passed.

@jhiemstrawisc Shall I wait until this is done?

jhiemstrawisc:
> > One to-do that came from a discussion with Brian -- go back through this and restore any exported APIs from the client to have the same signature. There were a few places where I changed whether a URL or a string were passed.
> 
> @jhiemstrawisc Shall I wait until this is done?

Probably a good idea for now, I think I'll have to adjust a few things that are non-trivial

pelican_url/query.go
```diff
@@ -0,0 +1,131 @@
...
+
+// Like url.Values.Get, this only gets the first value for a given key
+func (pvs PelicanURLValues) Get(key string) string {
+	if values, ok := pvs[key]; ok {
+		return values[0]
+	}
+	return ""
+}
+
+func (pvs PelicanURLValues) Add(key string, val string) {
```
turetske, line 61: I might have missed this, but where in the code is 'Add' used?

pelican_url/discovery_test.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+}
+
+func TestDiscoverFederation(t *testing.T) {
+	discServer := getTestDiscoveryServer(t)
+	defer discServer.Close()
+	discUrl, err := url.Parse(discServer.URL)
+	require.NoError(t, err)
+
+	t.Run("TestVanilla", func(t *testing.T) {
```
turetske, line 84: The original TestDiscoverFederation also has an invalid URL test. Is it removed because the PelicanUrl validation should ensure that before Discover Federation is run?

turetske (changes requested):
A few changes and clarifications requested.

pelican_url/query.go
jhiemstrawisc: I don't recall if/where it's currently used, but it's something I've used enough with regular URL parsing that I wanted to be forward thinking.

pelican_url/discovery_test.go
jhiemstrawisc: In theory, yeah. The problem is that `DiscoverFederation` is still exported, so I should add a few error checks. Thanks for catching that.

turetske (approved):
LGTM

PR #1591, opened 2024-09-19
Auto-generate token if the issuer is available

bbockelm:
This tweak will cause the client to search for the issuer's private key in its local configuration and auto-generate a token as needed.  (On simple origin deployments, it's not uncommon for the origin to be run by a single individual who has the private key)

This prevents the user (in this specific case) from needing to know how to create tokens to use Pelican: simply drop the private key into `~/.config/pelican/issuer.jwk` and go!

### DIFF
--- a/client/acquire_token.go
+++ b/client/acquire_token.go
@@ -20,6 +20,7 @@ package client
 
 import (
 	"context"
+	"crypto/ecdsa"
 	"encoding/json"
 	"fmt"
 	"io/fs"
@@ -33,6 +34,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/lestrrat-go/jwx/v2/jwk"
 	jwt "github.com/lestrrat-go/jwx/v2/jwt"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


client/acquire_token.go
```diff
@@ -706,3 +735,108 @@ func AcquireToken(destination *url.URL, dirResp server_structs.DirectorResponse,
 
 	return token.AccessToken, nil
 }
+
+// Given a URL and a known public key, determine whether the public key
+// is valid for the issuer URL.
+//
+// If valid, returns the corresponding keyId and sets found to true.
+func findKeyId(url string, ecPubKey *ecdsa.PublicKey) (keyid string, found bool) {
```
jhiemstrawisc, line 743: This seems generically useful enough that it might be better suited as an exported function in one of our utils or token packages. Moving it to utils would try a bit of restructuring because we generally can't import config there due to cyclic dependencies, so I won't push back if you think the function should remain unexported in `client`.

bbockelm: Yes -- I'd like to leave the restructuring until we have a second use case.

Since this is a non-public interface, there won't be any compatibility concerns if we wait until later.

jhiemstrawisc (approved):

Issue #1368, opened 2024-06-03
Remove top-level capabilities in ServerAd from `server_structs.go`


jhiemstrawisc:
This issue is related to work from PR #1365, which started moving Pelican towards using our `Capabilities` struct for serverAds, in recognition that origins are like namespaces in their need to enumerate capabilities. As we span this transition across releases to avoid backwards compatibility issues, we should eventually deprecate and remove the top-level capabilities in `ServerAd`. For example, we should get rid of `ServerAd.Writes` and replace it with `ServerAd.Capabilities.Writes`.

Issue #1367, opened 2024-06-03
Finish removing `namespaceadv2.PublicRead`


jhiemstrawisc:
See PR #1366. This should be targeted for the 7.10 or 7.11 release or thereabouts to make sure this doesn't immediately break backwards compatibility.

PR #1595, opened 2024-09-24
Use capabilities structs in various types of server ads

jhiemstrawisc:
This moves us away from using something like `ad.Writes` in favor of `ads.Caps.Writes` by using our capabilities struct across the board where appropriate.

The work in this PR has been spread out over 3 releases to minimize potential impact to backwards compatibility issues, with the 7.11 release targeted for final breaking changes.

Closes issue #1367 
Closes issue #1368

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -49,10 +49,6 @@ func consolidateDupServerAd(newAd, existingAd server_structs.ServerAd) server_st
 	consolidatedAd.Caps.Writes = existingAd.Caps.Writes || newAd.Caps.Writes
 	consolidatedAd.Caps.Listings = existingAd.Caps.Listings || newAd.Caps.Listings
 
-	consolidatedAd.DirectReads = existingAd.DirectReads || newAd.DirectReads
-	consolidatedAd.Writes = existingAd.Writes || newAd.Writes
-	consolidatedAd.Listings = existingAd.Listings || newAd.Listings
-
 	return consolidatedAd
 }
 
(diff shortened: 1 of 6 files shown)


jhiemstrawisc:
@CannonLock, this PR likely has a few changes you should be aware of. In particular, if you look at changes in `server_structs/director.go`, I've removed a few fields that the Web UI likely ingests. Can you take a look and let me know where I need to make additional changes?

turetske (approved):
Approving, but I want to ensure @CannonLock approved before merging.

bbockelm:
@CannonLock - this one is waiting on you.  Thoughts?

CannonLock:
@jhiemstrawisc I think the issue here is that nothing changes in the UI?

I was told that the server capabilities struct should be:

```
"capabilities": {
    "PublicReads": false,
    "Read": false,
    "Write": false,
    "Listing": false,
    "FallBackReads": false
}
```

but currently as of this PR it is 

```
"capabilities": {
    "PublicRead": false,
    "Read": false,
    "Write": false,
    "Listing": false,
    "FallBackRead": false
}
```

jhiemstrawisc:
That _should_ be unrelated to this PR. One thing I'll note is that we've long since updated the names for a few of these internally and in the docs, but we've avoided updating the JSON representation for backwards compatibility issues. If there's a way for us to start staging this change in the UI without it being a hard break, we'd like the values to be:
```
"capabilities": {
    "PublicReads": false,
    "Reads": false,
    "Writes": false,
    "Listings": false,
    "DirectReads": false
}
```
This is how we store them internally at this point.

Is it possible on your end to write a wrapper that can use both JSONs to create the same web UI elements with those values? If we can do that, I can also update the produced JSON to match the internal values.

Does any of that make sense?

CannonLock:
@jhiemstrawisc Makes perfect sense. I actually non-discriminately regurgitate the key value pairs that the capabilities json provides to me. 

So the UI will work the same without changes so long as you stick to the below standard where the key is a string and its value is a boolean.

"capabilities": {
    "literally any string key": boolean
}

jhiemstrawisc:
That's great to know, and at some point we should chat about where else that assumption is true! Be that the case, I'll update a few more things in this PR.

PR #1596, opened 2024-09-24
Enable debug symbols in image builds of pelican client

jthiltges:
This changes cmake settings to use the RelWithDebInfo build type when building the pelican client for the docker images.

For the price of slightly less optimization (O3 to O2), and a slightly larger output binary (1.3M to 7.3M), we get line level debugging, which will make investigating segfaults much easier.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -71,7 +71,7 @@ RUN \
     cd xrdcl-pelican && \
     git checkout v0.9.4 && \
     mkdir build && cd build && \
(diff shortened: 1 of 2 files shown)


matyasselmeci:
Good call; in fact, RelWithDebugInfo is how we build the xrdcl-pelican RPM.

jhiemstrawisc (approved):

jhiemstrawisc:
Approved, pending passing tests. I'll check back in when they're all done running.

PR #1598, opened 2024-09-24
Switched the 'ls -L' and the logging '-l' flags

### LINKED ISSUE #1550: Switch logging and ls long flags
For 7.11, now that we've warned the users that the logging and ls long flags will be switched, we need to implement the fix.

turetske:
closes #1550


jhiemstrawisc (approved):
LGTM

PR #1602, opened 2024-09-27
Moved web-config.yaml to be in InitServer()

turetske:
Also fixed the other use of $HOME in InitConfig() as CHTC users will not always have a $HOME set

Fixes the chtc bug that the team ran into on the condor pool this week.


jhiemstrawisc (changes requested):
It looks like both `InitServer` and `InitClient` still call the helper function `initConfigDir`. Is it possible to move any logic we need into that helper and then either call it once in `InitConfig` (shared by all Pelican instances) or once in both `InitServer` and `InitClient` so that we don't need to do special case handling directly in those functions?

turetske:
> It looks like both `InitServer` and `InitClient` still call the helper function `initConfigDir`. Is it possible to move any logic we need into that helper and then either call it once in `InitConfig` (shared by all Pelican instances) or once in both `InitServer` and `InitClient` so that we don't need to do special case handling directly in those functions?

So, we can't call it in initConfig because it will hit the same issue we are trying to fix (where there is sometimes no $HOME directory when run as a transfer plugin). Which is why I added back in the weird special case in InitConfig, to avoid that specific instance (but it still needs to exist). Other than that, it's called once at the beginning of `InitClient` and `InitServer`, so I'm unsure which special case handling you are referring to.

jhiemstrawisc (approved):
LGTM

PR #1603, opened 2024-09-29
Add prestage support to the client

bbockelm:
The new `pelican object prestage` command (and the corresponding
public APIs) allow a user to trigger the prestaging of data to a
specific cache.

The prestage itself is not intelligent: it checks to see if a file
is already present and, if not, it triggers a full download.  Future
work may extend this on the Pelican-side so there's a control
interface exposed that triggers prestaging without data movement all
the way to the client.

Includes unit tests and some modest refactoring.  Given the current
LIGO origins don't have the collections URL set (and I wanted to
test it out), this also provides the ability to override the
collections URL on the CLI.


bbockelm:
Note: patch requires the following XRootD fix to work: https://github.com/xrootd/xrootd/pull/2348

bbockelm:
Closing in favor of the reworked version in #1923.

PR #1604, opened 2024-09-30
Add "third party copy" mode

bbockelm:
This adds initial support for so-called "third party copy" mode, where the client (the "third party") triggers a transfer from a Pelican data federation to an origin.  This will eventually support large-scale bulk transfers but, for now, it simply rounds out the `client.DoCopy` method: one can copy from a `pelican://` URL to a `pelican://` URL.

Given the client churn in #1603 (and given #1603 similarly adds new client functionality, working in the same part of the code), this was built on top of that PR.  We should merge this only after that has gone through.


jhiemstrawisc:
@bbockelm this PR has been sitting without updates for ~8 months and has accrued quite a few merge conflicts. If there aren't plans to finish it soon, can we close it?

bbockelm:
Yeah, let's close it for now.  I'll leave the branch not-deleted and hope we can come back when it resurfaces.

jhiemstrawisc:
@bbockelm just wanted to leave a ping that @alexandertuna was recently asking about third party copy to support one of our users. He's cooked up this in the meantime:
```
pelican object put $(curl -s <origin>/<namespace + object>) osdf:///<destination> -t token.txt
```

alexandertuna:
> @bbockelm just wanted to leave a ping that @alexandertuna was recently asking about third party copy to support one of our users. He's cooked up this in the meantime:
> 
> ```
> pelican object put $(curl -s <origin>/<namespace + object>) osdf:///<destination> -t token.txt
> ```

@jhiemstrawisc could you replace `$(...)` with `<(...)` ? I think process substitution will work for this purpose and command substitution won't work

PR #1607, opened 2024-09-30
Fixing the webconfig file issue when there isn't a $HOME directory set

turetske:


jhiemstrawisc (approved):

bbockelm:
Hey -- we need to revisit this:
- InitConfig goes from not being able to fail to calling `os.Exit` directly.  A library should never, ever exit the process.  That's the job of the upper level tool.
- initConfigDir is now public, meaning there's an additional step to setup the configuration.  That's indicative of something else gone wrong.
- There's now been multiple regressions around handling of `$HOME` in the last year.  Let's establish a unit test to get the right behavior in place (`$HOME` unset, `$HOME` set but missing, etc) to make sure we understand what we want.

So we'll probably need to revert this, or at least do a follow-up fix.

turetske:
> Hey -- we need to revisit this:
> 
> * InitConfig goes from not being able to fail to calling `os.Exit` directly.  A library should never, ever exit the process.  That's the job of the upper level tool.
> * initConfigDir is now public, meaning there's an additional step to setup the configuration.  That's indicative of something else gone wrong.
> * There's now been multiple regressions around handling of `$HOME` in the last year.  Let's establish a unit test to get the right behavior in place (`$HOME` unset, `$HOME` set but missing, etc) to make sure we understand what we want.
> 
> So we'll probably need to revert this, or at least do a follow-up fix.

To address point number 2: That was solely to make the unit tests work. Anything that called InitServer in a test (which then bypassed the InitConfig that always runs outside of tests) needed to call InitConfigDir to work and it needed to be public in order to do so. The other solution would be to call InitConfig each time for the tests, but then that would require more viper setup before the tests. I can fix that.

Point 1: I'll fix that.

Point 3: I think we want home set unless it's a transfer plugin? Just so I understand the test requirement.

PR #1608, opened 2024-10-01
Assign unknown client IPs a random geo ip in the contiguous US & cache the result

### LINKED ISSUE #1587: Generate & cache geolocation for uknown client IPs
Right now, the Director generates a completely random list of caches when redirecting a client whose IP has no GeoIP resolution. As our understanding of how cache selection should work evolves, we've identified two primary issues with this:

1. The same GeoIP-less client making repeated requests to the Director gets sent to different caches every time, severely affecting performance (more or less negating the benefits of caching in the first place). Instead of taking this approach, the Director should pick a random GeoIP location (assumed to be in the US) and cache that result for ~20 minutes. This allows repeated requests to get the same list of caches. Note that this moves a lot of the randomness out of the cache selection, and into the GeoIP assignment the Director makes for the client.
2. GeoIP-less clients may sometimes be redirected to caches outside the US, but we'd rather everyone get sent to something in the US. By fixing 1, we also fix 2.

jhiemstrawisc:
This PR aims to improve the Director's cache sorting algorithm, specifically in the case of clients whose GeoIP resolution can't be determined.

With these changes, client IPs that don't have an entry in MaxMind entry (or that can't be parsed as an IP in the first case) are assigned a random location in the contiguous United States. Doing this makes the Director behave as though it knows the client's true location by delivering caches sorted based on the new location. This differs from the previous algorithm, where those IP addresses resulted in completely random cache sorting.

When a client IP is assigned a random location, that location is cached for 20 minutes, unless refreshed by further client interactions. This is done so that repetitive client requests get a consistent cache ordering (barring any other load information or the sort algorithm's stochastic jiggle), which should further improve cache hits.

Closes #1587

### DIFF
--- a/director/sort_test.go
+++ b/director/sort_test.go
@@ -45,6 +45,10 @@ var yamlMockup string
 
 func TestCheckOverrides(t *testing.T) {
 	viper.Reset()
+	t.Cleanup(func() {
+		viper.Reset()
+		geoIPOverrides = nil
+	})
 
 	// We'll also check that our logging feature responsibly reports
 	// what Pelican is telling the user.
@@ -138,8 +142,6 @@ func TestCheckOverrides(t *testing.T) {
 		require.Equal(t, expectedCoordinate.Lat, coordinate.Lat)
 		require.Equal(t, expectedCoordinate.Long, coordinate.Long)
 	})
-
-	viper.Reset()
 }
 
 func TestSortServerAdsByTopo(t *testing.T) {
@@ -192,6 +194,7 @@ func TestSortServerAds(t *testing.T) {
 	viper.Reset()
 	t.Cleanup(func() {
 		viper.Reset()
+		geoIPOverrides = nil
 	})
 
 	// A random IP that should geo-resolve to roughly the same location as the Madison server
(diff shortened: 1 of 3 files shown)


director/sort_test.go
```diff
@@ -450,11 +450,20 @@ func TestGetClientLatLong(t *testing.T) {
...
 		coord := getClientLatLong(clientIp)
 
 		assert.True(t, coord.Lat <= usLatMax && coord.Lat >= usLatMin)
 		assert.True(t, coord.Long <= usLongMax && coord.Long >= usLongMin)
 		assert.Contains(t, logOutput.String(), "Unable to sort servers based on client-server distance. Invalid client IP address")
+		assert.NotContains(t, logOutput.String(), "Retrieving pre-assigned lat/long")
+
+		// Get it again to make sure it's coming from the cache
+		coord = getClientLatLong(clientIp)
+		assert.True(t, coord.Lat <= usLatMax && coord.Lat >= usLatMin)
```
turetske, line 468: Rather than asserting that it's within the same  US bound, assert that the two coordinates are the same.

turetske (changes requested):
One minor change to a test requested.

director/sort_test.go
jhiemstrawisc: Good call. Done :)

turetske (approved):
LGTM!

Issue #1302, opened 2024-05-20
Replace viper.Reset with config.Reset


bbockelm:
In unit tests, we often call `viper.Reset()` to create fresh configuration for the unit tests.

However, this doesn't clear _all_ the global configuration state.  There's also:
- The cached preferred prefix.
- The cached transport object (though this is regenerated if you call `InitClient` or `InitServer`)
- Federation metadata
- Origin exports (note: this is in the `server_utils` module, meaning some refactoring may be required for this idea to work).

Various unit tests reset the above state -- but sometimes this is missed, causing the state from one unit test to leak to another.  This can cause difficult-to-debug ordering dependencies between unit tests (worst case is that one unit test must be run before another in order for the second to pass!).

It'd be better to have a `config.Reset()` helper function that can blow away all the global state instead of having to clear it one-by-one.

PR #1619, opened 2024-10-04
New Reset function 

h2zh:
This new Reset function addresses the issue #1302, integrating:

1. viper.Reset()
2. reset cached preferred prefix
3. reset cached transport object 
4. reset Federation metadata
5. reset Origin exports

If the use case needs to reset all of them, use `server_utils.Reset()`
If it needs to reset 1-4, use `config.ResetConfig()`
If only origin exports need to be reset, use `server_utils.ResetOriginExports()`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1504,7 +1504,7 @@ func InitClient() error {
 	return nil
 }
 
-func Reset() {
+func ResetConfig() {
 	viper.Reset()
 
 	// Clear cached preferred prefix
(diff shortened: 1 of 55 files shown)


config/config.go
```diff
@@ -1503,3 +1503,21 @@ func InitClient() error {
 
 	return nil
 }
+
+func ResetConfig() {
```
bbockelm, line 1507: Please add a comment / docstring to the function.

bbockelm (changes requested):
Small changes requested, inline.

The most significant is to have a more precise name for the function in `server_utils`.  Will cause a lot of code churn - but I think it's worthwhile to get more precise as a start.

Please fix the linter failure as well.  You can use pre-commit to run the linter prior to commit to avoid pushing that will result in early failures.

config/config.go
h2zh: on it : )

h2zh:
address the above issues, could you take another look @bbockelm

bbockelm (approved):
With the last set of changes, all looks good!

Issue #1605, opened 2024-09-30
GeoIP Error Metric


patrickbrophy:
Issues #1548 and #1512 highlight areas for improving GeoIP error handling and failure visualization. After consulting with @bbockelm, we’ve determined that implementing a new Prometheus metric to track GeoIP failures would be ideal. This metric would include several labels:
- `network`, representing the client’s network.
- `source`, indicating whether the error originated from the client or the server.
- `project`, for identifying which project the error relates to.

The overall objective is to generate a ranking of errors by project, allowing for better insight into the frequency and distribution of GeoIP failures.

bbockelm:
@patrickbrophy and I chatted quickly:
- `error_source` is repetitive; let's change it to just `source`.
- Let's propagate the information about the source and project down the call stack in a context object.

Issue #1548, opened 2024-08-09
GeoIP Failure web-interface display


turetske:
This is a branch off of issue #1512 that's being separated out into it's own issue for easier tracking.

For GeoIP lookups If the accuracy returned is ever greater than 100KM, then we should:

* Ensure a warning is displayed on both the director and service (cache/origin) web interface.
* Display an admin-only widget in the director's web interface that is the list of the 5 most recent GeoIP lookup failures

PR #1620, opened 2024-10-07
New GeoIP Error Metric

patrickbrophy:
This PR aims to fix issue #1605. This PR implements a new metric called `pelican_director_geoip_errors`. This metric has the following labels:

- `network` - this is the client's network
- `source` - this is the source of the error, it can take the value of `client` or `server`
- `project` - this is the project that the client sends along in the User-Agent

The goal is to integrate this metric into the WebUI, as addressed in #1548.

### DIFF
--- a/director/sort.go
+++ b/director/sort.go
@@ -159,7 +159,7 @@ func getLatLong(ctx context.Context, addr netip.Addr) (lat float64, long float64
 	labels := prometheus.Labels{
 		"network": "",
 		"source":  "",
-		"project": "",
+		"proj":    "",
 	}
 
 	network, ok := utils.ApplyIPMask(addr.String())
@@ -172,11 +172,11 @@ func getLatLong(ctx context.Context, addr netip.Addr) (lat float64, long float64
 	project, ok := ctx.Value(ProjectContextKey{}).(string)
 	if !ok || project == "" {
 		log.Warningf("Failed to get project from context")
-		labels["project"] = "unknown"
+		labels["proj"] = "unknown"
 		labels["network"] = "unknown"
 		labels["source"] = "server"
 	} else {
-		labels["project"] = project
+		labels["proj"] = project
 	}
 
 	reader := maxMindReader.Load()
--- a/director/sort.go
+++ b/director/sort.go
@@ -165,6 +165,7 @@ func getLatLong(ctx context.Context, addr netip.Addr) (lat float64, long float64
 	network, ok := utils.ApplyIPMask(addr.String())
 	if !ok {
 		log.Warningf("Failed to apply IP mask to address %s", ip.String())
+		labels["network"] = "unknown"
 	} else {
 		labels["network"] = network
 	}
@@ -173,7 +174,6 @@ func getLatLong(ctx context.Context, addr netip.Addr) (lat float64, long float64
 	if !ok || project == "" {
 		log.Warningf("Failed to get project from context")
 		labels["proj"] = "unknown"
-		labels["network"] = "unknown"
 		labels["source"] = "server"
 	} else {
 		labels["proj"] = project
--- a/director/sort.go
+++ b/director/sort.go
@@ -30,10 +30,13 @@ import (
 
 	"github.com/jellydator/ttlcache/v3"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 
+	"github.com/pelicanplatform/pelican/metrics"
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/server_structs"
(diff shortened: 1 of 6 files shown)


director/sort.go
```diff
@@ -144,21 +148,41 @@ func checkOverrides(addr net.IP) (coordinate *Coordinate) {
...
+		log.Warningf("Failed to get project from context")
+		project = "unknown"
+	}
 	reader := maxMindReader.Load()
 	if reader == nil {
 		err = errors.New("No GeoIP database is available")
+		metrics.PelicanDirectorGeoIPErrors.With(prometheus.Labels{
+			"network": network,
+			"source":  "server",
+			"project": project,
```
jhiemstrawisc, line 175: You already know the question I'm going to ask about this 🤓 How can we make sure a malicious client doesn't blow up cardinality by setting bogus projects in the user agent? This might be harder to solve, since the Director doesn't have a list of known projects ahead of time (or at any point). Is there a way to set a cardinality limit so that it's safe to pass user-defined input through to Prometheus?

director/sort.go
```diff
@@ -144,21 +148,41 @@ func checkOverrides(addr net.IP) (coordinate *Coordinate) {
...
+		log.Warningf("Failed to apply IP mask to address %s", ip.String())
+	}
+
+	project, ok := ctx.Value(ProjectContextKey{}).(string)
+	if !ok {
+		log.Warningf("Failed to get project from context")
+		project = "unknown"
+	}
 	reader := maxMindReader.Load()
 	if reader == nil {
```
jhiemstrawisc, line 170: This should probably be treated differently than the other cases because it indicates "the director can't load a database it should have". If this error is triggered by a client request, I'd be surprised if it isn't triggered by _every_ client request. In that case, it doesn't make a lot of sense to include the network/project information because they don't tell us much. If nothing else, we can be more explicit by setting the source to "director".

director/sort.go
```diff
@@ -144,21 +148,41 @@ func checkOverrides(addr net.IP) (coordinate *Coordinate) {
...
 
+	network, ok := utils.ApplyIPMask(addr.String())
+	if !ok {
+		log.Warningf("Failed to apply IP mask to address %s", ip.String())
+	}
+
+	project, ok := ctx.Value(ProjectContextKey{}).(string)
+	if !ok {
+		log.Warningf("Failed to get project from context")
+		project = "unknown"
```
jhiemstrawisc, line 167: The function `extractProjectFromUserAgent` returns an empty string if no project exists in the user agent. Does that get picked up here and assigned to "unknown", or does that need to happen explicitly? My hunch is it needs to be explicit: `if !ok || project == ""`

metrics/director.go
```diff
@@ -91,4 +91,9 @@ var (
 		Name: "pelican_director_client_version_total",
 		Help: "The total number of requests from client versions.",
 	}, []string{"version", "service"})
+
+	PelicanDirectorGeoIPErrors = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "pelican_director_geoip_errors",
+		Help: "The total number of errors encountered while downloading the GeoIP database",
```
jhiemstrawisc, line 97: I don't think the help message is quite correct -- it's not really an issue downloading the database (which happens on Director startup, and then twice on Wednesdays and Fridays), but an issue resolving an IP using the database. How about "The total number of errors encountered trying to resolve coordinates using the GeoIP MaxMind database."

jhiemstrawisc (changes requested):

patrickbrophy (commented):
Implemented the changes requested.

director/sort.go
```diff
@@ -156,33 +156,40 @@ func getLatLong(ctx context.Context, addr netip.Addr) (lat float64, long float64
...
+	} else {
+		labels["network"] = network
 	}
 
 	project, ok := ctx.Value(ProjectContextKey{}).(string)
 	if !ok || project == "" {
 		log.Warningf("Failed to get project from context")
-		project = "unknown"
+		labels["project"] = "unknown"
+		labels["network"] = "unknown"
```
jhiemstrawisc, line 176: Can you move setting the unknown network into the `!ok` check after running `utils.ApplyIPMask`, or is the intended behavior that we'd never set the network if there's no project?

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
The final changes look good. @patrickbrophy, I'll approve now so you can merge but let's wait until the tests finish. There are some tests that have turned exceedingly flaky (mostly a few MacOS ftransfer tests related to large files), so if those fail and they're all that fail, merge away

patrickbrophy:
It looks like I messed up a merge conflict merge so I am going to fix those first.

patrickbrophy:
LGTM

PR #1623, opened 2024-10-08
Director interface

CannonLock:
- Add Origin metrics interface v1
- Add Director metrics interface v1
- Add Graph Overlay and Context for control
- Move time context to root layout
- Update the Nginx Proxy to auto reload credentials
- Add test framework and tests
- Add Boxplot graphs
- Add usage Bar graphs


CannonLock:
@h2zh I made the changes we discussed. Do you mind approving this PR when you get a chance?

h2zh (approved):

h2zh (changes requested):
seems like the ubuntu check failure might be a problem. btw I re-run the ubuntu check in the latest merged PR (#1633) and it still succeeds. I consulted Emma for this and she suggested not to merge it in in case breaking the test completely in main

h2zh (approved):
Emma caught the merge that causes the ubuntu test failure 😎 and estimated the fix might take a bit. So other PRs can go through at the moment

Issue #1502, opened 2024-07-22
Cleanup origin startup stdout


bbockelm:
When watching folks during the PEARC tutorial, it was quite apparent how confusing our startup has become -- lots of "stuff" gets printed to stdout not from Pelican.

Here's what we have currently:

```
Initializing run environment for Pelican...
Jul 22, 2024 6:45:54 PM edu.uiuc.ncsa.security.core.util.MyLoggingFacade info
INFO: scitokens-server:Logging to file /dev/pts/0
Jul 22, 2024 6:45:54 PM edu.uiuc.ncsa.security.core.util.MyLoggingFacade info
INFO: scitokens-server:OAuth 2 for MyProxy, version 5.4.1 startup on Mon Jul 22 18:45:54 UTC 2024
Jul 22, 2024 6:45:54 PM edu.uiuc.ncsa.security.core.util.MyLoggingFacade info
INFO: scitokens-server:loading configuration.
Jul 22, 2024 6:45:54 PM edu.uiuc.ncsa.security.core.util.MyLoggingFacade info
INFO: oa4mp:Error loading authorization configuration. Disabling use of headers
Jul 22, 2024 6:45:54 PM edu.uiuc.ncsa.security.core.util.MyLoggingFacade info
INFO: oa4mp:ping enabled
Jul 22, 2024 6:45:54 PM edu.uiuc.ncsa.security.core.util.MyLoggingFacade info
INFO: oa4mp:No claim source configuration found.
Jul 22, 2024 6:45:54 PM edu.uiuc.ncsa.security.core.util.MyLoggingFacade info
INFO: oa4mp:Starting to load LDAP configuration.
No chain present for host cert; trying to derive one
No chain present; will use empty file
Starting Pelican...
Running pelican with arguments: origin serve -p 8444
Version: 7.9.3-next
Build Date: 2024-06-25T16:40:15Z
Build Commit: a024f4636b25ecccb26308769c0256d488d62392
Built By: goreleaser
2024/07/22 18:45:54 OK   20240528174529_create_db_tables.sql (2.76ms)
2024/07/22 18:45:54 goose: successfully migrated database to version: 20240528174529

Pelican admin interface is not initialized
```

I think we should dump:
- All the oa4mp/Java stuff
- The "No chain present" lines (may be related to the above?)
- The Build date / commit /built by
- Output from goose / ORM.

Marking as critical as I think this might be a simple fix but bigger payoff for new people.

[...]

PR #1625, opened 2024-10-08
Origin startup stdout cleanup - Incorporating Goose and build info output into the logging system

Saartank:
This PR fixes issue #1502 

- It incorporates Goose output into the logging system. Goose has two levels of output: `info` and `fatal`, which are now mapped to the corresponding levels in `logrus`.
- The build information at server startup is now logged as part of the `debug` logs.


config/config.go
```diff
@@ -919,6 +931,13 @@ func PrintPelicanVersion(out *os.File) {
 	fmt.Fprintln(out, "Built By:", GetBuiltBy())
 }
 
+func LogPelicanVersion() {
+	log.Debugf("Version: %s", GetVersion())
```
bbockelm, line 935: If all we're doing is concatenating strings together, the `Debugln` variant is more elegant.  Additionally given the information used to be printed all the time by default; let's move it to `Info`:

```suggestion
	log.Infoln("Version:", GetVersion())
```

(and the subsequent lines as well.)

bbockelm (changes requested):
Just one small comment, then I think we're ready to go.

bbockelm:
@Saartank - looks like there's still a CI failure; can you investigate?

Also, once done, the commit history is a bit scattered on this one.  Please squash it to one commit per piece of functionality.

Saartank:
Closing this PR because I made a mistake while changing the commit history. Created a new PR with a clean commit history: [https://github.com/PelicanPlatform/pelican/pull/1639](https://github.com/PelicanPlatform/pelican/pull/1639).

Issue #1593, opened 2024-09-23
Allow deleting registrations from the web interface


matyasselmeci:
Pelican Service:
- [ ] Client
- [ ] Plugin
- [x] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

Right now in order to delete an origin or cache from the registry (so that you can use a different key, for example), you have to hand-edit the database.  You should be able to do that from the web interface instead.

PR #1626, opened 2024-10-08
frontend for deleting register namespace

h2zh:
image
This feature enables user to delete their approved namespace in the registry.
Address issue #1593

### DIFF
--- a/web_ui/frontend/app/registry/page.tsx
+++ b/web_ui/frontend/app/registry/page.tsx
@@ -216,7 +216,10 @@ export default function Home() {
               <NamespaceCardList<CardProps>
                 data={approvedOriginData}
                 Card={Card}
-                cardProps={{ authenticated: user }}
+                cardProps={{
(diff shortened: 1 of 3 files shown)


CannonLock:
@h2zh Looks good, I pushed the two changes as they were quick. 

Two things:

- I don't have the alert yet in the global context so you have to create the alertportal to set an alert and have it shown

- Run `npm run format:fix` before pushing to fix any js linter issues

CannonLock (approved):

CannonLock (approved):
Approved good catch thank you!

CannonLock:
@h2zh Can you run `npm run format:fix` at web_ui/frontend so that the linter error is resolved?

h2zh:
sure, on it @CannonLock

h2zh:
I ran that command but seems like nothing changes. Could I visit your cubicle to see what's the issues displayed on your side? @CannonLock

PR #1629, opened 2024-10-09
Reduce the threshold for fuse search

CannonLock:


CannonLock:
@h2zh Super trivial one, @bbockelm mentioned that the search was a bit to allowing so I cut it down to a more reasonable value to help limit the matches to something reasonable. 

The .2 value is just trial and error for what felt like the best amount of fuzziness to allow. .2 provides only values that clearly match when you look through by hand. No more `hot` matching `https:**`.

h2zh (approved):

PR #1631, opened 2024-10-09
Stop stomping out the previous tag to update NPM

CannonLock:
Previously we retagged commits so that the npm tag would be updated with the tag used. Now we will just update that npm tag manually before the initial tag to prevent messing things up.


CannonLock:
@jhiemstrawisc Not sure where the release document is but can you add to the step before you push the tag:

```
- Update the `verison` key in the package.json file located at `web_ui/frontend/package.json` to align with the new tag.
```

jhiemstrawisc (approved):
LGTM

jhiemstrawisc:
> @jhiemstrawisc Not sure where the release document is but can you add to the step before you push the tag:
> 
> ```
> - Update the `verison` key in the package.json file located at `web_ui/frontend/package.json` to align with the new tag.
> ```

Done :)

Issue #1622, opened 2024-10-07
Director Redirects Metric


patrickbrophy:
After chatting with @bbockelm, we talked about creating metrics to better understand clients used by the community. This issue is to propose a new metric that will track the number of redirects from the director. 

The labels for this metric are:
- `destination`: Where the director redirected the client to. This label can take on the values of `origin` or `cache`
- `status_code`: The HTTP status code of the redirection
- `version`: The version of the requesting client

PR #1632, opened 2024-10-09
Director Redirection Prometheus Metric

patrickbrophy:
This PR addresses issue #1622. This PR implements a new prometheus counter metric called `pelican_director_redirections_total`. This metric has the following labels:
- `destination`: This is the redirection destination. Either `origin` or `cache`
- `status_code`: This is the HTTP status code of the redirection
- `version`: This is the verision of the client. If we can't extract the version from the User-Agent header, we set this to `unknown`
- `network`: This is the masked IP address of the client


bbockelm:
It looks like the idea to use a deferred function works out nicely here!

Looking at the metric makes me think -- should we be also adding the client IP network as a label?

patrickbrophy:
Yeah that seems like it would be an easy change. I'll add that in.

director/director.go
```diff
@@ -1321,6 +1323,32 @@ func collectClientVersionMetric(reqVer *version.Version, service string) {
...
+	if version == nil {
+		labels["version"] = "unknown"
+		return
+	}
+	labels["version"] = version.String()
+
+	maskedIp, ok := utils.ApplyIPMask(ctx.ClientIP())
+	if ok {
+		labels["network"] = maskedIp
+	}
```
jhiemstrawisc, line 1348: Should we set this to "unknown" if the IP mask comes back as "ok"? An "unknown" in the metric will look better than an empty string because it'll look less like we didn't handle something.

jhiemstrawisc (changes requested):

jhiemstrawisc:
Last note on this -- I think you'll want to rebase against the current main to fix the failing "prettier-linter" test.

jhiemstrawisc (approved):
LGTM

PR #1642, opened 2024-10-10
Install sssd-client in dev/prod containers

jhiemstrawisc:
@williamnswanson, as you requested

When this is patched into 7.10, it should also cause the container builds to pick up xrootd 5.7.1 (unless there's some weird caching going on that I'm unaware of).


bbockelm (approved):
LGTM.  Does exactly what we were aiming for.

PR #1643, opened 2024-10-10
correct OA4MP storage directory permission

h2zh:
see the linked issue for the goal of this PR

How I test it:
Screenshot 2024-10-10 at 3 51 12 PM
(1-4 corresponds the blue dots in the screenshot)
1. delete `/opt/scitokens-server/var/storage` directory if it exists
2. build and run the pelican project. verify if the above directory gets created
3. check the permissions of parent directory is 0755
4. check the permissions of leaf directory `/opt/scitokens-server/var/storage` is 0700

### DIFF
--- a/oa4mp/serve.go
+++ b/oa4mp/serve.go
@@ -308,24 +308,23 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 
 	// Ensure the OA4MP storage directory exists and has correct permissions and ownership
 	OA4MPStoragePath := "/opt/scitokens-server/var/storage"
-	if _, err := os.Stat(OA4MPStoragePath); os.IsNotExist(err) {
-		log.Debugln("OA4MP storage directory does not exist. Creating: ", OA4MPStoragePath)
-		if err = os.MkdirAll(OA4MPStoragePath, 0755); err != nil {
-			log.Errorln("Failed to create directory:", OA4MPStoragePath, ": ", err)
+
+	if err := os.Chmod(OA4MPStoragePath, 0700); err != nil {
+		if os.IsNotExist(err) {
+			log.Debugln("OA4MP storage directory does not exist. Creating", OA4MPStoragePath)
+			if err = os.MkdirAll(OA4MPStoragePath, 0755); err != nil {
+				log.Fatal("Failed to create directory", OA4MPStoragePath, ":", err)
+			}
+			if err = os.Chmod(OA4MPStoragePath, 0700); err != nil {
+				log.Errorln("Failed to change permissions of", OA4MPStoragePath, "after creating it:", err)
+			}
+		} else {
+			log.Errorln("Failed to change permissions of", OA4MPStoragePath, ":", err)
 		}
 	}
 
-	if err := os.Chmod(OA4MPStoragePath, 0700); err != nil {
-		log.Errorln("Failed to change permissions of: ", OA4MPStoragePath, ": ", err)
-	}
-
-	if user.Username != "tomcat" {
-		log.Errorln("Username should be 'tomcat', but is: ", user.Username)
-	}
-	uid, gid := user.Uid, user.Gid
-
-	if err := os.Chown(OA4MPStoragePath, uid, gid); err != nil {
-		log.Errorln("Failed to change ownership of ", OA4MPStoragePath, " to tomcat: ", err)
+	if err := os.Chown(OA4MPStoragePath, user.Uid, user.Gid); err != nil {
+		log.Errorln("Failed to change ownership of", OA4MPStoragePath, "to tomcat:", err)
 	}
 
 	qdlBoot := filepath.Join(param.Issuer_QDLLocation.GetString(), "var", "scripts", "boot.qdl")
--- a/oa4mp/serve.go
+++ b/oa4mp/serve.go
@@ -313,21 +313,21 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 		if os.IsNotExist(err) {
 			log.Debugln("OA4MP storage directory does not exist. Creating", OA4MPStoragePath)
 			if err = os.MkdirAll(OA4MPStoragePath, 0755); err != nil {
-				err = errors.Wrap(err, "Failed to create OA4MP storage directory")
+				err = errors.Wrap(err, "failed to create OA4MP storage directory")
 				return
 			}
 			if err = os.Chmod(OA4MPStoragePath, 0700); err != nil {
-				err = errors.Wrap(err, "Failed to change the permissions of OA4MP storage directory after creating it")
+				err = errors.Wrap(err, "failed to change the permissions of OA4MP storage directory after creating it")
 				return
 			}
 		} else {
-			err = errors.Wrap(err, "Failed to change the permissions of OA4MP storage directory")
+			err = errors.Wrap(err, "failed to change the permissions of OA4MP storage directory")
 			return
 		}
 	}
 
 	if err = os.Chown(OA4MPStoragePath, user.Uid, user.Gid); err != nil {
-		err = errors.Wrap(err, "Failed to change the ownership of OA4MP storage directory")
+		err = errors.Wrap(err, "failed to change the ownership of OA4MP storage directory")
 		return
 	}
 
--- a/oa4mp/serve.go
+++ b/oa4mp/serve.go
@@ -309,22 +309,26 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 	// Ensure the OA4MP storage directory exists and has correct permissions and ownership
 	OA4MPStoragePath := "/opt/scitokens-server/var/storage"
 
-	if err := os.Chmod(OA4MPStoragePath, 0700); err != nil {
+	if err = os.Chmod(OA4MPStoragePath, 0700); err != nil {
 		if os.IsNotExist(err) {
 			log.Debugln("OA4MP storage directory does not exist. Creating", OA4MPStoragePath)
 			if err = os.MkdirAll(OA4MPStoragePath, 0755); err != nil {
-				log.Fatal("Failed to create directory", OA4MPStoragePath, ":", err)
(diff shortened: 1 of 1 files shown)


oa4mp/serve.go
```diff
@@ -306,6 +306,28 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 		return
 	}
 
+	// Ensure the OA4MP storage directory exists and has correct permissions and ownership
+	OA4MPStoragePath := "/opt/scitokens-server/var/storage"
+	if _, err := os.Stat(OA4MPStoragePath); os.IsNotExist(err) {
```
bbockelm, line 311: Skip the `Stat`; just go ahead and do a `os.Chmod` and catch the `os.IsNotExist` as you do currently.

If you do create the directory creation, do a follow-up `os.Chmod`.

oa4mp/serve.go
```diff
@@ -306,6 +306,28 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
...
+		log.Debugln("OA4MP storage directory does not exist. Creating: ", OA4MPStoragePath)
+		if err = os.MkdirAll(OA4MPStoragePath, 0755); err != nil {
+			log.Errorln("Failed to create directory:", OA4MPStoragePath, ": ", err)
```
bbockelm, line 314: On failure, bail-out.  There's no point to log and continue.

Also, tune up the error message; a space is automatically inserted between components and we should only use `:` when concatenating distinct errors.

```suggestion
			log.Errorln("Failed to create directory", OA4MPStoragePath, ":", err)
```

oa4mp/serve.go
```diff
@@ -306,6 +306,28 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 		return
 	}
 
+	// Ensure the OA4MP storage directory exists and has correct permissions and ownership
+	OA4MPStoragePath := "/opt/scitokens-server/var/storage"
+	if _, err := os.Stat(OA4MPStoragePath); os.IsNotExist(err) {
+		log.Debugln("OA4MP storage directory does not exist. Creating: ", OA4MPStoragePath)
```
bbockelm, line 312:
```suggestion
		log.Debugln("OA4MP storage directory does not exist. Creating", OA4MPStoragePath)
```

oa4mp/serve.go
```diff
@@ -306,6 +306,28 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
...
+		if err = os.MkdirAll(OA4MPStoragePath, 0755); err != nil {
+			log.Errorln("Failed to create directory:", OA4MPStoragePath, ": ", err)
+		}
+	}
+
+	if err := os.Chmod(OA4MPStoragePath, 0700); err != nil {
+		log.Errorln("Failed to change permissions of: ", OA4MPStoragePath, ": ", err)
```
bbockelm, line 319:
```suggestion
		log.Errorln("Failed to change permissions of", OA4MPStoragePath, ":", err)
```

also return on error.

oa4mp/serve.go
```diff
@@ -306,6 +306,28 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
...
+	}
+
+	if user.Username != "tomcat" {
```
bbockelm, line 322: This check isn't necessary -- assume it's already correct (we may allow folks to override it via config later.

oa4mp/serve.go
```diff
@@ -306,6 +306,28 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
...
+		log.Errorln("Username should be 'tomcat', but is: ", user.Username)
+	}
+	uid, gid := user.Uid, user.Gid
+
+	if err := os.Chown(OA4MPStoragePath, uid, gid); err != nil {
```
bbockelm, line 327: Eliminate a line:

```suggestion
	if err := os.Chown(OA4MPStoragePath, user.Uid, user.Gid); err != nil {
```

also, below, tune up the log line and return.

bbockelm (changes requested):
A few minor things.  Feel free to squash the resulting commit.

oa4mp/serve.go
```diff
@@ -306,6 +306,27 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
...
 	}
 
+	// Ensure the OA4MP storage directory exists and has correct permissions and ownership
+	OA4MPStoragePath := "/opt/scitokens-server/var/storage"
+
+	if err := os.Chmod(OA4MPStoragePath, 0700); err != nil {
+		if os.IsNotExist(err) {
+			log.Debugln("OA4MP storage directory does not exist. Creating", OA4MPStoragePath)
+			if err = os.MkdirAll(OA4MPStoragePath, 0755); err != nil {
+				log.Fatal("Failed to create directory", OA4MPStoragePath, ":", err)
```
jhiemstrawisc, line 316: Let's avoid `log.Fatal`, which terminates the program by calling `os.Exit(1)`. Instead, when there's an error, it should be returned. Notice the rest of this function does something like this when it encounters an error:
```
if err != nil {
		err = errors.Wrap(err, "Failed to load the private issuer key for running issuer")
		return
	}
```
Because the names of the function's return variables are declared in the function signature, this return is actually returning the `err` variable that's set in the conditional.

That is, rather than log the error, use `errors.Wrap` or `errors.Wrapf` to set `err` and then return.

oa4mp/serve.go
```diff
@@ -306,6 +306,27 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
...
+			}
+			if err = os.Chmod(OA4MPStoragePath, 0700); err != nil {
+				log.Errorln("Failed to change permissions of", OA4MPStoragePath, "after creating it:", err)
```
jhiemstrawisc, line 319: Same here. Instead of logging the error and continuing, do something like:
```
if err = os.Chmod(OA4MPStoragePath, 0700); err != nil {
    err = errors.Wrapf(err, "failed to change permissions of %s after creating it")
    return
}
```

oa4mp/serve.go
```diff
@@ -306,6 +306,27 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
...
+			}
+		} else {
+			log.Errorln("Failed to change permissions of", OA4MPStoragePath, ":", err)
```
jhiemstrawisc, line 322: Same concept here.

oa4mp/serve.go
```diff
@@ -306,6 +306,27 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
...
+		}
+	}
+
+	if err := os.Chown(OA4MPStoragePath, user.Uid, user.Gid); err != nil {
+		log.Errorln("Failed to change ownership of", OA4MPStoragePath, "to tomcat:", err)
```
jhiemstrawisc, line 327: And here.

oa4mp/serve.go
```diff
@@ -306,6 +306,27 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 		return
 	}
 
+	// Ensure the OA4MP storage directory exists and has correct permissions and ownership
+	OA4MPStoragePath := "/opt/scitokens-server/var/storage"
```
jhiemstrawisc, line 310: Is it possible to make this filepath configurable? Presumably OA4MP somehow knows to use this path, so if it's a path Pelican tells it to use (as opposed to it being hardcoded by OA4MP or in a Dockerfile), it'd be better to add this as a configuration option and document what it's for and what gets stored here in `docs/parameters.yaml`. I get a little nervous seeing undocumented posix filepaths hardcoded in packages like this. For example, if this the stuff that gets stored here needs to persist between Docker container restarts, users will have a heck of a time figuring that out unless they go grepping through our source code.

jhiemstrawisc (changes requested):

oa4mp/serve.go
h2zh: Thanks for your reviews! I'm starting to address these issues you mentioned

oa4mp/serve.go
h2zh: The OA4MPStoragePath is a fixed directory controlled by OA4MP. Pelican doesn't tell it to do anything other than changing the permissions here. So I think there is no necessity to make it configurable.

oa4mp/serve.go
```diff
@@ -309,22 +309,26 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 	// Ensure the OA4MP storage directory exists and has correct permissions and ownership
 	OA4MPStoragePath := "/opt/scitokens-server/var/storage"
 
-	if err := os.Chmod(OA4MPStoragePath, 0700); err != nil {
+	if err = os.Chmod(OA4MPStoragePath, 0700); err != nil {
 		if os.IsNotExist(err) {
 			log.Debugln("OA4MP storage directory does not exist. Creating", OA4MPStoragePath)
 			if err = os.MkdirAll(OA4MPStoragePath, 0755); err != nil {
-				log.Fatal("Failed to create directory", OA4MPStoragePath, ":", err)
+				err = errors.Wrap(err, "Failed to create OA4MP storage directory")
```
jhiemstrawisc, line 316: I think this is the last thing I see -- the preferred go syntax is to keep wrapped error messages lowercase, so we prefer `errors.Wrap(err, "failed to create OA4MP ...")` instead of `errors.Wrap(err, "Failed to create OA4MP ...")`. Once that's changed in these error messages, I think this is ready 😀

jhiemstrawisc (changes requested):

oa4mp/serve.go
h2zh: ohhh my bad. on it rn

h2zh: ok. I pushed a commit to punch these capital letters @jhiemstrawisc

jhiemstrawisc (approved):
👍

jhiemstrawisc:
Merging this since I think Brian's comments have already been covered.

PR #1649, opened 2024-10-11
Cleanup MacOS installation instructions

### LINKED ISSUE #1534: Mention "M" chips in installation page for MacOS to point to ARM
We say "If you have ARM64, use this binary", but people may not know that M.X chips mean ARM. Simple enough to mention this in our docs.

### LINKED ISSUE #1535: MacOS installation commands need some love
Issues found:
- Confusing `// ARM` comments in bash things. Rachel suggests removing these
- Untarring the downloaded artifact does not produce the directory our docs says it does. We produce `pelican-<version>`, but the docs say it'll still be `pelican_Darwin_<blah>`
- The "Next Steps" links at the bottom of the installation page are all 404s
- Confusing "which pelican" instructions. We should just say "if `which pelican` produces a path, it's installed"
- Some letters (notably ds in the bash fields) don't render correctly

jhiemstrawisc:
@aowen-uwmad, hoping you have time to review these updates for clarity, grammar, etc. I tried to clean up a bit more than was requested in the issues linked below so that I could hit a few things while they were in my context window 😄 

Closes #1535
Closes #1534

### DIFF
--- a/docs/pages/federating-your-data.mdx
+++ b/docs/pages/federating-your-data.mdx
@@ -1,7 +1,7 @@
 import ExportedImage from "next-image-export-optimizer";
 import { Key } from '@mui/icons-material';
 
-# Federation Your Data via Pelican Origin
+# Federating Your Data via a Pelican Origin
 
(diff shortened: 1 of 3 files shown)


bbockelm:
@aowen-uwmad - will you be able to review this in the next day or two?

aowen-uwmad (approved):
Lgtm, as far as I can tell.
I do not have a Mac, so I can't test that the Mac instructions are correct.

PR #1651, opened 2024-10-11
Stop specifying Federation.DirectorUrl in osdf-cache config

matyasselmeci:
The issue causing caches to have to be told the location of the "origins" director has long since been solved.


jhiemstrawisc (commented):
👍

jhiemstrawisc (approved):

PR #1653, opened 2024-10-14
Switch to use new Caps struct but still handle old JSON from Director

jhiemstrawisc:
This change fixes the stuff I broke in commit ba186d2 by handling previously- unconsidered backwards compatibility issues. Now, unmarshalling the NSAdV2 struct will correctly populate capabilities even if the source JSON uses the old style of capabilities.

This also fixes the capabilities displayed for each origin at the Director.

In addition to passing unit/CI tests, the reviewer should verify that they can start a federation with an Origin and Director based off of this branch AND an Origin based off of a different branch. Both origins should successfully advertise to the Director at the same time, and they should display the correct, updated capability names (e.g. no FallbackRead).

### DIFF
--- a/server_structs/director.go
+++ b/server_structs/director.go
@@ -45,13 +45,21 @@ type (
 		CredentialIssuer url.URL      `json:"issuer"`
 	}
 
-	// Note that the json are kept in uppercase for backward compatibility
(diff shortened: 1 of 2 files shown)


turetske (approved):
Tested with the v7.9.9 version of Pelican and it all worked as expected

PR #1661, opened 2024-10-17
Align OSDF RunLocation hierarchy with Pelican components

brianhlin:
The asymmetry was bothering me. Though it's unclear to me /why/ we need separate run dirs? Because we can't serve different federations with single origin/cache services now?

### DIFF
--- a/systemd/osdf-cache.yaml
+++ b/systemd/osdf-cache.yaml
@@ -20,4 +20,4 @@ XRootD:
   Sitename: TOPOLOGY_RESOURCE_NAME
 Cache:
   DataLocation: "/mnt/osdf"
-  RunLocation: /run/pelican/osdf-cache/xrootd
+  RunLocation: /run/pelican/xrootd/osdf-cache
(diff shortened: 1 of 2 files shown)


turetske:
> The asymmetry was bothering me. Though it's unclear to me /why/ we need separate run dirs? Because we can't serve different federations with single origin/cache services now?

Well, they were originally made different run locations so that we could do the pelican federation in a box testing without the origin and cache clobbering each other.

brianhlin:
> Well, they were originally made different run locations so that we could do the pelican federation in a box testing without the origin and cache clobbering each other.

Oh sorry, I should clarify. What annoyed me was `/run/pelican/xrootd/origin`, the default when running the `pelican` binary in origin mode vs OSDF origin configuration, `/run/pelican/osdf-origin/xrootd`.

What's not particularly clear to me is why we even need an `osdf-origin` run dir vs an `origin` run dir, for example.

turetske (approved):

Issue #1664, opened 2024-10-18
Docs for `Cache.Datalocations`, `Cache.MetaLocations` and `Xrootd.Mount` are confusing and conflicting


jhiemstrawisc:
Documentation for cache config options `Cache.Datalocations/Metalocations` says that these values should not be placed under `Xrootd.Mount` or unintended file exposure will occur. However, `Xrootd.Mount` doesn't appear to be used in the cache's generated xrootd configuration. In fact, the docs for `Xrootd.Mount` only list the origin as an affected service.

We should:
1. Explain what `Cache.Datalocations/Metalocations` do in more user-friendly language, maybe with an explanation of the cache directory structure that looks (something) like:
```
.
└── /xcache/
    ├── data/
    │   ├── 00000000
    │   ├── 00000001
    │   └── 00000002
    ├── meta
    └── namespaces/
        ├── foo --> 00000000
        ├── bar --> 00000001
        └── baz --> 00000002
```
along with why a user would want to configure these variables and what happens when they do/don't. It may also be useful to specify the types of disks that would work well for these (e.g. do they all need fast disk? Or does only `/xcache/data` need fast disk?)

2. Remove references to `Xrootd.Mount` if we determine for sure that caches don't use this variable.

Issue #1227, opened 2024-05-06
Drop privileges in API server


matyasselmeci:
Pelican Service:
- [ ] Client
- [ ] Plugin
- [x] Registry
- [x] Director
- [x] Origin
- [x] Cache
- [ ] Other (please give the detail)

The Pelican web/API server runs as root; this means that a security vulnerability would result in the attacker having root privileges on the server, which is not good.  This is especially an issue since we instruct people to open the API server to the world.

Have the process running the API server drop privileges.  Allow the admin to specify the Unix user and group in their config; by default both should be named 'pelican'.  The native packages (RPMs and Debs) should create this user and group if they does not exist.  The Dockerfiles should pre-create the user and group with a fixed UID/GID.

Ignore the user:group setting if Pelican is not started as root.  Exit with an error if Pelican is started as root but cannot drop privileges (e.g. due to the user not existing).  Give a big warning but continue if the admin specifies root:root as the user:group.

PR #1666, opened 2024-10-18
Add default config.d locations to RPM packaging

### LINKED ISSUE #1660: RPM packaging should use config.d by default
Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [X] Origin
- [X] Cache

Ahead of the OSG 24 upgrade, we should settle on a more admin-friendly configuration layout using multiple `config.d` dirs

- Treat `/etc/pelican/pelican.yaml` as the entrypoint. Should be marked as config but will be overwritten by package updates
- Create `/usr/share/pelican/config.d`: this is where default config will go and any edits will be destroyed upon package update
- Create `/etc/pelican/config.d`: this is where admin config will go, which will remain untouched by packaging. It's a good location for providing default, commented-out example configurations as long as they're marked `config,noreplace`

brianhlin:
Fixes #1660

This changes the default configuration of the Pelican and OSDF services, clearing up ownership 

1.  The Pelican project owns each service config entrypoint, e.g. `/etc/pelican/pelican-origin.yaml`
1.  Federation admins can dump and own files in `/usr/share/pelican/config.d`. I don't actually expect this to see a ton
    of use but since we own (1), making this a directory will allow federation admins to package federation-specific
    config without needing our intervention
1.  Service admins own files in `/etc/pelican/config.d`

(1) and (3) are particularly important as they allow us to more effectively change defaults, including alignment of
`pelican` and `osdf` binaries at the server level and potentially consolidating down to a single `pelican.yaml`
entrypoint.

N.B. this does somewhat break the current paradigm where users could feasibly run multiple different services on a
single host. I think it's fine to do this because:

1.  This was never supported in the proto-Pelican days. It's not clear to me that anyone actually wants it
1.  If someone really wants to do this, they can use their own custom config files + systemd overrides
1.  If we really wanted to, we could add another dir layer per service type (e.g., `/etc/pelican/config.d/origin`) in
    the future

I did not add an example file for caches since it's not clear exactly what we should be telling users (see #1664).

Note to self the OSG Software Team needs to modify the `osdf-server` RPMs to pull in the following files:

-   `systemd/10-osdf-defaults.yaml`
-   `systemd/examples/15-osdf.yaml`
-   `systemd/examples/50-webui.yaml` (though I'm not sure we should be disabling this by default?)


matyasselmeci:
I only skimmed the description so far, but this caught my eye:

> (though I'm not sure we should be disabling [the Web UI] by default?)

I think enabling it by default is a security hazard until #1227 is implemented.

matyasselmeci (approved):
LGTM. Is there a Jira ticket for updating the osdf-server RPMs?

bbockelm:
@brianhlin - there's a fix for various race conditions in the Mac OS X testing in main.  Can you merge or rebase so we get a clean bill of health?

brianhlin:
Done

@matyasselmeci https://opensciencegrid.atlassian.net/issues/SOFTWARE-6028

Issue #1640, opened 2024-10-10
Add Prom Retention Configuration


CannonLock:
Prometheus is hardcoded to retain data for 15 days. It would be great to be able to hold more if you elected.

The existing 15d of data results in 4gb for the director.

Assigned to Brian for triage.

PR #1670, opened 2024-10-21
Configurable Prometheus Retention

patrickbrophy:
This PR addresses issue #1640. This PR adds a new configuration called `Retention`. This PR defines the default retention time as 15 days.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2259,7 +2259,7 @@ name: Monitoring.Retention
 description: |+
   The duration of which Prometheus should retain the monitoring data.
 type: duration
-default: "15d"
+default: 360h
 components: ["origin", "cache", "director", "registry"]
 ---
 ############################
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2258,7 +2258,7 @@ components: ["origin", "cache", "director", "registry"]
 name: Monitoring.Retention
 description: |+
   The duration of which Prometheus should retain the monitoring data.
-type: string
+type: duration
 default: "15d"
 components: ["origin", "cache", "director", "registry"]
 ---
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2260,6 +2260,7 @@ description: |+
   The duration of which Prometheus should retain the monitoring data.
 type: string
 default: "15d"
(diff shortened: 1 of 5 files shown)


docs/parameters.yaml
```diff
@@ -2255,6 +2255,13 @@ type: bool
 default: true
 components: ["origin", "cache", "director", "registry"]
 ---
+name: Monitoring.Retention
+description: |+
+  The duration of which Prometheus should retain the monitoring data.
+type: string
```
CannonLock, line 2261:
```suggestion
type: duration
```

CannonLock (changes requested):
Can't commit on the soundness except for how it relates to the UI, but this tweak makes sure the duration UI is used in the config rather then the default string UI.

docs/parameters.yaml
patrickbrophy: I am not opposed to this, however using the duration type requires that the units are in the following: "ns", "us" (or "µs"), "ms", "s", "m", "h".

CannonLock: Oh gosh, prometheus extends time.Duration units. 

Let's either add a new type that defines this new `Prometheus Duration` or set it to `Duration`. I want to be able to check all inputs in the config UI and setting them as string prevents that. 

I slightly prefer just using duration for now, and if admins come back and say that it sucks to convert a month into hours then we can create a new type.

patrickbrophy: Ok for now I'll just change it to `duration`.

CannonLock (approved):
Syntax LGTM

docs/parameters.yaml
```diff
@@ -2255,6 +2255,13 @@ type: bool
 default: true
 components: ["origin", "cache", "director", "registry"]
 ---
+name: Monitoring.Retention
```
jhiemstrawisc, line 2258: Can we make the variable a little more self descriptive? What about something like "Monitoring.DataRetention" or "Monitoring.TimeSeriesRetention". Those are a bit on the long side, but I like that they help describe more about what's being retained.

jhiemstrawisc (changes requested):
One small change to request a user-facing config name, otherwise looks straight forward and correct

jhiemstrawisc (approved):
LGTM

PR #1673, opened 2024-10-22
Persist server downtime

h2zh:
This PR enables director to persist the origin/cache server downtime set by the admin in a sqlite db. Fundamentally, it syncs the existing variable `filteredServers` in cache with the `server_downtimes` table in db.

In order to test it, you need to add Director.DbLocation: <directory_of_your_choice> (e.g. /workspaces/pelican/director.sqlite) in `pelican.yaml`

A little bit interpretation of the existing code:
`filteredServers` has a data field `filterType`, which have the following types:
> permFiltered: "Permanently Disabled via the director configuration"
> tempFiltered: "Temporarily disabled via the admin website"
> topoFiltered: "Disabled via the Topology policy"
> tempAllowed: "Temporarily enabled via the admin website"
> "": the empty value at the UI side
> default: "Unknown Type"
Different filter types decide the server downtime is temporary or permanent. The new db in this PR coheres this logic.

### DIFF
--- a/director/director_api.go
+++ b/director/director_api.go
@@ -225,7 +225,7 @@ func LaunchServerCountMetric(ctx context.Context, egrp *errgroup.Group) {
 	})
 }
 
-// Populate internal filteredServers map by Director.FilteredServers
+// Populate internal filteredServers map by Director.FilteredServers and director db
 func ConfigFilterdServers() {
 	filteredServersMutex.Lock()
 	defer filteredServersMutex.Unlock()
@@ -236,6 +236,15 @@ func ConfigFilterdServers() {
 
 	for _, sn := range param.Director_FilteredServers.GetStringSlice() {
 		filteredServers[sn] = permFiltered
+	}
+
+	persistedServerStatuses, err := GetAllServerStatuses()
+	if err != nil {
+		log.Error("Failed to read persisted ServerStatuses from director db", err)
+		return
+	}
+	for _, serverStatus := range persistedServerStatuses {
+		filteredServers[serverStatus.Name] = serverStatus.FilterType
 	}
 }
 
--- a/director/director_api.go
+++ b/director/director_api.go
@@ -66,7 +66,7 @@ func listAdvertisement(serverTypes []server_structs.ServerType) []*server_struct
 func checkFilter(serverName string) (bool, filterType) {
 	filteredServersMutex.RLock()
 	defer filteredServersMutex.RUnlock()
-
+	log.Debugln("Checking the filter applied on server", serverName, "Current filteredServers map:", filteredServers)
 	status, exists := filteredServers[serverName]
 	// No filter entry
 	if !exists {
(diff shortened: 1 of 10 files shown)


director/director_db.go
```diff
@@ -0,0 +1,133 @@
...
+// Set filterType of a given server. If the server doesn't exist in director db, create a new entry for it
+func SetServerDowntime(name string, filterType filterType) error {
+	var serverDowntime ServerDowntime
+	// slience the logger for this query because there's definitely an ErrRecordNotFound when a new downtime info inserted
+	err := db.Session(&gorm.Session{Logger: db.Logger.LogMode(logger.Silent)}).First(&serverDowntime, "name = ?", name).Error
+
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return CreateServerDowntime(name, filterType)
+	} else if err != nil {
+		return errors.Wrap(err, "Error retrieving Server Status")
```
jhiemstrawisc, line 116: "Error" should also be lowercased because this is a wrapped/returned error.

This is also a small nitpick, but I think the more canonical "Go" way of doing this error handling would be:
```
if err != nil {
    if errors.Is(err, gorm.ErrRecordNotFound) {
        return CreateServerDowntime(name, filterType)
    }
    
    return errors.Wrapf(err, "unable to retrieve downtime status for server %s", name)
}
```

director/director_ui.go
```diff
@@ -251,8 +251,14 @@ func handleFilterServer(ctx *gin.Context) {
 	// If we previously temporarily allowed a server, we switch to permFiltered (reset)
 	if filterType == tempAllowed {
 		filteredServers[sn] = permFiltered
+		if err := SetServerDowntime(sn, permFiltered); err != nil {
+			log.Errorln("Failed to persist the downtime in director db", err)
```
jhiemstrawisc, line 255: These two error log lines should have a colon to separate the your error handle and the error produced by the function.

docs/parameters.yaml
```diff
@@ -1243,6 +1243,14 @@ components: ["cache"]
 ############################
 #  Director-level configs  #
 ############################
+name: Director.DbLocation
+description: |+
+  A filepath to the intended location of the director's database.
```
jhiemstrawisc, line 1248: Let's also include some information about what's in the database.

director/director_db.go
```diff
@@ -0,0 +1,133 @@
...
+		UpdatedAt:  time.Now(),
+	}
+
+	if err := db.Create(serverDowntime).Error; err != nil {
+		return err
+	}
+	return nil
+}
+
+func GetServerDowntime(name string) (filterType, error) {
```
jhiemstrawisc, line 88: This comment applies to all of the exported functions in this file other than `InitializeDB`-- think hard about whether we need them to be exported, which implies it's a function that's meant to be called by other packages. In this case, I think the majority of these functions are meant to be used internally by the Director. If you can make the case that some other package may eventually need to call one of these functions(like `launchers` has to do to set up the Director database), keep them exported. Otherwise we should lowercase the first letter of the function.

jhiemstrawisc (changes requested):

director/director_api.go
```diff
@@ -66,7 +66,7 @@ func listAdvertisement(serverTypes []server_structs.ServerType) []*server_struct
 func checkFilter(serverName string) (bool, filterType) {
 	filteredServersMutex.RLock()
 	defer filteredServersMutex.RUnlock()
-
+	log.Debugln("Checking the filter applied on server", serverName, "Current filteredServers map:", filteredServers)
```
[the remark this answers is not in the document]
h2zh: thanks for these comments! I start on addressing them :) filteredServers should only keep in my local debug phase

h2zh:
I think this PR needs to be squashed then merge

director/director_api.go
```diff
@@ -232,19 +232,19 @@ func ConfigFilterdServers() {
...
+		log.Debugln("Loaded server downtime configuration from the Director.FilteredServers parameter:", filteredServers)
 	}
 
-	if param.Director_DbLocation.IsSet() {
-		persistedServerDowntimes, err := GetAllServerDowntimes()
+	if param.Director_DbLocation.GetString() != "" {
+		persistedServerDowntimes, err := getAllServerDowntimes()
 		if err != nil {
-			log.Error("Failed to read persisted ServerDowntimes from director db", err)
+			log.Error("Failed to read persisted server downtimes from director db", err)
```
jhiemstrawisc, line 241: This log line should have a colon to separate the two halves of the error message.

jhiemstrawisc (changes requested):
One very small change, otherwise things look good to me :)

director/director_api.go
h2zh: gotcha. thanks!

PR #1677, opened 2024-10-23
Tuning Pelican for running xrootd under valgrind

bbockelm:
Valgrind is incredibly useful C++ debugging tool to check for leaks and other memory usage errors.

However, it also makes the application incredibly slow: it's often a 20x (or more) performance penalty.  Xrootd caches hit the timeouts frequently.

This PR adds a few hidden tunables to allow us to increase timeouts enough for xrootd to run under valgrind.  It also adds some standard C++ debugging tools into the development container.

### DIFF
--- a/config/resources/defaults.yaml
+++ b/config/resources/defaults.yaml
@@ -60,6 +60,7 @@ Director:
   CachePresenceTTL: 1m
   CachePresenceCapacity: 10000
 Cache:
+  DefaultCacheTimeout: "9.5s"
   Port: 8442
   SelfTest: true
   SelfTestInterval: 15s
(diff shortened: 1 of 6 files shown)


jhiemstrawisc (changes requested):
I don't think we pull the actual defaults from `docs/parameters.yaml` yet, so the value we get from `param.Xrootd_MaxStartupWait.GetDuration()` is going to be 0 unless it's defined in the user's custom config. My guess is this is what's breaking the ticker.

Defaults can be defined in either `config/resources/defaults.yaml` or in our new `config/config.go::SetServerDefaults`

turetske (approved):

jhiemstrawisc (approved):

Issue #1636, opened 2024-10-09
Improve `pelican_director_server_count` metric


patrickbrophy:
After a brief chat with @bbockelm, we talked about updating the `pelican_director_server_count` metric. The first of the these improvements is cardinality. The cardinality is quite large with the `server_lat`, `server_long`, `server_web_url`, and `server_url`. All of these can be removed. We can also change this metric to a gauge to track the amount of currently connected servers to the director. This change is logical with the idea that the amount of servers can change both positively and negatively. The final improvement would be to add the version of the server. This is so we can see when to break backwards compatibility.

bbockelm:
@jhiemstrawisc / @CannonLock - you guys may be interested in watching this issue.

Having the version number in Prometheus for each of our services will allow us to answer questions like "can we drop backward compatibility with caches running version XYZ?"  I plan a follow-up issue to display this information in the director UI.

PR #1684, opened 2024-10-23
Improved `pelican_director_server_count` metric

patrickbrophy:
This PR addresses issue #1636. This PR removes unnecessary labels from the metric and changes the metric to a gauge. This metric should now reflect at any given moment, how many servers are connected to the director.

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -106,7 +106,7 @@ var (
 )
 
 func init() {
-	HookServerAdsCache()
+	hookServerAdsCache()
 }
 
 func getRedirectURL(reqPath string, ad server_structs.ServerAd, requiresAuth bool) (redirectURL url.URL) {
--- a/director/director.go
+++ b/director/director.go
@@ -105,6 +105,10 @@ var (
 	statUtilsMutex = sync.RWMutex{}
 )
 
+func init() {
+	HookServerAdsCache()
+}
+
 func getRedirectURL(reqPath string, ad server_structs.ServerAd, requiresAuth bool) (redirectURL url.URL) {
 	var serverURL url.URL
 	if requiresAuth && ad.AuthURL.String() != "" {
--- a/launchers/director_serve.go
+++ b/launchers/director_serve.go
@@ -45,7 +45,7 @@ func DirectorServe(ctx context.Context, engine *gin.Engine, egrp *errgroup.Group
 
 	director.LaunchMapMetrics(ctx, egrp)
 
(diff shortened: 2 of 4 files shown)


launchers/director_serve.go
```diff
@@ -45,7 +45,7 @@ func DirectorServe(ctx context.Context, engine *gin.Engine, egrp *errgroup.Group
 
 	director.LaunchMapMetrics(ctx, egrp)
 
-	director.LaunchServerCountMetric(ctx, egrp)
+	director.LaunchServerCountMetric()
```
jhiemstrawisc, line 48: Since this isn't actually launching a go routine anymore, I'm not sure I love keeping the setup here (which mostly deals with launching background tasks).

Maybe move this closer to where the `serverAds` cache is actually constructed? This may be a good time to use an `init()` function in the `director` package.

metrics/director.go
```diff
@@ -82,10 +82,10 @@ var (
 		Help: "The total stat queries the director issues. The status can be Succeeded, Cancelled, Timeout, Forbidden, or UnknownErr",
 	}, []string{"server_name", "server_url", "server_type", "result", "cached_result"}) // result: see enums for DirectorStatResult
 
-	PelicanDirectorServerCount = promauto.NewCounterVec(prometheus.CounterOpts{
+	PelicanDirectorServerCount = promauto.NewGaugeVec(prometheus.GaugeOpts{
 		Name: "pelican_director_server_count",
-		Help: "Total number of servers, delineated by pelican/non-pelican and origin/cache",
-	}, []string{"server_name", "server_type", "server_url", "server_web_url", "server_lat", "server_long", "from_topology"})
+		Help: "The number of servers, delineated by pelican/non-pelican and origin/cache",
```
jhiemstrawisc, line 87: Let's be even more explicit and say "The number of servers currently recognized by the Director, delineated by blah blah"

jhiemstrawisc (changes requested):
The meat is all there, my main concern is moving the metric setup out of `launchers` since it no longer relies on a background process

director/director.go
```diff
@@ -105,6 +105,10 @@ var (
 	statUtilsMutex = sync.RWMutex{}
 )
 
+func init() {
+	HookServerAdsCache()
```
jhiemstrawisc, line 109: I think this is the last change request from me -- now that this isn't being access by another package, there's no need to export. Can you lowercase the function name?

jhiemstrawisc (approved):

PR #1691, opened 2024-10-28
POSIX Sync Handling

turetske:
This code adjust the sync behavior for transfers to and from origins. Only fully tested on POSIX right now due to S3 PROPFIND/LS issues.

The cases and behavior that should occur are discussed here: #1638 - For the benefit of the reviewer, here is the list of test commands I ran with the various caveats about the files/dirs/collections/objects. Be sure you match the various different existence cases of each test.

| Left Argument Type | Right Argument Type | Command | Expected Result | Description |
| ----------  | -------------------- | --------------------- | ------ | ------------ |
| Collection | File (which exists) | `/pelican object sync pelican://3695d3c19055:8445/my-origin/test origin.go` | FAILURE | Trying to sync a remote collection with a file is expected to fail|
| Collection | Directory (does exist) | `./pelican object sync pelican://3695d3c19055:8445/my-origin/test $PWD/test` | SUCCESS |  Everything in `my-origin/test` is recursively moved into `$PWD/test` |
| Collection | Directory (does not exist | `./pelican object sync pelican://3695d3c19055:8445/my-origin/test $PWD/test` | SUCCESS | Creates `$PWD/test` and everything in `my-origin/test` is recursively moved into `$PWD/test` | 
| Object | Directory (exists) | `./pelican object sync pelican://3695d3c19055:8445/my-origin/locked-file2.txt tmp/ ` | SUCESS | Moves `locked-file2.txt` into `tmp/` |
| Object | FilePath (does not exist) w '/'| `./pelican object sync pelican://3695d3c19055:8445/my-origin/locked-file2.txt tmp/` | SUCCESS | Creates `tmp/` and moves `locked-file2.txt` into `tmp`|
| Object | DirPath (does not exist) w/o '/'| `./pelican object sync pelican://3695d3c19055:8445/my-origin/locked-file2.txt tmp` | SUCCESS | Copies `locked-file2.txt` into `tmp` |
| Object | FilePath (exists) | `./pelican object sync pelican://3695d3c19055:8445/my-origin/locked-file2.txt tmp.txt ` | SUCCESS | Overwites `tmp.txt` with `locked-file2.txt` |
| Directory | Namespace Prefix | `./pelican object sync $PWD/tmp/inner/ pelican://3695d3c19055:8445/my-origin/outer -t test-token.jwt` | SUCCESS | Copies contents of `$PWD/tmp/inner/` into `/my-origin/outer` Creating prefixes as needed |
| File | Collection (POSIX Origin) | `./pelican object sync $PWD/testfile.txt pelican://3695d3c19055:8445/my-origin/outer -t test-token.jwt` | FAILURE | Trying to sync a file with a POSIX Collection should fail |
| File | Collection (S3 Origin) | `./pelican object sync /workspaces/pelican/test-cache/xrootd_server.go pelican://3695d3c19055:8445/my-s3/testing/uwdf-test -d -t test-token.jwt` | SUCCESS | Trying to sync a file with an S3 collection name should create an object of that name (because S3 allows for both)|
| File | Object (POSIX Origin) | `./pelican object sync $PWD/testfile.txt pelican://3695d3c19055:8445/my-origin/test-file.txt -t test-token.jwt` | SUCCESS | Overwrites `/my-origin/test-file.txt` with `$PWD/testfile.txt`|
| File | non-existent namespace prefix | `./pelican object sync $PWD/testfile.txt pelican://3695d3c19055:8445/my-origin/fileput/oops/test.txt -t test-token.jwt` | SUCCESS | Writes the file to the location and creates prefixes as necessary|

### DIFF
--- a/client/fed_linux_test.go
+++ b/client/fed_linux_test.go
@@ -32,7 +32,6 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/pelicanplatform/pelican/client"
@@ -68,12 +67,12 @@ func TestRecursiveUploadsAndDownloads(t *testing.T) {
 	tokenConfig.AddResourceScopes(token_scopes.NewResourceScope(token_scopes.Storage_Read, "/"),
 		token_scopes.NewResourceScope(token_scopes.Storage_Modify, "/"))
 	token, err := tokenConfig.CreateToken()
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	tempToken, err := os.CreateTemp(t.TempDir(), "token")
-	assert.NoError(t, err, "Error creating temp token file")
+	require.NoError(t, err, "Error creating temp token file")
 	defer os.Remove(tempToken.Name())
 	_, err = tempToken.WriteString(token)
-	assert.NoError(t, err, "Error writing to temp token file")
+	require.NoError(t, err, "Error writing to temp token file")
 	tempToken.Close()
 
 	// Disable progress bars to not reuse the same mpb instance
@@ -81,9 +80,9 @@ func TestRecursiveUploadsAndDownloads(t *testing.T) {
 
 	// Make our test directories and files
 	tempDir, err := os.MkdirTemp("", "UploadDir")
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	innerTempDir, err := os.MkdirTemp(tempDir, "InnerUploadDir")
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	defer os.RemoveAll(tempDir)
 	defer os.RemoveAll(tempDir)
 	permissions := os.FileMode(0755)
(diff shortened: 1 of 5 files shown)


client/fed_linux_test.go
```diff
@@ -580,9 +579,67 @@ func TestSyncUpload(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, newTestFileContent, string(contentBytes))
+
+		smallTestFile, err := os.CreateTemp(newDir, "smallTest")
+		smallTestFileContent := "smaller test file"
+		require.NoError(t, err, "Error creating small test file")
+		_, err = smallTestFile.WriteString(smallTestFileContent)
+		require.NoError(t, err, "Error writing to small test file")
+
+		//Upload the file into the same location as the previous test file - should overwrite
```
jhiemstrawisc, line 621: Doesn't overwriting an object violate its immutability? Shouldn't that be something we disallow?

client/fed_linux_test.go
```diff
@@ -675,8 +732,54 @@ func TestSyncDownload(t *testing.T) {
...
+
+	t.Run("testSyncDownloadObject", func(t *testing.T) {
+		// Set path for object to upload/download
+		dirName := t.TempDir()
+		filename1 := filepath.Base(tempFile1.Name())
+		uploadUrlObj := fmt.Sprintf("%s/%s", uploadUrl, filename1)
+		downloadObjName := filepath.Join(dirName, filename1)
+
+		// Synchronize a download of a single file into an existing directory (should create the file)
+		transferDetailsDownload, err := client.DoGet(fed.Ctx, uploadUrlObj, dirName, true, client.WithTokenLocation(tempToken.Name()), client.WithSynchronize(client.SyncSize))
```
jhiemstrawisc, line 747: Given the direction of these transfers, can you use `downloadUrlObj` instead of `uploadUrlObj`?

client/handle_http.go
```diff
@@ -2464,7 +2506,42 @@ func (te *TransferEngine) walkDirUpload(job *clientTransferJob, transfers []tran
 	// Get our list of directory entries
 	infos, err := os.ReadDir(localPath)
 	if err != nil {
-		return err
+		info, err := os.Stat(localPath)
+		if err != nil {
+			return errors.Wrap(err, "failed to stat local path")
+		}
+		// If the path leads to a file and not a directory, create a job to uploade the file and return
```
jhiemstrawisc, line 2513: uploade --> upload

jhiemstrawisc (changes requested):
The test cases as described in the chart seem to work for me. However, I'm concerned about the cases where the behavior is for a local file to overwrite a remote object, which seems contrary to the object's immutability...

Another test case whose behavior bothered me was:
```
pelican object sync <local file> pelican://`hostname`/my-origin/some/path/
```

where the remote path ends in a trailing `/`. In this case, the remote object `/some/path` is created, which I agree is correct -- but I also think we should log a warning for the user that the trailing `/` in their command is being ignored and _not_ interpreted as a directory.

client/fed_linux_test.go
turetske: I'll talk with Brian in this case, but in our original discussion and my pictures of the whiteboard, he wanted an overwrite because the idea is that we function mostly like `rsync`. 

I think the case isn't that we don't allow overwrites, we just assume immutability and download behavior is undefined if the origin value is changed. 

Since the origin owner/admin could easily change the underlying files on a POSIX origin as well, I think preventing it in this case isn't worth it unless we could also protect the underlying file structure, which we can't.

turetske: Actually, even more importantly, if the sync fails midway, we need to be able to overwrite the file even if it still exists. Because the original failed. So yes, we need to be able to do this.

cmd/object_sync.go
```diff
@@ -181,6 +181,13 @@ func syncMain(cmd *cobra.Command, args []string) {
 		}
 	} else {
 		for _, src := range sources {
+			if srcStat, err := os.Stat(src); err != nil {
+				log.Errorln("Source does not exist")
```
jhiemstrawisc, line 185: Can you log the source here too?

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):

PR #1696, opened 2024-10-28
Macos poller fix

turetske:
Adding back in the accidentally removed poller fix for the MacOSX and the pelican protocol. Should hopefully make Mac tests more robust.


jhiemstrawisc (commented):
Sweetness, I look forward to this fixing our tests!

jhiemstrawisc (approved):

PR #1702, opened 2024-10-30
Correcting inconsistent use of x86_64 vs arm64

aowen-uwmad:
Assumed focus would be on x86_64, so made that the "main" naming used throughout.
Added note to PATH section about changing the folder name if using ARM, to mirror previous section.


jhiemstrawisc (approved):
👍 Thanks for the cleanup!

PR #1704, opened 2024-11-01
Update api-docs.mdx

CannonLock:
Exercise in getting this generated server side is futile since the SwaggerUI component is client only anyways.


jhiemstrawisc (approved):
LGTM -- let's see if re-running the failed unit tests gives us the green, and then we're good to merge

Issue #1706, opened 2024-11-01
Ungraceful Handling of Object Retrieval with Directory Path


Saartank:
When a directory path is passed to the `pelican object get` command without the recursive flag, it is not handled gracefully. Here is the current behavior:

- **Standard flow through Cache:**  
  An error is printed, and an empty file is created at the destination.

![image](https://github.com/user-attachments/assets/d523f836-d06b-4fd9-b8a0-c0a138a07a46)

- **Direct read through Origin:**  
  The command runs successfully. A file is created at the destination, containing HTML of a table listing the contents of the requested directory.
    
![image](https://github.com/user-attachments/assets/a96a76fc-150d-45ac-99c0-18989f55a878)

brianaydemir:
Closing as not planned. I don't want to change how we configure XRootD by default, e.g., by turning off its directory listings, and in the absence of reports from users, it's not clear it's worth the effort to update the client and caches to check for `get`s on directories.

aowen-uwmad:
To round out the scope, I was able to trigger this error a couple of ways.

**Pelican CLI**

```
$ pelican --version
Version: 7.22.0
Build Date: 2025-12-19T19:00:13Z
Build Commit: 074a9766f8c06e281c4fbfeb98a205f021288b56
Built By: goreleaser
```

```
$ pelican object get osdf:///pelicanplatform/test ./
$ cat test
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<meta http-equiv="content-type" content="text/html;charset=utf-8"/>
<link rel="stylesheet" type="text/css" href="/static/css/xrdhttp.css"/>
<link rel="icon" type="image/png" href="/static/icons/xrdhttp.ico"/>
<title>/pelicanplatform/test</title>
</head>
<body>
<h1>Listing of: /pelicanplatform/test</h1>
<div id="header"><table id="ft">
<thead><tr>
<th class="mode">Mode</th><th class="flags">Flags</th><th class="size">Size</th><th class="datetime">Modified</th><th class="name">Name</th></tr></thead>
[...]

Issue #1645, opened 2024-10-11
Ignore directories in the transfer plugin


bbockelm:
HTCSS, in certain configurations, ask Pelican to transfer a directory on stageout.  This is incorrect -- HTCSS already iterates through the directory and provides a list of objects.

While we are waiting on a fix from upstream, we should simply skip requests to upload a directory; that is, stat the name being given to the plugin and continue in the loop if the name refers to a directory.

PR #1707, opened 2024-11-01
IsDir Check to Prevent Creation of Empty Files on Object Uploads

Saartank:
## `IsDir` Check to Prevent Creation of Empty Files on Object Uploads

Currently, when trying to upload an object without the recursive option and a directory path is provided, Pelican does not handle this gracefully. It outputs an error because it attempts to read a directory, and it also creates an empty file at the destination in the process.

This issue affects both `pelican object put` and `pelican plugin` commands since the same underlying function manages object uploads.

### Related Issue #1645

When Condor attempts to transfer a directory, it iterates through it and places both directory paths and file paths in the infile. Attempting to transfer a directory without the recursive option causes Pelican to create an empty file at the destination, which breaks the code.

This PR adds an `IsDir` check before initiating an object transfer. This update resolves both issues:
1. Passing a directory in the infile provided by Condor will no longer have any effect.
2. Using `pelican object put` with a directory without the recursive option will not create an empty file at the destination and will only output an error.

A similar problem occurs while downloading an object as well. This is more complex since the object to be transferred is on a remote machine, and the behavior differs when downloading from a cache or directly from the origin using the `?directread` query. A separate Issue #1706 has been created for tracking this.

### DIFF
--- a/client/fed_linux_test.go
+++ b/client/fed_linux_test.go
@@ -835,3 +835,91 @@ func TestSyncDownload(t *testing.T) {
 		require.Equal(t, newTestFileContent, string(contentBytes))
 	})
 }
+
+// This test verifies the behavior when a directory path is passed to the object put command without the recursive option.
+// In this scenario, an appropriate error message should be logged, and no directory or file should be created at the destination.
+func TestObjectPutNonRecursiveDirPath(t *testing.T) {
+	// Create instance of test federation
+	server_utils.ResetTestState()
+
+	fed := fed_test_utils.NewFedTest(t, mixedAuthOriginCfg)
+	discoveryUrl, err := url.Parse(param.Federation_DiscoveryUrl.GetString())
+	require.NoError(t, err)
+
+	te, err := client.NewTransferEngine(fed.Ctx)
+	require.NoError(t, err)
+
+	// Create a token file
+	issuer, err := config.GetServerIssuerURL()
+	require.NoError(t, err)
+
+	tokenConfig := token.NewWLCGToken()
+	tokenConfig.Lifetime = time.Minute
+	tokenConfig.Issuer = issuer
+	tokenConfig.Subject = "origin"
+	tokenConfig.AddAudienceAny()
+	tokenConfig.AddResourceScopes(token_scopes.NewResourceScope(token_scopes.Storage_Read, "/"),
+		token_scopes.NewResourceScope(token_scopes.Storage_Modify, "/"))
+	token, err := tokenConfig.CreateToken()
+	require.NoError(t, err)
+	tempToken, err := os.CreateTemp(t.TempDir(), "token")
+	require.NoError(t, err, "Error creating temp token file")
+	defer os.Remove(tempToken.Name())
+	_, err = tempToken.WriteString(token)
+	require.NoError(t, err, "Error writing to temp token file")
+	tempToken.Close()
+
+	// Disable progress bars to not reuse the same mpb instance
+	viper.Set("Logging.DisableProgressBars", true)
+
+	tempDir, err := os.MkdirTemp("", "UploadDir")
+	require.NoError(t, err)
+	defer os.RemoveAll(tempDir)
+	permissions := os.FileMode(0755)
+	err = os.Chmod(tempDir, permissions)
+	require.NoError(t, err)
+
+	t.Run("testObjectPutNonRecursiveDirPath", func(t *testing.T) {
+
+		oldPref, err := config.SetPreferredPrefix(config.PelicanPrefix)
+		require.NoError(t, err)
+		defer func() {
+			_, err := config.SetPreferredPrefix(oldPref)
+			require.NoError(t, err)
+		}()
+
+		for _, export := range fed.Exports {
+			tempPath := tempDir
+			dirName := filepath.Base(tempPath)
+			uploadUrl := fmt.Sprintf("pelican://%s%s/%s/%s", discoveryUrl.Host,
+				export.FederationPrefix, "osdf_osdf", dirName)
+
+			// Upload the file with PUT with recursive bool set to false
+			_, err := client.DoPut(fed.Ctx, tempDir, uploadUrl, false, client.WithTokenLocation(tempToken.Name()))
+
+			require.Error(t, err, "Expected an error when passing a dir path to object put command without recursive option set")
+			expectedMessage := "the provided path '" + tempPath + "' is a directory, but a file is expected"
+			require.Contains(t, err.Error(), expectedMessage, "Error message did not match expected text")
+
+			// Check if any file is created at the destination by using get
+			if export.Capabilities.PublicReads {
+				_, err = client.DoGet(fed.Ctx, uploadUrl, t.TempDir(), false)
+			} else {
+				_, err = client.DoGet(fed.Ctx, uploadUrl, t.TempDir(), false, client.WithTokenLocation(tempToken.Name()))
+			}
+
+			require.Error(t, err, "Expected an error when attempting to find a non-existent object")
+			require.Contains(t, err.Error(), "server returned 404 Not Found", "Error message did not match expected text")
+
+		}
+	})
+
+	t.Cleanup(func() {
+		if err := te.Shutdown(); err != nil {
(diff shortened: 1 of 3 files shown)


client/handle_http.go
```diff
@@ -2153,6 +2153,14 @@ func uploadObject(transfer *transferFile) (transferResult TransferResults, err e
 		ioreader = ap
 		sizer = ap
 	} else {
+
+		if fileInfo.IsDir() {
+			log.Errorln("The given path", transfer.localPath, "is a directory. Use --recursive or -r flag if using the pelican object command for directories")
```
turetske, line 2158:
```suggestion
			log.Errorln("The given path", transfer.localPath, "is a directory. Use the sync command or the --recursive or -r flag if using the pelican object command for directories")
```

client/handle_http.go
```diff
@@ -2153,6 +2153,14 @@ func uploadObject(transfer *transferFile) (transferResult TransferResults, err e
...
+			err := errors.New("the provided path '" + transfer.localPath + "' is a directory, but a file is expected; use --recursive or -r flag if using the pelican object command for directories")
```
turetske, line 2159:
```suggestion
			err := errors.New("the provided path '" + transfer.localPath + "' is a directory, but a file is expected; use the sync command or the --recursive or -r flag if using the pelican object command for directories")
```

turetske (changes requested):
Two issues:

We don't want the transfer plugin to fail if it hits a directory, just ignore it. Could returning a transferResult with an error when the plugin attempts to do a bunch of transfers cause it to report the transfers having failed rather than simply realizing that this is a skipped transfer? What actually occurs with this change on condor end? Does a transfer of a whole directory finish correctly?

Can you add a unit test which will catch this? Something which tries to upload a directory when using pelican object put and gets the error?

client/handle_http.go
jhiemstrawisc, line 2159: One quick comment since I'm looking at this now -- instead of having two extremely similar messages, one that gets logged and one that gets returned, just return the error. Otherwise the debug logs at the client get very repetitive.

Saartank: Yes, I understand. I have updated this in my latest commit and am now only returning the error.

https://github.com/PelicanPlatform/pelican/pull/1707/files#diff-4b3ff5f5414bcb82345ca02dead65d984f23d0737392da402d2517b8f192ce61R2158-R2160

client/handle_http.go
Saartank: I simplified the message to: `the provided path <given-path> is a directory, but a file is expected`. I felt it was unnecessary to list all possible ways to handle directories, as new methods might be introduced in the future. Let me know if you disagree, and I can revise it.

https://github.com/PelicanPlatform/pelican/pull/1707/files#diff-4b3ff5f5414bcb82345ca02dead65d984f23d0737392da402d2517b8f192ce61R2158

Saartank:
The problem with Condor, as I understand, is that when it tries to upload a directory, it iterates through it and includes both the inner file paths and the directory path in the infile. This approach is incorrect it should only list the files (objects) in the infile and they are working on fixing this issue. The problem with Pelican is that when a directory path is passed for upload without the recursive option set, it creates an empty file at the destination with the same name as the directory. This Pelican behavior, combined with Condor’s listing of directory paths in the infile, causes problems. This PR adds a directory check so that when a directory path is passed (without the recursive option), no empty file with the directory name is created. I have added an end-to-end test to precisely mimic this scenario.

https://github.com/PelicanPlatform/pelican/pull/1707/files#diff-7047045de9e6b2f5fc15f5af465c47b95cde0a18745cd7e34ee61faba13267cbR316-R488

This is an end-to-end test that attempts to upload a directory containing files. In the infile, the test lists the paths of both the inner files and the directory (without any recursive option). The plugin is then initiated, and the test verifies the expected behavior: the two files should be successfully uploaded, while the directory itself should not be uploaded, and no empty file should be created with the directory name at the destination. An appropriate error should be returned in the directory upload's corresponding result-ad in the outfile.

An error was previously returned as well, but it was not the error we initiated. Instead, it was an error generated when attempting to read the directory. This PR gracefully handles the described scenario, returning an appropriate error message and preventing an empty file at the destination.

cmd/plugin_test.go
```diff
@@ -311,6 +313,179 @@ func TestStashPluginMain(t *testing.T) {
...
+// - The directory itself is not uploaded, and no empty file with the directory name is created at the destination.
+// - An appropriate message indicating the directory upload failure is returned in the corresponding classad.
+func TestInfileUploadWithDirAndFiles(t *testing.T) {
+
+	server_utils.ResetTestState()
+	t.Cleanup(func() {
+		server_utils.ResetTestState()
+	})
+
+	if os.Getenv("RUN_STASHPLUGIN") == "1" {
```
turetske, line 329: Why is this a check? What happens if RUN_STASHPLUGIN == 0? Is this a leaky environment variable? If that's the case, I think it would be better to find where the issue is and fix the leaked environment variable rather than code against it.

cmd/plugin_test.go
```diff
@@ -311,6 +313,179 @@ func TestStashPluginMain(t *testing.T) {
...
+	defer tempObject1.Close()
+
+	tempObject2Content := "temp object 2 content"
+	tempObject2, err := os.Create(filepath.Join(tempUploadDir, "tempObject2"))
+	assert.NoError(t, err, "Error creating temp file")
+	_, err = tempObject2.WriteString(tempObject2Content)
+	assert.NoError(t, err, "Error writing to temp file")
+	defer tempObject2.Close()
+
+	viper.Set("Origin.EnablePublicReads", true)
```
turetske, line 371: Move the "viper.Set" block closer to the beginning of the function. It's good practice to put the test state you are creating at the top, unless you are changing them mid way.

cmd/plugin_test.go
```diff
@@ -311,6 +313,179 @@ func TestStashPluginMain(t *testing.T) {
...
+	token, err := tokenConfig.CreateToken()
+	require.NoError(t, err)
+	tempToken, err := os.CreateTemp(t.TempDir(), "token")
+	require.NoError(t, err, "Error creating temp token file")
+	defer os.Remove(tempToken.Name())
+	_, err = tempToken.WriteString(token)
+	require.NoError(t, err, "Error writing to temp token file")
+	tempToken.Close()
+
+	viper.Set("Logging.DisableProgressBars", true)
```
turetske, line 402: See the above about putting the Viper.Set at the top if possible.

client/fed_linux_test.go
```diff
@@ -835,3 +835,91 @@ func TestSyncDownload(t *testing.T) {
 		require.Equal(t, newTestFileContent, string(contentBytes))
 	})
 }
+
+// This test verifies the behavior when a directory path is passed to the object put command without the recursive option.
```
turetske, line 839: As said in the other file, Use """ for header comments rather than //

turetske (changes requested):

cmd/plugin_test.go
Saartank: The function `stashPluginMain` calls `os.exit` and cannot be tested directly. Therefore, it is executed in its own process. The test itself sets this environment variable while starting itself in a different process, and the variable is set only for the duration of that process.

Below is the link to the snippet where the test runs itself in a separate process with the environment variable set.

https://github.com/PelicanPlatform/pelican/pull/1707/files#diff-7047045de9e6b2f5fc15f5af465c47b95cde0a18745cd7e34ee61faba13267cbR434-R442

cmd/plugin_test.go
```diff
@@ -311,6 +313,179 @@ func TestStashPluginMain(t *testing.T) {
 	assert.Contains(t, output, amountDownloaded)
 }
 
+// This test addresses the issue mentioned in issue #1645. It creates a directory containing two files.
```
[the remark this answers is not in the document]
Saartank: Addressed it in the latest commit: https://github.com/PelicanPlatform/pelican/pull/1707/commits/850b10759cd74b822f05b7d23f1aef043d032760

turetske (approved):

Issue #1635, opened 2024-10-09
`proj` field not exclusively projects


CannonLock:
When I am looking at the reported `xrootd_transfer_bytes` as aggregated by the director the `proj` that is reported does not align with the field description in the docs.

[I expect](https://docs.pelicanplatform.org/monitoring-pelican-services/prometheus#label-proj): 

`Client’s User-Agent header when requesting the file. This is used to label the project name that accesses the file.`

I get:
```
Python/3.12 aiohttp/3.9.5
check_http/v2.2 (monitoring-plugins 2.2)
curl/7.76.1
pelican-client/7.9.9 project/LIGO
...
```

It is not clear to me if I should be doing more parsing or if this field is falsely populated.

==========================================================================

Below I have included the query that revealed that and the `proj` field values as they currently exist. 

Query: `sum by (proj) (increase(xrootd_transfer_bytes{type!="write",proj!=""}[1d]))`

Result: 

```
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "proj": "xrdcl-pelican/1.0"
        },
        "value": [1728503234.765, "35516993256.308914"
        ]
      },
      {
        "metric": {
          "proj": "Python/3.12 aiohttp/3.9.4"
        },
        "value": [1728503234.765, "0"
        ]
      },
      {
        "metric": {
          "proj": "Go-http-client/1.1"
        },
        "value": [1728503234.765, "386735.14599450305"
        ]
      },
      {
        "metric": {
          "proj": "pelican-client/7.10.1"
        },
        "value": [1728503234.765, "0"
        ]
      },
      {
        "metric": {
          "proj": "pelican-client/7.10.1 project/eht"
        },
        "value": [1728503234.765, "1057450997.4816345"
        ]
      },
      {
        "metric": {
          "proj": "pelican-client/7.10.5"
        },
        "value": [1728503234.765, "0"
        ]
      },
      {
        "metric": {
          "proj": "pelican-client/7.10.5 project/xenon"
        },
        "value": [1728503234.765, "0"
        ]
      },
      {
        "metric": {
          "proj": "pelican-client/7.10.7"
        },
        "value": [1728503234.765, "42.00729341908078"
        ]
      },
      {
        "metric": {
          "proj": "pelican-client/7.10.7 project/KOTO"
        },
        "value": [1728503234.765, "0"
        ]
      },
      {
        "metric": {
          "proj": "pelican-client/7.9.9"
[...]

PR #1708, opened 2024-11-04
Fix `proj` label setting bug

patrickbrophy:
This PR addresses issue #1635. This PR moves the project extract logic outside of the director and adds to the `utils` package. This allows us to use the same logic to get the project from the `appinfo` provided by XRootD. The proj label will be empty if there is no project in the user-agent.


jhiemstrawisc (approved):
LTGM

PR #1709, opened 2024-11-04
Update/clarify `Cache.<Data/Meta>Locations` documentation

### LINKED ISSUE #1664: Docs for `Cache.Datalocations`, `Cache.MetaLocations` and `Xrootd.Mount` are confusing and conflicting

jhiemstrawisc:
This PR tries to clarify some of the language surrounding these configuration options and removes mention of `Xrootd.Mount` which has no effect on caches.

Closes #1664

### DIFF
--- a/cache/cache_api.go
+++ b/cache/cache_api.go
@@ -37,7 +37,7 @@ import (
 func CheckCacheSentinelLocation() error {
 	if param.Cache_SentinelLocation.IsSet() {
 		sentinelPath := param.Cache_SentinelLocation.GetString()
-		dataLoc := param.Cache_LocalRoot.GetString()
+		dataLoc := param.Cache_StorageLocation.GetString()
 		sentinelPath = path.Clean(sentinelPath)
 		if path.Base(sentinelPath) != sentinelPath {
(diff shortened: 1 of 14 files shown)


turetske (approved):
Minor comments on some of the comments, but otherwise seems to test correctly with both old and new naming system.

PR #1716, opened 2024-11-05
Unmarshal Capabilities correctly in all structs that use them

### LINKED ISSUE #1715: Incompatibility between 7.11 Director and <7.11 Origins/Caches
After changing a few JSON fields that get passed between servers in PR #1653, an incompatibility started occurring between new 7.11 Directors and older Origins/Caches. The issue comes from the fact that the new unmarshal function in that PR targeted only `NamespaceAdV2 Caps` without considering that other structs like `ServerAd` and `OriginAdvertiseV2` were prone to the same backwards compatibility breaks.

The relevant fix should target the `Capabilities` struct unmarshalling directly instead of creating solutions for each of the parent structs (which is the approach implemented in #1653)

jhiemstrawisc:
Instead of targeting NamespaceAdV2 like I did previously, this change moves the unmarshal logic to the level of the Capabilities struct. Why I didn't do that in the first place? I couldn't tell ya

In addition, I now test that JSON for ServerAd and OriginAdvertiseV2 (which may also arrive to the Director with old/new capabilities) unmarshal correctly.

Closes #1715

### DIFF
--- a/server_structs/director.go
+++ b/server_structs/director.go
@@ -187,48 +187,36 @@ type (
 	}
 )
 
-// A helper function to handle JSON->NSAdV2 unmarshalling across multiple deprecated JSON keys.
+// A helper function to handle JSON->Caps unmarshalling across multiple deprecated JSON keys.
 //
-// When the Director sends a list of NamespaceAdV2 structs as JSON, it may contain
-// the old capabilities struct. This function checks if the raw JSON contains the
-// "FallbackRead" field, indicating the old capabilities struct, and unmarshals
-// the JSON into the new struct accordingly. This is to ensure backwards compatibility
-// We can probably think about removing this function when we don't see origins/directors
+// Old Caches/Origins will send various bits of JSON to the Director containing the JSON representation
+// of the OldCapabilities struct. To make sure these old JSON representations are unmarshalled correctly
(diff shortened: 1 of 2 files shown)


jhiemstrawisc:
@turetske FWIW these new unit tests would have caught the issue we had with other types of structs unmarshalling. The issue I have with creating a regression test that sends various ads to a real director is that their failure to unmarshal is opaque -- if the structs don't line up, the Director doesn't throw any errors, it just doesn't record the right information. Instead, this simulates the actual JSON the Director _would_ receive and makes sure we end up the correct structs regardless of which Capabilities type the JSON encodes.

jhiemstrawisc:
To test, start a Director with these changes. Then start an origin with these changes as well and point at your Director. Check the Director's Web UI to see that the capabilities listed for the origin are correct.

Next you need to point an Origin with version <=7.10.12 at this Director. To do this, leave the Director running, but in a different VSCode tab checkout some older tag for Pelican with `git checkout tags/<version number>`. Luckily rebuilding won't affect the Director as long as it stays running, so when you rebuild a binary for this new tag you can restart your origin. I recommend restarting it with a different set of capabilities so it's obvious to you that the caps have changed and are still correct.

turetske (approved):
LGTM and I was able to run the test as expected.

PR #1729, opened 2024-11-12
Fix caps 7.11

jhiemstrawisc:
I'm hoping this is enough to revert any of the breaking changes introduced during the 7.11 rollout. As much as I'd like to move forward with these changes and fix things to make them backwards compatible, that started becoming more complicated than I'd like for something I consider urgent.


turetske (approved):
Tested and works between 7.11 and 7.10.

PR #1730, opened 2024-11-12
Update macos.mdx to make architectures match

alexandertuna:
In the current readme, `curl` and `tar` refer to different osx architectures. This PR makes them match.


jhiemstrawisc (approved):
Thanks for the cleanup!

Issue #1724, opened 2024-11-11
Web UI at Director displays Origin capabilities but not namespace capabilities


jhiemstrawisc:
When the Director receives an advertisement for an origin with multiple namespaces, it currently only displays the Origin capabilities, and not the capabilities of the namespaces themselves. Consider these two images:

![image](https://github.com/user-attachments/assets/a6f09219-d024-48ca-b2bb-6b9f97aada35)
![image](https://github.com/user-attachments/assets/7e83608c-b4a8-4a20-971f-305173bc9637)

The first shows the _namespace_ capabilities as they are configured at the origin. However, because the _Origin capabilities_ include writes and direct reads, the visual at the Director makes it seem like there's an error in advertising.

I think we need two fixes here -- first, to make it very clear at the Origin which capabilities are a property of the Origin, and which are properties of any namespaces. Second, we should be able to check this same information at the Director.

CannonLock:
@jhiemstrawisc What is the point of the Origins capabilities? Where do those factor in if we are defining them on the namespace?

jhiemstrawisc:
It gets down to how we manage the many-to-many relationship between origins and namespaces. Consider this situation:
```
.
└── Federation/
    ├── Origin1 # supports reads, not writes
    │   ├── /namespace1 # reads
    │   └── /namespace2 # reads, writes
    └── Origin2 # supports writes, not reads
        ├── /namespace2 # reads, writes
        └── /namespace3 # writes
```

Namespaces should be globally-defined (we're slowly working toward that), so any origin that serves a namespace should respect that namespace's capabilities. On the flip side, a particular origin might not want to support a capability, even if one of the namespaces it federates supports it.

When we're trying to pick which origin to send a request to, we do an AND operation on the origin's capabilities and the namespace capabilities -- that is, if you wanted to read from `/namespace2`, you'd be sent to origin1. If you wanted to write to `/namespace2`, you'd be sent to origin2.

CannonLock:
@jhiemstrawisc Cool, thanks for the image, that helps a lot. I will make it happen :)

CannonLock:
@jhiemstrawisc Is there any interaction when namespaces have a hierarchy?

For instance:
```
/major - Write
/major/minor - Read, Write
```

Or are they treated independently of each other?

CannonLock:
Closed by #1737

PR #1737, opened 2024-11-14
Namespace UI

CannonLock:
- Show namespace capabilities
- Show interaction between namespace and server capabilities


CannonLock:
Screenshot 2024-11-14 at 4 10 46 PM (2)

CannonLock:
Closes https://github.com/PelicanPlatform/pelican/issues/1724

CannonLock:
New namespace card with the nested capabilities. Should give easy insight to what caches and servers are open to this namespace. 

Screenshot 2024-11-18 at 3 36 39 PM (2)

h2zh:
> New namespace card with the nested capabilities. Should give easy insight to what caches and origins provide what capabilities for this namespace.
> 
> Screenshot 2024-11-18 at 4 12 43 PM (2)

There are multiple "issuer" showing up here. I think it'd be better to differentiate them here in case causing further confusion. Here are my two cents to avoid semantic overload:

1. Change "issuer" under "token generation" to another word, like "generator". Also change its occurrences everywhere else
2. Add tool-tip to "issuer" and "generator", explaining their difference in Pelican project. For example:
- Token Generator: a server that performs the technical task of creating the token
- Token Issuer: a server as a trusted authority that vouches for the authenticity of the token
 
Could you take a quick look and share your thoughts on the necessity of this change @jhiemstrawisc

h2zh (approved):

Issue #561, opened 2023-12-29
Allow origin issuer key to be rotated


bbockelm:
We currently keep the issuer private key around for the lifetime of the process.  If the key is changed then it should be reloaded and propagated throughout the internal structures.

Instead of having a time-based cache for this, I think it'd be better to have a maintenance goroutine that periodically checks/reloads/updates the internal data structures.

In scope for this issue:
- Rotating out private key from internal memory structures if the admin provides a new one.
- Registering new key with the registry.

Not in scope for this issue:
- Generating a new public JWKS file with both old and new keys.  It's up to the admin to get the public key set correct.
   - In the future, we should track when the last known token for a given key expires and automatically merge in prior public keys used by the origin.
- Delaying signing tokens with the new key until a public key propagation time period has passed.
- Actually generating the new private key itself.

That is, we should assume the origin administrator knows what they're doing when rotating keys.  A follow-up ticket can automate the key rotation process.

PR #1748, opened 2024-11-15
Server-side key refresh and update registered public keys from origin server

h2zh:
### Overview

This PR introduces 4 major features: 

1. Support multiple private key files in a new directory specified by a new config param `IssuerKeysDirectory`, while also incorporating the legacy key saved at `IssuerKey`
2. Periodically refresh above locations to maintain an up-to-date in-memory private keys map `AllKeys`. A token or payload signature is considered valid if any of these keys could have produced it. The key with highest lexicographical filename is set as the current issuer key `CurrentKey` to sign tokens and payload signatures. They live in a struct `IssuerKeys` protected by Go's atomic pointer from race condition.
3. If the server is an origin and keys change after the refresh, the registered public keys update for the corresponding namespace will be triggered. Use the public keys generated from the latest `issuerKeys.AllKeys` to overwrite the previous registered public keys on registry database.
4. Server's public keys are generated from `issuerKeys.AllKeys`, which means multiple public keys could be obtained at `<server_url>/.well-known/issuer.jwks`.

### New Key

In this PR, admin can create new private key by simply dropping the private key into into the new directory as a .pem file (the location where config param `IssuerKeysDirectory` points to, which by default is `/etc/pelican/issuer-keys`). Pelican program will pick up all .pem files (including the legacy one at `IssuerKey` if it exists) through a goroutine running every 5 mins (see details in 1-3 above).

### Security

How do we know the origin who updates the namespace's public key is the same origin registered this namespace? We authorize the origin through handshakes with nonce signatures: 

1. At the start of handshake, client (origin, same below) send nonce to the registry to start a handshakes authentication. We use extra fields in these handshakes to verify the client possess the rights to update the namespace's public key.
2. The registry sends all known public keys for the namespace registered in the database.
3. The client receives the list of public keys and selects a corresponding private key from that list (if one is available) to sign the challenge.
4. The client sends the server the updated set of public keys plus the proof of possession from one of the known keys.
5. The registry verify the signature using the corresponding previous registered public key. If it passes, overwrite the public keys of the registered namespace with the latest set from the client

During the handshakes, all the namespaces managed by the sender origin are sent in a list, and then their corresponding registered public keys get update at the same time. This feature reduces the number of HTTP requests. It builds on the fact that currently, registered public keys are managed "per origin" rather than "per namespace." In other words, all namespaces registered by a single origin share the same set of public keys. (The existing namespace registration process does the same thing too)

### Design thinking

Existing file hierarchy:

```
├── whatever-parent-dir/ (by default /etc/pelican)                             
│   ├── issuer.jwk            # Existing key location, specified by config param “IssuerKey”
```

To ensure backward compatibility, the new directory that stores the private keys (saved as .pem files) is mounted to a new config param `IssuerKeysDirectory`. User doesn’t need to change the value of `IssuerKey`, because we still incorporate the that file in the key refresh, if it exists. The new file hierarchy looks like this:

```
├── whatever-parent-dir/ (by default /etc/pelican)                             
│   ├── issuer.jwk            # (optional) Legacy key location, specified by config param “IssuerKey”
│   └── issuer-keys/          # The new directory storing multiple private keys where “IssuerKeysDirectory” refers to
│       ├── pelican_generated_<timestamp_1>_<randomChars>.pem    # If no private key is provided, new .pem file will be created
│       ├── pelican_generated_<timestamp_2>_<randomChars>.pem         
│       ├── ...                 
│       └── <admin's fav name>.pem
```

Algorithm efficiency tradeoff

For a map using atomic.pointer, it requires to copy the entire map before update any key-value pair. In the `loadPEMFiles` func, because it runs every 5 minutes to update the in-memory key map from the file, which is not frequent, we think the memory use by a map using atomic.pointer is still acceptable.

### Future work

At this moment, we have gone far away from the initial ticket and are pretty close to the finishing line. #561

The original final goal is: when multiple origins have the same namespace, the admin of these origins can append a common private key to all. Then these origin servers will register the corresponding shared public key in the registry. Currently this operation needs to manually done by OSDF admin.

If the goal hasn't changed, it's DONE!😃


config/init_server_creds_test.go
```diff
@@ -150,3 +154,44 @@ func TestLoadPrivateKey(t *testing.T) {
...
+		// because they are named after unix epoch timestamp
+		time.Sleep(1 * time.Second)
+
+		// Create another private key
+		secondKey, err := GeneratePEMandSetActiveKey(issuerKeysDir)
+		require.NoError(t, err)
+		require.NotNil(t, secondKey)
+		assert.NotEqual(t, key.KeyID(), secondKey.KeyID())
+
+		// Check if the active private key points to the lastest key
```
turetske, line 191:
```suggestion
		// Check if the active private key points to the latest key
```

config/init_server_creds.go
```diff
@@ -574,9 +799,26 @@ func loadIssuerPublicJWKS(existingJWKS string, issuerKeyFile string) (jwk.Set, e
 // Return the private JWK for the server to sign tokens
 func GetIssuerPrivateJWK() (jwk.Key, error) {
 	key := issuerPrivateJWK.Load()
-	if key == nil {
-		issuerKeyFile := param.IssuerKey.GetString()
-		newKey, err := loadIssuerPrivateJWK(issuerKeyFile)
+	issuerKeysDir := param.IssuerKeysDirectory.GetString()
+	currentIssuerKeysDir := getCurrentIssuerKeysDir()
+	// Handles runtime changes to the issuer keys directory (configured via "IssuerKeysDirectory" parameter).
```
turetske, line 804: Why would runtime changes occur?

The parameter is assigned/parsed on startup. If the server settings are changed in the web-ui it forces a restart. Can this actually occur?

config/config.go
```diff
@@ -1489,6 +1490,8 @@ func SetClientDefaults(v *viper.Viper) error {
 	configDir := v.GetString("ConfigDir")
 
 	v.SetDefault(param.IssuerKey.GetName(), filepath.Join(configDir, "issuer.jwk"))
+	v.SetDefault(param.IssuerKeysDirectory.GetName(), filepath.Join(configDir, "issuer-keys"))
```
turetske, line 1493: If IssuerKey is being replaced by IssuerKeysDirectory, that has a very specific implication that the former shouldn't be set and if it is, the replacement of the deprecation will be handled by the function on line 606 and line 1491 shouldn't be set.

What happens when they are both set? Does it cause issues? If it can't be nicely replaced, then remove the replaced by part in param.yaml.

origin/origin_ui.go
```diff
@@ -161,10 +161,32 @@ func handleExports(ctx *gin.Context) {
 	ctx.JSON(http.StatusOK, res)
 }
 
+// Create a new private key in the given directory, but without loading it immediately
+// This new key will be detected and loaded later by the ongoing goroutine `LaunchIssuerKeysDirRefresh`,
+// which periodically refreshes the keys in the issuer keys directory
+func createNewIssuerKey(ctx *gin.Context) {
+	issuerKeysDir := param.IssuerKeysDirectory.GetString()
```
turetske, line 168: My on the return still stands, but looking at the original issue, generating new private keys is specifically not in scope got this issue.

turetske (changes requested):
I want to do another run through of the server_creds and the actual checking and updating of the keys in the namespace to ensure security, but I have enough non-trivial changes requested that I'm giving this review now.

Major points:

1.) I'm worried about backwards compatibility. What happens when you run with an old origin and a new registry (what about the reverse)?
2.) Updating the private keys and the periodic origin registration are two different tasks. The periodic updating of private keys doesn't need to do a re-register, we already have that occurring. This is redundant.
3.) This PR removes the checks which determine whether an origin is already registered under a different key. This is important because we want to be able to properly communicate this and stop origin startup if the some other user is trying to use the same namespace prefix. This removal is a bad idea.
4.) This PRs tests checks that things work properly, but unit tests should also ensure that failures occur properly and give the expected error messages.

h2zh:
> 1.) I'm worried about backwards compatibility. What happens when you run with an old origin and a new registry (what about the reverse)?
For private key generation and retrieval, the new and old version will use `param.IssuerKey` and `param.IssuerKeysDirectory` respectively so there's no conflicts. 
In new origin, its json payload to registry will have two new data fields: "client_prev_signature" and "all_pubkeys". If there is an old origin and these two fields are missing, they will be an empty string and nil respectively and handled by registry (skip previous signature check; set pubkey to all_pubkeys). If there is an old registry receiving the json from new origin, these two extra data fields will be ignored, so nothing will happen.

config/init_server_creds.go
h2zh: Currently runtime changes occur happen in two test scenarios:
"invalid-token-sig-key" in `pelican/web_ui/prometheus_test.go`, "diff-secrets-yield-diff-result" in `pelican/config/encrypted_test.go`

config/config.go
h2zh: The reason why I set both `IssuerKey` and `IssuerKeysDirectory` is migrating existing private key file. In the `migratePrivateKey` function, it first checks if there is an key at `IssuerKey`, if so, it will move the key to `IssuerKeysDirectory`. In `docs/parameters.yaml`, I tag `IssuerKey` with 'deprecated' and 'replacedby: IssuerKeysDirectory' to let user know this transition.

h2zh:
This is the security checking workflow when updating the keys. Hope this could facilitate your reviews : ) @turetske 
registry-auth-auth drawio

registry/registry_pubkey_update.go
```diff
@@ -0,0 +1,296 @@
...
+							var prevRawkey interface{}
+							if err := existingKey.Raw(&prevRawkey); err != nil {
+								return false, nil, errors.Wrap(err, "failed to generate raw pubkey from client's previous pubkey")
+							}
+							// Get client's previous signature from payload
+							var prevKeyVerified bool
+							if data.ClientPrevSignature == "" {
+								prevKeyVerified = true
+							} else {
+								clientPrevSignature, err := hex.DecodeString(data.ClientPrevSignature)
```
turetske, line 171: This portion determining clientPrevSignature isn't dependent on the iteration of the existing keys or client keys, so move it outside of the doubly nested loop to keep the loop as inexpensive as possible.

launcher_utils/update_namespace_pubkey.go
```diff
@@ -0,0 +1,118 @@
...
+				}
+				log.Debugln("Private keys directory refreshed successfully. The active (latest) private key is", key.KeyID())
+				log.Debugln("Previous private key is", config.GetPreviousIssuerPrivateJWK().KeyID())
+
+				// Update public key in registry db with the new active private key
+				extUrlStr := param.Server_ExternalWebUrl.GetString()
+				extUrl, _ := url.Parse(extUrlStr)
+				namespace := server_structs.GetOriginNs(extUrl.Host)
+				if err := updateNamespacesPubKey(ctx, []string{namespace}); err != nil {
+					log.Errorf("Error updating the public key of the registered origin namespace %s: %v", namespace, err)
```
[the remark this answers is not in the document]
h2zh: You're right. I add a check here to make sure the `updateNamespacesPubKey` only happens when there is a new active private key.

docs/parameters.yaml
```diff
@@ -64,14 +64,27 @@ components: ["origin", "registry", "director"]
...
 name: IssuerKey
 description: |+
-  A filepath to the file containing a PEM-encoded ecdsa private key which later will be parsed
+  [Deprecated] A filepath to the file containing a PEM-encoded ecdsa private key which later will be parsed
   into a JWK and serves as the private key to sign various JWTs issued by this server.
 
   A public JWK will be derived from this private key and used as the key for token verification.
 type: filename
+deprecated: true
+replacedby: "IssuerKeysDirectory"
```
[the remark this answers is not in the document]
h2zh: I removed both `deprecated` and `replacedby ` fields for IssuerKey, while keep a [deprecated] tag in the description to let user know

docs/parameters.yaml
```diff
@@ -64,13 +64,13 @@ components: ["origin", "registry", "director"]
...
 description: |+
-  [Deprecated] A filepath to the file containing a PEM-encoded ecdsa private key which later will be parsed
+  [Deprecated] Use IssuerKeysDirectory instead. The key file set by this parameter will automatically move to IssuerKeysDirectory.
+  
+  A filepath to the file containing a PEM-encoded ecdsa private key which later will be parsed
   into a JWK and serves as the private key to sign various JWTs issued by this server.
 
   A public JWK will be derived from this private key and used as the key for token verification.
 type: filename
-deprecated: true
```
jhiemstrawisc: I think we still want `-deprecated: true` even if there's no direct replacement -- this being set is what generates a warning for the user. Otherwise they may never know to check the documentation for a deprecation warning.

docs/parameters.yaml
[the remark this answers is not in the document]
turetske: Ah, no, we want to keep `deprecated` just remove `replacedby`

docs/parameters.yaml
h2zh: seems like they need to be removed together, otherwise the Go Generate Check fails.
`panic: Parameter entry 'IssuerKey' is deprecated but missing 'replacedby' key. If there is no replacement, use 'none'`

jhiemstrawisc: Ah, gotcha. Then in this case, I think we want to do what the param error suggests -- use `ReplacedBy: "none"`

jhiemstrawisc:
One other quick comment, now that I think we've settled on an answer -- wherever there's some JSON that gets passed back and forth via API calls, we should stick to camel case:
https://github.com/orgs/PelicanPlatform/discussions/1734#discussioncomment-11621001

For example, the new `RegisteredPrefixUpdate` struct from `registry/registry_pubkey_update.go` uses snake case when it defines JSON tags like `json:"client_nonce"`. Unless we're worried about a specific backwards compatibility issue here, these should be defined like `json:"clientNonce"`

registry/client_commands_test.go
```diff
@@ -66,95 +61,252 @@ func registryMockup(ctx context.Context, t *testing.T, testName string) *httptes
...
 	server_utils.ResetTestState()
-}
 
-func TestRegistryKeyChainingOSDF(t *testing.T) {
 	ctx, cancel, egrp := test_utils.TestContext(context.Background(), t)
-	defer func() { require.NoError(t, egrp.Wait()) }()
-	defer cancel()
+	t.Cleanup(func() {
+		func() { assert.NoError(t, egrp.Wait()) }()
+		cancel()
```
bbockelm, line 304: This is clearly the deadlock and looks like some bad copy/pasting without understanding what the original code does.

The original was:
```
	defer func() { require.NoError(t, egrp.Wait()) }()
	defer cancel()
```

Since `defer` is executed LIFO, this will invoke `cancel()` followed by the `Wait()` (to wait on the canceled functions to clean up).

The revised code is effectively this:

```
         defer func() {
	        func() { require.NoError(t, egrp.Wait()) }()
	        cancel()
          }
```

Which *waits* on the group to stop, then shuts it down, a clear deadlock.

registry/registry_pubkey_update.go
```diff
@@ -0,0 +1,359 @@
...
+				existingNs, err := getNamespaceByPrefix(prefix)
+				if err != nil {
+					log.Errorf("Failed to get existing namespace to update: %v", err)
+					return nil, errors.Wrap(err, "Server encountered an error getting existing namespace to update")
+				}
+
+				// Check the origin is authorized to update (possessing the public key used for prefix initial registration)
+				// Parse all public keys of the sender into a JWKS
+				var clientKeySet jwk.Set
+				if data.AllPubkeys == nil { // backward compatibility - AllPubkeys only exists in the payload in Pelican 7.12 or later
```
bbockelm, line 165: I'm confused by this.

`data` is from the client -- why is there backward compatibility code in an API call introduced in 7.12?

registry/registry_pubkey_update.go
```diff
@@ -0,0 +1,359 @@
...
+					for clientKeysIter.Next(ctx) {
+						clientKey := clientKeysIter.Pair().Value.(jwk.Key)
+
+						clientKid, ok := clientKey.Get("kid")
+						if !ok {
+							log.Warnf("Skipping client key without 'kid'")
+							continue
+						}
+
+						if registryDbKid == clientKid {
```
bbockelm, line 215: This seems unnecessary.

The client should prove it possess one of the existing keys.

However, this *additionally* forces the client to keep the key it used for the challenge in the updated key set.  That seems unnecessary.  We should simply see if the key used to sign is one of the existing known keys.

registry/registry_pubkey_update.go
```diff
@@ -0,0 +1,359 @@
...
+				if data.ClientPrevSignature == "" {
+					prevKeyVerified = true
+				}
+				clientPrevSignature, err := hex.DecodeString(data.ClientPrevSignature)
+				if err != nil {
+					return nil, errors.Wrap(err, "Failed to decode the client's previous signature")
+				}
+
+				// Check if any key in `clientKeySet` matches a key in `registryDbKeySet`
+				registryDbKeysIter := registryDbKeySet.Keys(ctx)
```
bbockelm, line 193: These loops could be replaced by the `LookupKeyID` method.

registry/registry_pubkey_update.go
```diff
@@ -0,0 +1,359 @@
...
+		"client_nonce":     data.ClientNonce,
+		"server_payload":   hex.EncodeToString(serverPayload),
+		"server_signature": hex.EncodeToString(serverSignature),
+	}
+	return res, nil
+}
+
+// Update the public key of registered prefix(es) if the http request passed client and server verification for nonce.
+// It returns the response data, and an error if any
+func updateNsKeySignChallengeCommit(ctx *gin.Context, data *RegisteredPrefixUpdate) (map[string]interface{}, error) {
```
bbockelm, line 109: I think the design here is still a bit confused.  For example, it _removes_ the previous key -- not clear that is matching the intent here.

Let me suggest something cleaner:

1.  At the start of handshake, the registry sends all known public keys for the namespace.
2. The client receives the list of public keys and selects a corresponding private key from that list (if one is available) to sign the challenge.
3. The client sends the server the updated set of public keys plus the proof of possession from one of the known keys.

Thus, at the end of the handshake:
- The client has demonstrated it owns a private key from the existing known public keys.
- The client's desired set of keys is synchronized with the registry; the client can remove or add any arbitrary keys in the call.

registry/client_commands_test.go
```diff
@@ -66,95 +61,252 @@ func registryMockup(ctx context.Context, t *testing.T, testName string) *httptes
...
-	server_utils.ResetTestState()
-
-	svr := registryMockup(ctx, t, "serveregistry")
-	defer func() {
-		err := ShutdownRegistryDB()
-		assert.NoError(t, err)
-		svr.CloseClientConnections()
-		svr.Close()
-	}()
+// func getSortedKids(ctx context.Context, jsonStr string) ([]string, error) {
```
bbockelm, line 64: No commented out tests in the final version!

bbockelm (changes requested):
There's still a lot to do here.  Some minor stylistic / convention comments in the review -- plus a few major requests:

1.  Redo the handshake between origin and registry to allow arbitrary replacement of the public keys.
2. Move the key refresh logic into the config module and add unit tests.  Make the updates atomic and have things in lexicographical order.
3. Do not deprecate the existing issuer key (at least not in this version!) to allow external folks to adapt to the new setup.  Do not delete a key that an admin has provided.  Instead, behave as if the issuer key was found within the key directory.

bbockelm:
Oh!  One final item from the review -- please rebase and squash down the commits to a clean list.

Each commit should be self-contained (compiles, passes tests), covering a single piece of functionality (such as adding the key directory logic; or adding the API to update the keys), and have a detailed commit message.  Please refer to https://osg-htc.org/technology/software/git-software-development/#making-good-pull-requests-the-art-of-good-commits for some best practices.  Given the scope of the changes, I'd expect 4-5 commits to come out on this branch -- not 60.

I typically like keeping the git history in response to reviewer comments to help the reader understand the evolution of ideas.  However, at this point, the history of the PR is so messy that it'd be impossible to glean that knowledge -- more value is in a clean branch.

registry/client_commands_test.go
h2zh: Thanks for pointing out and fixing this lingering problem! Having said that, this is just an intermediate problem in the tests - the root problem is topology mock server doesn't response in `TestRegistryKeyChainingOSDF` and test timed out in `TestRegistryKeyChaining` (see the error logs in the [tests result](https://github.com/PelicanPlatform/pelican/actions/runs/12469465642/job/34804139018)). These two errors were suppressed if the other two tests in client_commands_test.go are commented out. Justin figured out there is a test state leak because commenting out the other tests allowed both TestRegistryKeyChaining and TestRegistryKeyChainingOSDF to pass. I'm still looking for the leak and hoping fixing the problems mentioned in your new PR comments could potentially fix the leak.

h2zh: <google-sheets-html-origin><style type="text/css"></style>
  | Test result | Logs
-- | -- | --
Comment Out ⬇️ | TestRegistryKeyChainingOSDF | TestRegistryKeyChaining |  
TestServeNamespaceRegistry | ❌ | ❌ | https://github.com/PelicanPlatform/pelican/actions/runs/12471626378/job/34808934516
TestMultiPubKeysRegisteredOnNamespace | ❌ | ❌ | https://github.com/PelicanPlatform/pelican/actions/runs/12471246608/job/34807853287
Both | ✔️ | ✔️ | https://github.com/PelicanPlatform/pelican/actions/runs/12447058854/job/34750148518

cmd/generate_keygen.go
```diff
@@ -39,7 +39,7 @@ func keygenMain(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "failed to get the current working directory")
 	}
 	if privateKeyPath == "" {
-		privateKeyPath = filepath.Join(wd, "issuer.jwk")
+		privateKeyPath = filepath.Join(wd, "issuer-keys")
```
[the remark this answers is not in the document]
h2zh: Actually, the private/public generation logic lies in `config.GetIssuerPublicJWKS()`, which was already updated to incorporate the new keys in IssuerKeysDirectory instead of only one key file at IssuerKey

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
...
+
+	// Get the key id of the existing key
+	key, err := loadSinglePEM(legacyPrivateKeyFile)
+	if err != nil {
+		log.Warnf("Failed to load key %s: %v", key.KeyID(), err)
+		return err
+	}
+
+	// Rename the existing private key file and set destination path
+	fileName := fmt.Sprintf("migrated_%d_%s.pem",
```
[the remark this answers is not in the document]
h2zh: Done. I also use the mkstemp logic in another similar function `GeneratePEM`

origin/origin_ui_test.go
```diff
@@ -0,0 +1,99 @@
...
+	viper.Set("IssuerKeysDirectory", filepath.Join(tDir, "test-issuer-keys"))
+	viper.Set("ConfigDir", t.TempDir())
+	config.InitConfig()
+
+	router := gin.Default()
+	router.GET("/api/v1.0/origin_ui/newIssuerKey", createNewIssuerKey)
+
+	// Mock GeneratePEM
+	originalGeneratePEM := GeneratePEM
+	GeneratePEM = func(dir string) (jwk.Key, error) {
```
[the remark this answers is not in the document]
h2zh: I created a new protected API endpoint in origin, where GeneratePEM is called, so that new issuer key could be added through origin Web UI by origin admin. To do that I have to set GeneratePEM as publicly-exported. That API endpoint is behind web_ui.AdminAuthHandler so I think it is safe…? Is there a better way to achieve the same functionality without make GeneratePEM publicly exported?

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
...
-	// Store the key in the in-memory cache
-	issuerPrivateJWK.Store(&key)
+	latestKeys := getIssuerPrivateKeysCopy()
+	for _, file := range files {
+		if filepath.Ext(file.Name()) != ".pem" {
+			continue
+		}
+		// parse the private key in this file and add to the in-memory keys map
+		keyPath := filepath.Join(dir, file.Name())
+		key, err := loadSinglePEM(keyPath)
```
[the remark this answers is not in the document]
h2zh: It seems that one .pem is designed to hold only one single cryptographic key. This file format itself doesn't inherently support multiple keys within a single file.

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
...
+		log.Warnf("Failed to load key %s: %v", keyPath, err)
+	}
+
+	log.Debugf("Generated private key %s", key.KeyID())
 	return key, nil
 }
 
+// Generate a new .pem file and then set the private key it contains as the active one
+// This function is also used in origin API to let admin generate new private key
+func GeneratePEMandSetActiveKey(dir string) (jwk.Key, error) {
```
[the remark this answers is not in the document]
h2zh: For this and the following relevant comments, I remove the origin API that creates new IssuerKey to eliminate the public exposure of GeneratePEM, GeneratePEMandSetActiveKey funcs.

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
 	return nil
 }
 
-// Helper function to load the issuer/server's private key to sign tokens it issues.
-// Only intended to be called internally
-func loadIssuerPrivateJWK(issuerKeyFile string) (jwk.Key, error) {
-	// Check to see if we already had an IssuerKey or generate one
-	if err := GeneratePrivateKey(issuerKeyFile, elliptic.P256(), false); err != nil {
-		return nil, errors.Wrap(err, "Failed to generate new private key")
+// Helper function to move the existing single key file to new directory
```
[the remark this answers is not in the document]
h2zh: Reality distortion field created for legacy key file at the IssuerKey location 🛸

registry/client_commands.go
```diff
@@ -281,3 +282,154 @@ func NamespaceDelete(endpoint string, prefix string) error {
...
+	privateKeyRaw := &ecdsa.PrivateKey{}
+	if err = privateKey.Raw(privateKeyRaw); err != nil {
+		return errors.Wrap(err, "failed to get an ECDSA private key")
+	}
+	signature, err := signPayload([]byte(clientPayload), privateKeyRaw)
+	if err != nil {
+		return errors.Wrap(err, "failed to sign payload")
+	}
+	// Also sign the payload with the previous private key
+	prevPrivateKey := config.GetPreviousIssuerPrivateJWK()
```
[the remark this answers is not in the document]
h2zh: I removed the previous active key field in the communication between origin and registry.

registry/client_commands.go
```diff
@@ -281,3 +282,154 @@ func NamespaceDelete(endpoint string, prefix string) error {
...
+		}
+		if err = allKeysSet.AddKey(pubKey); err != nil {
+			return errors.Wrap(err, "failed to add public key to all keys JWKS")
+		}
+	}
+
+	// Create data for the second POST request
+	unidentifiedPayload := map[string]interface{}{
+		"pubkey":           keySet,
+		"prev_pubkey":      prevKeySet,
```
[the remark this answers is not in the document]
h2zh: I removed the previous active key field in the communication between origin and registry.

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
...
+	keysCopy := getIssuerPrivateKeysCopy()
+	keysCopy[newKey.KeyID()] = newKey
+	issuerPrivateKeys.Store(&keysCopy)
+	SetActiveKey(newKey)
+
+	return newKey, nil
+}
+
+// Re-scan the disk to load the issuer/server's private keys, return the active key to sign tokens it issues
+func LoadIssuerPrivateKey(issuerKeysDir string) (jwk.Key, error) {
```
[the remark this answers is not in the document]
h2zh: I changed this function name to loadIssuerPrivateKey

config/init_server_creds.go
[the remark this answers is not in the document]
h2zh: The sensitive key setting logic is now encapsulated within private functions (generatePEMandSetIssuerKey and setIssuerKey) to prevent direct external access.
In tests, I used a combination of `GeneratePEM(directory)` and `RefreshKeys` to simulate generating a new key and assigning it as the issuer key. This is inspired by the existing public functions `GeneratePrivateKey` and `GetIssuerPrivateJWK`. Both of them don't touch the sensitive key setting logic so I think it's ok to make them public.

config/init_server_creds.go
```diff
@@ -45,16 +47,101 @@ import (
...
 
+func GetPreviousIssuerPrivateJWK() jwk.Key {
+	previousKey := previousIssuerPrivateJWK.Load()
+	if previousKey == nil {
+		return nil
+	}
+	return *previousKey
+}
+
+func UpdatePreviousIssuerPrivateJWK() {
```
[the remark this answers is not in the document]
h2zh: In the new handshakes logic, the concept of previous key is completely removed

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
...
+
+	return key, nil
+}
+
+// Helper function to load/refresh all .pem files from specified directory and find the most recent modified private key
+func loadPEMFiles(dir string) (jwk.Key, error) {
+	var mostRecentKey jwk.Key
+	var mostRecentModTime time.Time
+
+	files, err := os.ReadDir(dir)
```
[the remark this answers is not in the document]
h2zh: Done.

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
...
+			continue
+		}
+		latestKeys[key.KeyID()] = key
+
+		// find the most recent modified file
+		fileInfo, err := file.Info()
+		if err != nil {
+			continue
+		}
+		if fileInfo.ModTime().After(mostRecentModTime) {
```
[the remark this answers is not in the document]
h2zh: The issuerPrivateJWK now represents the private key with the highest lexicographical filename, used for signing tokens and payloads. The issuerPrivateKeys holds all valid private keys, enabling verification of token and payload signatures against one of them. Clarifying comments have been added to the usage of these variables in init_server_creds.go.

registry/client_commands_test.go
h2zh: Update: all failed tests on macOS env are fixed.

registry/registry_pubkey_update.go
h2zh: Done. Also apply this nice function in other places previous using for loop.

registry/registry_pubkey_update.go
h2zh: Done. For now, client's entire on-disk key set is synchronized with the registry at the end the handshakes.

registry/registry_pubkey_update.go
h2zh: To reduce the for-loop and avoid verify the signature with each public key in the registry database, I attach the `MatchedKeyId` in the payload sent back from the client. Then the registry can verify the signature using the right key (if it exists) in one step.
```
// Look for the matched key sent back from the client in `registryDbKeySet`
matchedKey, found := registryDbKeySet.LookupKeyID(data.MatchedKeyId)
if !found {
  return nil, permissionDeniedError{
	  Message: fmt.Sprintf("The client that tries to update prefix '%s' cannot be authorized because it doesn't contain any public key matching the existing namespace's public key in db", prefix),
  }
}
rawkey := &ecdsa.PublicKey{}
if err := matchedKey.Raw(rawkey); err != nil {
  return nil, errors.Wrap(err, "failed to generate the raw key from the matched key")
}
keyUpdateAuthzVerified := verifySignature(clientPayload, keyUpdateAuthzSignature, rawkey)
```

config/init_server_creds_test.go
```diff
@@ -150,3 +154,44 @@ func TestLoadPrivateKey(t *testing.T) {
...
+		tempDir := t.TempDir()
+		issuerKeysDir := filepath.Join(tempDir, "issuer-keys")
+
+		key, err := LoadIssuerPrivateKey(issuerKeysDir)
+		require.NoError(t, err)
+		require.NotNil(t, key)
+
+		// Wait for 1 second to avoid duplicated private key filenames
+		// because they are named after unix epoch timestamp
+		time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
h2zh: Done. I modify the test to make sure it passes without this line of `sleep`

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
...
+	issuerPrivateKeys.Store(&latestKeys)
+	SetActiveKey(mostRecentKey)
+	log.Debugf("Set private key %s as active", mostRecentKey.KeyID())
+
+	return mostRecentKey, nil
+}
+
+// Create a new .pem file (combining GeneratePrivateKey and LoadPrivateKey functions)
+func GeneratePEM(dir string) (jwk.Key, error) {
+	filename := fmt.Sprintf("pelican_generated_%d_%s.pem",
```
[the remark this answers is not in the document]
h2zh: I use go's temp file management `os.CreateTemp` to create unique filename

launcher_utils/update_namespace_pubkey.go
```diff
@@ -0,0 +1,142 @@
...
+
+				originExports, err := server_utils.GetOriginExports()
+				if err != nil {
+					return err
+				}
+				originExportsNs := make([]string, len(originExports))
+				for i, export := range originExports {
+					originExportsNs[i] = export.FederationPrefix
+				}
+				if err := updateNamespacesPubKey(ctx, originExportsNs); err != nil {
```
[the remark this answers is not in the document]
h2zh: Registry public key update is similar to the registry namespace registration, so I put them in the same package. The existing variable `modules` in the launcher is used to identify the origin server.
```
// Check the directory containing .pem files regularly, load new private key(s)
// For origin server, if new file(s) are detected, then register the new public key
func LaunchIssuerKeysDirRefresh(ctx context.Context, egrp *errgroup.Group, modules server_structs.ServerType) {
	egrp.Go(func() error {
		ticker := time.NewTicker(5 * time.Minute)
		defer ticker.Stop()

		for {
			select {
			case <-ctx.Done():
				log.Debugln("Stopping periodic check for private keys directory.")
				return nil
			case <-ticker.C:
				// Refresh the disk to pick up any new private key
				keysChanged, err := config.RefreshKeys()
				if err != nil {
					return nil
				}

				// Update public key registered with namespace in registry db when the private key(s) changed in an origin
				if modules.IsEnabled(server_structs.OriginType) && keysChanged {
					if err = triggerNamespacesPubKeyUpdate(ctx); err != nil {
						return err
					}
				}
			}
		}
	})
}
```

launcher_utils/update_namespace_pubkey.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	if err = registry.NamespacesPubKeyUpdate(key, prefixes, siteName, url); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Check the issuer key directory containing .pem files every 5 minutes, load new private key(s)
+// if new file(s) are detected, then register the new public key
+func LaunchIssuerKeysDirRefresh(ctx context.Context, egrp *errgroup.Group) {
```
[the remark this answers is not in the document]
h2zh: Done. I figured out a straightforward way to trigger the registry update without introducing callback functions or one more go routine. https://github.com/PelicanPlatform/pelican/pull/1748#discussion_r1905756957

config/init_server_creds.go
```diff
@@ -503,44 +590,200 @@ func GenerateCert() error {
...
+	if len(latestKeys) == 0 {
+		newKey, err := GeneratePEMandSetActiveKey(dir)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create a new .pem file to save private key")
+		}
+		return newKey, nil
+	}
 
+	// Save up-to-date private keys and the active key in memory
+	issuerPrivateKeys.Store(&latestKeys)
```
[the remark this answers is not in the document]
h2zh: Done. 
```
type IssuerKeys struct {
	// CurrentKey is the private key used to sign tokens and payloads. It corresponds to the
	// private key with the highest lexicographical filename among the legacy key file
	// (if present) and all .pem files in IssuerKeyDirectory.
	CurrentKey jwk.Key

	// AllKeys holds all valid private keys as a [keyID:key] map, including those from .pem files
	// in IssuerKeyDirectory and legacy key file at IssuerKey (if exists). A token or
	// payload signature is considered valid if any of these keys could have produced it.
	AllKeys map[string]jwk.Key
}
```

launcher_utils/update_namespace_pubkey.go
```diff
@@ -0,0 +1,142 @@
...
+	egrp.Go(func() error {
+		ticker := time.NewTicker(5 * time.Minute)
```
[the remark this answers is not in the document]
h2zh: I refered other periodic Go routines (e.g., LaunchServerIOQuery, LaunchPeriodicAdvertise, RegisterNamespaceWithRetry) and observed that they all use inline fixed time intervals. To maintain a consistent coding style, I've adopted the same approach here. Additionally, I intentionally made this function highly abbreviated to facilitate testing of RefreshKeys() and NamespacesPubKeyUpdate(ctx) in registry/client_commands_test.go.

config/encrypted.go
```diff
@@ -380,17 +377,18 @@ func SaveConfigContents_internal(config *OSDFConfig, forcePassword bool) error {
 // Take a hash, and use the hash's bytes as the secret.
 func GetSecret() (string, error) {
 	// Use issuer private key as the source to generate the secret
-	issuerKeyFile := param.IssuerKey.GetString()
-	err := GeneratePrivateKey(issuerKeyFile, elliptic.P256(), false)
+	privateKey, err := GetIssuerPrivateJWK()
```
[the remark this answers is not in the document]
bbockelm: One thought after some reflection --

The encrypted config is used almost exclusively on the client side for now.  Thus, this becomes a blocker for removing *old* keys but not adding new keys.  If the `config` module had an API to get the oldest key in addition to the current issuer, the issue is avoided until we want to tackle automatic key rotation.

So, one way forward:
- Add this new API to `config` to get the "oldest" signing key.
- Update this call to use the new API.
- File an issue to introduce a stable secret key to be used for encrypting client configuration.  This way, we don't have to do the full solution now.

config/init_server_creds.go
```diff
@@ -44,15 +46,77 @@ import (
...
+	}
+
+	return (*keysPtr).AllKeys
+}
+
+// Helper function to create a directory and set proper permissions to save private keys
+func createDirForKeys(dir string) error {
+	gid, err := GetDaemonGID()
+	if err != nil {
+		return errors.Wrap(err, "failed to get deamon gid")
```
bbockelm, line 114:
```suggestion
		return errors.Wrap(err, "failed to get daemon gid")
```

config/init_server_creds.go
```diff
@@ -503,44 +567,186 @@ func GenerateCert() error {
...
+			return nil, errors.Wrapf(err, "failed to create directory and set permissions: %s", dir)
+		}
+	}
+	// Traverse the directory for .pem files in lexical order
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !d.IsDir() && filepath.Ext(d.Name()) == ".pem" {
```
bbockelm, line 639: Two items:
- I want to avoid single-character variable names.  Humans struggle with understanding them from context.  something like `dirEntry` is much better than `d`.
- Use [`SkipDir`](https://pkg.go.dev/io/fs#SkipDir) to avoid recursing into a subdirectory.

config/init_server_creds.go
```diff
@@ -503,44 +567,186 @@ func GenerateCert() error {
...
+func loadPEMFiles(dir string) (jwk.Key, error) {
+	var mostRecentKey jwk.Key
+	var mostRecentFileName string
+	latestKeys := getIssuerPrivateKeysCopy()
+
+	// Load legacy private key if it exists - parsing the file at IssuerKey act as if it is included in IssuerKeysDirectory
+	issuerKeyPath := param.IssuerKey.GetString()
+	if issuerKeyPath != "" {
+		if _, err := os.Stat(issuerKeyPath); err == nil {
+			issuerKey, err := loadSinglePEM(issuerKeyPath)
```
bbockelm, line 614: Please file a follow-up issue to read *all* PEM objects in a file, not just the first one.

config/init_server_creds.go
```diff
@@ -503,44 +567,186 @@ func GenerateCert() error {
...
+	tempFile, err := os.CreateTemp(dir, filenamePattern)
 	if err != nil {
-		return nil, errors.Wrap(err, "Failed to assign key ID to private key")
+		return nil, errors.Wrap(err, "failed to remove temp file")
+	}
+	keyPath := tempFile.Name()
+	if err := tempFile.Close(); err != nil {
+		return nil, errors.Wrap(err, "failed to close temp file")
+	}
+	if err := os.Remove(keyPath); err != nil {
```
bbockelm, line 700: This introduces a race condition -- another file can be created after the removal but before it's rewritten.

Consider having a helper function take an open file handle (still making sure the mode is correct) and generate the key and refactor the existing `GeneratePrivateKey` to use that.

config/init_server_creds.go
```diff
@@ -700,3 +919,36 @@ func LoadSessionSecret() ([]byte, error) {
 	}
 	return rest, nil
 }
+
+func areKeysDifferent(a, b map[string]jwk.Key) bool {
```
bbockelm, line 923: Please add a documentation string to the function.

config/init_server_creds.go
```diff
@@ -700,3 +919,36 @@ func LoadSessionSecret() ([]byte, error) {
...
+	for key := range b {
+		if _, exists := a[key]; !exists {
+			return true
+		}
+	}
+
+	return false // All keys are the same
+}
+
+func RefreshKeys() (bool, error) {
```
bbockelm, line 943: Please add documentation string to the exported function

bbockelm, line 943: Please add test coverage for this function.

docs/parameters.yaml
```diff
@@ -64,14 +64,29 @@ components: ["origin", "registry", "director"]
 ---
 name: IssuerKey
 description: |+
+  [Deprecated] Use IssuerKeysDirectory instead. The key file set by this parameter will automatically move to IssuerKeysDirectory.
```
bbockelm, line 67: I think this documentation is currently wrong, no?  That's the prior behavior of the patch series.

launcher_utils/key_refresh_and_update_namespace_pubkey.go
```diff
@@ -0,0 +1,141 @@
+/***************************************************************
+ *
+ * Copyright (C) 2024, Pelican Project, Morgridge Institute for Research
```
bbockelm, line 3: It's not 2025.

launchers/launcher.go
```diff
@@ -196,6 +196,13 @@ func LaunchModules(ctx context.Context, modules server_structs.ServerType) (serv
 		lc.Register(ctx, rootGroup)
 	}
 
+	// Start a routine to periodically refresh the private key directory
+	// This ensures that new or updated private keys are automatically loaded and registered
+	if modules.IsEnabled(server_structs.RegistryType) || modules.IsEnabled(server_structs.OriginType) ||
```
bbockelm, line 201: I think we can drop the conditional -- for servers, we always want to keep the private keys refreshed.

launcher_utils/key_refresh_and_update_namespace_pubkey.go
```diff
@@ -0,0 +1,141 @@
...
+	egrp.Go(func() error {
+		ticker := time.NewTicker(5 * time.Minute)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ctx.Done():
+				log.Debugln("Stopping periodic check for private keys directory.")
+				return nil
+			case <-ticker.C:
```
bbockelm, line 125: Switch this routine to using `server_utils.LaunchWatcherMaintenance` instead.  The `server_utils.LaunchWatcherMaintenance` does effectively the same thing *except* it also sets up a watcher on the directory and will get notified immediately about changes.

registry/client_commands.go
```diff
@@ -281,3 +283,165 @@ func NamespaceDelete(endpoint string, prefix string) error {
 	fmt.Println(string(respData))
 	return nil
 }
+
+func NamespacesPubKeyUpdate(privateKey jwk.Key, prefixes []string, siteName string, namespacePubKeyEndUpdatepoint string) error {
```
bbockelm, line 287: Please add documentation string to exported function.

registry/registry_db.go
```diff
@@ -382,7 +382,6 @@ func AddNamespace(ns *server_structs.Namespace) error {
 	if ns.AdminMetadata.Status == "" {
 		ns.AdminMetadata.Status = server_structs.RegPending
 	}
-
```
bbockelm, line 385: In the future, please reduce unnecessary whitespace churn.

registry/registry_pubkey_update.go
```diff
@@ -0,0 +1,270 @@
...
+					Message: fmt.Sprintf("The client that tries to update prefix '%s' cannot be authorized because it fails to pass the proof of possession verification", prefix),
+				}
+			}
+
+			// Process origin's latest public key(s)
+			allPubkeyData, err := json.Marshal(data.AllPubkeys)
+			if err != nil {
+				return nil, errors.Wrapf(err, "Failed to convert the latest public key(s) from json to string format for the prefix %s", prefix)
+			}
+			clientPubkeyString := string(allPubkeyData)
```
bbockelm, line 179: Relying on the string representation seems dicey here.  Do a logical comparison of the public keys, as you do in the `config` module.

bbockelm (changes requested):
Getting quite close here.

I think the biggest issue is something that got dropped along the way -- when is it safe to use the latest key?

Right now, the latest key is used immediately upon discovery -- even though the key takes time to propagate through the system.  A prior attempt kept a copy of the "previous key" and used that for some undefined period (basically, until the timer on the key update loop fired again).  Further, on this review, I noticed that we also are rotating the keys for the encrypted config file logic which also has issues (see inline comments).

I think we will need follow-up work (I'll writeup an issue) to handle when it is safe to use the newest key.  However, in the meantime, I think we can move ahead with this PR with a small change: always use the "oldest" (where "old" is actually based on the filename sorting you currently have) key.  This means that, as new keys get added, what we use for signing stays stable, avoiding both issues above (for now).  That will allow us to get the bulk of the PR in without having to make it even larger.

bbockelm (approved):
I think we're at the "good enough" point that we can get this into a release without making the PR larger.

Follow-up PRs to finish this work include:
- https://github.com/PelicanPlatform/pelican/issues/1898
- https://github.com/PelicanPlatform/pelican/issues/1899
- https://github.com/PelicanPlatform/pelican/issues/1924
- https://github.com/PelicanPlatform/pelican/issues/1925

Issue #1725, opened 2024-11-11
Add version information to the ServerAd structure


turetske:
Pelican Service:
- [X] Origin
- [X] Cache

As part of getting origin versions from the prometheus metrics at the director, the director needs to have some idea of origin versions.

Generally, the way the Director gets its information from the Origin is via the Origin advertising information to the director. So additional information needs to be added to the ServerAd structure. This will also allow us to advertise version information from Cache's as well, since this would be general server information

Short version: Add a field in the ServerAd struct (in `server_structs/director.go`) that reports a version in the form of `v<x>.<y>.<z>` (eg. v7.11.10) and update the Origin and Cache code to populate this information on startup. Look in `launchers/` to see where the startup code is.

PR #1753, opened 2024-11-18
Add Version info to ServerAd

patrickbrophy:
This PR addresses issue #1725. This PR adds a version field to the server ad. The goal is use this version field in the future for monitoring purposes.

### DIFF
--- a/director/director_test.go
+++ b/director/director_test.go
@@ -874,6 +874,153 @@ func TestDirectorRegistration(t *testing.T) {
 		assert.False(t, getAd.DisableDirectorTest)
 		teardown()
 	})
+
+	t.Run("origin-advertise-with-version-and-ua-test", func(t *testing.T) {
+		c, r, w := setupContext()
+		pKey, token, _ := generateToken()
+		publicKey, err := jwk.PublicKeyOf(pKey)
+		assert.NoError(t, err, "Error creating public key from private key")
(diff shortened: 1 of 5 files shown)


turetske (changes requested):
Two things.

1.) If the adV2 that the director gets from the origin/cache doesn't contain a version (so from anything older than this commit), we still want to be able to get the version.  I know I said we'd discuss it later, but I think it's better as part of this PR, in hindsight. So in that case, get the version from the 'User-Agent' string.

2.) Write a unit test to ensure it gets the version properly in either case.

director/director_test.go
```diff
@@ -874,6 +874,153 @@ func TestDirectorRegistration(t *testing.T) {
 		assert.False(t, getAd.DisableDirectorTest)
 		teardown()
 	})
+
+	t.Run("origin-advertise-with-version-and-ua-test", func(t *testing.T) {
```
turetske, line 878: Can you add one more test. One where the header has a version that doesn't match the advertisement's version. And ensure that the version that is in getAd.Version matches the advertisement. Or just change the version in the advertisements you do have to not be the same that's in the headers so that it can be tested just within the following tests.

turetske (changes requested):

turetske (approved):
Looks good to me, I'll merge once the tests have passed.

PR #1756, opened 2024-11-19
Attempt longer ctx deadline in flaky localcache test

jhiemstrawisc:
This has been a flaky test for too long, and I've even seen it start failing locally when I'm running heavy Docker processes in the background. Clearly something just needs a bit more time to do its thing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -132,7 +132,7 @@ jobs:
         # either 'goreleaser' (default) or 'goreleaser-pro'
         distribution: goreleaser
         version: latest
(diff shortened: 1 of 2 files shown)


turetske (approved):

jhiemstrawisc:
Adding a note for posterity -- this was fixed by early commits, but the fix wasn't noticed because of our test infrastructure. When we calculate the previous code coverage, we checkout main and re-run tests. Because the bug existed in main, I was misinterpreting the failure as "my fixes aren't working" instead of "this was broken beforehand".

See issue #1768

PR #1759, opened 2024-11-21
Removed the "network" labels from the metrics

turetske:


jhiemstrawisc (approved):
Seems okay to me. Hopefully this does the trick

PR #1769, opened 2024-11-24
Implement Prohibited Caches for Federation Prefixes

Saartank:
This PR implements a step towards enabling users to configure a list of caches where data should not propagate for a given federation prefix.

## Changes Introduced:

1. **Modify Registry Database**:
   - Added a new column, `prohibited_caches`, in the `namespace` table to store prohibited caches for federation prefixes. This column stores a list of prohibited cache hostnames.

2. **Adjust Registry Endpoints**:
   - Created a new endpoint: `GET /api/v1.0/registry/namespaces/prohibitedCaches` that returns a map of federation prefixes to their corresponding list of prohibited cache hostnames.
   - Modified existing `registry` and `registry_ui` namespace endpoints to include prohibited caches data in their responses.

3. **Add Goroutine in Director**:
   - Added a new goroutine in the director that periodically makes requests to the registry endpoint and maintains the director’s knowledge of prohibited caches in memory.
   - Added a new configuration parameter, `Director.ProhibitedCachesRefreshInterval`, to control the time interval at which this goroutine sends requests to the registry endpoint.

4. **Update Matchmaking Logic**:
   - Updated the director matchmaking logic to respect the configured prohibited caches.


Saartank:
There have been major design changes requested for this task, and it will now be completed as part of a different PR altogether.

PR #1770, opened 2024-11-25
Pulling latest xrootd to see if that fixed the problem

turetske:


bbockelm:
Is this meant to be merged?  If so, we need to tidy up the commits (no commented out lines; that's what history is for) and explain what the problem is instead of being vague.

turetske:
Nope, forgot to make it a draft and delete after testing.

Issue #1717, opened 2024-11-06
OSDF director startup creates warnings for missing `project` during GeoIP resolution


jhiemstrawisc:
The current approach to recording GeoIP resolution metrics has the issue that it creates a lot of scary noise on startup for OSDF Directors. Consider the logs I get when I run things locally:
```   
INFO[2024-11-06T14:16:28Z] Generating/advertising server ads from OSG topology service... 
INFO[2024-11-06T14:16:28Z] The following servers are currently configured in downtime: map[string]director.filterType{"CARDIFF_UK_OSDF_CACHE":"topologyFiltered", "CHTC_STASHCACHE_CACHE":"topologyFiltered", "FZU_STASH":"topologyFiltered", "MGHPCC_NRP_OSDF_ORIGIN":"topologyFiltered", "PDX-Coeus-CE1":"topologyFiltered", "SDSC_NRP_OSDF_ORIGIN":"topologyFiltered", "SDSC_NRP_OSDF_XENON_ORIGIN":"topologyFiltered"} 
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] GeoIP resolution of the address 128.105.69.247 resulted in a suspiciously large accuracy radius of 1000 km. This will be treated as GeoIP resolution failure and result in random server sorting. Setting lat/long to null. 
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] GeoIP resolution of the address 128.105.69.4 resulted in a suspiciously large accuracy radius of 1000 km. This will be treated as GeoIP resolution failure and result in random server sorting. Setting lat/long to null. 
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
WARNING[2024-11-06T14:16:28Z] Failed to get project from context           
[...]

PR #1772, opened 2024-11-25
Fix OSDF Director startup warnings and separate GeoIP metric collection

patrickbrophy:
This PR addresses issue #1717. This PR fixes the unnecessary logging and separates out the GeoIP error metric collection. This is achieved via using a custom error type. This new error type carries along the prometheus labels. When we observe this new error type, we then update the metric with the passed along labels.

### DIFF
--- a/director/sort.go
+++ b/director/sort.go
@@ -61,13 +61,13 @@ type (
 		Coordinate Coordinate `mapstructure:"Coordinate"`
 	}
 
-	GeoIPError struct {
+	geoIPError struct {
 		labels   prometheus.Labels
 		errorMsg string
 	}
 )
 
-func (e GeoIPError) Error() string {
+func (e geoIPError) Error() string {
 	return e.errorMsg
 }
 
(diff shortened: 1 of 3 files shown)


director/sort.go
```diff
@@ -286,7 +296,19 @@ func sortServerAds(ctx context.Context, clientAddr netip.Addr, ads []server_stru
 	weights := make(SwapMaps, len(ads))
 	sortMethod := param.Director_CacheSortMethod.GetString()
 	// This will handle the case where the client address is invalid or the lat/long is not resolvable.
-	clientCoord := getClientLatLong(ctx, clientAddr)
+	clientCoord, err := getClientLatLong(clientAddr)
+	if err != nil {
```
jhiemstrawisc, line 300: Same comment about this section -- error should be printed regardless of error type

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM, and I no longer see the errors pop up when spinning up an OSDF Director. Thanks!

PR #1781, opened 2024-11-27
Force labelling of PRs + connecting to an issue and reopen unlabeled issues

jhiemstrawisc:
This PR comes with two new actions -- one that forces every PR to be connected to at least one issue, as well as forcing at least one label on every PR, and one that re-opens issues if they're closed without a label.

After some consideration, I decided not to work on an action that prevents closing issues without a linked PR because some issues may be closed as `won't-fix`, which inherently means they won't link to anything.

One caveat is that the action for re-opening issues currently requires that the `GITHUB_TOKEN` has write permissions, because it sends a PATCH request through the github api. Not sure how the reviewer will test this ahead of time, since I think it needs to be merged into main. For my own testing, I merged these into my fork's main (which I later reset to track upstream/main again) and then opened+closed an unlabeled issue. I was able to successfully observe that it gets re-opened until a label is assigned.

### DIFF
--- a/.github/workflows/enforce-PR-labelling.yml
+++ b/.github/workflows/enforce-PR-labelling.yml
@@ -1,4 +1,4 @@
-name: PR and Issue Validation
+name: PR Validation
 
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


CannonLock (changes requested):
One Tweak. Other then this it looks good and works on my local as expected.

CannonLock (approved):
LGTM

PR #1782, opened 2024-11-27
Toggle dl webhook on release

CannonLock:
Hits the release webhook toggling dl.pelicanplatform.org to pull down the newest release.

### DIFF
--- a/.github/workflows/post-release.yaml
+++ b/.github/workflows/post-release.yaml
@@ -1,4 +1,5 @@
 # Toggle webhook to pull latest release onto pelicanplatform.org and update the download offerings there
+# Post release this will result in the new release being available and the Major/Minor pointers being moved/created accordingly
 name: post-release
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/post-release.yaml
```diff
@@ -0,0 +1,14 @@
+# Toggle webhook to pull latest release onto dl.pelicanplatform.org
```
jhiemstrawisc, line 1: Can you make this comment a little more verbose and explain what sending a post request to that endpoint does? I tried the curl command locally and can't figure out what's happening other than that it spits out `{"message":"success"}` lol

jhiemstrawisc (changes requested):

.github/workflows/post-release.yaml
CannonLock: How is this?

# Toggle webhook to pull latest release onto pelicanplatform.org and update the download offerings there

Hitting this endpoint toggles dl.pelicanplatform.org to update. Which post release means pulling in the new release and updating the symlinks accordingly.

CannonLock:
```
# Toggle webhook to pull latest release onto pelicanplatform.org and update the download offerings there
# Post release this will result in the new release being available and the Major/Minor pointers being moved/created accordingly
```

jhiemstrawisc (approved):
LGTM

PR #1788, opened 2024-12-03
(Draft) Add Cache metrics page

CannonLock:
( Draft for now until I can get my hands on a running cache to view real data )

- Same as Origin as they have the same metrics
- Centralize shared comps


h2zh (commented):
The code looks good. I’m curious why Origin metrics page has StorageGraph but Cache’s doesn’t? Is it possible to share the layout file across Origin and Cache’s metrics pages?

CannonLock:
@h2zh 

> The code looks good. I’m curious why Origin metrics page has StorageGraph but Cache’s doesn’t? Is it possible to share the layout file across Origin and Cache’s metrics pages?

Good call I have shared the layout. 

I dropped the storage because this is not a metric that populated for a cache. As it is the amount of free space I would expect it to always be 0 when the cache is hot?

h2zh (approved):
LGTM! One nitpicky thing: the commit messages and commit content are mixed up in the last two commits. You may want to squash them.

PR #1791, opened 2024-12-03
Replace os.stat sentinel file check with XRootD http GET

mwestphall:


server_utils/origin.go
```diff
@@ -758,11 +784,29 @@ func CheckOriginSentinelLocations(exports []OriginExport) (ok bool, err error) {
 			if path.Base(sentinelPath) != sentinelPath {
 				return false, errors.Errorf("invalid SentinelLocation path for StoragePrefix %s, file must not contain a directory. Got %s", export.StoragePrefix, export.SentinelLocation)
 			}
-			fullPath := filepath.Join(export.StoragePrefix, sentinelPath)
-			_, err := os.Stat(fullPath)
+
+			fullPath := filepath.Join(export.FederationPrefix, sentinelPath)
+			tkn, err := generateFileTestScitoken(sentinelPath)
```
jhiemstrawisc, line 789: Do you know if it would work to generate a token scoped specifically to the sentinel object instead of a token that could be used to read anything from the namespace?

server_utils/origin.go
```diff
@@ -758,11 +784,29 @@ func CheckOriginSentinelLocations(exports []OriginExport) (ok bool, err error) {
...
+			if err != nil {
+				return false, errors.Wrap(err, "Failed to generate self-auth token for sentinel file check")
```
jhiemstrawisc, line 791: "file" --> "object"

Another note is that I think we want the wrapped error to begin with a lowercase "f". The error message here gets concatenated with other error strings as it bubbles back up through the stack, so the golang convention is to start error strings without capitalization. This comment applies to other error messages in this section, too.

server_utils/origin.go
```diff
@@ -750,6 +753,29 @@ from S3 service URL. In this configuration, objects can be accessed at /federati
...
+	fTestTokenCfg.Issuer = issuerUrl
+	fTestTokenCfg.Subject = "origin"
+	fTestTokenCfg.Claims = map[string]string{"scope": fmt.Sprintf("storage.read:/%v", resourceScope)}
+	// For self-tests, the audience is the server itself
+	fTestTokenCfg.AddAudienceAny()
+
+	// CreateToken also handles validation for us
+	tok, err := fTestTokenCfg.CreateToken()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create file test token")
```
jhiemstrawisc, line 773: Here and elsewhere, can you use the word "object" instead of "file"?

server_utils/origin.go
```diff
@@ -750,6 +753,29 @@ from S3 service URL. In this configuration, objects can be accessed at /federati
 	return originExports, nil
 }
 
+// Generate a test auth token for checking the sentinel location
+func generateFileTestScitoken(resourceScope string) (string, error) {
+	issuerUrl := param.Server_ExternalWebUrl.GetString()
+	if issuerUrl == "" { // if both are empty, then error
+		return "", errors.New("failed to create token: invalid iss, Server_ExternalWebUrl is empty")
```
jhiemstrawisc, line 760: The error message here is a little Mountain Dewey for the average user. How about something like "failed to create a sentinel test token because a required configuration 'Server.ExternalWebUrl" is unset." 

One thing to note here is the formatting of `Server.ExternalWebUrl`, which is our user-facing formatting because it translates more cleanly into the yaml configurations -- the underscore is an internal representation.

server_utils/origin.go
```diff
@@ -758,11 +784,29 @@ func CheckOriginSentinelLocations(exports []OriginExport) (ok bool, err error) {
...
+			fullPath := filepath.Join(export.FederationPrefix, sentinelPath)
+			tkn, err := generateFileTestScitoken(sentinelPath)
+			if err != nil {
+				return false, errors.Wrap(err, "Failed to generate self-auth token for sentinel file check")
+			}
+
+			sentinelUrl := fmt.Sprintf("%v%v", param.Origin_Url.GetString(), fullPath)
+			req, err := http.NewRequest(http.MethodGet, sentinelUrl, nil)
+			if err != nil {
+				return false, errors.Wrap(err, "Failed to create GET request for sentinel file check")
```
jhiemstrawisc, line 797: "file" --> "object"

server_utils/origin.go
```diff
@@ -758,11 +784,29 @@ func CheckOriginSentinelLocations(exports []OriginExport) (ok bool, err error) {
 			if path.Base(sentinelPath) != sentinelPath {
 				return false, errors.Errorf("invalid SentinelLocation path for StoragePrefix %s, file must not contain a directory. Got %s", export.StoragePrefix, export.SentinelLocation)
 			}
-			fullPath := filepath.Join(export.StoragePrefix, sentinelPath)
-			_, err := os.Stat(fullPath)
+
```
jhiemstrawisc, line 787: After testing some of this by hand, I'm realizing we don't have a single log statement saying that this check is being performed. Can you add something along the lines of `log.Infof("Checking that sentinel object %s is accessible for federation prefix %s", sentinelPath, federationPrefix)` to this loop so the user can see that each object is being queried?

server_utils/origin.go
```diff
@@ -758,11 +784,29 @@ func CheckOriginSentinelLocations(exports []OriginExport) (ok bool, err error) {
...
+			req, err := http.NewRequest(http.MethodGet, sentinelUrl, nil)
+			if err != nil {
+				return false, errors.Wrap(err, "Failed to create GET request for sentinel file check")
+			}
+			req.Header.Set("Authorization", "Bearer "+tkn)
+
+			client := http.Client{Transport: config.GetTransport()}
+			resp, err := client.Do(req)
 			if err != nil {
 				return false, errors.Wrapf(err, "fail to open SentinelLocation %s for StoragePrefix %s. Collection check failed", export.SentinelLocation, export.StoragePrefix)
```
jhiemstrawisc, line 804: I know you didn't write this error message, but would you mind changing it to:
"failed to open sentinel object %s for federation prefix %s."

I don't think the "collection check failed" portion of the message actually means anything, since this is an object and not a collection. I also think it makes more sense to speak in terms of the federation prefix and not the storage prefix (but you could probably change my mind on this if you feel differently).

server_utils/origin.go
```diff
@@ -758,11 +784,29 @@ func CheckOriginSentinelLocations(exports []OriginExport) (ok bool, err error) {
...
-			fullPath := filepath.Join(export.StoragePrefix, sentinelPath)
-			_, err := os.Stat(fullPath)
+
+			fullPath := filepath.Join(export.FederationPrefix, sentinelPath)
+			tkn, err := generateFileTestScitoken(sentinelPath)
+			if err != nil {
+				return false, errors.Wrap(err, "Failed to generate self-auth token for sentinel file check")
+			}
+
+			sentinelUrl := fmt.Sprintf("%v%v", param.Origin_Url.GetString(), fullPath)
```
jhiemstrawisc, line 794: I'd be a bit more comfortable if this URL was constructed using the `net/url` package, unless we're certain we won't encounter funkiness with missing  or duplicate `/`s between the joined strings.

jhiemstrawisc (changes requested):

server_utils/origin.go
mwestphall: I believe the scitoken generation code is already specifically scoped to the sentinel object, does this log message look right for that?
```
time="2024-12-05T17:11:11Z" level=info msg="241205 17:11:11 8650 scitokens_Access: New valid token mapped_username=, subject=origin, issuer=https://a6c42d6578d1:33709, authorizations=/test/mock_sentinel:read,dir,stat;/pelican/monitoring/mock_sentinel:read,dir,stat" daemon=xrootd.origin
```

mwestphall:
@jhiemstrawisc left one comment asking for clarification, otherwise implemented the suggested changes

server_utils/origin.go
jhiemstrawisc: Ah, you're right then, I saw that we weren't passing the "full path", but forgot we shouldn't be including the federation prefix in the token scope. Resolved!

server_utils/origin.go
```diff
@@ -750,18 +754,64 @@ from S3 service URL. In this configuration, objects can be accessed at /federati
 	return originExports, nil
 }
 
+// Generate a minimally scoped auth token that allows the origin
+// to query itself for its sentinel file
+func generateSentinelCheckScitoken(resourceScope string) (string, error) {
```
jhiemstrawisc, line 759: One last minor nitpick -- this function technically generates a WLCG token, which isn't the same as a scitoken. How about "generateSentinelReadToken"?

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM. Thanks Matt!

Issue #1520, opened 2024-07-30
Namespace Registration Completion Polling


CannonLock:
Handle the case where a user registers a namespace in the registry but comes back to the origin and sees that the namespace is not yet registered.

PR #1792, opened 2024-12-03
Add a refresh interval to the origin exports

CannonLock:
- UI and UI backing updates to make sure its up to date

The current issue is that when you redirect from a corrected registration in the registry back to the origin you were registering the initial query from the origin is so fast that it misses the changes. 

This adds a UI loading bar that waits 5 seconds for a requery. If for some reason it takes longer then 5 seconds I also auto refresh these metrics every 5 seconds in the background so things will just switch over automatically in the UI.

### DIFF
--- a/web_ui/frontend/components/DataExportTable.tsx
+++ b/web_ui/frontend/components/DataExportTable.tsx
@@ -23,6 +23,7 @@ import { getErrorMessage } from '@/helpers/util';
 import { Capabilities } from '@/types';
 import { CapabilitiesDisplay } from '@/components';
 import CircularProgress from '@mui/material/CircularProgress';
+import { useSearchParams } from 'next/navigation';
 
 type RegistrationStatus =
   | 'Not Supported'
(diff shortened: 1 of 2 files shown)


CannonLock:
Uploading Screen Recording 2024-12-03 at 3.52.09 PM.mov…

CannonLock:
Closes https://github.com/PelicanPlatform/pelican/issues/1520

CannonLock:
@h2zh I didn't think to remove it all the time. I pushed a couple commits to address this. Now it nows when it is being returned to from the registry and will only poll/show the loading bar then.

web_ui/frontend/components/DataExportTable.tsx
```diff
@@ -413,7 +413,7 @@ export const DataExportTable = ({ boxProps }: { boxProps?: BoxProps }) => {
       {from_registry && pending &&
         <Box display={"flex"} flexDirection={"column"}>
           <LinearProgress sx={{mb:1, w: "100%"}} />
-          <Typography variant={'subtitle2'} color={grey[400]} mx={"auto"}>Checking Registry for Updates</Typography>
+          <Typography variant={'subtitle2'} color={grey[400]} mx={"auto"}>Starting Update Watcher; Will Poll Every 10 Seconds;</Typography>
```
h2zh: Maybe mentioning Registry here would be more straightforward?
like `Polling from Registry for Updates Every 10 Seconds`

h2zh (approved):
LGTG

PR #1795, opened 2024-12-04
Redirect Director Metrics if Not Admin

CannonLock:
- Redirect director metrics if not admin
- Hide navigation options if not admin
- Update AuthenicatedContent to consume static lists to enable easier SSR


CannonLock:
https://github.com/user-attachments/assets/afc969ac-517b-468a-9754-9fe9b9c5ad03

Issue #1183, opened 2024-04-26
Add a linter to check typos in non-Go files 


haoming29:
We added a linter to golang-ci to check typos in Go source code #1159, It would be beneficial if we do the same check for files outside of Go, like docs.

There seems to be one actively maintained tool that supports pre-commit integration for typo check: https://github.com/crate-ci/typos

PR #1798, opened 2024-12-05
Run additional linters as pre-commit hooks

brianaydemir:
Despite the original motivation for #1183, it turns out that most of the changes in this PR are to Go files. Also, while I'm here, I decided to add [ShellCheck](https://www.shellcheck.net/) as another linter.

My intent behind the new configuration in `.pre-commit-config.yaml` and `_typos.yaml` is to reduce the amount of code that needs "fixing" in this PR. Developers are always welcome to run `typos` and `shellcheck` on their own, but given that this is an existing codebase, I don't think it's useful to flag every potential issue.

Note that `typos` does _not_ run a spell check but instead has a built-in database of common spelling corrections. As a result, there might still be spelling mistakes in the code base. Furthermore, while I've applied the changes suggested by `typos`, I didn't put much effort into reviewing every modified comment for clarity and grammar.

I did confirm that `typos` seems to be willing to run on the vast majority of relevant files in this repository.

**For reviewers:**

* Be aware that some of the changes in this PR are to executable code, and not just comments. I'm relying on the GO compiler to make sure that executable code has been modified consistently. Shell scripts will require more careful review, e.g., there's some certificate mangling in `images/entrypoint.sh` that I'm guessing has never been used in the wild.


brianaydemir:
@jhiemstrawisc Assigning to you to find a reviewer. (I'll rebase this once 7.12 is out the door.)

images/entrypoint.sh
jhiemstrawisc, line 1: Since this is a critical section of bash scripting that goes beyond testing, can you double check the changes to make sure we think they're okay?

lotman/lotman_test.go
```diff
@@ -267,11 +267,11 @@ func TestUpdateLot(t *testing.T) {
 	require.True(t, success)
 
 	// Update the test-1 lot
-	dedGB := float64(999.0)
+	dedicatedGB := float64(999.0)
```
jhiemstrawisc, line 270: Is there a way to prevent the checker from blocking useful abbreviations like this? I guess it's not a huge issue since most of us tab-complete, but I can see this being annoying in the future if these spell checkers don't let us choose variable names we want because it thinks it knows better...

metrics/xrootd_metrics_test.go
```diff
@@ -358,7 +358,7 @@ func TestHandlePacket(t *testing.T) {
 		expectedLinkConnectIncDup := strings.NewReader(mockPromLinkConnectInc)
 		expectedLinkByteXferIncDup := strings.NewReader(mockPromLinkByteXferInc)
 
-		// First time received a summmary packet
```
jhiemstrawisc, line 361: And here I had assumed this was just an extra tasty packet!

jhiemstrawisc (approved):
Things here look okay to me and I wouldn't mind merging at is, although I had a few comments/questions I'm hoping you can address before we do that.

Changes notwithstanding, once this is rebased it should be okay to merge. Thanks Brian!

lotman/lotman_test.go
brianaydemir: I'm not sure why that specific variable name kept getting flagged despite the following in `_typos.toml`:
```
extend-ignore-identifiers-re = [
    ## Ignore TLAs. Two and three letter acronyms are not worth fixing.
    "^[A-Za-z]{2,3}$",
    ...
```
There were definitely quite a few more names getting flagged before I added that bit of config….

jhiemstrawisc: Since _typos.toml accounts for most of these (and it's unclear why this one case isn't being filtered), I don't think I'm too worried.

images/entrypoint.sh
brianaydemir: I believe we (the royal "we") are the only current users of the built-in OA4MP issuer, and the previous version of that chunk of scripting wouldn't have functioned correctly anyway, as far as I can tell. Honestly, we'll likely find out the first time someone needs OA4MP to query an LDAP server…

Issue #1494, opened 2024-07-18
Group information not correctly escaped (OA4MP)


bbockelm:
When the embedded OA4MP issuer is used, it does not correctly escape the generated group path.

For example, I created the following authorization:

```
   - actions: ["read"]
     prefix: /groupdir/$GROUP
```

and COManage has me in a group named "CO:COU:Topology Contacts:members:active".

The resulting scope was:

```
storage.read:/groupdir/CO:COU:Topology Contacts:members:active
```

Spaces are significant for the scope list -- we should adjust the QDL logic so the resulting scopes are URL-escaped.

PR #1800, opened 2024-12-06
Escape user and group names in issued tokens (#1494)

brianaydemir:
Both the SciTokens and WLCG profiles require that a scope's path components be URL encoded. When substituting for `$USER` and `$GROUP` in any authorization templates, we should ensure that the value being substituted in is URL encoded.

References:

* [QDL function reference](https://qdl-lang.org/pdf/function-reference.pdf)
* [SciTokens profile](https://scitokens.org/technical_docs/Claims)
* [WLCG profile](https://github.com/WLCG-AuthZ-WG/common-jwt-profile/blob/master/profile.md)

### DIFF
--- a/oa4mp/resources/policies.qdl
+++ b/oa4mp/resources/policies.qdl
@@ -35,12 +35,12 @@ scopes := {};
 while [has_value(key, group_list.)]
 [
     group_scopes := { {{- range $idx, $action := .Actions }}{{- if eq $idx 0 -}}'{{- $action -}}:'{{else}}, '{{- $action -}}:'{{- end -}}{{ end -}} } + '{{- .Prefix -}}';
-    scopes := scopes \/ |^replace(~group_scopes, '$GROUP', key);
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
LGTM. I haven't tested (as I'm not sure how to do that in this case), but the `encode()` function is exactly as described in the QDL docs for URL-encoding these values.

brianaydemir:
@jhiemstrawisc I suspect that in practice, most or all "real" values being substituted into `$USER` and `$GROUP` will not contain any characters that need or would get URL-encoded, so I'm not too worried there.

What this will do is ensure that:

* If we, say, substitute in `http://cilogon.org/serverA/users/000000` for `$USER`, we won't accidentally produce a scope with some strange-but-potentially-valid path in the namespace.
* If we, say, substitute in `Topology Contacts` for `$GROUP`, we won't accidentally produce a second scope `Contacts` (which is not valid any profile that we care about).

PR #1819, opened 2024-12-11
Docker images: Put binaries into standard locations

### LINKED ISSUE #1744: Images: don't put pelican binaries into /pelican
We had one site bind-mount their storage dir over `/pelican`, which caused an error at startup because the binaries were no longer there.

We do not prescribe a location for where data should be mounted, and IMO `/pelican` is a perfectly reasonable choice; we should avoid the issue by moving the binaries to some different location.

It is good practice to follow the Filesystem Hierarchy Standard, even in containers.  The `pelican` and `osdf` binaries should be moved into `/usr/local/bin`.

matyasselmeci:
- pelican and osdf go into /usr/local/bin
- pelican-server and osdf-server go into /usr/local/sbin

Fixes #1744

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -195,10 +195,10 @@ COPY --from=xrootd-plugin-builder /usr/lib64/libnlohmann_json_schema_validator.a
 COPY images/entrypoint.sh /entrypoint.sh
 
 # Copy here to reduce dependency on the pelican-build stage in the final-stage and x-base stage
-COPY --from=pelican-build /pelican/dist/pelican_linux_amd64_v1/pelican /pelican/pelican
-COPY --from=pelican-build /pelican/dist/pelican_linux_amd64_v1/pelican /pelican/osdf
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (approved):
LGTM. Thanks for the cleanup, Mat!

bbockelm:
@matyasselmeci / @jhiemstrawisc -- are there any backward compatability items to note here?

jhiemstrawisc:
I don't think there should be. We start containers by passing them a list of arguments, not by invoking any binary name directly. This preserves the formatting, but stores the binary in a separate location, so I'd be surprised if it didn't translate directly to our deployments.

I'll double check that our (minimal) existing documentation doesn't need to be updated to reference the new path.

jhiemstrawisc:
The previous binary paths `/pelican/pelican` and `/pelican/osdf` appear to be undocumented in our [container setup docs](https://docs.pelicanplatform.org/install/docker).

PR #1825, opened 2024-12-12
Rename Server.TLSCertificate to Server.TLSCertificateChain

### LINKED ISSUE #1574: Rename Server.TLSCertificate to Server.TLSCertificateChain
Pelican Service:
- [ ] Client
- [ ] Plugin
- [x] Registry
- [x] Director
- [x] Origin
- [x] Cache

The service certificate specified in `Server.TLSCertificate` needs to be a full certificate chain instead of just a certificate, because clients do not generally have intermediate CAs in their CA certificate bundles.  Therefore the option name `TLSCertificate` is misleading since it implies a single cert.

It would be nice to rename it to point out that you need a cert chain; I suggest `Server.TLSCertificateChain` as the new name.

Saartank:
Fixes #1574

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -1786,6 +1786,20 @@ components: ["registry"]
 ############################
 #   Server-level configs   #
 ############################
+name: Server.TLSCertificate
+description: |+
+  A filepath to a file containing an X.509 host certificate to use for TLS
+  authentication when running server components of Pelican.
+
+  If you override this filepath, you need to provide the matched-pair private key
+  via Server.TLSKey and a Certificate Authority (CA) certificate via Server.TLSCACertificateFile.
+type: filename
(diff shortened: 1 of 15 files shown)


docs/parameters.yaml
```diff
@@ -1786,10 +1786,10 @@ components: ["registry"]
 ############################
 #   Server-level configs   #
 ############################
-name: Server.TLSCertificate
+name: Server.TLSCertificateChain
```
turetske, line 1789: Instead of renaming this, please create a separate parameter and deprecate this one while labeling Server.TLSCertificateChain as the replacement. Then ensure that our deprecated assignment function in the config does the reassignment properly.

turetske (changes requested):

turetske (approved):

PR #1827, opened 2024-12-12
Trigger release actions if tag matches new `-rc.X` prerelease tags

jhiemstrawisc:
Since we now plan to give release candidates tags suffixed with `-rc.X`, where X indicates the release candidate number, we also need to run the release actions for these tags.

The release.yaml action generates artifacts for our github repo and marks things as a pre-release, so we definitely want it to trigger on release candidate tags. The build-prod-container.yaml action creates our containers, and since these are used by integrations to test release candidates, we also need those too. It looks like the triggers for dev container builds will already match against release candidate tags (if I can mentally parse the regex) so we should be fine there.

### DIFF
--- a/.github/workflows/build-prod-container.yml
+++ b/.github/workflows/build-prod-container.yml
@@ -94,7 +94,7 @@ jobs:
         id: latest_version
         run: |
           git fetch --tags
-          tags=$(git tag -l 'v*.*.*' | sort -V)
+          tags=$(git tag -l 'v*.*.*' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V)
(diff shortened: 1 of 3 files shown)


jhiemstrawisc:
@CannonLock and @matyasselmeci, I assigned you both as reviewers because I'm probably missing some other things that need to be updated in Pelican to handle the new tag structure. If you think of anything, can you work on changes and commit them directly to this branch?

matyasselmeci (commented):
I can't make a line comment for this because it's not in the lines you changed, but you should make sure that images for "rc" versions do not get tagged "latest". That should be around line 97 of build-prod-container.yml.

PR #1831, opened 2024-12-16
Enable auto-generation of Lotman policies

jhiemstrawisc:
This gives caches the ability to inform Lotman of known namespaces, which Lotman can use to automatically create a (currently toothless) purge policy.

A lot of the work here involved figuring out how to specify named lot policies through our yaml config. Examples of what I came up with can be found in docs/parameters.yaml. Outside of that, some changes involved figuring out how to merge local config with potentially-discovered config.

Until Pelican actually uses the new purge code with the Lotman purge plugin, none of this really does much other than set up a database. For now, this lets us set up most of the lot config in a cache through Pelican, and then the last bits for testing purge stuff need to be configured manually. Until then, the actual purge ordering portion of a policy isn't plumbed into anything -- eventually it will modify XRootD config directly.

Finally, I still haven't worked out completely how to manage expiration/deletion of lots. I'll save that for another PR, potentially in the next release cycle.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2722,34 +2722,126 @@ type: bool
 default: false
 components: ["cache"]
 ---
-name: Lotman.Lots
+name: Lotman.PolicyDefinitions
 description: |+
-  Declarative configuration for LotMan.  This is a list of objects, each of which describes a "lot". Every lot
-  can be defined with the following:
+  A list of named Lotman purge policy definitions that may be enabled by the cache administrator through setting the `Lotman.EnabledPolicy`
+  configuration. Each policy definition is an object with the following fields:
+  - `PolicyName`: The name of the policy. This is used to identify the policy in the `Lotman.EnabledPolicy` configuration.
+  - `PurgeOrder`: An ordered list of strings indicating the order in which lots should be purged. The strings should be one of the following:
+    - `del`: Purge lots that have passed their deletion time.
+    - `exp`: Purge lots that have passed their expiration time.
+    - `opp`: Purge lots that have passed their opportunistic storage quota.
+    - `ded`: Purge lots that have passed their dedicated storage quota.
+  - `DiscoverPrefixes`: A boolean indicating whether Lotman should automatically discover prefixes from the Director. If true, Lotman will
+    attempt to create lots for all discovered federation prefixes. Locally-defined lots will take precedence over discovered lots if the two have
+    the same name.
+  - `MergeLocalWithDiscovered`: A boolean indicating whether Lotman should merge locally-defined lot configurations with discovered namespaces.
+    Most Lot configuration fields will take precedence from local configuration, but the `Paths` and `Parents` fields are additive.
+  - `DivideUnallocated`: A boolean indicating whether Lotman should attempt to make intelligent decisions regarding management policy attributes
+    for lots that have not provided explicit values. These decisions are based on the cache's total storage capacity and the number of lots
+    that have been explicitly configured, and are intended to maximize potential cache utilization. This should be set to "true" in most cases.
+  - `Lots`: A list of lot objects, each of which describes a "lot". Every lot can be defined with the following:
+    - `LotName`: REQUIRED. The name of the lot.  This is used to identify the lot in the LotMan database.
+    - `Owner`: REQUIRED. A string identifying the owner of the lot's data (as opposed to someone who can modify the lot itself).
+      The Owner field should generally be set to the issue for the lot's namespace path. For example, if the lot
+      tracks namespace `/foo/bar`, the owner might be set to `https://registry.com/api/v1.0/registry/foo/bar`.
+    - `Paths`: OPTIONAL. A list of path objects, each of which describes a path that should be managed by the lot.
+        - `Path`: REQUIRED. The path to be managed by the lot.
+        - `Recursive`: REQUIRED. A boolean indicating whether the path should be managed recursively. If true, the lot will
+          manage all files and directories under the specified path.
+    - `ManagementPolicyAttrs`: REQUIRED. The lot's management policy attributes object. This contains information about resources the lot should
+      be allocated, and how it should be managed.
+        - `DedicatedGB`: REQUIRED. The amount of storage, in GB, that should be dedicated to the lot. This means the lot can assume it
+          always has access to this quantity.
+        - `OpportunisticGB`: REQUIRED. The amount of opportunistic storage, in GB, the lot should have access to, when storage is available.
+        - `MaxNumObjects`: REQUIRED. The maximum number of objects a lot is allowed to store.
+        - `CreationTime`: REQUIRED. A unix timestamp indicating when the lot should begin being considered valid. Times in the future indicate
+          the lot should not be considered valid until that time.
+        - `ExpirationTime`: REQUIRED. A unix timestamp indicating when the lot expires. Lots may continue to function after expiration, but lot
+          data owners should recognize the storage is at-will and may be pre-empted at any time.
+        - `DeletionTime`: REQUIRED. A unix timestamp indicating when the lot and its associated data should be deleted.
 
-  - `LotName`: REQUIRED. The name of the lot.  This is used to identify the lot in the LotMan database.
-  - `Owner`: REQUIRED. A string identifying the owner of the lot's data (as opposed to someone who can modify the lot itself).
-    The Owner field should generally be set to the issue for the lot's namespace path. For example, if the lot
-    tracks namespace `/foo/bar`, the owner might be set to `https://registry.com/api/v1.0/registry/foo/bar`.
-  - `Paths`: OPTIONAL. A list of path objects, each of which describes a path that should be managed by the lot.
-      - `Path`: REQUIRED. The path to be managed by the lot.
-      - `Recursive`: REQUIRED. A boolean indicating whether the path should be managed recursively. If true, the lot will
-        manage all files and directories under the specified path.
-  - `ManagementPolicyAttrs`: REQUIRED. The lot's management policy attributes object. This contains information about resources the lot should
-    be allocated, and how it should be managed.
-      - `DedicatedGB`: REQUIRED. The amount of storage, in GB, that should be dedicated to the lot. This means the lot can assume it
-        always has access to this quantity.
-      - `OpportunisticGB`: REQUIRED. The amount of opportunistic storage, in GB, the lot should have access to, when storage is available.
-      - `MaxNumObjects`: REQUIRED. The maximum number of objects a lot is allowed to store.
-      - `CreationTime`: REQUIRED. A unix timestamp indicating when the lot should begin being considered valid. Times in the future indicate
-        the lot should not be considered valid until that time.
-      - `ExpirationTime`: REQUIRED. A unix timestamp indicating when the lot expires. Lots may continue to function after expiration, but lot
-        data owners should recognize the storage is at-will and may be pre-empted at any time.
-      - `DeletionTime`: REQUIRED. A unix timestamp indicating when the lot and its associated data should be deleted.
+  For example, Lotman could be configured with the "my-policy" policy with the following:
+  ```yaml
+  Lotman:
+    EnabledPolicy: "my-policy"
+    PolicyDefinitions:
+    - PolicyName: "my-policy"
+      DivideUnallocated: true
+      PurgeOrder: ["del", "exp", "opp", "ded"]
+      DiscoverPrefixes: true
+      MergeLocalWithDiscovered: true
+      Lots:
+      - LotName: "/foo/bar"
+        Owner: "https://registry.com/api/v1.0/registry/foo/bar"
+        Paths:
+          Path: "/foo/bar"
+          Recursive: true
+        ManagementPolicyAttrs:
+          DedicatedGB: 100
+          OpportunisticGB: 100
+          MaxNumObjects: 1000
+          CreationTime: 1614556800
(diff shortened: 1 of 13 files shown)


jhiemstrawisc:
For testing this PR, I'd recommend configuring a cache with some lotman policy and a few defined lots, of which at least one should should be an existing namespace. The cache should start up and create the lotman sqlite database, and the tester should be able to verify the values are correctly merged, that a lot is created for each top-level namespace (e.g. for `/chtc` but not for `/chtc/PROTECTED`), and that cache disk space is evenly across lots' `dedicated_GB` field for lots that didn't provide a config override.

jhiemstrawisc:
@CannonLock, this one also requires a ping for you -- there's a new `PolicyDefinitions` object from parameters.yaml that replaces the old `Lots` object.

jhiemstrawisc:
Also note that tests will continue to fail until the Harbor outage is fixed.

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+			}
+			for key, value := range policyLotMap {
+				lotMap[key] = value
+			}
+		}
+	} else {
+		lotMap = policyLotMap
+	}
+
+	cacheDisks := param.Cache_DataLocations.GetStringSlice()
```
turetske, line 825: Is it possible that Cache_DataLocations could somehow be nested folders? Have we ensure that isn't the case? Because otherwise this could potentially double count space.

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+	if err != nil {
+		return errors.Wrap(err, "error converting high watermark to byte value for Lotman")
+	}
+	remainingToHwmB := hwm
+
+    // first iterate through all lots and subtract from our total space any amount
+    // that's already been allocated. Note which lots have an unset value, as we'll
+    // need to return to them.
+    returnToKeys := make([]string, 0, len(*lotMap))
+	for key, lot := range *lotMap {
```
turetske, line 622: During this loop, isn't it possible that remainingToHwmB could become negative? What occurs there? You say you don't ever want to dedicate more space than we have available, but I'm not seeing any check that is occurring here. 

EDIT: Ah, I see the check is later in Validation function.  Either add a comment saying this, as the above is misleading (you specify it's bad but don't clarify it's checked elsewhere) OR take this opportunity to see that the values conflict here (if remainingToHwmB ever gets < 0) and return the appropriate error and exit early.

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+		// now iterate through the lots that need space allocated and assign them the
+		// remaining space
+		spacePerLotRemainingB := remainingToHwmB / uint64(len(returnToKeys))
+		for _, key := range returnToKeys {
+			lot := (*lotMap)[key]
+			if lot.MPA == nil {
+				lot.MPA = &MPA{}
+			}
+			dGb := bytesToGigabytes(spacePerLotRemainingB)
+			oGb := bytesToGigabytes(hwm - spacePerLotRemainingB)
```
turetske, line 648: if in the above case, remainingToHwmB is negative, then spacePerLotRemainingB is negative which would make oGB higher than the watermark in this case.

lotman/lotman_test.go
```diff
@@ -363,3 +394,602 @@ func TestDeleteLotsRec(t *testing.T) {
...
+		"lot4": {
+			LotName: "lot4",
+			MPA: &MPA{
+				OpportunisticGB: &oppGB, // hardcoded values should be respected
+			},
+		},
+	}
+
+	totalDiskSpaceB := uint64(30000000000) // 30GB
+	viper.Set("Cache.HighWaterMark", "25g")
```
turetske, line 804: What happens when the Cache.HighWaterMark is poorly set and lower than the combined total of the dedicatedGB?

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+}
+
+// One limitation in Lotman is that a lot cannot be created unless all of its parents exist. Unfortunately,
+// this means we have to sort our lots topologically to ensure that we create them in the correct order.
+// Failure to do so appears to result in a segfault in Lotman.
+func topoSort(lotMap map[string]Lot) ([]Lot, error) {
+    sorted := make([]Lot, 0, len(lotMap))
+    visited := make(map[string]bool)
+
+	// Recursively visit each lot and its parents, DFS-style
```
turetske, line 735: Something's off with the indentation in this function. Can you fix it?

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+	// Set up lot timestamps (creation, expiration, deletion) if needed
+	configLotTimestamps(&lotMap)
+
+	internalLots, err = topoSort(lotMap)
+	if err != nil {
+		return internalLots, errors.Wrap(err, "error sorting lots prior to instantiation")
+	}
+
+	log.Tracef("Internal lot configuration: %+v", internalLots)
+	err = validateLotsConfig(internalLots, totalDiskSpaceB)
```
turetske, line 899: Is there any reason why this validation of lots config is happening after the topoSort? Since this check is faster O(N) versus O(N+M), it should be run first to avoid wasting time on a more expensive check in case this fails.

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+            result[key] = value
+        }
+    }
+
+    return result, nil
+}
+
+// Grab a map of policy definitions from the config file, where the policy
+// name is the key and its attributes comprise the value.
+func getPolicyMap() (map[string]PurgePolicy, error) {
```
turetske, line 433: Is there any sort of validation check that the values set in Lotman.PolicyDefinitions make sense, conform to the structure defined in parameters.yaml? That just says it's expecting and Object, so couldn't a use put anything there an it would propagate?

It would probably break things later, but would it give a meaningful/understandable error?

turetske (changes requested):
Mostly asking questions and clarification. A few test additions as well as some validation questions.

docs/parameters.yaml
```diff
@@ -2722,34 +2722,126 @@ type: bool
...
+          the lot should not be considered valid until that time.
+        - `ExpirationTime`: REQUIRED. A unix timestamp indicating when the lot expires. Lots may continue to function after expiration, but lot
+          data owners should recognize the storage is at-will and may be pre-empted at any time.
+        - `DeletionTime`: REQUIRED. A unix timestamp indicating when the lot and its associated data should be deleted.
+
+  For example, Lotman could be configured with the "my-policy" policy with the following:
+  ```yaml
+  Lotman:
+    EnabledPolicy: "my-policy"
+    PolicyDefinitions:
```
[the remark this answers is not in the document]
CannonLock: Yup, I will post the fix within the hour.

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+	if _, exists := lotMap["default"]; !exists {
+		lotMap["default"] = Lot{
+			LotName: "default",
+			Owner:   federationIssuer,
+			Parents: []string{"default"},
+			MPA: &MPA{
+				// Set default values to 0 and let potential reallocation happen later.
+				DedicatedGB:     &zero,
+				OpportunisticGB: &zero,
+				MaxNumObjects:   &Int64FromFloat{Value: 0}, // Purge plugin doesn't yet use this, set to 0.
```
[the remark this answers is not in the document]
jhiemstrawisc: I'm a bit fuzzy on why I decided to make the Lot struct take all these pointers (it had something to do with the conversion of JSON in and out of Lotman), but the issue here is that I needed to define the float variable `zero` before I can reference it because Go doesn't let you reference literals. The other fields that take an int64 pointer can be instantiated directly, so they don't need the same setup.

That being said, I've already got the variable floating around there (ha) so I'm not opposed to using it if you'd prefer the consistency.

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+
+    return result, nil
+}
+
+// Grab a map of policy definitions from the config file, where the policy
+// name is the key and its attributes comprise the value.
+func getPolicyMap() (map[string]PurgePolicy, error) {
+	policyMap := make(map[string]PurgePolicy)
+	var policies []PurgePolicy
+	err := viper.UnmarshalKey("Lotman.PolicyDefinitions", &policies)
```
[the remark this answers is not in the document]
jhiemstrawisc: This unmarshals the values directly into the `policies` slice. I didn't go the route of `viper.Get`  because that returns the raw data as an `interface{}`, which would require extra validation and a cast.

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+	policyName := param.Lotman_EnabledPolicy.GetString()
+	if _, exists := policies[policyName]; !exists {
+		return internalLots, errors.Errorf("enabled policy %s is not defined in the configuration", policyName)
+	}
+	policy := policies[policyName]
+
+	discoverPrefixes := policy.DiscoverPrefixes
+	shouldMerge := policy.MergeLocalWithDiscovered
+	if shouldMerge && !discoverPrefixes {
+		return internalLots, errors.New("MergeLocalWithDiscovered is set to true, but DiscoverPrefixes is set to false. This is not a valid configuration")
```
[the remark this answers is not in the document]
jhiemstrawisc: I view it as a more assertive check. True, if nothing is discovered, the merge is a no-op, but such configuration probably means the administrator doesn't really understand what these options do, or doesn't realize they have it set up this way.

I'm not _entirely_ opposed to making this a warning instead of an error, but my experience working with the folks who run our software in production is that none of them bother to look at startup logs unless there's an issue that prevents startup. I still see individuals using `Origin.NamespacePrefix` despite the fact that we deprecated nearly a year ago with a very easy drop-in replacement. 

If you still want to see the warning instead of a fatal error, I'll make the change, but I think this will wind up going unnoticed.

lotman/lotman_linux.go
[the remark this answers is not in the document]
jhiemstrawisc: Scratch that, there's a type error if I do that without yet another cast. I'd like to keep things the way they are for now if that's okay.

lotman/lotman_linux.go
jhiemstrawisc: This would probably produce a non-functioning cache in the first place. I double checked and xrootd doesn't throw an error if started with nested data disks, but the handling of that shouldn't be in this lotman code. I can add a check higher up in the cache.

lotman/lotman_linux.go
jhiemstrawisc: I agree it's a good idea to catch it early. I'll add a check here in addition to referencing the validation that happens later.

lotman/lotman_test.go
jhiemstrawisc: This was a good point, and I'm glad I poked at the code here. Because `hwm` is represented as a uint64, there was an uncaught underflow that caused it to wrap and had unexpected behavior. Thanks for pointing it out!

lotman/lotman_linux.go
jhiemstrawisc: No, you're right. I thought of this too and punted because of time constraints. That'll only hurt me in the long run, so I'll put together some validation code.

lotman/lotman_linux.go
```diff
@@ -314,9 +333,580 @@ func getFederationIssuer() (string, error) {
...
+	mergedLot := lot1
+
+	// Prefer lot1's owner
+    if lot1.Owner == "" {
+        mergedLot.Owner = lot2.Owner
+    }
+
+	// Calculate union between the parents -- if this gets us in trouble by introducing cycles,
+	// lotman will tell us on startup (hopefully...).
+	parentSet := make(map[string]bool)
```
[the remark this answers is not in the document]
jhiemstrawisc: I'm of the opinion that the cache administrator should get the final say in most of the configuration here. You actually made me realize that the current code prefers the discovered namespaces over the locally-configured namespaces (which I think was an oversight, corrected in one of my commits by swapping the input order of `mergeLotMaps`).

While the local admin should have final say, it'd be nice if the we still got some of the configuration from the upstream federation. Lots can only have one proper owner tied to them in the lotman database (probably a mistake at the lotman level), so I can't merge those or I think I probably would. Parents are a list so I don't have to choose between one set of config or another.

One side effect here is that the owners of a lot's parents _can_ be retrieved from lotman, which lets us treat them like proper owners. This means merging parents let us use signed tokens from parent owners to operate on lower level lots. Because we merge paths with the discovered lot namespaces, this guarantees the federation issuer still has some ability to modify lot configuration (because this setup makes every lot a child of the root lot, which is owned by the federation issuer). If this wasn't a merge, the local cache admin could effectively bar the fed administrators from managing lots, which is maybe a good thing, maybe a bad thing. Ultimately we assume a trust relationship between caches and federations, so I thought this was probably the right way to go.

turetske:
@jhiemstrawisc I've looked through your changes and they seem correct, but I need to do another run and look at the sql database

jhiemstrawisc:
Cannon is going to open a followup PR with the new parameter validation code, so I'm merging even though this currently fails that action.

PR #1832, opened 2024-12-16
Cleanup changes made from $HOME fixes from before

### LINKED ISSUE #1647: Refactor/Redo $Home directory fix for the transfer plugin
-- InitConfig went from not being able to fail to calling os.Exit directly. A library should never, ever exit the process. That's the job of the upper level tool. Remove all os.Exit accesses in InitConfig
-- initConfigDir is now public, meaning there's an additional step to setup the configuration. That's indicative of something else gone wrong. We should make this private again and update the tests to call InitConfig correctly
-- There's now been multiple regressions around handling of $HOME in the last year. Let's establish a unit test to get the right behavior in place ($HOME unset, $HOME set but missing, etc) to make sure we understand what we want.

turetske:
Note that this won't address part 3 - making InitConfigDir private - since that function was used outside the module for the config_printer work and the fix isn't trivial.

Also, InitConfigDir and InitConfig only occur on linux machines (configDir being needed for servers and not the client) and in the plugin, so these tests only run on Linux because they deal with hard coded paths.


config/config_test.go
```diff
@@ -180,6 +181,92 @@ func TestInitConfig(t *testing.T) {
...
+		// Save the original environment variables
+		oldConfigRoot := isRootExec
+		resetEnv := setHomeDirEnv(t, mockHomeDir)
+
+		defer func() {
+			resetEnv()
+			isRootExec = oldConfigRoot
+		}()
+
+		t.Run("RootUserNoConfigDir", func(t *testing.T) {
```
jhiemstrawisc, line 222: Since each of these test cases share a very similar structure, I think they can be cleaned up by iterating over a slice of testing cases. Something along the lines of:
```
    type testCase struct {
        name         string
        isRootExec   bool
        configDir    string
        homeEnv      string
        expectedDir  string
    }

    testCases := []testCase{
        {
            name:        "RootUserNoConfigDir",
            isRootExec:  true,
            configDir:   "",
            homeEnv:     "",
            expectedDir: "/etc/pelican",
        },
        {
            name:        "NonRootWithConfigDir",
            isRootExec:  false,
            configDir:   "/test/configDir",
            homeEnv:     "",
            expectedDir: "/test/configDir",
        },
        ... (rest of test cases)
    }

    for _, tc := range testCases {
            isRootExec = tc.isRootExec
            viper.Reset()

            if tc.configDir != "" {
                viper.Set("ConfigDir", tc.configDir)
            }

            if tc.homeEnv != "" {
                os.Setenv("HOME", tc.homeEnv)
            } else {
                os.Unsetenv("HOME")
            }

            InitConfigDir(viper.GetViper())

            cDir := viper.GetString("ConfigDir")
            require.Equal(t, tc.expectedDir, cDir)
        })
    }
}
```

We use this structure elsewhere, and I think it's really useful to highlight what each individual test is checking for.

jhiemstrawisc (changes requested):

config/config.go
```diff
@@ -789,23 +790,19 @@ func SetBaseDefaultsInConfig(v *viper.Viper) {
...
 	configDir := v.GetString("ConfigDir")
 	if configDir == "" {
 		if IsRootExecution() {
 			configDir = "/etc/pelican"
 		} else {
-			configTmp, err := getConfigBase()
-			if err != nil {
-				return err
-			}
+			configTmp := getConfigBase()
```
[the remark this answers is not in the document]
turetske: Because that's how the function was originally written. I'll change it, though.

config/config_test.go
```diff
@@ -180,6 +181,86 @@ func TestInitConfig(t *testing.T) {
 	assert.Equal(t, "", param.Federation_DiscoveryUrl.GetString())
 }
 
+func TestHomeDir(t *testing.T) {
+	if runtime.GOOS != "windows" {
```
jhiemstrawisc, line 185: One final small change that just occurred to me. Instead of wrapping all this in the `if runtime.GOOS != "windows"`, can you do:
```
if runtime.GOOS == "windows" {
    t.Skip("Skipping test on Windows, as it's not expected to work")
}

// Rest of the function
```

This gets rid of one layer of nesting that makes the test a bit harder to read, and also prints a message when the test is run on Windows.

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM. Thanks Emma!

PR #1841, opened 2024-12-18
Move /pelican/monitoring into the "namespace" directory

jhiemstrawisc:
The /pelican/monitoring namespace is where we write various test files that are used to determine whether an XRootD service is currently healthy. After updating our default locations for various cache directories, this was missed, and the monitoring namespace was written to an inaccessible location. This commit corrects the locations.


jhiemstrawisc:
@h2zh, feel free to ignore the failing "Validate Parameters File" test -- that one's not related to this PR

jhiemstrawisc:
The failing ubuntu test for `repeated-stat-access` is one I've seen failing semi-frequently across all PRs, so I don't think that affects this.

PR #1850, opened 2024-12-23
Drop privileges to user `pelican` if requested

bbockelm:
This drafts out an approach to having the pelican server process run unprivileged after initial startup.  Components:

0. This is enabled/disabled by configuration parameters, `Server.DropPrivileges` and currently defaults to false.
1. The configuration now tracks a new user object, the "Pelican" user (defaulting to `pelican`).  This represents the target daemon user the process will be run as once the permissions are dropped.
2. The beginnings of marking the various configuration & runtime directories/files of the service as belonging to the `pelican` user.
3. Infrastructure to signal to the xrootd process to either send itself a signal or make a copy of the file descriptor.
4. Image build infrastructure for the new plugin.  To make the final PR smaller, this item could be pulled out into a separate PR.
5. Finally, actually dropping the privileges to the `pelican` user after launch is almost complete.

TODO items:
- Open a separate issue to handle the RPM-side of the changes.  We've now grown a new dependency on an XRootD plugin and need to get the OSG packaging in a similar condition.
- Actually test the signaling between `pelican` and the new xrootd plugin.  I'm simply showing this as the draft of the approach so it can be handed off to @h2zh.  I expect this is going to need bugfixes.
- More thorough investigation of file ownership.
- Write a unit test that, when run as root, will fork off a separate process and run a "federation in a box" with drop privileges enabled.  Check the functionality of host certificate updates, shutdown signaling, and the ownership of directories is as expected.


images/Dockerfile
```diff
@@ -57,7 +57,7 @@ RUN yum -y update \
 ####
 FROM dependency-build AS xrootd-plugin-builder
 # Install necessary build dependencies
-RUN  yum install -y --enablerepo=osg-testing xrootd-devel xrootd-server-devel xrootd-client-devel curl-devel openssl-devel git cmake3 gcc-c++ sqlite-devel
+RUN  yum install -y --enablerepo=osg-testing xrootd-devel xrootd-server-devel xrootd-client-devel xrootd-private-devel curl-devel openssl-devel git cmake3 gcc-c++ sqlite-devel
```
matyasselmeci, line 60: Not a big fan of depending on a private API. If we absolutely need that class, we should get the developers to make it public, otherwise this is going to break when we least expect it.

xrootd/xrootd_config.go
```diff
@@ -561,32 +590,54 @@ func CopyXrootdCertificates(server server_structs.XRootDServer) error {
...
+		return errors.Wrapf(err, "Failure when moving key pair for xrootd")
 	}
 
-	if _, err = destFile.Write([]byte{'\n', '\n'}); err != nil {
-		return errors.Wrap(err, "Failure when writing into copied key pair for xrootd")
+	return nil
+}
+
+// After privileges have been dropped, copy the server certificates
+// to the xrootd process.
```
h2zh, line 605: Based on the overall logic and contextualization, this might be "to the pelican process"?

images/Dockerfile
bbockelm: Yeah -- I've got similar feelings.  I'll put it on the TODO list.

That said, there's been no changes here in a long, long time and the changes occur far less often than we have to tweak/debug xrootd anyway.

I think the big motivation for the private/public split is to maintain ABI compatibility for EPEL and Debian (who want to hold things stable for multiple years).

xrootd/xrootd_config.go
bbockelm: No -- pelican is sending a file descriptor to xrootd here.

h2zh:
I created a PR https://github.com/bbockelm/pelican/pull/1 to merge the unit test to `bbockelm:drop_privs`

xrootd/drop_privilege_test.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func TestDropPrivilegeSignaling(t *testing.T) {
+	_, cancel, egrp := test_utils.TestContext(context.Background(), t)
+	t.Cleanup(func() {
+		cancel()
+		require.NoError(t, egrp.Wait())
+		server_utils.ResetTestState()
+	})
+
```
Saartank, line 152: Should we do a `server_utils.ResetTestState()` before running the test as well so that we work with a clean viper state.

config/config.go
```diff
@@ -1251,13 +1251,51 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
...
+		pelicanLocationsNoRecursive := []string{
+			param.Shoveler_AMQPTokenLocation.GetString(),
+		}
+		if err = setFileAndDirPerms(pelicanLocationsNoRecursive, 0750, 0640, puser.Uid, 0, false); err != nil {
+			return errors.Wrap(err, "failure when setting up the file permissions for pelican")
+		}
+		pelicanDirs := []string{
+			param.LocalCache_RunLocation.GetString(),
+			param.Lotman_DbLocation.GetString(),
+			param.Lotman_LotHome.GetString(),
```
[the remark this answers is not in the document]
h2zh: For the first question, it doesn't matter because `Lotman_LotHome` is by default set to the same directory as other db locations (under `/var/run/pelican` directory). Adding a is_lotman_enabled condition here won't change anything.

For the second one, I'd appreciate it if @bbockelm could suggest on how to deal with this dilemma.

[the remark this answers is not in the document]
Saartank: Also, I just noticed that the same problem will occur with other parameters as well, since we are currently not checking whether that feature is required or not. For example, if I run the registry, it would create the directory for `Origin.GlobusConfigLocation`.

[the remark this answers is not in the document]
h2zh: You're right. The current approach creates a bunch of directories that are irrelevant to the active running server, or services that are not enabled. I'm thinking of adding a check before adding each entry to pelicanLocations/pelicanDirs. Here is a sample pseudo code:
```
if Cache.EnableLotman is true
     add Lotman_LotHome to pelicanDirs
```
@bbockelm Do you think this would be a good idea?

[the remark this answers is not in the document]
Saartank: Also we must check if the cache server is enabled using `currentServers.IsEnabled(server_structs.CacheType)`

config/init_server_creds.go
```diff
@@ -222,7 +209,7 @@ func GeneratePrivateKey(keyLocation string, curve elliptic.Curve, allowRSA bool)
 	log.Warningf("Will generate a new private key at location: %v", keyLocation)
 
 	keyDir := filepath.Dir(keyLocation)
-	if err := MkdirAll(keyDir, 0750, -1, gid); err != nil {
+	if err := MkdirAll(keyDir, 0750, -1, user.Gid); err != nil {
```
h2zh, line 212: This line is my reference for PR #2062

PR #1856, opened 2024-12-26
Packaging: create pelican user and group and some directories for it

matyasselmeci:
#1860 ; this creates a system user and group named `pelican`; it also pre-creates some directories that Pelican won't be able to create after dropping privs to the `pelican` user.


jhiemstrawisc (approved):
LGTM

PR #1857, opened 2024-12-26
Install xrdcl-pelican from osg-testing instead of building it from source

### LINKED ISSUE #1592: Use `xrdcl-pelican` RPMs in images instead of building source
I think the plan we settled on was to build from source for the dev image (faster ability to test against a specific branch) and to install via RPM in the production image.

matyasselmeci:
Since we're not building it from source, pelican-plugin.conf, which we copy into a later image, is not in /usr/local so that path needed to be fixed too.

Closes #1592

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -64,15 +64,9 @@ RUN  yum install -y --enablerepo=osg-testing xrootd-devel xrootd-server-devel xr
 # the image
 ADD https://api.github.com/repos/PelicanPlatform/xrdcl-pelican/git/refs/heads/main /tmp/hash-xrdcl-pelican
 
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
LGTM.

On a side note, I can't get the PR validation test to pass because the issue was linked with the "Closes #" key in a comment instead of linked through the "Development" section of the Github UI -- I haven't found a way to detect this kind of linking through the Github PR history API, and it prevents any other kind of linking. That being said, using "Closes" is enough to keep our metrics generation happy, so this test failure is really just noise in this case. There are some obvious improvements to be had with these new actions...

Issue #1870, opened 2025-01-08
Deprecation Handling Fails with Default Values from defaults.yaml


Saartank:
If there is a parameter `NewParameter` which replaces the now-deprecated `OldParameter` and the default value of `NewParameter` is coming from `defaults.yaml`, then our deprecation handling logic won't work in this scenario. That is, if someone only sets `OldParameter`, the program will not propagate the value to `NewParameter`.

This is because we use the function `viper.MergeConfig` to load values from default YAMLs. These values, however, are not treated as defaults but rather explicitly set.

PR #1871, opened 2025-01-08
Correct deprecated handling logic

Saartank:
This PR corrects the handling of deprecated configuration parameters.

Before this PR, we handled the deprecation of parameters in the `InitConfig` function by overriding the default of the new parameter with the value of the old parameter if it was set. However, with that setup, the default value was being overridden in the `SetClientDefaults` and `SetServerDefaults` functions when the client and server were initialized.

With this PR, the deprecation handling logic is moved to the end of the `SetClientDefaults` and `SetServerDefaults` functions, after the defaults for the client and server are set, ensuring they are not overridden.

Also, even with these changes in place, there is an issue where any configuration parameter that replaces a deprecated parameter and whose value comes from `defaults.yaml` causes the deprecation handling logic to break. This is described in issue #1870.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -615,21 +615,9 @@ func handleDeprecatedConfig() {
 			} else {
 				for _, rep := range replacement {
 					log.Warningf("The configuration key '%s' is deprecated. Please use '%s' instead.", deprecated, rep)
-					// value := viper.Get(deprecated)
-					// viper.SetDefault(rep, value)
+					value := viper.Get(deprecated)
+					viper.SetDefault(rep, value)
 
-					// Retrieve current values
-					currentValue := viper.Get(deprecated)
-					previousValue := viper.Get(rep)
-
-					// Set the new value
-					viper.SetDefault(rep, currentValue)
-
-					// Log the event
-					log.Warningf(
-						"Transferred value from deprecated key '%s' to new key '%s'. Deprecated value: '%v'. Previous value of new key: '%v'.",
-						deprecated, rep, currentValue, previousValue,
-					)
 				}
 			}
 		}
--- a/config/config.go
+++ b/config/config.go
@@ -610,22 +610,26 @@ func handleDeprecatedConfig() {
 	deprecatedMap := param.GetDeprecated()
 	for deprecated, replacement := range deprecatedMap {
 		if viper.IsSet(deprecated) {
-			if len(replacement) == 1 {
-				if replacement[0] == "none" {
-					log.Warningf("Deprecated configuration key %s is set. This is being removed in future release", deprecated)
-				} else {
-					log.Warningf("Deprecated configuration key %s is set. Please migrate to use %s instead", deprecated, replacement[0])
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -1043,13 +1032,7 @@ func SetServerDefaults(v *viper.Viper) error {
...
 		v.SetDefault(param.Origin_RunLocation.GetName(), filepath.Join("/run", "pelican", "xrootd", "origin"))
 		v.SetDefault(param.Cache_RunLocation.GetName(), filepath.Join("/run", "pelican", "xrootd", "cache"))
-
-		// Several deprecated keys point to Cache.StorageLocation, and by the time we reach this section of code, we should
-		// have already mapped those keys in handleDeprecatedConfig(). To prevent overriding potentially-mapped deprecated keys,
-		// we only re-set he default here if this key is not set.
-		if !v.IsSet(param.Cache_StorageLocation.GetName()) {
-			v.SetDefault(param.Cache_StorageLocation.GetName(), filepath.Join("/run", "pelican", "cache"))
-		}
+		v.SetDefault(param.Cache_StorageLocation.GetName(), filepath.Join("/run", "pelican", "cache"))
```
jhiemstrawisc, line 1035: Hmm, there's a bigger issue here as well, and this section of code has now broken. The issue is that the parameters `Cache.NamespaceLocation`, `Cache.DataLocations` and `Cache.MetaLocations` are supposed to be scoped under `Cache.StorageLocation`, if the value is set.

Moving the `handleDeprecatedConfig` to run after all these defaults are set breaks this mapping, because these secondary values inherit the default `Cache.StorageLocation` and are unaware that `Cache.DataLocation` is being used as the value (because that hasn't happened yet).

I think this means we need to re-evaluate the approach and that we should also have a test for this inheritance to make sure it doesn't break.

config/config.go
```diff
@@ -610,22 +610,14 @@ func handleDeprecatedConfig() {
...
+			if replacement[0] == "none" {
+				log.Warningf("Deprecated configuration key %s is set. This is being removed in future release", deprecated)
 			} else {
-				log.Warningf("Deprecated configuration key %s is set. This is being replaced by %s instead", deprecated, replacement)
-				log.Warningf("Setting default values of '%s' to the value of %s.", replacement, deprecated)
-
-				value := viper.Get(deprecated)
 				for _, rep := range replacement {
+					log.Warningf("The configuration key '%s' is deprecated. Please use '%s' instead.", deprecated, rep)
+					value := viper.Get(deprecated)
```
jhiemstrawisc, line 618: I think I'd still like this log message to mention the expected behavior where we attempt to map deprecated keys to to their replaced values. Can you restore the second warning or lump it into a single warning?

jhiemstrawisc (changes requested):

Saartank:
Marking this PR as closed since the work has been addressed in PR #1872.

Saartank:
Marking this PR as closed since the work has been addressed in PR #1872.

PR #1872, opened 2025-01-09
Fix logic for handling deprecated configuration parameters

Saartank:
This PR fixes the logic of handling for deprecated parameters. Before this PR, we were setting the values from the deprecated config parameters as the default for the values of the new config parameters. However, these defaults from deprecated config params were overwritten when the server or client were initialized.

This PR adds changes so that rather than setting the value from the deprecated param as `viper.SetDefault` on the new config parameter, it uses `viper.Set` if the new config param is not set.

There still remains the issue that if the default of the new config param comes from `defaults.yaml`, the deprecation handling logic would fail. Another issue #1870 is created to address this.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -610,22 +610,17 @@ func handleDeprecatedConfig() {
 	deprecatedMap := param.GetDeprecated()
 	for deprecated, replacement := range deprecatedMap {
 		if viper.IsSet(deprecated) {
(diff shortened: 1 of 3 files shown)


jhiemstrawisc (approved):
LGTM. Thanks for all the work piecing this together :)

PR #1877, opened 2025-01-10
Use alma9 container for building prod pelican-server binaries

jhiemstrawisc:
Prior to the introduction of the pelican-server binary, pelican was always statically-linked. However, pelican-sever needs to be dynamic so it can access the lotman shared library.

This caused an issue in our production cache containers because the pelican-server binary was built in an alpine container (inherited from the old goreleaser container) which uses libmusl for linking, but then copied into an Alma container in the final stage, where libmusl is unavailable.

In this fix, the base container used to build pelican binaries is switched from the goreleaser container to a raw alma9 container. A side effect of this is that we now have to install a few things we previously didn't. Another thing I can't readily explain is that making this change appears to have altered the path these binaries are built under from
`/pelican/dist/pelican_linux_amd64_v1/pelican`
to
`/pelican/dist/linux_amd64/pelican_linux_amd64_v1/pelican`

This doesn't appear to be an issue and was accounted for by changing a few paths in the Dockerfile. Famous last words?

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -20,19 +20,47 @@ ARG BASE_YUM_REPO=release
 ARG BASE_OSG_SERIES=23
 ARG BASE_OS=el9
 
-FROM goreleaser/goreleaser:v1.21.0 AS pelican-build
+###########################
+# Pelican Build container #
+###########################
+# Building Pelican requires go, goreleaser, and packages for the web build
+FROM almalinux:9 AS pelican-build
 ARG IS_NONRELEASE_BUILD="true"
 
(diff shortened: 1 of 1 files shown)


jhiemstrawisc:
I think it's safe to ignore the linter test here. It's complaining about typos in frontend code comments -- I didn't touch these, and I'd rather not correct them here.

williamnswanson (approved):
I tested a osdf-cache built with these changes and I was able to successfully run the server (which I couldn't do with the previous goreleaser-built image due to dynamic-linking issues).

Issue #1311, opened 2024-05-20
Safety net for director Prometheus instance


haoming29:
We experienced a Prometheus cardinality spike recently as our `gin` exporter has a `path` label for requests it serves, causing half a million time series in less than a week. Consequently, the director memory went up til 5GiB. 

Since the director handles all object access requests and it scrapes all other servers, it's expected to have large cardinality for various metric labels. Although a fix is in effect to reduce known metric cardinality in #1276, to prevent Prometheus process accidentally explodes the director, we should add various limits to Prometheus to control the max memory/time series/labels it can use.

There's a good reference of what knobs to turn from Cloudflare blog https://blog.cloudflare.com/how-cloudflare-runs-prometheus-at-scale and we should adapt them accordingly

PR #1887, opened 2025-01-13
Add configuration for Prometheus to limit cardinality

patrickbrophy:
This PR addresses issue #1311. This PR uses the default values as described in [this design doc](https://docs.google.com/document/d/1sVJiZWKj8g3k7t_cy07aA_P7Es3Cg39mklrDKe_xo3c/edit?usp=sharing). This PR adds the ability to configure the label name and value length limits, the number of labels per time series, and the scrape limit. More information about these parameters can be found [here](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)

### DIFF
--- a/config/resources/defaults.yaml
+++ b/config/resources/defaults.yaml
@@ -105,7 +105,7 @@ Monitoring:
   PromQLAuthorization: true
   AggregatePrefixes: ["/*"]
   DataRetention: 360h
-  LabelLimt: 64
+  LabelLimit: 64
   LabelNameLengthLimit: 128
   LabelValueLengthLimit: 2048
   SampleLimit: 200
--- a/config/resources/defaults.yaml
+++ b/config/resources/defaults.yaml
@@ -105,6 +105,10 @@ Monitoring:
   PromQLAuthorization: true
   AggregatePrefixes: ["/*"]
   DataRetention: 360h
+  LabelLimt: 64
+  LabelNameLengthLimit: 128
+  LabelValueLengthLimit: 2048
+  SampleLimit: 200
 Shoveler:
   MessageQueueProtocol: amqp
   PortLower: 9930
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2496,6 +2496,36 @@ type: duration
 default: 360h
 components: ["origin", "cache", "director", "registry"]
 ---
+name: Monitoring.LabelLimit
+description: |+
+  The maximum number of labels that can be attached to a single metric. 0 means no limit.
+type: int
+default: 64
+components: ["origin", "cache", "director", "registry"]
+---
+name: Monitoring.LabelNameLengthLimit
+description: |+
+  The maximum length of a label name. 0 means no limit. The default value is 128 bytes, which allows for up to 32 characters.
+type: int
+default: 128
+components: ["origin", "cache", "director", "registry"]
+---
+name: Monitoring.LabelValueLengthLimit
+description: |+
+  The maximum length of a label value. 0 means no limit. The default value is 2048 bytes, which allows for up to 512 characters.
+type: int
+default: 2048
+components: ["origin", "cache", "director", "registry"]
+---
+name: Monitoring.SampleLimit
+description: |+
+  Per-scrape limit on the number of scraped samples that will be accepted.
+  If more than this number of samples are present after metric relabeling
+  the entire scrape will be treated as failed. 0 means no limit.
+type: int
+default: 200
+components: ["origin", "cache", "director", "registry"]
+---
 ############################
 #   Shoveler-level configs   #
 ############################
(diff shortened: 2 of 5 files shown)


docs/parameters.yaml
```diff
@@ -2496,6 +2496,36 @@ type: duration
 default: 360h
 components: ["origin", "cache", "director", "registry"]
 ---
+name: Monitoring.LabelLimit
+description: |+
+  The maximum number of labels that can be attached to a single metric. 0 means no limit.
+type: int
+default: 64
```
jhiemstrawisc, line 2503: These defaults don't get set automatically by adding them here. You'll want to add them all to `config/resources/defaults.yaml` or in `config/config.go::InitServerDefaults()`. I think the defaults.yaml is probably the better place in this case because the defaults are simple ints and don't depend on the detected runtime environment.

docs/parameters.yaml
```diff
@@ -2496,6 +2496,36 @@ type: duration
...
+components: ["origin", "cache", "director", "registry"]
+---
+name: Monitoring.LabelNameLengthLimit
+description: |+
+  The maximum length of a label name. 0 means no limit. The default value is 128 bytes, which allows for up to 32 characters.
```
jhiemstrawisc, line 2508: It'd be helpful to include a quick example of what a single "metric" looks like, and how the "label" and "value" keys work in it. Maybe a triple-backtick formatted JSON would work to show that here?

It'd also be useful to provide some insight about the consequences of picking a "bad" value here. You can say something about how these numbers may directly affect the memory consumption of our services, and caution the user from overriding the defaults.

web_ui/prometheus.go
```diff
@@ -228,6 +228,13 @@ func configDirectorPromScraper(ctx context.Context) (*config.ScrapeConfig, error
 		RefreshInterval:  model.Duration(15 * time.Second),
 		HTTPClientConfig: sdHttpClientConfig,
 	}
+
+	// Cardinality limits for Prometheus, configured via Pelican configuration file
+	scrapeConfig.LabelLimit = uint(param.Monitoring_LabelLimit.GetInt())
```
jhiemstrawisc, line 233: For all of these parameters, what happens if the user sets them to a negative or huge value? Can you do some input validation before passing them to Prometheus?

web_ui/prometheus.go
```diff
@@ -396,6 +403,13 @@ func ConfigureEmbeddedPrometheus(ctx context.Context, engine *gin.Engine) error
 			}},
 		},
 	}
+
+	// Cardinality limits for Prometheus, configured via Pelican configuration file
+	scrapeConfig.LabelLimit = uint(param.Monitoring_LabelLimit.GetInt())
```
jhiemstrawisc, line 408: I haven't spent much time with this code in general. Why does it become necessary to set these scrape limits in three different places?

jhiemstrawisc (changes requested):

web_ui/prometheus.go
patrickbrophy: I had a hard time wrapping my head around it as well, but I figured if we set it to often it too often it wouldn't be an issue.

web_ui/prometheus.go
```diff
@@ -84,6 +84,28 @@ func init() {
...
 
+const (
+	MaxLabelLimit            = 10
+	MaxLabelNameLengthLimit  = 256
+	MaxLabelValueLengthLimit = 4096
+	MaxSampleLimit           = 800
+)
+
+func setCardinalityLimits(cfg *config.ScrapeConfig) {
+	if param.Monitoring_LabelLimit.GetInt() > 0 && param.Monitoring_LabelLimit.GetInt() < MaxLabelLimit {
```
jhiemstrawisc, line 95: Instead of failing to set values when user provided configuration is bad, we should
a) fall back to the default and use that to configure the limits
b) log a warning for the user that their provided value of %d is bad and that we're falling back to use the default of %d.

jhiemstrawisc (changes requested):
One small touchup, and then I think we're good to go (pending some failing tests).

web_ui/prometheus.go
```diff
@@ -19,6 +19,7 @@ package web_ui
 import (
 	"context"
 	"fmt"
+	pelican_log "log"
```
jhiemstrawisc, line 22: We already import `"github.com/sirupsen/logrus"`, which is the logger we use throughout Pelican, so let's use that instead. You should do:
```
log "github.com/sirupsen/logrus"

...

log.Warningf("Here's some message that says %s", <your specifier>)
```

jhiemstrawisc (changes requested):

config/resources/defaults.yaml
```diff
@@ -105,6 +105,10 @@ Monitoring:
   PromQLAuthorization: true
   AggregatePrefixes: ["/*"]
   DataRetention: 360h
+  LabelLimt: 64
```
jhiemstrawisc, line 108: I think the failing tests are from this typo:
`limt` should `limit`

jhiemstrawisc (approved):

PR #1890, opened 2025-01-13
Retry client requests to Director if we detect the Director is rebooting

jhiemstrawisc:
This PR tries to improve our clients' behavior in the event that the Director is being rebooted, or was recently rebooted and has not yet repopulated its cache of server ads.

The primary issue we're trying to work around is that our production OSDF Director lives behind a Traefik ingress proxy. When the Director is rebooting, the ingress proxy still responds to Director requests, but almost always with a 404, 500, or 502. If the client gets an error response and but isn't actually talking to the Director, it should follow a backoff retry strategy to give the Director time to finish rebooting.

On the other hand, the Director may have recently rebooted, but hasn't had time to receive all the server advertisements it needs to issue the correct redirect. In this case, the Director detects that it recently rebooted and sends an HTTP 429 (too many requests) to the client. The client detects this and starts retrying while it waits for the Director to receive the needed server ads.

The core change here is that the Director now sends a [`Server`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server) header to clients, which lets clients determine who they're talking to and what they should do in the event they encounter certain errors.

This is mostly built on top of Brian's draft PR #1565, and this PR should replace that one.


jhiemstrawisc:
Finally, tests all passing. This one should be ready for a review!

turetske (approved):
Mostly approved with some comment request changes.

PR #1891, opened 2025-01-13
K8SPhysicalHostName Error Message

turetske:
If `K8SPhysicalHostName` set in the machine ad, have the plugin output "Hostname: <name>" in the error message.

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -745,9 +745,12 @@ func writeTransferErrorMessage(currentError string, transferUrl string) (errMsg 
 
 	errMsg += (" (Version: " + config.GetVersion())
 
-	siteName := parseMachineAd()
+	siteName, hostName := parseMachineAd()
 	if siteName != "" {
-		errMsg += "; Site: " + siteName + ")"
+		errMsg += "; Site: " + siteName
+	}
+	if hostName != "" {
+		errMsg += "; Hostname: " + hostName + ")"
 	} else {
 		errMsg += ")"
 	}
@@ -757,13 +760,13 @@ func writeTransferErrorMessage(currentError string, transferUrl string) (errMsg 
 
 // This function parses the machine ad present with a condor job to get the site name.
 // Only really needed on the ospool, otherwise this will return ""
-func parseMachineAd() string {
+func parseMachineAd() (string, string) {
 	var filename string
 	//Parse the .job.ad file for the Owner (username) and ProjectName of the callee.
 	if _, err := os.Stat(".machine.ad"); err == nil {
 		filename = ".machine.ad"
 	} else {
-		return ""
+		return "", ""
 	}
 
 	// https://stackoverflow.com/questions/28574609/how-to-apply-regexp-to-content-in-file-go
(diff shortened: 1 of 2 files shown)


cmd/plugin.go
```diff
@@ -789,11 +793,32 @@ func parseMachineAd() string {
...
+				siteMatchValue = strings.TrimSpace(matchParts[1])
+				siteMatchValue = strings.Trim(siteMatchValue, "\"") //trim any "" around the match if present
+				break
+			}
+		}
+	}
+
+	classadRegex, e = regexp.Compile(`^*\s*(K8SPhysicalHostName)\s=\s"(.*)"`)
+	if e != nil {
+		log.Fatal(e)
```
Saartank, line 805: I see we are using `log.Fatal` here, which will terminate the program after logging. Would this be a good idea? If this happens, the plugin won't produce the result ad and won't pass the information to Condor.

Saartank (changes requested):

cmd/plugin.go
turetske: I was following what was done for GlideIn_Site. Also this is for the creation of the regexp, so if that's causing an error, then there are larger problem.s

turetske:
@Saartank In order to test, setup a local condor pool and then run ```echo "K8SPhysicalHostName = \"TestHostName\"
" | condor_update_machine_ad```

Since I can't get the error to occur, you need to adjust the source code to force an error in `cmd/plugin.go`, the easiest way would be set isConfigErr to be true on line 112.

Then build the code and create a pelican_plugin binary and create a condor_submit file which uses that built binary to transfer some public file, I used ```pelican://osdf-itb.osg-htc.org/chtc/itb/osdf-pelican/TESTFILE_26.txt```

Then run and when it goes into a hold status, run condor_q -hold -long and see that the output sets (Hostname: EmmaTestCondor)

cmd/plugin_test.go
```diff
@@ -1177,3 +1177,64 @@ func TestParseDestination(t *testing.T) {
...
 }
+
+func TestWriteTransferErrorMessage(t *testing.T) {
+	server_utils.ResetTestState()
+	defer server_utils.ResetTestState()
+
+	currentDir, err := os.Getwd()
+	require.NoError(t, err)
+
+	tmpFile := currentDir + "/.machine.ad"
```
Saartank, line 1188: Would it be a good idea to create a temporary dir or file for this instead of creating a new file in the current working directory?

Saartank (changes requested):

cmd/plugin_test.go
turetske: Ah, in this case it has to be there because that's where the pelican_plugin looks for the machine ad and there isn't really a way for us to adjust that.

turetske: But I can create a defer task to remove the file

Saartank: Oh but you had already added a defer. 

https://github.com/PelicanPlatform/pelican/pull/1891/files#diff-7047045de9e6b2f5fc15f5af465c47b95cde0a18745cd7e34ee61faba13267cbR1235

Saartank (changes requested):

Saartank (approved):
Looks good!

PR #1896, opened 2025-01-15
Put the dynamically-linked pelican-server binary in the pelican-server RPM

### LINKED ISSUE #1855: Package pelican-server binary
Now that we have a separate "pelican-server" binary, we should put it in our packaging (RPMs, Debs, APKs).  Awkwardly, we already have a package called "pelican-server", but that just includes systemd files and config files for the individual components (e.g. `/etc/pelican/pelican-cache.yaml`).  Presumably we will be modifying the systemd files to use the pelican-server binary instead of the pelican binary.

I see three options for packaging (I'm using RPM as an example but we should also do Debs):

1. Put the pelican-server binary in the pelican RPM (same as the pelican binary).

2. Put the pelican-server binary in the pelican-server RPM (same as the systemd configs).

3. Put the pelican-server binary in its own RPM (say, "pelican-server-progs").

The advantage of choice 1 is that neither the pelican-server RPM nor the osdf-{cache,origin,director,registry} RPMs will gain any dependencies.  The downside is that admins won't be able to upgrade the client and server separately (which was one of the things we were hoping to get from splitting out pelican-server).

The advantage of choice 2 is that it's obvious (pelican-server contains pelican-server), and the pelican-server RPM won't gain any dependencies.  Unfortunately, the osdf-{cache,origin,director,registry} RPMs will gain a dependency on pelican-server, which will result in people installing, say, just osdf-cache, but also getting systemd files for pelican-origin, pelican-cache, pelican-director, pelican-registry.

The advantage of choice 3 is that it's a clean separation and doesn't have the previously mentioned downsides.  The disadvantage is that it's yet another RPM in our already large list of artifacts.  The dependencies of pelican-server and osdf-* will have to be updated.

I personally think 3 is the best choice here.

matyasselmeci:
Also updates the SystemD files for the cache and origin to use `pelican-server` instead of `pelican` or `osdf`.  For consistency with the current images, the director and registry .service files continue to use `pelican` or `osdf`.  Note that osdf-cache and osdf-origin no longer run a program with the magic `osdf` prefix, so they will be getting generic config unless overridden.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -430,7 +430,7 @@ nfpms:
           - "pelican-registry = 7"
           - "pelican-director = 7"
         dependencies:
-          - "pelican >= 7.4.0"
+          - "pelican >= 7.11.0"
           - "xrootd-server >= 1:5.7.0"
           - "xrootd-scitokens"
           - "xrootd-voms"
@@ -441,7 +441,7 @@ nfpms:
           - "pelican-registry (= 7)"
           - "pelican-director (= 7)"
         dependencies:
-          - "pelican (>= 7.4.0)"
+          - "pelican (>= 7.11.0)"
           - "xrootd-server (>= 5.7.0)"
           - "xrootd-scitokens-plugins"
           - "xrootd-voms-plugins"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -348,18 +348,20 @@ nfpms:
   # end package pelican-osdf-compet
 
   - package_name: pelican-server
-    builds: []
+    builds:
+      - pelican-server
     file_name_template: "{{ .ConventionalFileName }}"
     id: pelican-server
     vendor: OSG Consortium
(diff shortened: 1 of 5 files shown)


bbockelm:
@matyasselmeci - this one is still marked as draft.  Is it clear what we're waiting for?  Did it just get lost in the clutter?  Seems like the right thing to do...

matyasselmeci:
IIRC this is draft because initial tests on sc-origin ran into failures. I'm guessing this was because the binary prefix changed, but I didn't get to trying it with the OSDF configs as a YAML file.

matyasselmeci:
I tested this on osdf-itb-origin2000 and it works fine. Note that it's an osdf-itb origin, not an osdf origin, so we do have a config file specifying the federation instead of relying on the magic argv[0].

jhiemstrawisc (approved):
Looks okay to me. Is there any additional validation testing I should be aware of before a merge?

PR #1910, opened 2025-01-17
Update dependencies for OSX GitHub Action

bbockelm:
Switch OS X to the consolidated Pelican branch for xrootd and update various dependencies (xrdcl-pelican, xrootd-s3-http) to the latest desired release.


bbockelm:
Marking this as draft as the issue on Mac OS X is a real issue with the purge plugin patch.

jhiemstrawisc (approved):

PR #1915, opened 2025-01-17
Do not advertise HTTP caches

turetske:
For security reasons, we want to stop advertisement of HTTP caches.

This only occurs with non-pelican caches, which should be phased out by April. As such this shouldn't cause a large disruption (there are only a few caches where this occurs)

I decided to stop those caches from advertising to the director because I don't want them to be listed in the federation if they aren't available to the user.

### DIFF
--- a/director/advertise.go
+++ b/director/advertise.go
@@ -270,13 +270,15 @@ func AdvertiseOSDF(ctx context.Context) error {
 
 		for _, cache := range ns.Caches {
 			cacheAd := parseServerAdFromTopology(cache, server_structs.CacheType, server_structs.Capabilities{})
-			if existingAd, ok := cacheAdMap[cacheAd.URL.String()]; ok {
-				existingAd.NamespaceAds = append(existingAd.NamespaceAds, nsAd)
-				consolidatedAd := consolidateDupServerAd(cacheAd, existingAd.ServerAd)
-				existingAd.ServerAd = consolidatedAd
-			} else {
-				// New entry
-				cacheAdMap[cacheAd.URL.String()] = &server_structs.Advertisement{ServerAd: cacheAd, NamespaceAds: []server_structs.NamespaceAdV2{nsAd}}
+			if cacheAd.URL.Scheme != "http" {
+				if existingAd, ok := cacheAdMap[cacheAd.URL.String()]; ok {
+					existingAd.NamespaceAds = append(existingAd.NamespaceAds, nsAd)
+					consolidatedAd := consolidateDupServerAd(cacheAd, existingAd.ServerAd)
+					existingAd.ServerAd = consolidatedAd
+				} else {
+					// New entry
+					cacheAdMap[cacheAd.URL.String()] = &server_structs.Advertisement{ServerAd: cacheAd, NamespaceAds: []server_structs.NamespaceAdV2{nsAd}}
+				}
 			}
 		}
 	}
--- a/director/director_test.go
+++ b/director/director_test.go
@@ -1770,6 +1770,28 @@ func TestRedirects(t *testing.T) {
 		assert.NotContains(t, c.Writer.Header().Get("Link"), "pri=2")
 	})
 
+	t.Run("no-redirect-to-http-caches", func(t *testing.T) {
+		server_utils.ResetTestState()
+		t.Cleanup(func() {
+			server_utils.ResetTestState()
+		})
+
+		viper.Set("Director.CacheSortMethod", "random")
+
+		// Make sure the http cache from topology isn't included in the cache list
+		req, _ := http.NewRequest("GET", "/my/server/3", nil)
+		req.Header.Add("User-Agent", "pelican-client/7.6.1")
+		req.Header.Add("X-Real-Ip", "128.104.153.60")
+
+		recorder := httptest.NewRecorder()
+		c, _ := gin.CreateTestContext(recorder)
+		c.Request = req
+
+		redirectToCache(c)
+		assert.Contains(t, c.Writer.Header().Get("Link"), "pri=2")
+		assert.NotContains(t, c.Writer.Header().Get("Link"), "pri=3")
+	})
+
 	// Make sure collections-url is correctly populated when the ns/origin comes from topology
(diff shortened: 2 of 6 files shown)


bbockelm:
I'm not sure this is the best approach here.

Are these caches really HTTP-only?  Seems like we only need to avoid redirecting to HTTP but we don't need to remove all the caches that speak both protocols.

For example, the LIGO cache at Caltech has both https://stashcache.ligo.caltech.edu:8443 and http://stashcache.ligo.caltech.edu:8000; not redirecting for HTTP will have minimal impact but removing support for HTTPS will have a huge impact at their site.

director/advertise.go
```diff
@@ -270,13 +270,15 @@ func AdvertiseOSDF(ctx context.Context) error {
...
 		for _, cache := range ns.Caches {
 			cacheAd := parseServerAdFromTopology(cache, server_structs.CacheType, server_structs.Capabilities{})
-			if existingAd, ok := cacheAdMap[cacheAd.URL.String()]; ok {
-				existingAd.NamespaceAds = append(existingAd.NamespaceAds, nsAd)
-				consolidatedAd := consolidateDupServerAd(cacheAd, existingAd.ServerAd)
-				existingAd.ServerAd = consolidatedAd
-			} else {
-				// New entry
-				cacheAdMap[cacheAd.URL.String()] = &server_structs.Advertisement{ServerAd: cacheAd, NamespaceAds: []server_structs.NamespaceAdV2{nsAd}}
+			if cacheAd.URL.Scheme != "http" {
```
jhiemstrawisc, line 273: Rather than filter the entire ad, would it work to just set the `cacheAd.URL` to `nil`? Then we could still use these caches for authed redirects:
```
if cacheAd.URL.Scheme == "http" {
   cacheAd.URL = nil
}
if existingAd, ok := cacheAdMap[cacheAd.URL.String()]; ok {
    existingAd.NamespaceAds = append(existingAd.NamespaceAds, nsAd)
	consolidatedAd := consolidateDupServerAd(cacheAd, existingAd.ServerAd)
	existingAd.ServerAd = consolidatedAd
} else {
    // New entry
	cacheAdMap[cacheAd.URL.String()] = &server_structs.Advertisement{ServerAd: cacheAd, NamespaceAds: []server_structs.NamespaceAdV2{nsAd}}
}
```

If you went that route, it'd be worth double checking that override won't interfere with ad consolidation.

director/resources/mock_topology.json
```diff
@@ -125,6 +130,39 @@
                 "readhttps": true,
                 "usetokenonread": false,
                 "writebackhost": null
+            },
+            {
+                "caches": [
+                    {
+                            "auth_endpoint": "http://cache8.com",
+                            "endpoint": "http://cache8.com",
```
jhiemstrawisc, line 138: I think we should still be setting `endpoint` to an `http` URL and `auth_endpoint` to `https` -- that's the setup we'd get expect from real topology. Then the test can verify that those caches are still used for authed redirects but not for public reads.

director/resources/multi_export_topology.json
```diff
@@ -4,8 +4,8 @@
      {
         "caches":[
          {
-            "auth_endpoint": "osdf-uw-cache.svc.osg-htc.org:8443",
-            "endpoint": "osdf-uw-cache.svc.osg-htc.org:8443",
+            "auth_endpoint": "https://osdf-uw-cache.svc.osg-htc.org:8443",
+            "endpoint": "https://osdf-uw-cache.svc.osg-htc.org:8443",
```
jhiemstrawisc, line 8: Good catch -- these were missing schemes before. I think we still want to have an http/https split though.

director/director_test.go
```diff
@@ -1770,6 +1770,28 @@ func TestRedirects(t *testing.T) {
...
+		req, _ := http.NewRequest("GET", "/my/server/3", nil)
+		req.Header.Add("User-Agent", "pelican-client/7.6.1")
+		req.Header.Add("X-Real-Ip", "128.104.153.60")
+
+		recorder := httptest.NewRecorder()
+		c, _ := gin.CreateTestContext(recorder)
+		c.Request = req
+
+		redirectToCache(c)
+		assert.Contains(t, c.Writer.Header().Get("Link"), "pri=2")
```
jhiemstrawisc, line 1791: I'd prefer to see a more concrete check that the test succeeds here. If the code breaks and you get one HTTP and one HTTPS redirect, this might still pass.

jhiemstrawisc (changes requested):

director/advertise.go
turetske: I prefer this approach. Going to run some tests to make sure nothing breaks if we do that.

director/resources/mock_topology.json
turetske: Can do. We should probably set our tests to mimic that when possible.

director/director_test.go
turetske: Fair point.

director/director_test.go
```diff
@@ -1771,6 +1771,76 @@ func TestRedirects(t *testing.T) {
...
+		c, _ := gin.CreateTestContext(recorder)
+		c.Request = req
+
+		redirectToCache(c)
+
+		assert.Contains(t, c.Writer.Header().Get("Link"), "pelcache.test.edu")
+		assert.NotContains(t, c.Writer.Header().Get("Link"), "http:")
+	})
+
+	t.Run("redirect-to-topology-caches-auth-reads", func(t *testing.T) {
```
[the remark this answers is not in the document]
turetske: Possibly, but I think that should be its own separate issue.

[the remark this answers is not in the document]
turetske: Actually, looking over this. I can get it done. I definitely see a few areas to split and refactor.

jhiemstrawisc:
Wow, it's actually a real relief that removing the schemes from mock_topology didn't break a single test 😅

jhiemstrawisc (approved):
LGTM. Thanks for all the cleanup :)

Issue #1868, opened 2025-01-08
Create Database Package


patrickbrophy:
As mentioned in issue #1796, @bbockelm would like to move from a per-server database to a single database. In spirit of this, we should migrate our interactions with any database to its own package so that there is a single central place to look at how we interact with the database. This package would include the migrations, model schemas, and other functionality required.

Issue #1796, opened 2024-12-04
Grafana API Token Access


patrickbrophy:
Currently accessing Prometheus from Grafana is tricky. To mitigate this difficulty, we want to introduce an API token that when provided, allows access to Prometheus. This token should be persisted in a database. The tokens should be salted and hashed before being stored into the database.

Token management should be provided via a CRUD API. This new API would then be hooked up to the Pelican Web UI so that admins can manage these new tokens. This API should only be accessible to administrators.

bbockelm:
A few thoughts on the implementation:
- I really dislike how we've come to have per-server SQLite database locations (`Origin.DbLocation`, `Registry.DbLocation`, `Director.DbLocation`).  There should be a single database, for Pelican.  Don't solve the problem in this issue -- however, let's also not make it any worse.  Instead, let's create a new parameter, `Server.DbLocation`, that the others can eventually migrate to.
- *Do not store raw secrets*: Use `golang.org/x/crypto/bcrypt` to hash the token on disk.
- Unfortunately, verifying the hashed API key is computationally expensive (by design).  On success, keep a TTL cache of verified API keys.  This cache will need to be updated whenever changes to the API tokens are needed (e.g., invalidate entries when a token deletion is done).
   - Rate limit the verification of API keys.  There's an example of doing this already for the login API, I think.
- Given we don't store secrets, the token needs an ID (a handle we can refer to when invalidating it, for example).  One approach is to have the ID be the first 5 characters of the SHA-256 checksum of the token.  Since you're generating this, there's a chance of collision: on collision with an existing API key in the DB, simply retry the generation.
   - The contents of the token could be serialized in the form `ID.SECRET` (as in, `ab258.12b7fa82190d`).  That way, even though we currently generate the ID, it's not _forced_ to be the value of the checksum and we could change the algorithm in the futuer.
- Attributes of an API key that likely need to be persisted:
   - (Unique) ID.  Do not use a monotonically-increasing integer.  Indexed.
   - Human-readable name.
   - Expiration time.
   - List of scopes.  May need to be normalized out to a separate table (probably indexed by token ID).
   - Creation time & identity of creator.
   - Last change time & identity.
   - Salted token value.

So, if the token is `ab258.12b7fa82190d`, then we would:
1.  Look at the TTL cache of verified tokens for `ab258.12b7fa82190d` and use that entry.
[...]

PR #1917, opened 2025-01-17
Grafana Token API Access

patrickbrophy:
This PR addresses #1796. This PR currently implements a new token with its purpose to simplify access to Prometheus metrics. This token is accessible via the `Authorization` header. There are a new HTTP route at the director for creating and deleting tokens: 
 - `/api/v1.0/director/createGrafanaToken` 
 - `/api/v1.0/director/deleteGrafanaToken`

This PR also creates a database package as a work around to dependency cycles create between the director package and token package. Its is very rudimentary and will be improved in #1868.

### DIFF
--- a/database/director.go
+++ b/database/director.go
@@ -15,15 +15,22 @@ import (
 
 var DirectorDB *gorm.DB
 
-type ApiKey struct {
-	ID          string `gorm:"primaryKey;column:id;type:text;not null;unique"`
-	Name        string `gorm:"column:name;type:text"`
-	HashedValue string `gorm:"column:hashed_value;type:text;not null"`
-	Scopes      string `gorm:"column:scopes;type:text"`
-	ExpiresAt   time.Time
-	CreatedAt   time.Time
-	CreatedBy   string `gorm:"column:created_by;type:text"`
-}
+type (
+	ApiKey struct {
+		ID          string `gorm:"primaryKey;column:id;type:text;not null;unique"`
+		Name        string `gorm:"column:name;type:text"`
+		HashedValue string `gorm:"column:hashed_value;type:text;not null"`
+		Scopes      string `gorm:"column:scopes;type:text"`
+		ExpiresAt   time.Time
+		CreatedAt   time.Time
+		CreatedBy   string `gorm:"column:created_by;type:text"`
+	}
+
+	ApiKeyCached struct {
+		Token        string // "$ID.$SECRET_IN_HEX" string form
+		Capabilities []string
+	}
+)
 
 func generateSecret(length int) (string, error) {
 	bytes := make([]byte, length)
@@ -39,44 +46,51 @@ func generateTokenID(secret string) string {
 	return string(hash[:])[:5]
 }
 
-func VerifyApiKey(apiKey string, verifiedKeysCache *ttlcache.Cache[string, ApiKey]) (bool, error) {
+func VerifyApiKey(apiKey string, verifiedKeysCache *ttlcache.Cache[string, ApiKeyCached]) (bool, []string, error) {
+	item := verifiedKeysCache.Get(apiKey)
+	if item != nil {
+		// Cache hit
+		return true, item.Value().Capabilities, nil
+	}
 	parts := strings.Split(apiKey, ".")
 	if len(parts) != 2 {
-		return false, errors.New("invalid API key format")
+		return false, nil, errors.New("invalid API key format")
 	}
 	id := parts[0]
 	secret := parts[1]
-
-	item := verifiedKeysCache.Get(id)
-	if item != nil {
-		cachedToken := item.Value()
-		beforeExpiration := time.Now().Before(cachedToken.ExpiresAt)
-		matches := bcrypt.CompareHashAndPassword([]byte(cachedToken.HashedValue), []byte(secret)) == nil
-		if beforeExpiration && matches {
-			return true, nil
-		}
-	}
 
 	var token ApiKey
 	result := DirectorDB.First(&token, "id = ?", id)
 	if result.Error != nil {
 		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
-			return false, nil // token not found
+			return false, nil, nil // token not found
 		}
-		return false, errors.Wrap(result.Error, "failed to retrieve the API key")
+		return false, nil, errors.Wrap(result.Error, "failed to retrieve the API key")
 	}
 
 	if !token.ExpiresAt.IsZero() && time.Now().After(token.ExpiresAt) {
-		return false, nil
+		return false, nil, nil
 	}
 
 	err := bcrypt.CompareHashAndPassword([]byte(token.HashedValue), []byte(secret))
 	if err != nil {
-		return false, nil
+		return false, nil, nil
 	}
 
-	verifiedKeysCache.Set(id, token, ttlcache.DefaultTTL)
-	return true, nil
+	cacheTTL := ttlcache.DefaultTTL
+	if !token.ExpiresAt.IsZero() {
+		timeUntilExpiration := time.Until(token.ExpiresAt)
+		if timeUntilExpiration < cacheTTL {
+			cacheTTL = timeUntilExpiration
+		}
+	}
+
+	cached := ApiKeyCached{
+		Token:        apiKey,
+		Capabilities: strings.Split(token.Scopes, ","),
+	}
+	verifiedKeysCache.Set(id, cached, cacheTTL)
+	return true, cached.Capabilities, nil
 }
 
 func CreateApiKey(name, createdBy, scopes string, expiration time.Time) (string, error) {
@@ -115,7 +129,7 @@ func CreateApiKey(name, createdBy, scopes string, expiration time.Time) (string,
 	}
 }
 
-func DeleteApiKey(id string, verifiedKeysCache *ttlcache.Cache[string, ApiKey]) error {
+func DeleteApiKey(id string, verifiedKeysCache *ttlcache.Cache[string, ApiKeyCached]) error {
 	result := DirectorDB.Delete(&ApiKey{}, "id = ?", id)
 	if result.Error != nil {
 		return errors.Wrap(result.Error, "failed to delete the API key")
--- a/database/director.go
+++ b/database/director.go
@@ -50,7 +50,7 @@ func VerifyApiKey(db *gorm.DB, apiKey string, verifiedKeysCache *ttlcache.Cache[
 		cached := item.Value()
 		if cached.Token == apiKey { // check the cached token matches the one we are trying to verify
 			// check if the token has expired
-			if !cached.ExpiresAt.IsZero() && time.Now().After(cached.ExpiresAt) {
+			if !cached.ExpiresAt.IsZero() && time.Now().UTC().After(cached.ExpiresAt) {
 				return false, nil, errors.New("Token has expired")
 			}
 			return true, cached.Capabilities, nil
@@ -73,7 +73,7 @@ func VerifyApiKey(db *gorm.DB, apiKey string, verifiedKeysCache *ttlcache.Cache[
 
 	// Check if the token has expired
 	// If the token has an expiration time and the current time is after the expiration time, the token is invalid
-	if !token.ExpiresAt.IsZero() && time.Now().After(token.ExpiresAt) {
+	if !token.ExpiresAt.IsZero() && time.Now().UTC().After(token.ExpiresAt) {
 		return false, nil, errors.New("Token has expired")
 	}
 
@@ -128,8 +128,8 @@ func CreateApiKey(name, createdBy, scopes string, expiration time.Time) (string,
 			Name:        name,
 			HashedValue: string(hashedValue),
 			Scopes:      scopes,
-			ExpiresAt:   expiration,
-			CreatedAt:   time.Now(),
+			ExpiresAt:   expiration.UTC(),
+			CreatedAt:   time.Now().UTC(),
 			CreatedBy:   createdBy,
 		}
 		result := DirectorDB.Create(apiKey)
--- a/database/director.go
+++ b/database/director.go
@@ -67,15 +67,23 @@ func VerifyApiKey(db *gorm.DB, apiKey string, verifiedKeysCache *ttlcache.Cache[
 		return false, nil, errors.New("Failed to retrieve the API key")
 	}
 
+	// Check if the token has expired
+	// If the token has an expiration time and the current time is after the expiration time, the token is invalid
 	if !token.ExpiresAt.IsZero() && time.Now().After(token.ExpiresAt) {
 		return false, nil, errors.New("Token has expired")
 	}
 
+	// We compare the hashed value of the secret with the stored hashed value
+	// If there is a match, the API key is valid
+	// Otherwise, the API key is invalid
 	err = bcrypt.CompareHashAndPassword([]byte(token.HashedValue), []byte(secret))
 	if err != nil {
 		return false, nil, errors.New("Invalid API token")
 	}
 
+	// Cache the verified API key
+	// Keys that have an expiration time are cached with a TTL equal to the time until expiration
+	// Keys that don't have an expiration time are cached with the default TTL
 	cacheTTL := ttlcache.DefaultTTL
 	if !token.ExpiresAt.IsZero() {
 		timeUntilExpiration := time.Until(token.ExpiresAt)
--- a/database/director.go
+++ b/database/director.go
@@ -49,6 +49,10 @@ func VerifyApiKey(db *gorm.DB, apiKey string, verifiedKeysCache *ttlcache.Cache[
 		// Cache hit
 		cached := item.Value()
 		if cached.Token == apiKey { // check the cached token matches the one we are trying to verify
+			// check if the token has expired
+			if !cached.ExpiresAt.IsZero() && time.Now().After(cached.ExpiresAt) {
+				return false, nil, errors.New("Token has expired")
+			}
 			return true, cached.Capabilities, nil
 		} // otherwise the api token doesn't match the one in the cache so we do a hard check
 	}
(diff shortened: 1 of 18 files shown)


bbockelm (changes requested):
Bah!  I reviewed this last week but apparently didn't submit the review :(

See inline comments.

database/director.go
```diff
@@ -0,0 +1,129 @@
...
+	ID          string `gorm:"primaryKey;column:id;type:text;not null"`
+	Name        string `gorm:"column:name;type:text"`
+	HashedValue string `gorm:"column:hashed_value;type:text;not null"`
+	Scopes      string `gorm:"column:scopes;type:text"`
+	ExpiresAt   time.Time
+	CreatedAt   time.Time
+	CreatedBy   string `gorm:"column:created_by;type:text"`
+}
+
+var verifiedKeysCache *ttlcache.Cache[string, GrafanaApiKey] = ttlcache.New[string, GrafanaApiKey]()
```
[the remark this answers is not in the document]
patrickbrophy: I relocated the verified keys cache to the `token` package to avoid a dependency cycle that would occur if it were placed in the `web_ui` package. Keeping the cache in the token package eliminates dependency issues.

database/director.go
```diff
@@ -0,0 +1,129 @@
...
+	_, err := rand.Read(bytes)
+	if err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(bytes), nil
+}
+
+func generateTokenID(secret string) string {
+	hash := sha256.Sum256([]byte(secret))
+	return hex.EncodeToString(hash[:])[:5]
```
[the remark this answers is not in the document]
patrickbrophy: Possible unique token IDs (N): N = 2^40 = 1.1 * 10^12

Collision Probability (n = 1000):
P = n^2 / 2N = (1000^2) / (2 * 1.1 * 10^12) = 4.55 * 10^-7 = 0.000045%

database/director.go
```diff
@@ -0,0 +1,129 @@
...
+
+var verifiedKeysCache *ttlcache.Cache[string, GrafanaApiKey] = ttlcache.New[string, GrafanaApiKey]()
+
+func generateSecret(length int) (string, error) {
+	bytes := make([]byte, length)
+	_, err := rand.Read(bytes)
+	if err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(bytes), nil
```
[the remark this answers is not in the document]
patrickbrophy: Without hex encoding the API token looks like this
```
50ae6.g\ufffd\u0010Oo\ufffd\u001e\ufffd\ufffdJz@\\\ufffd\ufffd,z\ufffd͔\ufffd\ufffd\ufffdU\u003c̍\ufffd\u000b\ufffd\ufffdh
```

With the hex encoding the API token looks like this
```
d2c9e.f0a2673880101b43bc24fdff19cbcc3c37d48fd517774af98f06365875c1688a
```

I was just thinking from a usability standpoint the latter is easier to handle than the former.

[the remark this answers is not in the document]
bbockelm: You definitely want the hex encoding of the secret in the token.

However, currently, the hex encoded version _is_ the secret.  It's better to make the binary version the secret (as that makes it more clear what the key is and just use the encoded part as part of the transport.  Using the encoded version of the secret also creates confusion about the strength of the key -- it looks (incorrectly) stronger because the encoded version is longer.

So, you want to return the raw secret here and encode / decode as appropriate.

web_ui/authorization.go
```diff
@@ -89,7 +89,7 @@ func promQueryEngineAuthHandler(av1 *route.Router) gin.HandlerFunc {
 			authOption := token.AuthOption{
 				// Cookie for web user access and header for external service like Grafana to access
 				Sources: []token.TokenSource{token.Cookie, token.Header},
-				Issuers: []token.TokenIssuer{token.LocalIssuer},
+				Issuers: []token.TokenIssuer{token.LocalIssuer, token.GrafanaTokenIssuer},
```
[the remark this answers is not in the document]
patrickbrophy: I agree! Did you have any ideas on where we should accept it?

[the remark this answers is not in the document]
bbockelm: I suspect any place that accepts the `LocalIssuer` (the one associated with the server) should accept the `ApiTokenIssuer`.  I essentially view the `LocalIssuer` and `ApiTokenIssuer` as interchangeable, differing only in the credential formats.

database/director.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(parts) != 2 {
+		return false, nil, errors.New("invalid API key format")
+	}
+	id := parts[0]
+	secretHex := parts[1]
+
+	item := verifiedKeysCache.Get(id)
+	if item != nil {
+		// Cache hit
+		return true, item.Value().Capabilities, nil
```
bbockelm, line 61: Missing the check that the secret is correct, no?  Seems to only verify the public portion here

director/director.go
```diff
@@ -1571,6 +1571,7 @@ func RegisterDirectorAPI(ctx context.Context, router *gin.RouterGroup) {
 		// so that director can be our point of contact for collecting system-level metrics.
 		// Rename the endpoint to reflect such plan.
 		directorAPIV1.GET("/discoverServers", discoverOriginCache)
+
```
bbockelm, line 1574: Please avoid white space noise!

director/director_db.go
```diff
@@ -51,8 +50,8 @@ func InitializeDB() error {
 	if err != nil {
 		return errors.Wrap(err, "failed to initialize the Director's sqlite database")
 	}
-	db = tdb
-	sqldb, err := db.DB()
```
bbockelm, line 55: Are there any upgrade/compatibility concerns from this change?

web_ui/ui.go
```diff
@@ -540,6 +655,9 @@ func waitUntilLogin(ctx context.Context) error {
 //
 // You need to mount the static resources for UI in a separate function
 func ConfigureServerWebAPI(ctx context.Context, engine *gin.Engine, egrp *errgroup.Group) error {
+	// start the cache for verified API keys
+	go token.VerifiedKeysCache.Start()
```
bbockelm, line 659: Connect the go routine to an errgroup object -- that ensures the routine is shut down along with the rest of the server (especially relevant for starting up / shutting down unit tests).

bbockelm (changes requested):

bbockelm:
Oh - forgot to mention!  Please add unit test coverage.  The fact that it appears one could authenticate with an incorrect secret ID (only knowing the public ID) in the current patch series suggests that unit test coverage is going to be of great value.

director/director_db.go
patrickbrophy: I don't believe so. This change just moves things around.

patrickbrophy:
This is frustrating because I can't seem to replicate the test failures on my machine.

web_ui/ui_test.go
```diff
@@ -496,3 +503,261 @@ func TestServerHostRestart(t *testing.T) {
...
+
+	tempFile, err := os.CreateTemp("", "web-ui-passwd-api-token")
+	if err != nil {
+		fmt.Println("Failed to setup web-ui-passwd file")
+		os.Exit(1)
+	}
+	defer os.Remove(tempFile.Name())
+
+	dirName := t.TempDir()
+	server_utils.ResetTestState()
```
jhiemstrawisc, line 540: Best practice is to put this at the top of your test:
```
server_utils.ResetTestState()
defer server_utils.ResetTestState()
```

This makes sure you start clean and end clean.

web_ui/ui_test.go
```diff
@@ -496,3 +503,261 @@ func TestServerHostRestart(t *testing.T) {
...
+	viper.Set("ConfigDir", dirName)
+	config.InitConfig()
+	viper.Set("Server.UIPasswordFile", tempFile.Name())
```
jhiemstrawisc, line 543: For these special config params, can you set the name using our `param` package, e.g.:
```
viper.Set(param.Server_UIPasswordFile.GetName(), tempFile.Name())
```

This guarantees we see a test failure here if we ever change the name of the config parameter, where we might otherwise just end up setting a param that no longer exists.

web_ui/ui_test.go
```diff
@@ -496,3 +503,261 @@ func TestServerHostRestart(t *testing.T) {
...
+	content := "admin:password\n"
+	_, err = tempFile.WriteString(content)
+	assert.NoError(t, err, "Error writing to temp password file")
+
+	err = tempFile.Sync()
+	require.NoError(t, err)
+
+	//Configure UI
+	err = configureAuthDB()
+	assert.NoError(t, err)
```
jhiemstrawisc, line 556: It might be worth taking a second look at which of these should be a `require` and which should be an `assert`. The `require` word means "if this doesn't pass, the test is doomed and there's no additional information I might get by running other functions", while the `assert` means "the test should fail if this doesn't pass, but there might be useful information about why the test is broken in other failed assertions".

I might be wrong, but it feels like failure to configure the auth DB is pretty fatal.

web_ui/ui_test.go
```diff
@@ -496,3 +503,261 @@ func TestServerHostRestart(t *testing.T) {
...
+	t.Run("create-and-delete-token", func(t *testing.T) {
+		// Create a request
+		req, err := http.NewRequest("POST", "/api/v1.0/createToken", nil)
+		assert.NoError(t, err)
+
+		req.AddCookie(&http.Cookie{Name: "login", Value: cookieValue})
+
+		createTokenReq := CreateApiTokenReq{
+			Name:       "test-token",
+			CreatedBy:  "admin",
```
jhiemstrawisc, line 598: Is this `CreatedBy` field totally arbitrary? That is, do we actually trust the value it contains, or can someone generate a token and put whatever they want in the field?

web_ui/ui_test.go
```diff
@@ -496,3 +503,261 @@ func TestServerHostRestart(t *testing.T) {
...
+
+		recorder = httptest.NewRecorder()
+		route.ServeHTTP(recorder, req)
+
+		// Check unauthorized http response
+		assert.Equal(t, http.StatusForbidden, recorder.Code)
+
+		assert.Contains(t, recorder.Body.String(), "Invalid API token")
+	})
+}
```
jhiemstrawisc, line 763: In general, a lot of these tests feel like they could be streamlined/deduplicated by using a testing struct/slice. Can you investigate refactoring these to look something like:
```
func MyTest(t *testing.T) {
    testCases := []struct {
        name                string
        requestPath     string
        cookie              string
        expectedBody string
        expectErr         bool
        // otherCommonStuf
    } {
        {
            // fill out values for each test case
        },
        // other test cases
    }
    
    for _, tc := range testCases {
            // A function that handles the logic for each test
        })
    }

}
```

In general, I've found ChatGPT really useful for this kind of thing quickly -- if you give it a copy-paste of your tests and tell it what you want to do, it's pretty good at approximating the test struct and giving you the start of a test driver function. If you go that route, be sure to double check that your tests are still doing what you want them to!

jhiemstrawisc (changes requested):
Some of the comments are actual change requests, some are requests for you to verify my assumptions/reading of the code, and some are just me asking for more information about something I didn't understand.

web_ui/ui.go
```diff
@@ -375,6 +385,123 @@ func handleWebUIResource(ctx *gin.Context) {
...
+	if len(req.Scopes) == 0 {
+		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    "Scopes is required",
+		})
+		return
+	}
+
+	var expirationTime time.Time
+	if req.Expiration == "" || req.Expiration == "never" { // if expiration is not provided, set token to never expire
```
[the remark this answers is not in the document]
patrickbrophy: We have a way to delete tokens via the deleteApiToken method. Also Brian had mentioned that this is something that he wanted: https://github.com/PelicanPlatform/pelican/pull/1917#discussion_r1920918617

web_ui/ui.go
```diff
@@ -375,6 +385,123 @@ func handleWebUIResource(ctx *gin.Context) {
...
+
+func deleteApiToken(ctx *gin.Context) {
+	authOption := token.AuthOption{
+		Sources: []token.TokenSource{token.Cookie},
+		Issuers: []token.TokenIssuer{token.LocalIssuer},
+		Scopes:  []token_scopes.TokenScope{token_scopes.WebUi_Access},
+	}
+	status, ok, err := token.Verify(ctx, authOption)
+	if !ok {
+		log.Warningf("Cannot verify token: %v", err)
```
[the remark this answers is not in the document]
patrickbrophy: We follow the same pattern throughout the rest Pelican with other calls to token.Verify. Also looking through the token.Verify it seems that the error messages returned are all very sane.

database/director.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	id := parts[0]
+	secretHex := parts[1]
+
+	item := verifiedKeysCache.Get(id)
+	if item != nil {
+		// Cache hit
+		cached := item.Value()
+		if cached.Token == apiKey { // check the the cached token matches the one we are trying to verify
+			return true, cached.Capabilities, nil
```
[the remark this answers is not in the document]
bbockelm: FWIW -- the issue we hit previously was that the TTLCache's purge goroutine wasn't running.

I think that resulted in the cache entries never being removed (memory bloat).  The cache expirations were still occurring (so old data wasn't being served).

So, I'm OK to skip this personally.

web_ui/ui_test.go
```diff
@@ -496,3 +503,261 @@ func TestServerHostRestart(t *testing.T) {
...
+	route.DELETE("/api/v1.0/deleteApiToken/:id", deleteApiToken)
+	route.GET("/privilegedRoute", func(ctx *gin.Context) {
+		authOption := token.AuthOption{
+			Sources: []token.TokenSource{token.Header},
+			Issuers: []token.TokenIssuer{token.APITokenIssuer},
+			Scopes:  []token_scopes.TokenScope{token_scopes.Monitoring_Scrape},
+		}
+
+		status, ok, err := token.Verify(ctx, authOption)
+		if !ok {
```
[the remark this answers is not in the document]
patrickbrophy: I feel that this is an overreach for the context of this PR, however I do agree. I think refactoring to just return the status and an error seems much better.

web_ui/ui_test.go
patrickbrophy: We don't care whats put here. More or less just as a notion of record.

jhiemstrawisc (approved):
LGTM. I think the failing test is flaky and unrelated, but let's re-run it to be safe.

jhiemstrawisc:
> I'm not sure the test suite is failing. I am unable to replicate the failure locally.

One of the unit tests was failing in GHA, but I just reran it to see what it does.

jhiemstrawisc (approved):

director/migrations/20250103161929_create_db_tables.sql
```diff
@@ -0,0 +1,27 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
```
h2zh, line 3:
```suggestion
```
The SELECT 'up SQL query' and SELECT 'down SQL query' statements are placeholders and should be replaced with actual SQL statements or just remove them.

director/migrations/20250103161929_create_db_tables.sql
```diff
@@ -0,0 +1,27 @@
...
+-- +goose StatementEnd
```
h2zh, line 4: This line should be placed after line 22.
Quote from [goose docs:](https://pressly.github.io/goose/documentation/annotations/#multiple-statements)
> The -- +goose statementbegin and -- +goose statementend annotations can also be used to combine multiple statements so they get sent as a single query string instead of being executed individually.

director/migrations/20250103161929_create_db_tables.sql
```diff
@@ -0,0 +1,27 @@
...
+    hashed_value TEXT NOT NULL,
+    scopes TEXT,
+    expires_at DATETIME,
+    created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
+    created_by TEXT
+);
+
+-- +goose Down
+-- +goose StatementBegin
+SELECT 'down SQL query';
```
h2zh, line 26:
```suggestion
DROP TABLE IF EXISTS server_downtimes;
DROP TABLE IF EXISTS api_keys;
```

director/migrations/20250103161929_create_db_tables.sql
```diff
@@ -0,0 +1,27 @@
...
+    created_at DATETIME NOT NULL,
+    updated_at DATETIME NOT NULL
+);
+
+CREATE TABLE api_keys (
+    id TEXT PRIMARY KEY UNIQUE,
+    name TEXT,
+    hashed_value TEXT NOT NULL,
+    scopes TEXT,
+    expires_at DATETIME,
```
h2zh, line 19: It'd be better to ensure this uses UTC throughout the entire PR

director/migrations/20250103161929_create_db_tables.sql
```diff
@@ -0,0 +1,27 @@
...
+SELECT 'up SQL query';
+-- +goose StatementEnd
+
+CREATE TABLE IF NOT EXISTS server_downtimes (
+    uuid TEXT PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE,
+    filter_type TEXT NOT NULL,
+    created_at DATETIME NOT NULL,
+    updated_at DATETIME NOT NULL
+);
```
h2zh, line 12: Since server_downtimes is already created in the existing db migration file, we can this block out to avoid redundancy. (It won't cause problem if we have this because it creates server_downtimes table on if it doesn't exist - but it does exist). https://github.com/PelicanPlatform/pelican/blob/main/director/migrations/20241017135850_create_db_tables.sql 
```suggestion
```

h2zh (changes requested):
After talking to Justin, my review is focusing on the db. Here are some trivial things we can improve:

h2zh (approved):

PR #1919, opened 2025-01-17
Refactor & Cleanup Origin Exports code

jhiemstrawisc:
This refactors a lot of the suburban sprawl from `server_utils/origin.go` in a way that I hope is much more maintainable and much less likely to have hard-to-diagnose errors.

In particular, I tried out some OOP go-style (and left lots of comments for others to follow along with the setup). I couldn't _quite_ do things the way I expected, but I think I got pretty close to something that's clean and easy to follow.

### DIFF
--- a/server_utils/origin.go
+++ b/server_utils/origin.go
@@ -152,11 +152,6 @@ type Origin interface {
 	configureExports(Origin) ([]OriginExport, error)
 	validateStoragePrefix(sPrefix string) error
 
-	// Some methods are passed the Origin interface so they can call the appropriate
-	// method on a concrete origin type. This is necessary for overriding the base methods.
-	// The methods taking an Origin interface define common setup that needs to happen for _every_ origin.
-	// The "Extra" methods are used for origin-specific validation, configuration, etc.
-	//
 	// handleVolumeMounts is primarily used for configuring exports from the -v flag
 	handleVolumeMounts(Origin) error
 	handleVolumeMountsExtra() error
--- a/server_utils/origin.go
+++ b/server_utils/origin.go
@@ -154,7 +154,7 @@ type Origin interface {
 
 	// Some methods are passed the Origin interface so they can call the appropriate
 	// method on a concrete origin type. This is necessary for overriding the base methods.
-	// The methods taking an Origin interface define common setup that needs to happey for _every_ origin.
+	// The methods taking an Origin interface define common setup that needs to happen for _every_ origin.
 	// The "Extra" methods are used for origin-specific validation, configuration, etc.
 	//
 	// handleVolumeMounts is primarily used for configuring exports from the -v flag
--- a/server_utils/origin.go
+++ b/server_utils/origin.go
@@ -159,36 +159,13 @@ func validateFederationPrefix(prefix string) error {
 	if !strings.HasPrefix(prefix, "/") {
 		return errors.Wrapf(ErrInvalidOriginConfig, "prefix %s must begin with '/'", prefix)
 	}
-	if strings.Contains(prefix, "//") {
-		return errors.Wrapf(ErrInvalidOriginConfig, "prefix %s contains invalid '//' characters", prefix)
-	}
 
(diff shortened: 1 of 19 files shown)


jhiemstrawisc:
For the reviewer, my recommendation is to start in `origin.go::GetOriginExports` and follow that into `origin.configureExports`. That should give a pretty clear picture of the common setup each origin has.

server_utils/origin.go
```diff
@@ -136,59 +138,113 @@ func StringListToCapsHookFunc() mapstructure.DecodeHookFuncType {
...
-	if strings.Contains(prefix, "//") {
-		return errors.Wrapf(ErrInvalidOriginConfig, "prefix %s contains invalid '//' characters", prefix)
-	}
+type Origin interface {
+	configureExports(Origin) ([]OriginExport, error)
+	validateStoragePrefix(sPrefix string) error
+
+	// Some methods are passed the Origin interface so they can call the appropriate
+	// method on a concrete origin type. This is necessary for overriding the base methods.
+	// The methods taking an Origin interface define common setup that needs to happey for _every_ origin.
```
turetske, line 157:
```suggestion
	// The methods taking an Origin interface define common setup that needs to happen for _every_ origin.
```

turetske, line 157: Also, this is just the definition of an interface, which seems redundant.

turetske (changes requested):
Taking a look at this, this doesn't quite feel like the right approach.

What you are doing doesn't really lend itself to an interface, but more class Inheritance. The way Golang does this is with composition. This allows you avoid the very awkwardly named "....Extra()" functions in your code and rather call the embedded composed struct method followed by the same named current method. This might actually be a good place to use both composition and interfaces.

If we are doing this refactor, I also think OriginExport should be adjusted to be class/origin type specific (definitely ripe for composition with the common fields in the top level and the specific pieces like S3Bucket, S3AccessKeyFile, and S3SecretKeyfile).

jhiemstrawisc:
Let's do some churn mitigation here, since I'm effectively receiving a review that requests I redo the entirety of a 1700 line diff barely a day before the scheduled release, despite this PR being up for review for over a month now. 

Here's what I propose:
1. Move forward with this approach for now, which I did base heavily on lots of other Golang examples/tutorials. One feature I really like of Go is that it makes possible a variety of different approaches coding, and this structure is not a pattern I created, but one I've seen elsewhere. I'm a bit confused by your comment about using "composition and interfaces" -- this is already a large part of how I've structured this, where each individual origin struct is composed of the base struct, e.g.:
```
type Origin interface {}

type baseOrigin struct {
    // fulfills Origin Interface
}

type S3Origin struct {
    // composed of baseOrigin so we inherit its base methods
    baseOrigin
}
```
Can you explain more about how you don't see this setup as making use of both interfaces and composition?

2. I'm certainly opening to refactoring this again in the future, but I don't think the approach I chose warrants an outright rejection. The diff is functional, comes with bug fixes and previously-nonexistent tests, and the "extra" functions explicitly highlight the differences between the necessary configuration for each individual origin, which I find very useful in understanding specific the fundamental differences in our origin backends. However, for those future refactors if you'd rather see a particular programming approach, can you sketch out what that would look like in something I can explore in a [go playground](https://go.dev/play/)? I tried a number of different strategies when I was deciding how best to consolidate this stuff without having to overhaul every spot we use a slice of origin exports, and this was what I found to be the approach that best cuts down on duplication while still making it very easy to walk through the steps common to every origin's setup.
3. After we get this merged, we can open an issue to keep track of further refactoring/development tasks that make use of concrete suggestions arising from (2).

turetske:
@jhiemstrawisc 

That's completely reasonable. Especially since these are purely internal issues. I don't like the "Extra" functions at all, but I'll live with them for now. Most of the work would stay the same, but the function calls would use composition to avoid the above.

I will create a demonstration of what I mean for you in a separate issue.

turetske (approved):

PR #1921, opened 2025-01-18
Simplify Linux install instructions

### LINKED ISSUE #1939: Improve Linux install instructions
The Linux "single binary" install instructions can be drastically simplified given our new download page (which allows us to use `uname` to generate the correct URL).

Take a simplification pass.

bbockelm:
Downloading and making Pelican available is currently a one-liner in bash on almost all distributions (would be curious if someone found one where this doesn't work).

However, we somehow made this one-liner into a daunting multi-page document.  This is my attempt to simplify as much as possible.

### DIFF
--- a/docs/pages/install/linux-binary.mdx
+++ b/docs/pages/install/linux-binary.mdx
@@ -1,73 +1,107 @@
-# Install Pelican as a standalone binary on Linux
+# Install Pelican as unprivileged Linux user
 
-This document explains how to install Pelican on a Linux operating system as a standalone binary.
+This document explains how to install Pelican on a Linux operating system as a standalone binary without any special administrator privileges.
+
+## Quickstart
+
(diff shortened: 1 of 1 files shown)


bbockelm:
May be of note to:
- @CannonLock because the nice work with `dl.pelicanproject.org` makes the simplified links available.  Look how great it is!
- @jhiemstrawisc who has been tidying up documentation otherwise.

aowen-uwmad (approved):
Edits look good.
One-liner works as expected, if `$HOME/.local/bin` is already in the path.
Rest of the instructions should cover those where that may not be true.
(the Ubuntu 22.04 in my Windows Subsystem for Linux does not have that in path by default..)

bbockelm:
Alright, here we go!

I see the same thing on Ubuntu -- that's a bit disappointing.  It's unclear if that's just due to it being a minimal container environment or if it's shared among all the Debian family tree.

PR #1923, opened 2025-01-21
Add Prestage Functionality to Client

Saartank:
This PR is a rebase of PR #1603 on the latest `main` branch. 

### Changes Introduced:
1. Adds the ability to prestage objects (files/directories) to caches:
   - Currently, this is done by downloading the object to the client machine but disregarding the object instead of saving it to the disk.

2. Adds a collections URL override for the `object ls` command:
   - This was originally part of PR #1603.

### DIFF
--- a/client/fed_test.go
+++ b/client/fed_test.go
@@ -976,3 +976,78 @@ func TestTokenGenerate(t *testing.T) {
 		assert.Equal(t, transferResultsDownload[0].TransferredBytes, transferResultsUpload[0].TransferredBytes)
 	}
 }
+
+
+// A test that spins up a federation, and tests object get and put
+func TestPrestage(t *testing.T) {
+	viper.Reset()
+	server_utils.ResetOriginExports()
+	fed := fed_test_utils.NewFedTest(t, bothAuthOriginCfg)
+
+	te, err := client.NewTransferEngine(fed.Ctx)
+	require.NoError(t, err)
+
+	// Other set-up items:
+	// The cache will open the file to stat it, downloading the first block.
+	// Make sure we are greater than 64kb in size.
+	testFileContent := strings.Repeat("test file content", 10000)
+	// Create the temporary file to upload
+	tempFile, err := os.CreateTemp(t.TempDir(), "test")
+	assert.NoError(t, err, "Error creating temp file")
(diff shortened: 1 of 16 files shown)


client/fed_test.go
```diff
@@ -976,3 +976,98 @@ func TestTokenGenerate(t *testing.T) {
...
+
+		// assert.Equal(t, -1, age)
+
+		// Temporarily asserting age as 0 to avoid unused variable error.
+		assert.Equal(t, 0, age)
+
+		age, size, err = tc.CacheInfo(fed.Ctx, innerFileUrl)
+		require.NoError(t, err)
+		assert.Equal(t, int64(len(testFileContent)), size)
+		// Temporarily commenting out this assertion because a GET request
```
turetske, line 1047: Why these comments? Is it temporary? Just keep the changes assert and remove the confusing comments.

client/handle_http.go
```diff
@@ -1821,6 +1990,10 @@ func downloadHTTP(ctx context.Context, te *TransferEngine, callback TransferCall
 		unpacker = newAutoUnpacker(dest, behavior)
 		if req, err = grab.NewRequestToWriter(unpacker, transferUrl.String()); err != nil {
 			return 0, 0, -1, "", errors.Wrap(err, "Failed to create new download request")
+		} else if dest == "/dev/null" {
```
turetske, line 1993: In Brian's version, this is an else for the `if transfer.PackOption != "" {` (one level up) and not `if req, err = grab.NewRequestToWriter(unpacker, transferUrl.String()); err != nil {` so I'm worried this might be a rebase error.

turetske (changes requested):

client/fed_test.go
Saartank: Currently, due to a limitation in Xrootd, a GET request is used in the `CacheInfo` function, which ideally should have been a HEAD request. Once that limitation is resolved, the `CacheInfo` function will use a HEAD request, and the assertion in the test will need to be updated accordingly. When that change is being worked on, this comment will provide a reminder and background explanation of the necessary modification.

Saartank: Based on an offline discussion: provided a more clear message in the comments.

turetske (approved):

PR #1926, opened 2025-01-22
Install custom XRootD directly from Kojihub RPMs

jhiemstrawisc:
While we wait for certain riskier patches to be ingested upstream (either by XRootD itself or by the OSG repos), we need to install XRootD directly from Kojihub RPMs.

This commit primarily does that, with some other cleanup along the way, such as parameterizing versioning of various installed dependencies.


matyasselmeci (approved):
LGTM. I think there's improvements we can make in the future but that shouldn't hold this up.

PR #1927, opened 2025-01-22
Move `ticker.Stop()` into goroutine

bbockelm:
In the `Launch*` family of methods, if we stop the ticker outside the launched thread, it will never fire in the goroutine and hence never actually update.

From within the goroutine, it'll stop at shutdown and clean up the resources appropriately.

This needs to get backported to the v7.11 and v7.12 series.


jhiemstrawisc (approved):
LGTM, apologies for not catching this bug in my initial review! I spun up a federation and was able to observe that both authfiles and scitokens config were updated after 2 minutes.

PR #1934, opened 2025-01-23
Skip Log File for Interactive Commands

Saartank:
This PR adds changes to skip the creation of a log file for certain interactive commands:

- `config`
- `credentials`
- `generate`
- `namespace`
- `object`

These commands are designed to be interactive and should not create a log file as a side effect.


bbockelm (commented):
Why not just move the log file generation code to the `InitServer` routines?

We already have sorted all the sub-commands into `InitClient` and `InitServer`; it seems counter-productive to sort them a second time into "interactive" and "not interactive", especially as setting up the logging appears to be one of the last things done in `InitConfig`.

If you want to get fancy, you buffer all the log messages until the logging configuration is done (although you'd have to be careful to emit them in case of an error) -- but even just printing them to stdout for the duration between `InitConfig` and `InitXYZ` may suffice.

Saartank:
Based on the discussion this morning, we would not like to make `InitConfig` more complex. A better way moving forward would be to create the log file in `InitServer`. However, to achieve this, we would have to buffer all the logs generated before `InitServer` is executed.

Another problem I see with this approach is determining what to do if a fatal error occurs before the log buffer is emptied. How do we decide if the log buffer should be written to a file? These questions would have to be thought about.

Saartank:
A rework based on **log buffering** is currently being developed as part of the following development branch:

[fb10809 — Commit Link](https://github.com/Saartank/pelican/commit/fb10809b1f09a8b319070c068b151ecd9ca09677)

PR #1935, opened 2025-01-23
Plumb Lotman config through to XRootD purge plugin

jhiemstrawisc:
This finally gives Lotman some teeth in the cache. Now that we're running the correct patched version of xrootd, we can configure XRootD to use the lotman purge plugin.


jhiemstrawisc:
For the reviewer, I think the bar for approval of this PR is that:
a) When Lotman is disabled, everything functions as normal
b) When Lotman is enabled and the appropriate extra configuration is provided, the cache fully initializes and loads the purge plugin.

To test these, you'll need to make sure you're running in a dev container that was built from a very recent `latest-itb` tag (within the last two days). We only got the new XRootD patches and purge plugin installed in our containers this week, and both are required for anything to work!

You'll also need to make sure you're running a linux `pelican-server` binary, which is the only place this lotman code doesn't get compiled out.

After that, you should be able to set:
```
Cache:
  EnableLotman: true
```
in your configuration. Doing so must result in an extra line of xrootd configuration being populated, which will look something like
```
pfc.purgelib libXrdPurgeLotMan.so /var/lib/pelican del exp opp ded
```
If you want to change the location of the lotman database from its default `/var/lib/pelican` you can further add to your config:
```
Lotman:
  DbLocation: <your path>
```
which should update the xrootd configuration.

If this extra xrootd line doesn't appear when lotman is enabled, or it appears when lotman _isn't_ enabled, something is wrong. The presence of this extra configuration must not prevent the cache from fully initializing/booting.

If all that happens, I think I'm happy at this stage. There may be some extra future work to validate certain types of parameters for Lotman, but as you'll see in a comment, I'm punting on that because it's complicated and XRootD will handle it anyway.

config/config.go
```diff
@@ -1326,18 +1286,63 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
...
+		if lwmIsAbs == hwmIsAbs && lowWm >= highWm {
+			// Use config strings in error to present the configured values (as opposed to whatever conversion we get from validation)
+			return errors.Errorf("invalid Cache.HighWaterMark/LowWaterMark values. The high watermark must be greater than the low "+
+				"watermark. Got %s (low) and %s (high)", param.Cache_LowWatermark.GetString(), param.Cache_HighWaterMark.GetString())
+		}
+	}
+
+	if param.Cache_FilesBaseSize.IsSet() || param.Cache_FilesNominalSize.IsSet() || param.Cache_FilesMaxSize.IsSet() {
+		// Must have high/low watermarks
+		if !param.Cache_LowWatermark.IsSet() || !param.Cache_HighWaterMark.IsSet() {
```
Saartank, line 1306: These checks are futile because we set the defaults for `Cache.LowWatermark` and `Cache.HighWaterMark`  (from `defaults.yaml`), `IsSet()` will always return true for them.

utils/utils.go
```diff
@@ -153,3 +160,55 @@ func MapToSlice[K comparable, V any](m map[K]V) []V {
...
+			return val * 1024 * 1024 * 1024, true, nil
+		case 't':
+			return val * 1024 * 1024 * 1024 * 1024, true, nil
+		default: // Should never hit this default because of the check above, but can't hurt to set a default
+			return 0, false, errors.Errorf("watermark value %s for config param '%s' is missing a valid byte suffix (k|m|g|t).", wmStr, paramName)
+		}
+
+	}
+
+	log.Infof("Interpreting watermark value %s for config param '%s' as a percentage.", wmStr, paramName)
```
Saartank, line 208: Would it be a good idea to move these logs to debug, or perhaps not log them at all? Especially since these logs are generated even when the user has not explicitly set these configuration parameters, as they originate from our defaults in `default.yaml`.

Saartank (changes requested):
Found some minor problems and created issues: #1955 and #1956. These can be handled separately and do not need to block this PR from being included in 7.13 release.

config/config.go
jhiemstrawisc: We discussed in person, but I'm including my reply here for posterity. I think this check is very inexpensive and protects us to some degree from future changes in our setup. For example if we someday change the way we handle default configuration and these somehow become unset, this check will give us a clear indication we've broken something.

utils/utils.go
jhiemstrawisc: I'm not opposed to single log lines at startup that are mostly ignored _until_ they become relevant. If you'd like, I can switch this to a debug statement.

Saartank: No lets keep them as they are as we discussed, in person, and move forward with 7.13.

Saartank (approved):

PR #1936, opened 2025-01-24
Fixes for the director tests

### LINKED ISSUE #1902: Allow public read of monitoring metric
While @patrickbrophy works on the API token to access parts of the monitoring, a short-term fix (which would be extremely useful for @williamnswanson) would be to provide a boolean (defaults to Off) that makes the monitoring metric publicly accessible.

That is, today `curl https://osdf-registry.osg-htc.org/api/v1.0/metrics/health` from our monitoring systems results in a 403.  Add a small knob to optionally make that public and make the pelican services easier to monitor.

### LINKED ISSUE #1930: Consistently use DiscoveryUrl for federation issuer
We've found that the director and origins disagree on the issuer name to use for a "federation token".  Particularly, it seems the director uses its external web URL to generate tokens -- causing failures when the discovery URL and director URL are not the same thing!

bbockelm:
This PR contains one fix for the director test (solving confusion between the discovery URL and the director URL -- the director tests should have been using the discovery URL).

An integration test is *not* included but bugfixes are included that will eventually be necessary for the integration test.  In discussion with @jhiemstrawisc, we realized that the "federation-in-the-box" tests are failing because the scitokens.cfg is written prior to XRootD startup but the correct audience URL is not known until _after_ startup.  Our best idea for a fix is to implement an "automatic" mode for XRootD to have it determine the correct audience (which will require an XRootD release and hence take some time).

In the meantime, the bugfixes for a future test include:
1. Correct temporary directory creation in the configuration code (this was causing state to leak between tests).
2. Make the "health" metrics publicly available (simplifies the test as the test then doesn't require a token to verify it worked).
3. Allow the director test to run when the self test is disabled (previously, if the self test is disabled and the director test is enabled, then the director test would be broken due to a missing export).

Fixes #1902
Fixes #1930

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1068,10 +1068,15 @@ func SetServerDefaults(v *viper.Viper) error {
 		v.SetDefault(param.Shoveler_QueueDirectory.GetName(), filepath.Join(configDir, "shoveler/queue"))
 		v.SetDefault(param.Shoveler_AMQPTokenLocation.GetName(), filepath.Join(configDir, "shoveler-token"))
 
-		runtimeDir := filepath.Join(os.TempDir(), "pelican-xrootd-*") // Construct the expected path
-
+		var runtimeDir string
 		if v == viper.GetViper() && os.Getenv("XDG_RUNTIME_DIR") != "" {
 			runtimeDir = filepath.Join(os.Getenv("XDG_RUNTIME_DIR"), "pelican")
+		} else {
+			var err error
+			runtimeDir, err = os.MkdirTemp("", "pelican-xrootd-*")
+			if err != nil {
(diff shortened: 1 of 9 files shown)


bbockelm:
@jhiemstrawisc - do you understand why the PR validation is failing?  It appears to correctly link things in the GitHub UI.

jhiemstrawisc:
We spoke in person, but for posterity I'm typing up the reason:
The github PR history API does not treat issues linked with keywords like "fixes" or "closes" the same way it treats manually linking issues. The advice for the team has been to avoid keyword linking until we can figure out a solution. I think you can fix this by editing your comment to remove the links and manually linking.

turetske (approved):
LGTM!

PR #1946, opened 2025-01-27
Don't build dev images on git tags, tag im with latest-itb and sha256

jhiemstrawisc:
The previous action had a side effect such that whenever someone created a new tag for some release/release candidate, a new container with latest-itb would get pushed to Harbor, but from the `dev.Dockerfile` as it exists for that tag.

This essentially means _some_ but not all latest-itb tags weren't actually tracking main, but other branches instead.

After a chat with Sarthak & Emma, we don't see much use in building dev containers that aren't produced from the main branch, since we don't do dev work like that on our release series branches.

This commit removes tags as a trigger from the workflow, and only uses the image's sha256 and latest-itb as tags

### DIFF
--- a/.github/workflows/build-dev-container.yml
+++ b/.github/workflows/build-dev-container.yml
@@ -3,10 +3,6 @@ name: Build and Push Dev Image
 on:
   pull_request:
   push:
-    tags:
-      # only build and publish container on v7.0.0 and up
-      - v[7-9]\.[0-9]+\.[0-9]+-**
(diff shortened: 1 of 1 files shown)


bbockelm (approved):
LGTM.

To be clear, I do want to see per-branch production containers -- and after Brian A's overhaul of the container install we can revisit whether per-branch development containers makes sense as well (it's reasonable to argue you want this for the occasional patch release testing).

But the current behavior is definitely wrong and this definitely fixes it!

Issue #1778, opened 2024-11-27
Map at `/view/director/map/` could benefit from extra labels


jhiemstrawisc:
I just discovered we have this map (e.g. https://osdf-director.osg-htc.org/view/director/map/) and I love that we display this info.

Taking inspiration from the map at https://osg-htc.org/docs/data/stashcache/overview/, could we also label each of the origins/caches with a different color and provide a count of each in the legend?

CannonLock:
https://github.com/PelicanPlatform/pelican/pull/1947

PR #1947, opened 2025-01-27
Add Cache and Server Cards to Map

CannonLock:
- Add more data onClick to the map server components

**Redacted to prevent scope creep** (_This is ready, but I would like to tack on a couple items before merge and need the director data to work first_)

https://github.com/PelicanPlatform/pelican/issues/1778

### DIFF
--- a/web_ui/frontend/components/Map/ServerCard.tsx
+++ b/web_ui/frontend/components/Map/ServerCard.tsx
@@ -1,0 +1,43 @@
+import { Box, Grid, Typography } from '@mui/material';
+import { InformationSpan } from '@/components';
+import { ServerCapabilitiesTable } from '@/components/ServerCapabilitiesTable';
+import React from 'react';
+import { ServerDetailed, ServerGeneral } from '@/types';
+
+const ServerCard = ({
+  server,
+}: {
+  server?: ServerGeneral | ServerDetailed;
+}) => {
+  // If there is no server, return null
+  if (!server) {
+    return null;
+  }
+
+  return (
+    <Box>
+      <Grid container spacing={1}>
+        <Grid item xs={12}>
+          <InformationSpan name={'Type'} value={server.type} />
+          <InformationSpan name={'Status'} value={server.healthStatus} />
+          <InformationSpan name={'URL'} value={server.url} />
+          <InformationSpan
+            name={'Longitude'}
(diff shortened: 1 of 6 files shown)


web_ui/frontend/components/Map/ServerCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Box>
+      <Grid container spacing={1}>
+        <Grid item xs={12}>
+          <InformationSpan name={'Type'} value={server.type} />
+          <InformationSpan name={'Status'} value={server.healthStatus} />
+          <InformationSpan name={'URL'} value={server.url} />
+          <InformationSpan
+            name={'Longitude'}
+            value={server.longitude.toString()}
```
h2zh, line 26: I vaguely remember from my React class that in JSX, React can automatically convert numbers to strings when rendering them. So `toString()` here and the one on line 30 can be waived.

h2zh (changes requested):
Looks good in general, few things could be done to further improve!

web_ui/frontend/components/Map/ServerCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+          />
+          <InformationSpan
+            name={'Latitude'}
+            value={server.latitude.toString()}
+          />
+        </Grid>
+      </Grid>
+      {server.type == 'Origin' && (
+        <Box sx={{ my: 1 }}>
+          <ServerCapabilitiesTable server={server} />
```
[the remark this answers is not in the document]
CannonLock: I fix this here: https://github.com/PelicanPlatform/pelican/commit/9a178b36d792a44ff405718f4cb7097281b7f96e

web_ui/frontend/components/Map/ServerCard.tsx
CannonLock: Yeah, it does but I have the InformationSpan typed like so. 

```
export const InformationSpan = ({
  name,
  value,
  indent = 0,
}: {
  name: string;
  value: string;
  indent?: number;
}) => {
```

So I just cast it to pass the type check. To be fair I could accept quite a few more values and have more lenient typing I suppose.

h2zh (approved):
🚀

Issue #1866, opened 2025-01-07
Can't tell what plugin versions are installed in the image


matyasselmeci:
Some of the plugins (e.g. xrootd-s3-http, and xrdcl-pelican in the dev container (and prod containers until #1857)) are built from source, and I haven't figured out how to tell from the finished image what versions were used - I had to go back to the logs.

At minimum, write a text file called `/installed-versions.txt` that lists the versions of the plugins (and other software built from source) that were installed.

bbockelm:
@matyasselmeci - instead of adding a separate software discovery mechanism, could we just build the RPM from scratch and install it?  This way we could use the "normal" tools to query installed software versions.  One less Weird Thing.

matyasselmeci:
Yep, even better.

matyasselmeci:
A reminder from #1897 is that the RPMs should be installed into the final container, not the intermediate build container that we copy files from and discard, otherwise we lose the changes to the RPMDB.

brianaydemir:
In https://github.com/PelicanPlatform/pelican/commit/72a371c1a201f4c7d9bdcbf1df49ef8c7ed8e71e (#1972), I updated the container image build process to install all XRootD packages from RPMs, either ones that already exist in various repositories or ones that were built during the container image build process itself.

```
> docker run -it --rm --entrypoint /bin/bash pelican-origin:local
[root@02ce0cff6845 pelican]# rpm -q xrootd-s3-http
xrootd-s3-http-0.2.1-1.el9.x86_64
[root@02ce0cff6845 pelican]# rpm -q xrdcl-pelican
xrdcl-pelican-1.0.5-1.osg24.el9.x86_64
```

PR #1959, opened 2025-01-28
Install pelican dependencies in final stage container

jhiemstrawisc:
We've run into a lot of issues with the way we install dependencies in our production container, including the inability to determine versions for what's installed, and shared objects that are copied without bringing along their dependencies.

Instead, we should just install whatever we can directly as an RPM into the final-stage container. This solves both issues in one fell swoop of a Pelican's beak.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -121,8 +121,6 @@ done
 yum install -y "${package_urls[@]}"
 EOT
 
-# Koji won't have the xrootd-multiuser package, so that still gets installed from the OSG repos
-RUN yum install -y --enablerepo=osg-testing xrootd-multiuser
 ########################
 # End Dependency Build #
 ########################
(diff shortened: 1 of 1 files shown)


jhiemstrawisc:
This also partially addresses some of the concerns raised in #1866.

brianaydemir (approved):
LGTM.

brianaydemir:
Brief overview of my testing:

1. Check out the PR locally.
2. `docker build --target final-stage -t pelican:latest -f images/Dockerfile .`
3. Start my bespoke Docker Compose setup using this image.
4. In the dev environment container, run `pelican serve --module director,registry,origin,cache --config config.yaml`.
5. Log into the Web UI, select the registry, and accept all the pending registrations.
6. In the user environment container:

    * Try an `object ls`.
    * Try an `object get` as a direct read.
    * Try an `object get` via the cache, using `-c`. (Cache miss.)
    * Try the same exact thing. (Cache hit.)
    * Try an `object get`.
    * Try an `object put`. (Requires a token.)

Files:

* [docker-compose.txt](https://github.com/user-attachments/files/18588353/docker-compose.txt)
* [config.txt](https://github.com/user-attachments/files/18588352/config.txt)

Issue #1971, opened 2025-01-30
Run tests in the production container


bbockelm:
The GHA unit tests should run in the production container, not the development container as is [done here](https://github.com/PelicanPlatform/pelican/blob/main/.github/workflows/test-template.yml#L20).  This is to ensure there's not something special in the development container (such as a hidden dependency on the development RPMs) that causes Pelican to be non-functional in production.

*Note* the ongoing discussions about containers and branching.  It may be that the test template needs to have a parameter that allows us to use a branch-specific production container (so, say, the v7.12 branch is tested on the v7.12 container, not whatever random things we are preparing in `main`).

Issue #1964, opened 2025-01-29
Revamp Dockerfile


bbockelm:
Let's cleanup the container image definitions!

Right now, the development and production container are built alongside each other -- meaning the Dockerfile's have about 90% overlap and there's danger of getting out-of-sync.  We saw this recently when the production container was tweaked to install from RPM but ended up subtly breaking xrdcl-pelican (see https://github.com/PelicanPlatform/pelican/pull/1959).

It's unclear whether some container cleanup would have actually avoided that problem but it did highlight some best practices.

To the extent possible, let's cleanup!  Rebase the development container to inherit from the production container.  The development container, to the best of my understanding, should be the production container plus a few extra RPMs necessary for compiling, testing, and easier debugging of Pelican.

PR #1972, opened 2025-01-31
Clean up the container image build and testing process

brianaydemir:
The goal here is to address #1964, and hopefully set us up for addressing #1971.

The revamp of `images/Dockerfile` here aims to produce three sets of images:

1. Images for end-users.
2. An image for testing the above.
3. An image for developers.

These sets correspond to the results of the following build stages:

1. `director`, `registry`, `origin`, `cache`, and their `-osdf` variants.
2. `testing-container`.
3. `development-container`.

The build process has been reworked to:

- Produce end-user images that do not contain excessive amounts unnecessary software. For example, only the origin images contain an OA4MP install.
- Properly target both `linux/amd64` and `linux/arm64`. Except for `scitokens-oauth2-server`, everything is already capable of supporting both platforms; we just need to stop making assumptions. (And the bits we need from `scitokens-oauth2-server` should be platform agnostic, anyway.)
- **(New!)** Explicitly build an image for testing the build artifacts by basing the testing environment, by default, _on those exact build artifacts_. The default `osdf-origin` base for this image contains _all_ of the software for _all_ Pelican's components. By copying in the build directory from the build stage, the exact source code used and all the resulting build artifacts are present and available for examination.
- **(New!)** Base the dev container image, by default, on the production container images (specifically `osdf-origin`).


images/Dockerfile
```diff
@@ -16,350 +16,499 @@
...
 ARG BASE_OS=el9
+ARG OSG_SERIES=23
+ARG OSG_REPO=release
+
+# For compiling Pelican.
+ARG GO_VER=1.21.6
+ARG NODEJS_VER=20
+
+# For compiling LotMan.
+ARG JSON_VERSION="v3.11.3"
```
bbockelm, line 29: Looks like the OSG koji has basically this version (`v3.11.2`, at least) already built and available.  Let's use that instead.

images/Dockerfile
```diff
@@ -16,350 +16,499 @@
...
+ARG JSON_SCHEMA_VALIDATOR_VERSION="2.3.0"
+ARG LOTMAN_VERSION="v0.0.4"
```
bbockelm, line 31: Similarly, looks like we have lotman available in Koji as well: https://koji.osg-htc.org/koji/buildinfo?buildID=17613.  Another thing we could skip building.

It's less clear about `json-schema-validator`; there's a version in there but it's somewhat older.

Regardless, let's pin the build version where we can.

brianaydemir:
Notable changes:

- `images/dev.Dockerfile` is no more, having been folded into `images/Dockerfile`. Combined with a collection of new build `ARG`s, it should be substantially easier to ensure that production images and the `dev-container` install the same versions of software dependencies.

- There are new build `ARG`s for controlling whether some XRootD plugins are installed from source. Furthermore, when installing from source, we simply build an RPM and install it. As a result, the `Dockerfile` doesn't need to duplicate the build script, and the final container image has a record of which version of the plugin is installed.

- The GitHub Action and image build process are fairly aggressive about caching what they can _within_ the run, but we're no longer trying to share that cache _between_ runs. Saving and loading the build cache is not free, and saving the build cache will outright fail if it exceeds 10GB.

.github/workflows/build-prod-container.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and Push Prod Image
+name: Build and Push Container Images
```
jhiemstrawisc, line 1: Since the action no longer builds only the prod container, can you update the name of the file from "build-prod-container.yml" to sometihng like "build-images.yml"?

.github/workflows/build-prod-container.yml
```diff
@@ -16,82 +16,56 @@ on:
...
-
-      # Free disk space
-      - name: Free Disk space
-        shell: bash
+      - name: Determine build parameters
+        id: parameters
         run: |
-          sudo rm -rf /usr/local/lib/android  # will release about 10 GB if you don't need Android
+          PUSH_DEV=${{ github.repository == 'PelicanPlatform/pelican' && github.ref == 'refs/heads/main' }}
+          PUSH_SERVER=${{ github.repository == 'PelicanPlatform/pelican' && startsWith(github.ref, 'refs/tags/') }}
```
jhiemstrawisc, line 39: If I'm reading correctly, this would push for _any_ tag, no? I'm not sure if this is really worth fixing (I'll leave that to you), but should we only set `PUSH_SERVER` to true if the tag matches our semantic versioning scheme?

jhiemstrawisc (changes requested):
Overall, this looks like a really solid refactor of our container spaghetti, and I'm excited to see what it does in the wild.

My change requests are all very minor nitpicks. @brianaydemir, how do you feel about also adding the flow diagram of the container build strategy you showed me last week? That was really useful for me to understand the "why" behind a lot of these changes, and I think it'd be useful for others to have access to it.

images/Dockerfile
```diff
@@ -16,350 +16,614 @@
...
-####################
+# xrootd's UID and GID here are effectively set in stone because of the
+# existing data out in the wild that is owned by these IDs.
+RUN groupadd -g 10940 xrootd \
+    && useradd -u 10940 -g 10940 -d / -s /sbin/nologin xrootd
 
-FROM pelican-base AS director
+# We need to pin the install for many of the RPMs to "Koji" until all of
+# Pelican's patches are ingested into the OSG repositories.
+ARG XROOTD_VERSION="5.7.2"
```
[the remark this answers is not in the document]
brianaydemir: Done

.github/workflows/build-prod-container.yml
brianaydemir: Done.

.github/workflows/build-prod-container.yml
brianaydemir: The workflow is triggered only under the conditions listed at the top of the file. The only tags that trigger the workflow are of the form `v{semver}`, assuming the regex's are correct.

jhiemstrawisc: Ah, missed that

jhiemstrawisc (approved):
Love it.

PR #1980, opened 2025-02-03
Fix Director Capabilities Display

CannonLock:
- No longer display mis-representative cache capabilities
- Remove Cache capabilities
- Update Namespace Capabilities to list Caches that can host that namespace


web_ui/frontend/components/NamespaceCapabilitiesTable.tsx
```diff
@@ -55,41 +68,62 @@ export const NamespaceCapabilitiesTable = ({
...
 };
+
+const NamespaceCapabilitiesRow = ({
+  server,
+  namespace,
+}: {
+  server: ServerDetailed;
+  namespace: DirectorNamespace;
+}) => {
+  if (server.type === 'Origin') {
```
h2zh, line 115: Do we need to do the case formatting here?

I'm curious why you use `as` to perform type assertion on line 80, but use `===` here?
```
server.type.toLowerCase() as 'origin' | 'cache'
```

h2zh (approved):
LGTG

PR #1981, opened 2025-02-03
Allow for Shoveler messages to be understood by monitoring packet handler. 

### LINKED ISSUE #1916: Prometheus Monitoring data not showing up when the Shoveler is enabled for the component
Pelican Version:
sh-5.1# osdf-server --version
Version: 7.12.0
Build Date: 2025-01-14T22:55:31Z
Build Commit: b9d9f00b760544d29f2408d597f4b9d80c885b98
Built By: goreleaser

Pelican Service:
- [ ] Client
- [ ] Plugin
- [X] Registry
- [X] Director
- [X] Origin
- [X] Cache

Issue first detected in a cache, but we think it may affect any component trying to use both Shoveler and Prometheus monitoring at the same time.

**Describe the bug**
Prometheus doesn't seem to be getting any metrics for an active cache, but *only* if the Shoveler is also enabled and running at the same time.

**To Reproduce**
Enable both Shoveler and Prometheus monitoring for a component (under some load), notice that the component's `/metrics` endpoint doesn't have a value for `xrootd_server_bytes`.

**Expected behavior**
There should be a value for `xrootd_server_bytes`, since the component *is* in use.

 - OS: Linux

**Additional context**
We have a hypothesis that the Shoveler and Prometheus are scraping the same buffer for their metrics, and the Shoveler is clearing everything out before Prometheus can see and store it.

patrickbrophy:
This PR fixes [issue #1916](https://github.com/PelicanPlatform/pelican/issues/1916). During investigation with the local federation, we discovered that the XRootD monitoring packet handler wasn’t processing shoveler messages correctly. The shoveler was sending JSON-encoded packets, which the existing code couldn’t interpret. This update ensures that shoveler packets are now correctly parsed and understood.

### DIFF
--- a/metrics/shoveler.go
+++ b/metrics/shoveler.go
@@ -238,15 +238,6 @@ func LaunchShoveler(ctx context.Context, egrp *errgroup.Group, metricsPort int) 
 	// Create the UDP forwarding destinations
 	var udpDestinations []net.Conn
 
-	// By default, forward to metrics endpoint for Prometheus metrics
-	// TODO: integrate metrics to shoveler and remove the forwarding
-	metricsEndpoint := fmt.Sprint("127.0.0.1:", metricsPort)
-	udpConn, err := net.Dial("udp", metricsEndpoint)
-	if err != nil {
-		shovelerLogger.Warningln("Unable to connect to metrics endpoint:", metricsEndpoint, err)
-	}
-	udpDestinations = append(udpDestinations, udpConn)
-
 	if len(config.DestUdp) > 0 {
 		for _, dest := range config.DestUdp {
 			udpConn, err := net.Dial("udp", dest)
@@ -285,6 +276,12 @@ func LaunchShoveler(ctx context.Context, egrp *errgroup.Group, metricsPort int) 
 				continue
 			}
 			shoveler.PacketsReceived.Inc()
+
+			err = handlePacket(buf[:rlen])
+			if err != nil {
+				shovelerLogger.Errorln("Failed to handle packet:", err)
+				continue
+			}
 
 			if config.Verify && !shoveler.VerifyPacket(buf[:rlen]) {
 				shoveler.ValidationsFailed.Inc()
--- a/metrics/shoveler.go
+++ b/metrics/shoveler.go
@@ -184,7 +184,7 @@ func packageUdp(packet []byte, remote *net.UDPAddr) ([]byte, error) {
 	return b, nil
 }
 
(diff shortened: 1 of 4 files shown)


metrics/shoveler.go
```diff
@@ -286,6 +277,12 @@ func LaunchShoveler(ctx context.Context, egrp *errgroup.Group, metricsPort int)
 			}
 			shoveler.PacketsReceived.Inc()
 
+			err = handlePacket(buf[:rlen])
+			if err != nil {
+				shovelerLogger.Errorln("Failed to handle packet:", err)
```
jhiemstrawisc, line 282: Can you add a little more context to these `Failed to handle packet` log lines? I'm not 100% sure what they should say, but it'd be useful to log a) what pelican component should be responsible for handling the packet (shoveler? metrics engine? Not sure what's correct to say) and b) something about the packet we were expecting. Maybe the log should read something like:
```
shoveler.ErrorLn("Pelican failed to handle a monitoring UDP packet from the shoveler:", err)
```

metrics/xrootd_metrics.go
```diff
@@ -400,23 +400,27 @@ func ConfigureMonitoring(ctx context.Context, egrp *errgroup.Group) (int, error)
...
+	if !param.Shoveler_Enable.GetBool() {
+		go func() {
+			var buf [65536]byte
+			for {
+				// TODO: actually parse the UDP packets
+				plen, _, err := conn.ReadFromUDP(buf[:])
+				if errors.Is(err, net.ErrClosed) {
+					return
+				} else if err != nil {
+					log.Errorln("Failed to read from UDP connection", err)
```
jhiemstrawisc, line 414: Same gist here -- give this a bit more context, e.g. "Failed to read from UDP connection while aggregating xrootd metrics."

metrics/xrootd_metrics.go
```diff
@@ -400,23 +400,27 @@ func ConfigureMonitoring(ctx context.Context, egrp *errgroup.Group) (int, error)
...
+					continue
+				}
+				PacketsReceived.Inc()
+				if err = handlePacket(buf[:plen]); err != nil {
+					log.Errorln("Failed to handle packet:", err)
```
jhiemstrawisc, line 419: Same gist here

jhiemstrawisc (approved):
@patrickbrophy and I talked out of band to discuss testing. I was able to observe the expected results when I built the code and poked around for metrics.

I'll approve this even though I had a few change requests, because I don't think they should hold the PR back if you think the log lines are already sufficient. I know you didn't write them originally, but I think some of these could use a little more context. Otherwise, merge when ready!

PR #1983, opened 2025-02-04
Update xrootd release version to 1.5.purge.osg

Saartank:


jhiemstrawisc (changes requested):
Can you update the dev container as well?

PR #1984, opened 2025-02-04
Use Server.Hostname as the prefix for origin/cache namespace registration

h2zh:
This PR unifies the namespace registration for both origin and cache by using their server hostname (without port) as prefix. Before this PR, origin itself registers with "hostname:port" in its own `Server_ExternalWebUrl` as prefix; cache registers with its `Xrootd_Sitename` as prefix (no port). This inconsistency causes plenty of friction. The change in this PR will be a cornerstone for multiple subsequent tickets for Pelican developers.

When this comes into prod, the origin/cache server installed this Pelican version will send serverAd with the updated prefix. OSDF system admin needs to approve server's new advertisement and remove the old registration db entries if necessary.

### DIFF
--- a/cache/advertise.go
+++ b/cache/advertise.go
@@ -42,7 +42,7 @@ type (
 )
 
 func (server *CacheServer) CreateAdvertisement(name, originUrl, originWebUrl string) (*server_structs.OriginAdvertiseV2, error) {
-	registryPrefix := server_structs.GetCacheNS(param.Xrootd_Sitename.GetString())
+	registryPrefix := server_structs.GetCacheNS(param.Server_Hostname.GetString())
 	ad := server_structs.OriginAdvertiseV2{
 		Name:           name,
 		RegistryPrefix: registryPrefix,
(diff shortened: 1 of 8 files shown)


turetske (approved):

jhiemstrawisc:
Did one of you test how this functions when we require admin approval for a namespace? I'm worried that as is, this will invalidate all of the existing registrations we have for OSDF, requiring William to re-approve every namespace.

jhiemstrawisc:
Or potentially worse, could someone steal ownership/registration of a cache/origin namespace by being the first one to submit an advertisement with the new scheme?

h2zh:
Yes, I tested the new prefix admin approval process when running Pelican locally.

William won't need to re-approve all namespaces, but just the origin namespaces (whose prefix starts with `/origins/...`) and the cache namespaces (whose prefix starts with `/caches/...`) (approximate 49, 29 entries respectively, according to the [prod](https://osdf-registry.osg-htc.org/api/v1.0/registry_ui/namespaces)).

I don't think the cache/origin namespace hijack will happen because the workflow will be:  server installed this new Pelican version sends serverAds (e.g. "/origins/hawaii-its-origin.nationalresearchplatform.org") -> OSDF admin checks it and approves -> OSDF admin removes the old registration (e.g. "/origins/hawaii-its-origin.nationalresearchplatform.org:8080"). During this process, the old prefix will be replaced by the designated new prefix. OSDF admin can compare the new and old registration and use other methods to verify.

I hope this could address the concerns @jhiemstrawisc

PR #1985, opened 2025-02-04
Add federation token endpoint in Director, and implement routine for caches to fetch one

jhiemstrawisc:
I believe this implementation accomplishes the goal (and I can observe it working when I set things up locally), but there's still a lot of unit testing to do. I worked myself into a hole with the approach that now makes testing very hard due to cyclic imports.

Rather than finish all the refactoring to fix that up, I decided I'd open a draft to have others comment on the approach first. If the token management/verification checks out, I'll end up spending some time untangling things so I can run this through a proper test that spins up an entire fed-in-a-box.

Note that I started down this road because I assume Origins will also need fed tokens eventually, especially for third-party copies from origin to origin.

### DIFF
--- a/cache/advertise.go
+++ b/cache/advertise.go
@@ -271,7 +271,7 @@ func getTickerRate(tok string) time.Duration {
 
 func LaunchFedTokManager(ctx context.Context, egrp *errgroup.Group, cache server_structs.XRootDServer) {
 	// Do our initial token fetch+set, then turn things over to the ticker
-	tok, err := server_utils.GetFedTok(ctx, cache)
+	tok, err := server_utils.CreateFedTok(ctx, cache)
 	if err != nil {
 		log.Errorf("Failed to get a federation token: %v", err)
 	}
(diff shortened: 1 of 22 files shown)


jhiemstrawisc:
Okay, I think I provided a sensible set of tests in the latest round of commits. @h2zh, let's discuss this in person to make sure you understand the overall goals and scope of this PR -- I think that'll be important for a good review.

cache/advertise.go
```diff
@@ -212,3 +220,132 @@ func (server *CacheServer) GetNamespaceAdsFromDirector() error {
...
+				// Once again, parse the token, use it to set the next ticker fire
+				// while also building in a circuit breaker to set a min ticker rate
+				newTickerRate := getTickerRate(tok)
+				if newTickerRate != tickerRate {
+					fedTokTicker.Reset(newTickerRate)
+					tickerRate = newTickerRate
+				}
+
+				// Set the token in the cache
+				err = server_utils.SetFedTok(ctx, cache, tok)
```
h2zh, line 341: I feel like we never actually use the locally-saved fedtoken? We should use it by calling GetFedTokLocation in GetFedTok or somewhere?

h2zh (changes requested):

cache/advertise.go
jhiemstrawisc: You're correct, we don't actually use it anywhere inside core Pelican. However, Brian [worked on integrating](https://github.com/PelicanPlatform/xrdcl-pelican/pull/24) this token with `xrdcl-pelican`, which picks it up and forwards it to origins. His pull request assumes there's a token on disk, and this pull request puts that token on disk.

PR #1988, opened 2025-02-05
Fix xrootd log level mapping bug and inherit from Pelican log level

jhiemstrawisc:
This carries two fixes -- the first is for providing an xrootd log level that previously didn't have a definition, e.g. Logging.Origin.Http: trace. The new code should handle mapping that down if able, and up a level if needed

The new code also uses Pelican's configured Logging.Level as the default for the other xrootd values. This lets us crank up logging all at once without having to toggle each directive individually, while still allowing for individual overrides if so desired.


Saartank:
One problem I see is that because the default initialization is outside `SetServerDefaults` (in `config/config.go`), it is not being picked up the `pelican config` commands. Do you think it would be a good idea to move the default initialization code into `SetServerDefaults`?

Saartank (approved):
Looks good, thanks for the changes.

PR #1991, opened 2025-02-05
Pinned the goreleaser version to 2

turetske:
Very simple PR, fixes in the containers as well since they build from source.


jhiemstrawisc (approved):
LGTM. Thanks!

PR #1994, opened 2025-02-06
Remove tls authentication from caches

turetske:
Removed tls authentication from caches. This should stop browsers for asking users to choose a certificate even though it's not needed.

Removed the director x509 namespaces parameter since its no longer relevant.


xrootd/resources/xrootd-cache.cfg
```diff
@@ -62,9 +62,17 @@ pfc.ram 4g
 # Note that I assume we've already validated file directives -- if we have one defined, we _must_ have all three defined, which should be checked before we get here.
 pfc.diskusage {{if .Cache.LowWatermark}}{{.Cache.LowWatermark}}{{else}}0.90{{end}} {{if .Cache.HighWaterMark}}{{.Cache.HighWaterMark}}{{else}}0.95{{end}}{{if .Cache.FilesBaseSize}} files {{.Cache.FilesBaseSize}} {{.Cache.FilesNominalSize}} {{.Cache.FilesMaxSize}}{{end}} purgeinterval 300s
 xrootd.fslib ++ throttle # throttle plugin is needed to calculate server IO load
+<<<<<<< HEAD
```
nconrad, line 65: 👀

turetske: Top!

docs/parameters.yaml
```diff
@@ -1484,6 +1484,16 @@ default: 9.5s
 hidden: true
 components: ["cache"]
 ---
+name: Cache.TLSClientAuth
+description: |+
+  Turns client certificate authentication on or off in xrootd for the HTTPS protocol. When false (default) the
+  HTTPS protocol server will never request a certificate. When true, the HTTPS protocol server will always request
+  a always request a client certificate from the client.
```
jhiemstrawisc, line 1491: Duplicate "always request a"

docs/parameters.yaml
```diff
@@ -1484,6 +1484,16 @@ default: 9.5s
 hidden: true
 components: ["cache"]
 ---
+name: Cache.TLSClientAuth
+description: |+
```
jhiemstrawisc, line 1488: It might be helpful to be more explicit about who "client" and "server" are here - caches are often the server, but they can also be a client.

config/resources/osdf.yaml
```diff
@@ -27,8 +27,3 @@ Federation:
 Registry:
   RequireCacheApproval: true
   RequireOriginApproval: true
-Director:
-  X509ClientAuthenticationPrefixes:
-  - /xenon/PROTECTED/
-  - /user/ligo/
-  - /igwn/
```
jhiemstrawisc, line 34: How can I test that we don't break access for these namespaces? Or is the point that we're intentionally shutting things off for them? Do they know about it?

xrootd/resources/xrootd-cache.cfg
```diff
@@ -65,6 +65,11 @@ xrootd.fslib ++ throttle # throttle plugin is needed to calculate server IO load
 {{- if .Cache.LotmanCfg.Enabled}}
 pfc.purgelib libXrdPurgeLotMan.so {{.Cache.LotmanCfg.LotHome}} {{range  $value := .Cache.LotmanCfg.PurgeOrder}}{{$value}} {{end}}
 {{- end -}}
+{{if .Cache.TLSClientAuth}}
+http.tlsclientauth on
+{{else}}
+http.tlsclientauth off
+{{end}}
```
jhiemstrawisc, line 72: Since this is always configured, can you one-line the setup? Should be within epsilon of
```
http.tlsclientauth {{if .Cache.TLSClientAuth -}} on {{else}} off {{- end -}}
```

jhiemstrawisc (changes requested):

docs/parameters.yaml
```diff
@@ -1715,19 +1725,6 @@ type: bool
...
-
-  If a cache requests an object starting with one of these prefixes, then it will be instructed by the director
-  to use X.509 client authentication if available.
-
-  This setting allows for compatibility with specific legacy OSDF origins and is not needed for new origins.
-type: stringSlice
-default: []
-components: ["director"]
-hidden: true
----
```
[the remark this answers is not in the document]
turetske: Brian's removed the code that uses this from xrootd. I don't like having a parameter that can't do anything. I think that's more confusing to anyone reading the code.

xrootd/resources/xrootd-cache.cfg
turetske: Sure

docs/parameters.yaml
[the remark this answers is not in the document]
jhiemstrawisc: Do you know where X509 compatibility has been removed? I can't imagine xrootd proper has removed X509 auth, so I double checked the latest xrdcl-pelican source, and it looks like it [still supports the underlying features](https://github.com/search?q=repo%3APelicanPlatform%2Fxrdcl-pelican%20509&type=code) that were used by the code removed here.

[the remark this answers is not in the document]
turetske: He removed the code that has us specifiy specific namespaces where the cache will do deferred x509Authentication and the ability to specify those prefixes will be specified in xrootd cache config.

https://github.com/xrootd/xrootd/pull/2269

Now it only allows just turning it on or off, which is what my PR does.

turetske:
@jhiemstrawisc Addressed your PR comments.

The way to test this is to start up a federation including a cache and then open up a browser window and go to https://localhost:<cache-port>/<namespace prefix> and see that the annoying cert popup no longer appears.

jhiemstrawisc (changes requested):
I tested a browser request against main (which did throw the cert popup) and against this diff (which did not throw the popup), so I think the broad goal is accomplished.

The last thing I want to ask about is whether we can handle deprecating the `/listX509ClientPrefixes` API differently, instead of leaving it in place, responding with `HTTP 200` and sending an empty body. I had a quick chat with BrianA and Cannon about best practices for this, and they were of the opinion we should just get rid of the endpoint, which should be okay because the caches that previously used it still handle a 404 okay. Otherwise, an `HTTP 410` feels appropriate, as it signifies "The requested resource previously existed, but has been permanently removed." For simplicity and not letting this kind of thing linger in the codebase, I think I'm in favor of just removing it.

jhiemstrawisc (approved):
LGTM. I think it's good to go once the merge conflicts are cleared and the tests are passing.

PR #1996, opened 2025-02-06
Downtime set by origin and cache

### LINKED ISSUE #1251: Origin and cache should be able to control downtime
The director administrator can force a cache or origin into downtime centrally.

However, I don't want this power to only exist in the hands of the director administrator.  The origin/cache administrator should have a web UI (and a corresponding CLI) where they can put their own service in and out of downtime.

h2zh:
This PR enables origins and caches to configure their downtime periods in both their local database and the Director database. Downtime records persist only on their respective servers and propagate to Director through the existing advertisement mechanism that runs every minute. Director stores these downtime plans in memory from the received advertisements, while the Director database only reflects the servers _currently_ in downtime. This design aligns with our goal of transitioning Director to a stateless service in the near future.

downtime (6)
Diagram **updated** on Mar 13 to address issues brought up in the dev meeting

Rules between OSDF Federation admin and Origin/Cache admin: (checked with William)

- The OSDF admin can set any server to downtime.
- If the Origin/Cache administrator sets its own downtime, then OSDF admin cannot revoke it.

Algorithms explained:
When an origin or cache server sends its advertisement to Director, any included downtime information is evaluated immediately. If the advertisement indicates the server is currently in downtime, Director updates its database to reflect this status. If no active downtime is indicated, Director removes any existing downtime entries for that server from its database (if it was set by server admin, **while downtimes set by OSDF admin won't be affected**), effectively clearing outdated downtime records. In this way, Director doesn't need to persist the downtime set by server admin.

There is a potential optimization here (not implemented): Before traversing the list of downtimes in the server ad and writing to the database, the Director can consider it as a snapshot and compute a hash of it, and then compare it to an in‑memory hash from the previous snapshot for that server. Director proceeds the traversal and DB operations only if their hash values are different. I didn't implement this idea because I think it might be an overkill for this scenario, but I'm open to discussions.

This is a backend change. Frontend UI is not included.

~~Now origin and cache server can set their own downtime in the director by simply sending a request with payload `{"enableDowntime": true/false}` to API `/api/v1.0/downtime`. Then this PR will take care of the setting in the backend.~~


director/cache_ads.go
```diff
@@ -74,31 +74,16 @@ func (f filterType) String() string {
...
-			metrics.PelicanDirectorGeoIPErrors.With(labels).Inc()
-		}
-		log.Debugln("Failed to lookup GeoIP coordinates for host", sAd.URL.Host)
-	}
-
-	if sAd.URL.String() == "" {
-		log.Errorf("The URL of the serverAd %#v is empty. Cannot set the TTL cache.", sAd)
-		return
+// getServerAd returns the existing server ad of the given server url from the cache
+func getServerAd(serverUrl string) (*server_structs.Advertisement, error) {
```
h2zh, line 78: Here the diff looks chaotic but what I do is barely taking the logic of getServerAd out of the recordAd function, so that it could be reuse elsewhere. I'd suggest display the original code and the code in this PR side by side for review.

web_ui/ui.go
```diff
@@ -179,6 +184,152 @@ func getEnabledServers(ctx *gin.Context) {
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		ctx.JSON(resp.StatusCode, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    fmt.Sprintf("Director returned error: %s", directorResponse.Msg),
+		})
+		return
+	}
+	ctx.JSON(http.StatusOK, server_structs.SimpleApiResp{
+		Status: server_structs.RespFailed,
```
CannonLock, line 328:
```suggestion
		Status: server_structs.RespOK,
```

web_ui/ui.go
```diff
@@ -179,6 +184,152 @@ func getEnabledServers(ctx *gin.Context) {
...
+			Msg:    "Failed to parse Federation.DirectorURL: " + directorUrlStr,
+		})
+	}
+
+	directorUrl.Path = "/api/v1.0/director/downtime"
+
+	// Construct the request body
+	downtimeRequest := server_structs.DowntimeRequest{
+		ServerUrl:      serverUrl,
+		Hostname:       hostname,
```
CannonLock, line 241: Why pass in Hostname if you don't use it?

CannonLock (changes requested):
Two small items and one big one that might result in some refactoring. 

It is unwieldy to work with one off endpoints like this that only are there to mark a boolean. Is there any way that we can design this more RESTfully so that we are instead POSTing a body like {downtime: true/false} to /director/servers/<ServerName>.

This way if we ever want to adjust a servers state we can just tack it on the same POST request rather then creating a whole new endpoint.

I realize this is not easy as the ServerName in this case is a url which is a unwieldy url key. Is there anyway that we can get the ServerAd from the Servername?

bbockelm:
Additionally - when creating commit messages, please follow the advice starting in [this paragraph](https://osg-htc.org/technology/software/git-software-development/#write-a-succinct-subject-to-explain-what-each-commit-does):
- The subject should state what the commit does.
- The body should cover _why_ the change is needed.
- If the change is particularly complex, then you can cover the details of what the change is/how the change works.

Not every commit message is going to be perfect but we should always strive to better.

h2zh:
Proposed system design based on @bbockelm's comment and conversation with @CannonLock. I'd appreciate your feedbacks :)
downtime

Some design thoughts:
1. Need to create a db for Cache to persist the downtime there, while Cache server is stateless at the moment - thinking of implementing it in another PR in case this one goes too giant.
2. The OSDF admin can set any server to downtime. 
3. If the Origin administrator sets its own downtime, then OSDF admin cannot revert the downtime of that origin.

h2zh:
After talking to Cannon, we agree to replace the `toggle downtime` to `schedule downtime` in the Director. It means the Director admin can't toggle downtime anymore, but should submit a simple form to schedule downtime.  In this way, the new `Downtime` struct would be universal for director, origin, and cache. 
I'll create another PR for the work on the Director side, and get that done first.
```
Class    string
	Severity string
	// Downtime represents a server downtime event set by the admin of director, origin or cache
	Downtime struct {
		UUID        string   `json:"uuid" gorm:"primaryKey"`
		ServerName  string   `json:"serverName" gorm:"not null"` // Name of the server
		CreatedBy   string   `json:"createdBy" gorm:"not null"`  // Person who created this downtime
		Class       Class    `json:"class" gorm:"not null"`      // SCHEDULED or UNSCHEDULED
		Description string   `json:"description" gorm:"type:text"`
		Severity    Severity `json:"severity" gorm:"type:varchar(80);not null"`
		StartTime   int64    `json:"startTime" gorm:"not null;index"` // All time is in Epoch UTC milliseconds
		EndTime     int64    `json:"endTime" gorm:"index"`
		CreatedAt   int64    `json:"createdAt" gorm:"autoCreateTime:milli"`
		UpdatedAt   int64    `json:"updatedAt" gorm:"autoUpdateTime:milli"`
	}
)

const (
	SCHEDULED   Class = "SCHEDULED"
	UNSCHEDULED Class = "UNSCHEDULED"
)

const (
	Outage                      Severity = "Outage (completely inaccessible)"
	Severe                      Severity = "Severe (most services down)"
	IntermittentOutage          Severity = "Intermittent Outage (may be up for some of the time)"
	NoSignificantOutageExpected Severity = "No Significant Outage Expected (you shouldn't notice)"
)
```

h2zh:
Take a look at the doc on the top for the latest PR walkthrough!

database/server_api_test.go
```diff
@@ -0,0 +1,225 @@
...
+		Description: "Scheduled maintenance",
+		Severity:    "Outage (completely inaccessible)",
+		StartTime:   time.Now().UTC().UnixMilli(),
+		EndTime:     time.Now().UTC().Add(1 * time.Hour).UnixMilli(),
+		CreatedAt:   time.Now().UTC().UnixMilli(),
+		UpdatedAt:   time.Now().UTC().UnixMilli(),
+	}
+	pastDowntime := server_structs.Downtime{
+		UUID:        "01952a5a-fdc4-72a7-88e7-c98aaee5278d",
+		CreatedBy:   "Miko Brando",
```
bbockelm, line 90: Oddly enough, this appears to be a real person's name?  To avoid confusion, let's use a more placeholder value here.

database/server_api_test.go
```diff
@@ -0,0 +1,225 @@
...
+		r.ServeHTTP(w, req)
+
+		var resp server_structs.Downtime
+		require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+		assert.Equal(t, server_structs.NoSignificantOutageExpected, resp.Severity)
+	})
+
+	t.Run("update-downtime-with-invalid-uuid", func(t *testing.T) {
+		updatedDowntime := DowntimeInput{
+			Severity: "Outage (completely inaccessible)",
```
bbockelm, line 188: Can we use an enum here instead of a string?

database/server_test.go
```diff
@@ -0,0 +1,176 @@
...
+	t.Cleanup(func() {
+		teardownMockDowntimeDB(t)
+		config.ResetConfig()
+	})
+
+	t.Run("create-downtime-success", func(t *testing.T) {
+		mockDowntime := server_structs.Downtime{
+			UUID:        uuid.NewString(),
+			CreatedBy:   "test_user",
+			Class:       "SCHEDULED",
```
bbockelm, line 47: Similarly here, can we create an enum instead of a string?

director/director.go
```diff
@@ -1300,6 +1300,43 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
...
+		// Process received server(origin/cache) downtimes and toggle the director db accordingly when necessary
+		currentTime := time.Now().UTC().UnixMilli()
+		bringItDown := false
+		sn := adV2.Name
+		// Check if the server is currently in downtime
+		for _, downtime := range adV2.Downtimes {
+			if downtime.StartTime < currentTime && downtime.EndTime > currentTime || downtime.EndTime == 0 {
+				// Server is currently in downtime
+				// Backup the original filter type to revert in case of database failure
+				originalFilterType, hasOriginalFilter := filteredServers[sn]
```
bbockelm, line 1313: This needs to be managed by the `filteredServersMutex`.

director/director.go
```diff
@@ -1300,6 +1300,43 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
...
+
+				// Set the server to downtime in the director in-memory cache and database
+				filteredServers[sn] = serverFiltered
+				err := setServerDowntime(sn, serverFiltered)
```
bbockelm, line 1317: Ads should not override the director database!

The logic here also doesn't make sense:
- If there are multiple downtimes, then we'll set the state as down repeatedly.
- If there are multiple reasons why the server is down then they stomp on each other.

director/director.go
```diff
@@ -1300,6 +1300,43 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
...
+						delete(filteredServers, sn)
+					}
+				}
+				bringItDown = true
+				break
+			}
+		}
+		// If the server doesn't have an active downtime, it means Director's previously
+		// recorded downtime set by the server admin is stale and should be removed
+		if !bringItDown {
```
bbockelm, line 1332: Logic issue -- if no downtimes are sent at all, then potentially we mark the server as down.

bbockelm (dismissed):
This is a good step forward!

Several comments embedded.  Most important:
- Make this more generic about the web interface and not cache/origin specific.  Do not put the web APIs in the `database` module.
- Do not have the cache/origin admins override the wishes of the OSDF administrator.  If you want to serialize things, serialize it as a bitmask so we know multiple folks have set the downtime.  Long-term, however, we'll be moving the state to the registry so I'm OK if we just don't do database operations in the director.
- I'd like to see an end-to-end test where a downtime is declared by a server and we show that the director doesn't target it.

cache/cache.go
```diff
@@ -39,4 +41,14 @@ func RegisterCacheAPI(router *gin.Engine, ctx context.Context, egrp *errgroup.Gr
 	{
 		group.POST("/directorTest", func(ginCtx *gin.Context) { server_utils.HandleDirectorTestResponse(ginCtx, notificationChan) })
 	}
+
+	cacheDowntimeAPI := group.Group("/downtime")
```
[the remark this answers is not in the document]
h2zh: All downtime API endpoints have been relocated to web_ui/ui.go. A conditional statement has been implemented to enable these endpoints specifically when the server type is 'origin' or 'cache'.

database/server_test.go
h2zh: Same as above.

database/server_api_test.go
h2zh: This test is designed to match the input string format from the frontend. Currently, our backend and frontend don't have a centralized repository for shared type definitions. I've already created an enum-like constant (technically, there's no enum in Go) block for the Severity type in [code](https://github.com/h2zh/pelican/blob/9cbb79d3606806f9560d17d7d5a1beb4fc1b7c2b/server_structs/director.go#L206) for backend usage, but the frontend team will need to define their own enum and convert it to a string when making API calls.

database/server_api_test.go
h2zh: My bad :/ I was on Spanish 101 at that moment. I replace it with "John Doe".

database/server_api_test.go
```diff
@@ -0,0 +1,225 @@
...
+		assert.Equal(t, http.StatusOK, w.Code)
+		var resp []server_structs.Downtime
+		require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+		assert.Len(t, resp, 2)
+		assert.Contains(t, w.Body.String(), pastDowntime.UUID)
+	})
+
+	t.Run("create-active-downtime", func(t *testing.T) {
+		activeFutureDowntime := DowntimeInput{
+			CreatedBy:   "Pablo Afonso",
```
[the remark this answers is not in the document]
h2zh: I replace it with "John Doe Jr."

director/director.go
h2zh: [deleteServerDowntimeSetByServerAdmin](https://github.com/h2zh/pelican/blob/9cbb79d3606806f9560d17d7d5a1beb4fc1b7c2b/director/director_db.go#L169) function only deletes the downtime set by server (origin/cache) admin, but keeps the downtime set by us (OSDF admin)

director/director.go
h2zh: Ads would affect the director database, but not "override" - they have different priorities in different scenarios:

- The OSDF admin can set any server to downtime.
- If the Origin/Cache administrator sets its own downtime, then OSDF admin cannot revoke it.
- If the OSDF administrator sets a Origin/Cache to downtime, the Origin/Cache admin cannot revoke it.
- If the Origin/Cache administrator wants to set its own downtime, but this Origin/Cache is already being put in downtime by others (likely OSDF admin, or topology,...), the original downtime setting won't change.

If this server(origin/cache) has multiple active downtimes, this function will break the loop once the first one is set. Likewise, if they have different descriptions, it doesn't matter because the Director home page UI just displays a grey-out toggle button. If the Director admin wants to know the details (reasons) of multiple active downtimes, they can click on the server and find them in the expanded drawer UI ([downtimes data is loaded in API](https://github.com/h2zh/pelican/blob/9cbb79d3606806f9560d17d7d5a1beb4fc1b7c2b/director/director_ui.go#L250), just need frontend to add some UI displaying them).
However, if multiple active downtimes are set by different parties, see the last scenario above.

h2zh:
I addressed all issues except for the e2e test. No major logical change was made. So I think it's ready for the second review.

director/director.go
h2zh: For your reference, we use [`filterType`](https://github.com/h2zh/pelican/blob/a50a3b71bec40311393af24f0353b0c05bdab096/director/cache_ads.go#L44) to differentiate which entity creates a downtime. I add `serverFiltered` to represent the downtime set by server(origin/cache), while others were in the codebase.

CannonLock:
@jhiemstrawisc You finished the HA director piece right? How do you handle the in memory downtimes that are in the director?

director/director.go
```diff
@@ -1269,6 +1269,51 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
...
 	}
 
+	if adV2.Downtimes != nil {
+		// Process received server(origin/cache) downtimes and toggle the director db accordingly when necessary
+		currentTime := time.Now().UTC().UnixMilli()
+		bringItDown := false // Flag to indicate if this server is put in downtime in the traversal of
+		sn := adV2.Name
+		// Check if the server is currently in downtime
+		for _, downtime := range adV2.Downtimes {
+			if downtime.StartTime < currentTime && downtime.EndTime > currentTime || downtime.EndTime == 0 {
```
CannonLock, line 1279:
```suggestion
			if downtime.StartTime < currentTime && (downtime.EndTime > currentTime || downtime.EndTime == 0) {
```

Order of operations would allow this to proceed regardless, as long as EndTime was 0.

director/director.go
```diff
@@ -1269,6 +1269,51 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
...
+
+				// If the server is not in downtime, we need to set it to downtime in the director in-memory cache and database
+				filteredServers[sn] = serverFiltered
+				err := setServerDowntime(sn, serverFiltered)
+				if err != nil {
+					log.Warningf("Failed to set downtime for server %s in the director database: %v", adV2.Name, err)
+					if hasOriginalFilter {
+						filteredServers[sn] = originalFilterType
+					} else {
+						delete(filteredServers, sn)
```
CannonLock, line 1299: Same, needs a mutex

web_ui/server_api.go
```diff
@@ -0,0 +1,199 @@
...
+		existingDowntime.Description = downtimeInput.Description
+	}
+	if downtimeInput.Severity != "" {
+		existingDowntime.Severity = downtimeInput.Severity
+	}
+	if downtimeInput.StartTime != 0 {
+		existingDowntime.StartTime = downtimeInput.StartTime
+	}
+	if downtimeInput.EndTime != 0 {
+		existingDowntime.EndTime = downtimeInput.EndTime
```
CannonLock, line 182:
```suggestion
		existingDowntime.EndTime = downtimeInput.EndTime
```

Won't this break if I want to update the EndTime value to 0? 0 being a valid value indicating this does not end.

web_ui/ui.go
```diff
@@ -576,6 +576,20 @@ func configureCommonEndpoints(engine *gin.Engine) error {
...
 	engine.GET("/api/v1.0/listApiTokens", listApiTokens)
 	engine.GET("/api/v1.0/version", getVersionHandler)
+
+	isOriginOrCache := config.IsServerEnabled(server_structs.OriginType) || config.IsServerEnabled(server_structs.CacheType)
+	if isOriginOrCache {
+		downtimeAPI := engine.Group("/api/v1.0/downtime")
+		{
+			downtimeAPI.POST("/", AuthHandler, AdminAuthHandler, HandleCreateDowntime)
+			downtimeAPI.GET("/", AuthHandler, AdminAuthHandler, HandleGetIncompleteDowntime)
+			downtimeAPI.GET("/all", AuthHandler, AdminAuthHandler, HandleGetAllDowntime)
```
CannonLock, line 586:
```suggestion
			downtimeAPI.GET("/all", AuthHandler, AdminAuthHandler, HandleGetAllDowntime)
```

I would prefer this didn't exist. REST best practices indicates that filtering should be done via url parameters and not entirely new endpoints. 

I would prefer something more along the lines of a single GET endpoint that accepts `active=true` or active=false` to filter the response.

CannonLock (changes requested):

director/director.go
```diff
@@ -1269,6 +1269,51 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
...
+		bringItDown := false // Flag to indicate if this server is put in downtime in the traversal of
+		sn := adV2.Name
+		// Check if the server is currently in downtime
+		for _, downtime := range adV2.Downtimes {
+			if downtime.StartTime < currentTime && downtime.EndTime > currentTime || downtime.EndTime == 0 {
+				// Server is currently in downtime
+
+				// If this server is already put in downtime, we don't need to do anything
+				// Retrieve the original filter type to check and backup for revert in case of database failure
+				filteredServersMutex.Lock()
```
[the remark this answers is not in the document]
CannonLock: @h2zh Then I would convert this to a RLock as suggested, if you Lock then no one can read this well it is locked.

server_structs/director.go
```diff
@@ -78,6 +78,21 @@ type (
 		DirlistHost   string       `json:"dirlisthost"`
 	}
 
+	// Downtime represents a server downtime event
+	Class    string
+	Severity string
+	Downtime struct {
+		UUID        string   `json:"uuid" gorm:"primaryKey"`
```
CannonLock, line 85:
```suggestion
		UUID        string   `json:"id" gorm:"primaryKey"`
```

Sorry, this one is a bit of a knit pick but we use `id` in the api token and the namespace endpoints it is best to just keep things consistent and use id here.

web_ui/server_api.go
```diff
@@ -0,0 +1,209 @@
...
+		return
+	}
+	if !isValidSeverity(downtimeInput.Severity) {
+		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{Status: server_structs.RespFailed, Msg: "Invalid downtime severity"})
+		return
+	}
+
+	downtime := server_structs.Downtime{
+		UUID:        id.String(),
+		CreatedBy:   downtimeInput.CreatedBy,
```
CannonLock, line 86: I am surprised we are letting people set the value here based on their input. I think it would be better if we set this to our best user identifier. At this point the are logged in via the admin password or OAuth.

I noticed this same practice with the API token endpoint that Patrick created.

CannonLock (changes requested):

web_ui/server_api.go
h2zh: I was thinking of filling the `CreatedBy ` on the frontend code, because `web_ui.GetUserGroups(ctx)` gives me empty user name. Do you know which function could tell us the user's uuid? @CannonLock

CannonLock: I am surprised by that, by nature of this getting past the AuthHandler the context should have populated a value at "User".

Have you tried grabbing the value from there?

CannonLock (changes requested):
Couple code changes and one addition.

Currently your code doesn't advertise cache downtimes can you add that in?

Additionally I am going to push one update so that I can get the downtimes in the director api server response.

swagger/pelican-swagger.yaml
```diff
@@ -721,6 +727,54 @@ definitions:
...
+  Downtime:
+    type: object
+    properties:
+      id:
+        type: string
+        description: Unique identifier for the downtime entry
+        example: "019564c2-7893-73e1-ab74-c566972cf059"
+      createdBy:
+        type: string
+        description: The person who created the downtime entry
```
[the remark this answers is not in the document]
h2zh: I removed all `createdBy` in downtime-relevant user input fields. But it remains in the return json data fields.

web_ui/server_api.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	if downtimeInput.Description != "" {
+		existingDowntime.Description = downtimeInput.Description
+	}
+	if downtimeInput.Severity != "" {
+		existingDowntime.Severity = downtimeInput.Severity
+	}
+	if downtimeInput.StartTime != 0 {
+		existingDowntime.StartTime = downtimeInput.StartTime
+	}
```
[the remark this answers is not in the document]
h2zh: I think the current way is more straightforward and easy-to-maintain. For creation, we build a new object from stretch; for update, we check the existing fields before updating. There's no efficiency improvement if we combine them, but will increase the difficulty of understanding the code.

web_ui/server_api.go
```diff
@@ -0,0 +1,214 @@
...
+		downtimes, err = database.GetIncompleteDowntimes()
+	case "all":
+		downtimes, err = database.GetAllDowntimes()
+	default:
+		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    "Invalid 'status' query parameter. Must be 'incomplete' or 'all'.",
+		})
+		return
+	}
```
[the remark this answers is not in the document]
h2zh: Could I set "incomplete" as the default? I think in most cases this API endpoint is pinged for active and future downtimes. It can also reduce the payload size by default, though I guess nobody actually cares about such a small compression.

web_ui/server_api.go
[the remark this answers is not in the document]
CannonLock: I see this resolved above. I didn't note that the if `x != null` were to check what to update. I just wanted the verification function you have to be used in both places which it seems it done!

PR #1998, opened 2025-02-07
Spruce up description of `Shoveler.Enable` param

jhiemstrawisc:


docs/parameters.yaml
```diff
@@ -2677,7 +2677,11 @@ components: ["origin", "cache", "director", "registry"]
 ############################
 name: Shoveler.Enable
 description: |+
-  Enable XRootD monitoring shoveler: https://github.com/opensciencegrid/xrootd-monitoring-shoveler.
+  The shoveler gathers UDP monitoring messages from XRootD servers and sends them to a reliable message bus, such
```
brianhlin, line 2680:
```suggestion
  The shoveler gathers UDP monitoring messages from XRootD servers and sends them to a message bus, such
```

Ideally it's reliable but that's really up to the federation operators / in the eye of the beholder. In general, I'd stay away from value judgements in technical docs unless they're concrete requirements.

docs/parameters.yaml
```diff
@@ -2677,7 +2677,11 @@ components: ["origin", "cache", "director", "registry"]
...
+  as [RabbitMQ](https://www.rabbitmq.com/). It is used within Pelican for a variety of metrics aggregation when
+  those metrics are collected by the XRootD daemon and not by Pelican itself.
```
brianhlin, line 2682: This second sentence seems like an implementation detail. As an operator, why should I care which metrics are from XRootD vs Pelican?

brianhlin (commented):
Some nits

docs/parameters.yaml
jhiemstrawisc: Well, in a way it is an implementation detail, but isn't it one that matters if you care about using a shoveler? Only metrics sent from XRootD to the shoveler via UDP are available for use, which is a fraction of all the metrics Pelican can collect. If you're configuring a shoveler, I presume you care about the data that shows up there, including a description of the data that doesn't show up there.

I guess you're right that without a description of _which_ metrics this includes, the statement doesn't help much. Maybe @patrickbrophy can include a sampling of some of the metrics that show up here?

patrickbrophy: The shoveler just packages up monitoring messages from XRootD. The kinds of metrics that can appear are listed in the XRootD monitoring [docs](https://xrootd.web.cern.ch/doc/dev57/xrd_monitoring.htm). Some examples of metrics that could be included are things like the number of running threads, disc events, etc. Regardless of the shoveler being enabled, Pelican should ingest these monitoring messages and make them available via Prometheus.

brianhlin: > but isn't it one that matters if you care about using a shoveler? Only metrics sent from XRootD to the shoveler via UDP are available for use, which is a fraction of all the metrics Pelican can collect. If you're configuring a shoveler, I presume you care about the data that shows up there, including a description of the data that doesn't show up there.

Yeah but if I'm spinning up a federation from scratch, how do I make the decision whether or not to go through the trouble of setting up a shoveler + message bus? That being said, that's a higher level decision that probably doesn't deserve to be buried in the param docs. I'd say that we just strike this sentence

docs/parameters.yaml
```diff
@@ -2677,7 +2677,11 @@ components: ["origin", "cache", "director", "registry"]
 ############################
 name: Shoveler.Enable
 description: |+
-  Enable XRootD monitoring shoveler: https://github.com/opensciencegrid/xrootd-monitoring-shoveler.
+  The shoveler gathers UDP monitoring messages from XRootD servers and sends them to a message bus, such
```
brianhlin, line 2680:
```suggestion
  Enable the XRootD monitoring shoveler. The shoveler gathers UDP monitoring messages from XRootD servers and sends them to a message bus, such
```

I think it's still important to explicitly state what this knob does

brianhlin (approved):
LGTM

PR #2000, opened 2025-02-10
Run tests in more appropriate container images

brianaydemir:
The triggers for the unit tests have been moved into `build-and-test.yaml` (which was previously named `build-images.yaml`), because in order to run unit tests in a more appropriate container image, we might need to wait for that container to be built first.

The basic philosophy is this:

- Tags (`v{semvar}`) are always tested in a container based on the containers being pushed for that release or release candidate. This is the critical part.

- Pull requests are tested in whatever version of `pelican-test:latest-itb` happens to be available at the time. We can't push images during a pull request, and in the name of simplicity, I decided to forgo trying to find a "better" image. In practice, this should be fine so long as:

    1. Updates to Pelican's dependences are pushed to `main` prior to work that depends on those updates.
    2. All branches currently under development function correctly under `main`'s versions of those dependencies.

- Pushes to `main` trigger a new build of the `pelican-test` container image, which is then used to run the unit tests.

Other notable changes:

- I've attempted to restore the registry-based caching that was part of the dev container build process prior to #1972.

- I've removed the unrecognized `OriginUrl` parameter from `images/dev-config.yaml`. Along the way, I determined that `github_scripts/citests.sh` actually depended on that config file being in `/etc/pelican`, so I've added the relevant env var exports to the script. (Tests shouldn't be looking at or relying on `/etc/pelican`.)

Other remarks:

- The GitHub Actions web UI does a not-great job of visually laying out the jobs. You have to hover over each node to see its dependencies highlighted clearly.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,4 +1,4 @@
-name: Build and Push Container Images
+name: Build and Test
 
 on:
   pull_request:
@@ -15,9 +15,6 @@ on:
     types:
       - dispatch-build
   workflow_dispatch:
-
-permissions:
-  contents: write
 
 jobs:
   set-build-parameters:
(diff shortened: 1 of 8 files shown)


jhiemstrawisc (approved):
LGTM. For posterity, we discussed in person that we'd merge these changes and keep an eye open for anything that breaks. The challenge of GHA is that you don't know until you exercise all the triggers!

PR #2001, opened 2025-02-10
Deprecate Lotman.DbLocation in favor of Lotman.LotHome

jhiemstrawisc:
Elsewhere in Pelican, `DbLocation` params point to the the name of an actual file, not the base directory for a file. So as not to break that convention, I'm renaming this Lotman key to match the name Lotman actually uses to refer to the path.


Saartank (approved):
Looks good!

PR #2007, opened 2025-02-11
Added documentation for additional CLI commands

### LINKED ISSUE #1648: Add other command line information to the client documentation
We should enumerate and describe more client functionality such as 

-- sync
-- ls 
-- logging
-- stat

turetske:
Added descriptions of CLI commands for `ls`, `stat`, and `sync`

### DIFF
--- a/docs/pages/getting-data-with-pelican/client.mdx
+++ b/docs/pages/getting-data-with-pelican/client.mdx
@@ -53,7 +53,7 @@ When a Pelican client encounters a Pelican URL, it uses the hostname in the fede
 When using the `osdf:///` scheme, Pelican will load in the federation metadata automatically using the defaults for accessing the Open Science Data Federation (OSDF). This scheme should only be used if your federation or data can be accessed through the OSDF. When using this scheme, you should not specify the federation URL at all, as Pelican will handle this automatically. For example:
 
 ```bash
-pelican object get osdf:///<namespace-prefix></path/to/file> <local/path/to/file>
+pelican object get osdf:///<namespace-prefix></path/to/object> <local/path/to/file>
 ```
 
 >This scheme has three slashes (`///`) after the `osdf` because the hostname is left empty to be automatically populated therefore, just start the URL with the namespace prefix. Pelican currently recognizes the `osdf://` scheme (with two slashes) in case the user forgets to pass the third slash, but this is not recommended.
@@ -66,7 +66,7 @@ To use the pelican client to get public objects from a federation, use Pelican's
 To use the `pelican://` URL scheme, you need to specify the federation URL within the request URL. To do this, you would format the `object get` command like so:
 
 ```bash
-pelican object get pelican://<federation-url></namespace-prefix></path/to/file> <local/path/to/file>
+pelican object get pelican://<federation-url></namespace-prefix></path/to/object> <local/path/to/file>
 ```
 
 You can try this yourself by getting the public file that was mentioned earlier from the OSDF. Using the `object get` sub command, and providing the federation URL for the OSDF:
@@ -93,7 +93,7 @@ Protected namespaces require that a Pelican client prove it is allowed to access
 Some origins support authentication with CILogon's OIDC client. In these cases, the Pelican client is capable of guiding users through authentication with the origin, once you logged in via CILogon. To download protected objects from origins that support CILogon, run the same command as for downloading a public object:
 
 ```bash
-pelican object get pelican://<federation-url></namespace-prefix></path/to/file> <local/path/to/file>
+pelican object get pelican://<federation-url></namespace-prefix></path/to/object> <local/path/to/file>
 ```
 
 If you're doing this for the very first time, Pelican will create an encrypted token wallet on your system and you will be required to provide Pelican with a password for the wallet. If this isn't your first time, you will be asked to provide your already-configured password to unlock the token wallet.
@@ -132,7 +132,7 @@ Once you have the token, copy and paste it into a file on your machine, and pass
 
 ```bash
 # Pass the path to the file containing the token using -t flag
-pelican object get pelican://<federation-url></namespace-prefix></path/to/file> </local/path/to/file> -t </path/to/token/file>
+pelican object get pelican://<federation-url></namespace-prefix></path/to/object> </local/path/to/file> -t </path/to/token/file>
 ```
 
 For example, if a token is saved in a file named `my-token`, it can be used to get the object `/ospool/PROTECTED/auth-test.txt` by running:
@@ -146,7 +146,7 @@ $ pelican object get -f https://osg-htc.org /ospool/PROTECTED/auth-test.txt down
 Another powerful Pelican client command is the `pelican object put` command. This command does a simple PUT request to add your object to a data repository via the federation, and putting files into a data repository always requires a token. For the example, we will need a token to perform these requests (see the [previous section](#get-a-protected-object-from-your-federation) for more information). Here is how you can use `pelican object put`:
 
 ```bash
-pelican object put <path/to/local/file> pelican://<federation-url></namespace-prefix></path/to/file> -t </path/to/token/file>
+pelican object put <path/to/local/file> pelican://<federation-url></namespace-prefix></path/to/object> -t </path/to/token/file>
 ```
 
 > **Note:** you can also specify the federation url here with the `-f` flag, just be sure not to include it in the request URL as the host name if you decide to do so.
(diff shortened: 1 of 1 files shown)


docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -241,6 +241,91 @@ The Pelican client supports a variety of command line flags that modify the clie
...
+
+This will only be successful when listings are enabled in the Pelican origin. To learn more about origin capabilities, go to the [Origin docs](../federating-your-data/origin.mdx).
+
+If listings are not enabled, you will see the following error:
+
+```
+ERROR[2025-02-07T15:27:05-06:00] Failure getting pelican://osdf-director.osg-htc.org:443/nrao-ardg: failed to perform list request: Collections URL not found in director response. Are you sure there's an origin for prefix /<namespace-prefix> that supports listings?
+```
+
+## Getting Object Data From a Federation with Stat
```
jhiemstrawisc, line 260: Same comment about documenting when tokens are needed and how to provide one, or otherwise linking to documentation that explains it.

Also, I'd remove "From a Federation". I think it's enough to say "Getting Object Data with Stat".

docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -241,6 +241,91 @@ The Pelican client supports a variety of command line flags that modify the clie
...
+This will return data in the form of:
+
+```
+Name: </namespace-path></path/to/object>
+Size: <size>
+ModTime: <time-last-modified>
+IsCollection: false
+```
+
+This command can also be run on a collection:
```
jhiemstrawisc, line 277: I grepped through our docs, and I couldn't find anywhere that we actually define what it means to be a collection. Since this documentation explicitly describes how to work with collections, can you also take a stab at defining collections? I think the most appropriate place for the description will be in `docs'/pages/about-pelican/core-concepts.mdx`, right after the current definition of "Objects".

Once there's a definition, can you also link to it from this section?

docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -241,6 +241,91 @@ The Pelican client supports a variety of command line flags that modify the clie
...
+```bash
+pelican object stat pelican://<federation-url></namespace-prefix></path/to/object>
+```
+
+This will return data in the form of:
+
+```
+Name: </namespace-path></path/to/object>
+Size: <size>
+ModTime: <time-last-modified>
```
jhiemstrawisc, line 273: Can you document what time format this will be? (RFC 3339, e.g. 1970-01-01 00:00:00 +0000 GMT)

docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -241,6 +241,91 @@ The Pelican client supports a variety of command line flags that modify the clie
...
+```
+ERROR[2025-02-07T15:27:05-06:00] Failure getting pelican://osdf-director.osg-htc.org:443/nrao-ardg: failed to perform list request: Collections URL not found in director response. Are you sure there's an origin for prefix /<namespace-prefix> that supports listings?
+```
+
+## Getting Object Data From a Federation with Stat
+
+To use the pelican client to get object data, use Pelican's `object stat` sub-command.
+
+```bash
+pelican object stat pelican://<federation-url></namespace-prefix></path/to/object>
```
jhiemstrawisc, line 265: I think the only other flag worth documenting somewhere in this section is the `--json` flag:
```
Flags:
  -j, --json           Print results in JSON format
```

docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -241,6 +241,91 @@ The Pelican client supports a variety of command line flags that modify the clie
...
+```
+Name: </namespace-path></path/to/collection>
+Size: 0
+ModTime: <time-last-modified>
+IsCollection: true
+```
+
+## Syncing Objects with an Origin
+
+Pelican also has an `object sync` command which will sync objects from a source object or collections to a destination object or collection. If the objects already exist in the destination and are the same size or larger than the source objects of the same name, they will not be overwritten, but the objects which don't exist will be added.
```
jhiemstrawisc, line 294: Two things:
- Can you manually linewrap some of the longer lines so fit on an average monitor window without scrolling?
- I think this could be more explicit about the differences between `object copy/get/put` and `object sync`. Can you provide more information about when a user should use `sync` as opposed to the other transfer commands?

docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -241,6 +241,91 @@ The Pelican client supports a variety of command line flags that modify the clie
...
+
+> **Warning**: In the case where an object in the destination is smaller than an object with the same name in the source, the sync command assumes that a partial sync was interrupted and completes it by appending the necessary bytes. If it issn't the same file, but has the same name, this can result in the mashing of two files together.
+
+### Sync Objects from a Collection to a Local Directory
+
+To sync a local directory with a collection in the origin, use the following command:
```
jhiemstrawisc, line 300: I guess sync technically isn't a directional verb, but the source/dest here felt backwards to me when I read it. Maybe "To sync a remote collection with a local directory"?

docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -241,6 +241,91 @@ The Pelican client supports a variety of command line flags that modify the clie
...
+IsCollection: true
+```
+
+## Syncing Objects with an Origin
+
+Pelican also has an `object sync` command which will sync objects from a source object or collections to a destination object or collection. If the objects already exist in the destination and are the same size or larger than the source objects of the same name, they will not be overwritten, but the objects which don't exist will be added.
+
+> **Warning**: In the case where an object in the destination is smaller than an object with the same name in the source, the sync command assumes that a partial sync was interrupted and completes it by appending the necessary bytes. If it issn't the same file, but has the same name, this can result in the mashing of two files together.
+
+### Sync Objects from a Collection to a Local Directory
```
jhiemstrawisc, line 298: Few comments:
- Can you provide some discussion of or a link to how this can be done with tokens?
-  I think `sync` from remote to local also requires that `listings` is enabled, doesn't it? Can you explain any assumptions/requirements on Origin/prefix capabilities and how the user can verify those?
- I can't remember all the implementation details, but are there any other assumptions to document about any importance of trailing `/`s in anything? I know you had a big chart to explain all the operations `sync` had to perform. Maybe that table is even worth including here?

docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -53,7 +53,7 @@ When a Pelican client encounters a Pelican URL, it uses the hostname in the fede
 When using the `osdf:///` scheme, Pelican will load in the federation metadata automatically using the defaults for accessing the Open Science Data Federation (OSDF). This scheme should only be used if your federation or data can be accessed through the OSDF. When using this scheme, you should not specify the federation URL at all, as Pelican will handle this automatically. For example:
 
 ```bash
-pelican object get osdf:///<namespace-prefix></path/to/file> <local/path/to/file>
+pelican object get osdf:///<namespace-prefix></path/to/object> <local/path/to/file>
```
jhiemstrawisc, line 56: Thanks for cleaning all these up!

docs/pages/getting-data-with-pelican/client.mdx
```diff
@@ -241,6 +241,91 @@ The Pelican client supports a variety of command line flags that modify the clie
...
+and the data will look like:
+
+```
+Name: </namespace-path></path/to/collection>
+Size: 0
+ModTime: <time-last-modified>
+IsCollection: true
+```
+
+## Syncing Objects with an Origin
```
jhiemstrawisc, line 292: Same comment about reconsidering usage of "Origin" in the header here.

jhiemstrawisc (changes requested):
One extra comment that applies to each of the new sections - Can you provide a real example using our new canonical object: `pelican://osg-htc.org/pelicanplatform/test/hello-world.txt` or the `pelican://osg-htc.org/pelicanplatform/test` collection where appropriate?

docs/pages/getting-data-with-pelican/client.mdx
jhiemstrawisc: After a bit of discussion in the Pelican slack, I think one sentence per line would be better than the linewrap I previously suggested.

brianhlin:
I needed to direct a user to `pelican object sync` but we don't have any published documentation. After discussing this with Justin, the next steps should be to review the current state of this PR and make any necessary edits since things may have changed in the intervening year.

Reassigning to the original author, @turetske for this work.

turetske:
Since this got quite stale, closing it and creating an updated version of the documentation. See #3340

PR #2008, opened 2025-02-11
Handle `Lotman.LotHome` directory creation/permissions in Pelican

jhiemstrawisc:
This builds on the minimal changes in #2001. Let's get that merged first, and any change requests can be rebased here.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1066,7 +1066,7 @@ func SetServerDefaults(v *viper.Viper) error {
 		v.SetDefault(param.Director_DbLocation.GetName(), "/var/lib/pelican/director.sqlite")
 		v.SetDefault(param.Cache_DbLocation.GetName(), "/var/lib/pelican/cache.sqlite")
 		// The lotman db will actually take this path and create the lot at /path/.lot/lotman_cpp.sqlite
-		v.SetDefault(param.Lotman_LotHome.GetName(), "/var/lib/pelican")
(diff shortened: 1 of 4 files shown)


Saartank:
The drop in permissions would be collectively managed by PR #1850.

h2zh (changes requested):
The default value of config param `Lotman.LotHome` is `/var/run/pelican`. When drop privs is enabled, this directory is already set to be owned by the pelican user, because it is the place where Pelican server (registry/director/origin/cache) sqlite database and the monitoring directory live. It cannot be set to own by xrootd user.

A possible way to resolve this problem is changing the value of `Lotman.LotHome` to `/var/run/<ANOTHER_DIR>`. Just make sure this change wouldn't break anything in Lotman.
```
# namei -l /var/lib/pelican
f: /var/lib/pelican
drwxr-xr-x root    root    /
drwxr-xr-x root    root    var
drwxr-xr-x root    root    lib
drwxr-x--- pelican pelican pelican
```

jhiemstrawisc:
Adding @turetske as a second reviewer here so this doesn't have to gobble @h2zh's time away from his ongoing downtime interface PR.

turetske (approved):

jhiemstrawisc:
Going to bypass branch protections here since I have an approving review and I believe I accurately implemented the suggestions Howard asked for by moving the lot home directory under `/var/lib/lotman`.

PR #2010, opened 2025-02-11
Revert "Use Server.Hostname as the prefix for origin/cache namespace registration"

turetske:
Reverts PelicanPlatform/pelican#1984


h2zh (approved):
This reversion exactly matches the code before #1984

PR #2012, opened 2025-02-12
Make use of GitHub's `arm64` Action runners

brianaydemir:


brianaydemir:
Post-mortem: While it's easy enough to cache the `amd64` and `arm64` builds in parallel, I failed to find a way to restore those caches in a useful fashion. (Both caches need to be restored into the same job in order to build the final multi-architecture images using Docker Buildx.)

PR #2013, opened 2025-02-12
Install new plugin: xrdhttp-pelican v0.0.3

h2zh:
Since the `Dockerfile` has been revamped, I took the new xrdhttp-pelican plugin installation out of the dropping privileges PR, tweaked the installation steps to fit in the new container building process with the help of @brianaydemir, and created this separated PR to get xrdhttp-pelican installed in Pelican.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -296,7 +296,6 @@ RUN --mount=type=cache,id=dnf-${TARGETPLATFORM},target=/var/cache/dnf,sharing=lo
             sqlite-devel \
             tinyxml2-devel \
             zlib-devel \
-            xrootd-private-devel \
             \
             json-schema-validator-devel \
             nlohmann-json-devel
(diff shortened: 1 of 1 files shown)


images/Dockerfile
```diff
@@ -291,6 +296,7 @@ RUN --mount=type=cache,id=dnf-${TARGETPLATFORM},target=/var/cache/dnf,sharing=lo
             sqlite-devel \
             tinyxml2-devel \
             zlib-devel \
+            xrootd-private-devel \
```
brianaydemir, line 299: I don't think we need to add this package here, because we're installing it out of Koji in an earlier build stage.

brianaydemir (changes requested):
One minor nitpick. Looks good to me, otherwise.

images/Dockerfile
h2zh: Done! Thanks for pointing it out.

brianaydemir (approved):

Issue #1862, opened 2025-01-03
Create a Prometheus metric that captures time when a service crashed


CannonLock:
**Is your feature request related to a problem? Please describe.**
_A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]_

Over the break there was a issue where Pelican services were crashing and restarting themselves causing resulting in issues that did not leave a trace. 

**Describe the solution you'd like**
_A clear and concise description of what you want to happen._

In the sql lite db include a internal counter. Every time that pelican crashes increment this counter and publish this metric on prometheus.

bbockelm:
Off-ticket, @patrickbrophy asked:

> Should this be for each service or just the director?

Here, we're looking specifically for XRootD crashes -- so we'll be recording the information whenever the cache or origin component is enabled.

> [Are we] tracking if XRootD crashes or the entire pelican process?

Just XRootD.  Specifically, we probably want to record this information [here](https://github.com/PelicanPlatform/pelican/blob/v7.12.0/xrootd/launch_linux.go#L57), when pelican handles the death of an XRootD process by signal.


Additional thoughts:
- This should be in a generic "counters" table (feel free to tinker with the name, that one is a bit blah) that is string key / double value.  (I think you need to have the value be in the billions before floating point precision loss means `x + 1 == x`).  The key should be primary, unique, and indexed.
- Since this is for caches and origins -- and a single service could be both -- we're probably talking two keys.
- We don't need to blindly auto-publish all rows in the table to prometheus; rather, we can assume only known keys are published (so we can iterate through a list of keys set at compile time -- for this PR, there will be a single one).
- It may make sense to record "time of last crash" (unix timestamp) than a counter.  Looking at the Prometheus documentation, the [`changes()`](https://prometheus.io/docs/prometheus/latest/querying/functions/#changes) function will allow prometheus to determine the number of crashes within a certain time period.  Recording "time of last crash" avoids floating point precision issues to boot!

PR #2015, opened 2025-02-12
Prometheus Metric For When a Service Crashed

patrickbrophy:
This PR addresses #1862. This PR adds a new metric `pelican_server_xrootd_last_crash`. This metric is a gauge metric that's value is the timestamp of when XRootD last crashed. This PR also stores this data in sqlite database. This database is called `server.sqlite` because it is shared between both the cache and the origin.


daemon/launch_unix.go
```diff
@@ -244,6 +248,19 @@ func LaunchDaemons(ctx context.Context, launchers []Launcher, egrp *errgroup.Gro
 					metrics.SetComponentHealthStatus(metrics.HealthStatusComponent(metricName), metrics.StatusCritical,
 						launchers[chosen].Name()+" process failed unexpectedly")
 					err = errors.Wrapf(waitResult, "%s process failed unexpectedly", launchers[chosen].Name())
+					service_key := "origin"
+					if config.IsServerEnabled(server_structs.CacheType) {
+						service_key = "cache"
+					}
+					crashTimestamp := time.Now().Unix()
+					log.Debug("Recording xrootd crash at time: ", crashTimestamp)
```
jhiemstrawisc, line 256: Right now this records a unix timestamp, which isn't very friendly to read in a log. Can you instead convert the value into an RFC3339-conformant string?

docs/parameters.yaml
```diff
@@ -2196,6 +2196,13 @@ type: bool
 default: false
 components: ["cache", "director", "origin", "registry"]
 ---
+name: Server.DbLocation
+description: |+
+  A filepath to the intended location of the server's database.
+type: filename
+default: /var/lib/pelican/server.sqlite
+components: ["origin", "cache", "registry", "director"]
+---
```
jhiemstrawisc, line 2205: I think I was okay with this approach until I realized the Origin now has two sqlite databases. Previously we talked about using a common database package for Pelican, and we even talked about combining all our sqlite into a single database. I think that's outside the scope of this pull request, but at the very least let's (for now) make a separate `Cache.DbLocation` instead of introducing the separate `Server.DbLocation`.

When you introduce this, check the other `*.DbLocation` defaults, as we should set the default location to conform to that pattern. The actual instantiation of the default should happen in `config/config.go` and not in `config/resources/defaults.yaml`, which doesn't let you do dynamic defaulting.

daemon/launch_unix.go
```diff
@@ -244,6 +248,19 @@ func LaunchDaemons(ctx context.Context, launchers []Launcher, egrp *errgroup.Gro
...
+
+					dbErr := database.CreateOrUpdateCounter(service_key, int(crashTimestamp))
+					if dbErr != nil {
+						log.Debug("Error recording xrootd crash: ", dbErr)
+						return errors.Wrap(err, "Unable to record xrootd crash")
```
jhiemstrawisc, line 261: Is there any benefit in also wrapping/returning the database error? Also, since `dbErr` isn't used later, can you reformat to:
```
if dbErr := yourFunc(stuff); dbErr != nil {
    // your stuff
}
```

database/migrations/20250203165654_server.sql
```diff
@@ -0,0 +1,11 @@
...
+-- +goose StatementBegin
+CREATE TABLE IF NOT EXISTS counters (
+    key TEXT PRIMARY KEY,
+    value INTEGER NOT NULL DEFAULT 0
+);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+-- +goose StatementEnd
```
jhiemstrawisc, line 11: I don't actually know anything about gorm, or what these snippets of migration stuff do. I'm assuming you know what this does and that you've made sure it's okay!

fed_test_utils/fed.go
```diff
@@ -149,6 +149,7 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
 	viper.Set("Registry.RequireCacheApproval", false)
 	viper.Set("Director.CacheSortMethod", "distance")
 	viper.Set("Director.DbLocation", filepath.Join(t.TempDir(), "director.sqlite"))
+	viper.Set("Server.DbLocation", filepath.Join(t.TempDir(), "server.sqlite"))
```
jhiemstrawisc, line 152: Heads up, this will conflict with a small refactor I just merged. You should rebase on main, and once the database name is changed, this should end up being :
```
viper.Set(param.Cache_DbLocation.GetString(), filepath.Join(t.TempDir(), "server.sqlite"))`
```

metrics/server.go
```diff
@@ -0,0 +1,13 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	PelicanServerXRootDLastCrash = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
jhiemstrawisc, line 9: Why use a gauge vec? Time is monotonically increasing, so a counter feels more appropriate to me, but maybe this is something you've thought about more than I have.

jhiemstrawisc (changes requested):

config/config.go
```diff
@@ -1050,6 +1050,7 @@ func SetServerDefaults(v *viper.Viper) error {
 		v.SetDefault(param.Director_GeoIPLocation.GetName(), "/var/cache/pelican/maxmind/GeoLite2-City.mmdb")
 		v.SetDefault(param.Registry_DbLocation.GetName(), "/var/lib/pelican/registry.sqlite")
 		v.SetDefault(param.Director_DbLocation.GetName(), "/var/lib/pelican/director.sqlite")
+		v.SetDefault(param.Cache_DbLocation.GetName(), "/var/lib/pelican/cache.sqlite")
```
jhiemstrawisc, line 1053: I think this is still missing a default for non-root runners. Can you check around line 1064 to see how we set that up for other databases and add one?

docs/parameters.yaml
```diff
@@ -1541,6 +1541,13 @@ default: 9.5s
 hidden: true
 components: ["cache"]
 ---
+name: Cache.DbLocation
+description: |+
+  A filepath to the intended location of the cache's database.
+type: filename
+default: /var/lib/pelican/cache.sqlite
```
jhiemstrawisc, line 1548: Once you've added a default for non-root, you'll want something like:
```
root_default: /var/lib/pelican/cache.sqlite
default: $ConfigBase/<rest of path>
```

fed_test_utils/fed.go
```diff
@@ -168,6 +168,7 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
 	viper.Set(param.Server_EnableUI.GetName(), false)
 	viper.Set(param.Server_WebPort.GetName(), 0)
 	viper.Set(param.LocalCache_RunLocation.GetName(), filepath.Join(tmpPath, "local-cache"))
+	viper.Set(param.Cache_DbLocation.GetString(), filepath.Join(t.TempDir(), "cache.sqlite"))
```
jhiemstrawisc, line 171: Currently the other components are sorted by launcher type. Can you move this line to occur after the other Cache configs?

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
Looks solid to me!

PR #2019, opened 2025-02-14
Pin xrootd log levels in fed test and use typed params throughout

jhiemstrawisc:
This should cut down on the noise in our logs. I'm also open to further refinement of the log levels if the reviewer(s) think anything should be turned up or down. That being said, this sets everything to the same level they used before tying XRootD log levels to the global Pelican log level.

I also switched a bunch of viper variables to their typed `param` equivalents. This protects us from changing param names without remembering to update things in our tests.


brianaydemir (approved):
LGTM.

PR #2033, opened 2025-02-19
Build and test improvements

brianaydemir:
In addition to motivations listed in #2034, I've taken this opportunity to make the textual definitions for the various testing workflows more similar to each other.

### DIFF
--- a/.github/workflows/test-macos-windows.yml
+++ b/.github/workflows/test-macos-windows.yml
@@ -26,9 +26,10 @@ jobs:
     - name: Cache Next.js
       uses: actions/cache@v4
       with:
+        # Reference: https://nextjs.org/docs/pages/building-your-application/deploying/ci-build-caching#github-actions
         path: |
           ~/.npm
-          ${{ github.workspace }}/.next/cache
+          ${{ github.workspace }}/web_ui/frontend/.next/cache
         # Generate a new cache whenever packages or source files change.
         key: ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('**/*.js', '**/*.jsx', '**/*.ts', '**/*.tsx', '!**/node_modules/**') }}
         # If source files changed but packages didn't, rebuild from a prior cache.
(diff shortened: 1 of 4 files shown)


brianaydemir:
While it's challenging to acquire a statistically significant set of observations, I feel that addressing the two points in #2034 has resulted in some improvements to run times.

Before:

- https://github.com/PelicanPlatform/pelican/actions/runs/13375032926
- https://github.com/PelicanPlatform/pelican/actions/runs/13373772734

After:

- https://github.com/PelicanPlatform/pelican/actions/runs/13442513848

(When looking at those, do be aware that improvements to the Windows tests are 100% due to #2030.)

bbockelm:
What are the CI failures about?

.github/workflows/test-macos-windows.yml
```diff
@@ -46,51 +47,47 @@ jobs:
...
+    - name: Install macOS Dependencies
       run: ./github_scripts/osx_install.sh
       if: runner.os == 'macOS'
-    - name: Test Mac OS X
+    - name: Test macOS
       if: runner.os == 'macOS'
       run: |
         make web-build
-        go test -p=4 -v -coverpkg=./... -coverprofile=coverage.out -covermode=count ./...
+        go test -p=4 -timeout 15m -coverpkg=./... -covermode=count -coverprofile=coverage.out ./...
```
jhiemstrawisc, line 64: Do we actually need to be generating coverage information for Mac/Windows? I don't even think this information can be directly compared to the coverage report we generate in Linux because we compile out entire sections of code if we don't expect them to run on anything other than Linux.

jhiemstrawisc (approved):
LGTM. I added a comment/question about the need to be generating coverage reports for mac/windows, which I'm not convinced we really benefit from. If you disagree, feel free to merge when ready.

.github/workflows/test-macos-windows.yml
brianaydemir: I'd expect that the reports for platform X are compared only to other reports for platform X. A priori, you have to assume that there are differences between platforms that make cross-platform comparisons difficult.

That said, in terms of infrastructure, my preference is to maintain consistency between the platforms we support, wherever possible and achievable with reasonable effort. It doesn't seem like these coverage reports are the source of any woes, so if we find them useful on Linux, we might as well have them for macOS and Windows. It's one less difference for someone to stumble upon and wonder, "Why?"; or one less difference to document.

PR #2035, opened 2025-02-21
Init config refactor

turetske:
Moved InitConfig into InitServer and InitClient and adjusts the tests as needed.

Also made InitConfigDir platform agnostic.

Ideally, I would like to be able to make InitConfig private, but its use in the cmd/config_printer test makes that impossible.

I'm worried InitConfig will keep being used in tests until we figure out a solution for that if the author/reviewer don't catch it organically.


Saartank (commented):
With this change, the log file initialization code runs when `InitServer` is executed. However, some code is executed before `InitServer` is reached. Because of this, all logs generated before `InitServer` runs will be output to the console even if the log file is configured. This could be problematic if something goes wrong before `InitServer` is reached, as the log file would not contain the relevant information.

jhiemstrawisc (changes requested):

fed_test_utils/fed.go
```diff
@@ -198,7 +197,20 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
...
+		t.Fatalf("Error creating temporary config file for fed test: %v", err)
+	}
+	defer tmpFile.Close()
+
+	_, err = io.Copy(tmpFile, originConf)
+	if err != nil {
+		t.Fatalf("Error writing to temporary config file for fed test: %v", err)
+	}
+
+	viper.Set("config", tmpFile.Name())
```
jhiemstrawisc, line 213: I'm struggling to understand why this new section of code is necessary. A few observations I have:
- Is there a reason to use `os.CreateTemp()` here in favor of `t.TempDir()`? One gives us more explicit cleanup immediately after the test exits
- I'm not totally sure why we need to be writing this temp config in the first place, but what's getting written appears to be a broken config. For example, in one test I ran locally I see that the origin exports aren't being overridden:
```
Origin:
    Exports:
        - capabilities:
            - PublicReads
            - Writes
            - DirectReads
          federationprefix: /first/namespace
          storageprefix: /<SHOULD BE OVERRIDDEN>
```

It looks like you were trying to work around the fact that this call to `launchers.LaunchModules` blows out our old config from the previous `config.InitServer()`, which happens exactly because `InitServer` now invokes `InitConfigInternal()`.

I suspect something isn't working as expected here because starting the origins with the non-overriden config probably _should_ be resulting in test failures.

server_utils/origin.go
```diff
@@ -236,6 +236,7 @@ func (b *BaseOrigin) populateOriginCaps() {
 		Reads:       param.Origin_EnableReads.GetBool() || pubReads,
 		DirectReads: param.Origin_EnableDirectReads.GetBool(),
 	}
+
```
jhiemstrawisc, line 239: Why the whitespace changes in an otherwise untouched file?

xrootd/xrootd_config_test.go
```diff
@@ -46,17 +46,32 @@ import (
...
+
+	dirname, err := os.MkdirTemp("", "tmpDir")
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		os.RemoveAll(dirname)
+	})
+	viper.Set("ConfigDir", dirname)
+	viper.Set("Xrootd.RunLocation", dirname)
+	viper.Set("Cache.RunLocation", dirname)
+	viper.Set("Origin.RunLocation", dirname)
```
jhiemstrawisc, line 67: These values get set, then you clear the values, then you reset them and re-isolate the test using a different temp directory. I think this is probably just a bad copy-paste?

I think it's also generally better to prefer the version of this that uses `t.TempDir()` instead of `os.MkdirTemp`, unless there's a concrete reason to be using `os.MkdirTemp` (I think there are occasionally some permissions issues that creep up with `t.TempDir()`).

cmd/plugin_test.go
```diff
@@ -499,11 +501,7 @@ func TestPluginMulti(t *testing.T) {
 
 	dirName := t.TempDir()
 
-	viper.Set("Logging.Level", "debug")
-	viper.Set("Origin.StorageType", "posix")
-	viper.Set("Origin.ExportVolumes", "/test")
-	viper.Set("Origin.EnablePublicReads", true)
-	fed := fed_test_utils.NewFedTest(t, "")
+	fed := fed_test_utils.NewFedTest(t, publicTestOrigin)
```
jhiemstrawisc, line 504: Can we keep the test simple by setting these with the viper variables instead of creating an extra file?

jhiemstrawisc (changes requested):

config/config.go
```diff
@@ -1241,7 +1250,7 @@ func SetServerDefaults(v *viper.Viper) error {
 // Note not all configurations are supported: currently, if you enable both cache and origin then an error
 // is thrown
 func InitServer(ctx context.Context, currentServers server_structs.ServerType) error {
-	logging.FlushLogs(true)
```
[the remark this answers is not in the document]
turetske: I think this was a rebase error, I'll add it back.

config/config.go
```diff
@@ -786,9 +795,9 @@ func InitConfigDir(v *viper.Viper) {
 	v.SetConfigName("pelican")
 }
 
-// InitConfig sets up the global Viper instance by loading defaults and
+// InitConfigInternal sets up the global Viper instance by loading defaults and
 // user-defined config files, validates config params, and initializes logging.
-func InitConfig() {
+func InitConfigInternal() {
```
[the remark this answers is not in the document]
turetske: I think we should be good and the point of getting rid of the "InitConfig" and changing the name to "InitConfigInternal" was to make sure developers didn't explicitly use it. If we keep it available as a function, then that defeats the purpose.

director/discovery_test.go
```diff
@@ -213,9 +212,9 @@ func TestOidcDiscoveryHandler(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			server_utils.ResetTestState()
-			viper.Set("ConfigDir", t.TempDir())
-			viper.Set("Federation.DirectorUrl", tc.dirUrl)
-			config.InitConfig()
+			test_utils.InitClient(t, map[string]any{
+				"Federation.DirectorUrl": tc.dirUrl,
+			})
 			require.NoError(t, config.InitClient())
```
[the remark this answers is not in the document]
turetske: Just an error.

xrootd/xrootd_config_test.go
turetske: I think this was a rebase issue

config/config.go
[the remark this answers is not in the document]
turetske: I touched it at some point and VSCode adds the extra whitespace and I had touched so many files that I was just adding in everything that git was telling me was different.

cmd/plugin_test.go
turetske: Nope. With the changed to NewFedTest, it has to be a read in yaml file in order to properly override the exports.

xrootd/authorization_test.go
```diff
@@ -510,7 +509,6 @@ func TestMergeConfig(t *testing.T) {
 			err := os.WriteFile(scitokensConfigFile, []byte(configInput), fs.FileMode(0600))
 			require.NoError(t, err)
 
-			config.InitConfigDir(viper.GetViper())
```
jhiemstrawisc, line 513: Sorry for more goose chases -- it doesn't look like the this unit test isolates itself. Can you configure a temp dir somewhere by the other viper config up top?

fed_test_utils/fed.go
```diff
@@ -153,51 +154,89 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
 	directorStartTime := time.Now().Add(-6 * time.Minute)
 	director.SetStartupTime(directorStartTime)
 
-	config.InitConfig()
 	err = config.InitServer(ctx, modules)
 	require.NoError(t, err)
 
 	// Read in any config we may have set
+	var originConf interface{}
```
jhiemstrawisc, line 161: It gets a bit confusing to have both `originConf` and `originConfig` -- can you try another variable name? Also, I think we can use `any` instead of `interface{}` here, which is a bit easier to read in my opinion.

fed_test_utils/fed.go
```diff
@@ -153,51 +154,89 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
...
+		confMap := originConf.(map[string]interface{})
+
+		originRaw, exists := confMap["Origin"]
+
+		if exists {
+			originMap := originRaw.(map[string]interface{})
+
+			// Override the test directory from the config file with our temp directory
+			if exportsRaw, ok := originMap["Exports"]; ok {
+				for i, item := range exportsRaw.([]interface{}) {
```
jhiemstrawisc, line 179: Another place where I think we can use `[]any` instead of `[]interface{}`

fed_test_utils/fed.go
```diff
@@ -153,51 +154,89 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
...
+					originDir, err := os.MkdirTemp("", fmt.Sprintf("Export%d", i))
+					assert.NoError(t, err)
+					t.Cleanup(func() {
+						err := os.RemoveAll(originDir)
+						require.NoError(t, err)
+					})
+
+					exportMap := item.(map[string]interface{})
```
jhiemstrawisc, line 187: Another `any`

fed_test_utils/fed.go
```diff
@@ -153,51 +154,89 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
...
 		require.NoError(t, err, "error reading config")
+
+		err := yaml.Unmarshal([]byte(originConfig), &originConf)
+		require.NoError(t, err, "error unmarshalling into interface")
+
+		confMap := originConf.(map[string]interface{})
+
+		originRaw, exists := confMap["Origin"]
+
+		if exists {
```
jhiemstrawisc, line 174: This might be a bit cleaner as:
```
if originRaw, exists := confMap["Origin"]; exists {
    // blah
}
```

since `originRaw` is not used outside the scope of this condition.

I think it might also be a fatal error if we don't pass this existence check -- that would indicate the test thinks it was provided origin configuration, but that configuration doesn't contain `Origin:`.

fed_test_utils/fed.go
```diff
@@ -153,51 +154,89 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
...
+			originMap := originRaw.(map[string]interface{})
+
+			// Override the test directory from the config file with our temp directory
+			if exportsRaw, ok := originMap["Exports"]; ok {
```
jhiemstrawisc, line 178: In the top-level if, you use `exists` instead of `ok` to determine whether the entry is in the map -- I think `ok` is more canonically "go", but I honestly think "exists" is easier to read. Can you pick one or the other and use in both cases?

fed_test_utils/fed.go
```diff
@@ -153,51 +154,89 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
...
+				// Change ownership on the temporary origin directory so files can be uploaded
+				uinfo, err := config.GetDaemonUserInfo()
+				require.NoError(t, err)
+				require.NoError(t, os.Chown(originDir, uinfo.Uid, uinfo.Gid))
+
+				// Start off with a Hello World file we can use for testing in the StoragePrefix
+				err = os.WriteFile(filepath.Join(originDir, "hello_world.txt"), []byte("Hello, World!"), os.FileMode(0644))
+				require.NoError(t, err)
+			}
+		}
```
jhiemstrawisc, line 227: The majority of the code in the `if exportsRaw, ok := originMap["Exports"]; ok {` is duplicated in the subsequent `else` -- can you turn some of it into an inline helper function somewhere?

fed_test_utils/fed.go
```diff
@@ -153,51 +154,89 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
 	directorStartTime := time.Now().Add(-6 * time.Minute)
 	director.SetStartupTime(directorStartTime)
 
-	config.InitConfig()
 	err = config.InitServer(ctx, modules)
 	require.NoError(t, err)
 
 	// Read in any config we may have set
+	var originConf interface{}
 	if originConfig != "" {
```
jhiemstrawisc, line 162: This check shouldn't be needed anymore because a recent change added:
```
if originConfig == "" {
		originConfig = fedTestDefaultConfig
	}
```
to the start of the test

jhiemstrawisc (changes requested):

jhiemstrawisc (changes requested):
Getting close! I suspect both of the change requests are related to your epic rebase adventure.

jhiemstrawisc:
Bumping this to the 7.18 milestone since it's not critical for 7.17 (which we'd like to release today), but we can still try to merge soon before there's time to accumulate more merge conflicts.

config/config.go
```diff
@@ -1305,6 +1319,19 @@ func SetServerDefaults(v *viper.Viper) error {
...
+	if webConfigPath != "" {
+		if err := os.MkdirAll(filepath.Dir(webConfigPath), 0700); err != nil {
+			cobra.CheckErr(errors.Wrapf(err, "failed to create directory for web config file at %s", webConfigPath))
+		}
+	}
+	if err := setWebConfigOverride(viper.GetViper(), webConfigPath); err != nil {
+		cobra.CheckErr(errors.Wrapf(err, "failed to override configuration based on changes from web UI"))
+	}
+
+	logging.FlushLogs(true)
```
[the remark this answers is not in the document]
turetske: It was specifically failing a test without this. I'm unsure if the solution is to instead adjust logging so it doesn't do it's flush logs until further in the function.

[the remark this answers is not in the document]
jhiemstrawisc: Which test was failing?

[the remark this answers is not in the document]
jhiemstrawisc: I double checked my initial statement -- moving the web config override bit here from where it was previously _does_ break the ability of the server admin to set a new web config file path through the UI (this is because the override now happens before any server defaults have been initialized).

I also ran tests locally after restoring this block to where it was previously and didn't observe any failing tests. Can you take another look at why you thought this change was needed?

[This is the initial PR](https://github.com/PelicanPlatform/pelican/pull/2182) where I fixed some of this in the past if you need a bit more context.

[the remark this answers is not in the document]
turetske: Yeah. I'm unsure what changes. There was a test that was failing, and then I fixed it by moving it here, but now it's not. As long as it's working, now.

jhiemstrawisc (changes requested):

jhiemstrawisc:
By the way, the failing tests look like a bug I must have introduced to main recently. I'll dig into it.

jhiemstrawisc:
@turetske Assuming #2457 is approved, I've already checked that rebasing against it appears to fix these tests for me locally.

Issue #2036, opened 2025-02-21
Director UI is running into an unexpected error


Saartank:
![Image](https://github.com/user-attachments/assets/f1ff6603-483a-4aa7-b589-c26a22d3a2bd)

This happens when the Director UI is loaded for the first time.

PR #2037, opened 2025-02-24
Fix UI crashes

patrickbrophy:
This PR addresses #2036. This PR adds some fallbacks to some locations where data was null or undefined.

### DIFF
--- a/web_ui/frontend/app/director/components/GeoIpErrorTable.tsx
+++ b/web_ui/frontend/app/director/components/GeoIpErrorTable.tsx
@@ -51,7 +51,7 @@ const GeoIpErrorTable = () => {
       await alertOnError(getOverriddenGeoIps, 'Could not get config', dispatch)
   );
   const patchedIps = useMemo(() => {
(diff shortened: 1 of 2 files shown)


h2zh:
Hey @CannonLock, could you read through this PR? It seems like the problem is somehow stemmed from other frontend PR previously merged. Just want to see if you can recall anything else that need to amend too!

CannonLock (changes requested):
LGTM. 

Rather then casting undefined to '' in the sort in the StorageTable if would be better to remove Storage objects without a serverName in line 140. 

```
if(serverName === undefined){
  return acc
  }
```

h2zh:
After fixing the linter issue, could you squash all crumbly commits in this PR into one commit (e.g. using `git rebase -i HEAD~#NUMBER_OF_COMMITS`)?

h2zh (approved):
Nice work! 🚀

CannonLock (approved):

PR #2038, opened 2025-02-24
Add /version endpoint to expose version and build info

Saartank:
Added a new /version endpoint to expose version and build details, including build commit, build date, built by, and version number. Integrated this into the common set of APIs and updated the Swagger documentation accordingly.


h2zh (approved):
LGTM

PR #2049, opened 2025-02-25
Bump OS X Dependencies to avoid potential segfault

bbockelm:
Try to update the OS X dependencies to match Linux; we've been observing infrequent (but annoying) segfaults during unit tests.

Particularly, xrdcl-pelican v1.0.3 fixes a segfault in the director response caching code.

I'm not entirely sure this will fix the problem -- but it's useful to hit the "low hanging fruit"!


brianaydemir (approved):
LGTM, but for the record…

I've seen three failures in a row for `TestLargeFile`, so there might be a real issue there, even with these updates. On the other hand, that's a test for the "local cache," which I'm guessing isn't a critical use case for macOS.

PR #2050, opened 2025-02-26
Document restricted characters in fed prefixes

jhiemstrawisc:
Screenshot of the way this renders:
Prefix Restrictions

### DIFF
--- a/docs/pages/federating-your-data/choosing-namespaces.mdx
+++ b/docs/pages/federating-your-data/choosing-namespaces.mdx
@@ -1,4 +1,5 @@
 import ExportedImage from "next-image-export-optimizer";
+import { Table } from 'nextra/components';
 
 # Namespace Prefixes and How To Choose One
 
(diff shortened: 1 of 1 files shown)


docs/pages/federating-your-data/choosing-namespaces.mdx
```diff
@@ -130,3 +130,44 @@ other stores its images in posix. Namespacing allows the data owner to hide this
...
+those that are special inside URLs. In both cases, Pelican disallows these characters as a matter of design because they may cause issues with the way Pelican
+uses HTTP, or they may have unintended consequences when interacting with XRootD, where object names are treated like POSIX paths.
+
+For example, the character sequence `../` in POSIX means "up one directory", such that a filepath like `/foo/bar/../baz` is actually expanded to `/foo/baz`. This makes it
+potentially unsafe to allow a prefix like `/my-prefix/../`, which when interpreted as a filepath is just `/`. Because of the special meaning given to these characters, and because
+their use can lead to certain forms of computer attacks when handled incorrectly, Pelican disallows some characters and character combinations in prefixes.
+
+Other top-level prefixes are restricted because Pelican uses them internally for things like origin/cache registration, or in monitoring.
+
+In total, these are Pelican's namespace prefix restrictions:
```
aowen-uwmad, line 145:
```suggestion
The following table lists the characters that Pelican does not allow when defining a namespace prefix. 
```

aowen-uwmad (changes requested):
Recommend adding a paragraph (at the end?) that details what happens if you try to use one of the disallowed characters.
Does Pelican throw an error? What does it look like? Is it just a warning, but then the service won't be able to communicate with other Pelican services?

jhiemstrawisc:
@jhiemstrawisc First round of feedback is incorporated.

jhiemstrawisc:
These changes are approved, so I'm overriding the branch protection complaining that Andrew doesn't have write permissions while I sort that out separately.

Issue #2052, opened 2025-02-26
`pelican-server` binary is not being installed into `sbin`


brianaydemir:
Prior to #1972, the `pelican-server` binary, currently present only in the container images for running an XRootD cache, was installed into `/usr/local/sbin`:

https://github.com/PelicanPlatform/pelican/blob/8dbab6636dd18335068302e6df578c76b742db4f/images/Dockerfile#L286-L287

That PR unintentionally started putting the binary into `/usr/local/bin`, which breaks the container image's entrypoint script:

```
Starting Pelican...
Running pelican-server with arguments: cache serve
[FATAL tini (7)] exec /usr/local/sbin/pelican-server failed: No such file or directory
```

PR #2053, opened 2025-02-26
Install `pelican-server` into `/usr/local/sbin`

brianaydemir:
The move to `/usr/local/bin` was an unintentional change made when refactoring the container image build stages.

See #2052 for additional background.


jhiemstrawisc (approved):
LGTM

Issue #2056, opened 2025-02-26
The directory and registry incorrectly assume the existence of an `xrootd` user


brianaydemir:
When I refactored the container image build process (#1972), I removed the `xrootd` user from the director and registry images, on the assumption that neither component runs XRootD and thus should have no need of the user.

Unfortunately, it turns out that do they require the `xrootd` user be present on the system. If it is not, these components fail to start up.

Director failure:
```
time="2025-02-26T17:31:11Z" level=error msg="Fatal error occurred at the start of the program. Cleanup started: Failure when configuring the server: failure when creating a directory for the monitoring data: Unable to lookup the runtime user information; does the xrootd user exist?: user: unknown user xrootd"
```

Registry failure:
```
time="2025-02-26T20:36:03Z" level=error msg="Fatal error occurred at the start of the program. Cleanup started: Failure when configuring the server: failure when creating a directory for the shoveler on-disk queue: Unable to lookup the runtime user information; does the xrootd user exist?: user: unknown user xrootd"
```

(The fact that the registry is trying to create a directory for the shoveler queue in the first place is a separate issue: #2055.)

The layers of error messages implicate the following code:

https://github.com/PelicanPlatform/pelican/blob/98f55e4595425b06f3e62307007adc97184a900d/config/mkdirall.go#L78-L102

Essentially, in the process of creating directories, this function unconditionally attempts to look up the so-called "daemon" user, which when `pelican` is started `root`, turns out to be the system's `xrootd` user.

To start with, there is no reason to assume that the UID and GID provided to the function in the first place have any relationship to the "daemon" user, not least of all because there's nothing in the function's name or signature that would suggest a relationship to `xrootd`. We have the following call, for example:

https://github.com/PelicanPlatform/pelican/blob/98f55e4595425b06f3e62307007adc97184a900d/config/config.go#L1336-L1342

On Linux and macOS, this lookup serves no purpose other than to put a potentially incorrect group name into an error message. On Windows, we potentially add the wrong user to an ACL(?), in addition to potentially emitting an incorrect error message.

What this function _should_ be doing is a lookup based on the provided UID and GID.

Issue #2055, opened 2025-02-26
Director and registry unnecessarily create a directory for the shoveler queue on start-up


brianaydemir:
On start up, the director and registry components both attempt to create a directory for the shoveler queue, which is unnecessary given that neither component runs an XRootD server.

https://github.com/PelicanPlatform/pelican/blob/98f55e4595425b06f3e62307007adc97184a900d/config/config.go#L1344-L1346

For example, the following error from an attempt to start the registry attests to the attempt to create this directory:

```
time="2025-02-26T20:36:03Z" level=error msg="Fatal error occurred at the start of the program. Cleanup started: Failure when configuring the server: failure when creating a directory for the shoveler on-disk queue: Unable to lookup the runtime user information; does the xrootd user exist?: user: unknown user xrootd"
```

(The missing `xrootd` user is a separate issue: #2056.)

One can also simply start the container images normally for these components and observe the existence of `/var/spool/pelican/shoveler/queue`, a directory that's not present in the container image.

What this function _should_ be doing is checking the components being configured and conditionally creating the directory.

brianaydemir:
Note that this issue is masked (that is, might go unobserved) if an `xrootd` user exists.

PR #2057, opened 2025-02-27
Do not assume the existence of XRootD (`xrootd`) when it is not necessary

brianaydemir:
See #2055 and #2056 for this PR's background.

In `config/mkdirall.go`, I've decided that it's preferable to avoid doing a lookup on the provided `uid` and `gid` if at all possible. `MkdirAll` is a sufficiently low-level function that it seems reasonable that callers should bear the responsibility for providing correct UIDs and GIDs.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1341,9 +1341,12 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
 		return errors.Wrapf(err, "failure when creating a directory for the monitoring data")
 	}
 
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (approved):
LGTM. Can you open an issue for the embedded `FIXME` to make sure we don't lose track of this?

brianaydemir:
I re-ran the macOS tests and they passed. There's another PR out there that might or might not address some of these occasional failures that we've been seeing: #2049.

PR #2061, opened 2025-02-27
Use DiscoveryURL as token issuer for Prometheus scraping in Director

jhiemstrawisc:
Previously, the checks failed in the actual OSDF because the token encoded the Director's URL as the issuer, but the token consumer expected the issuer to be that of the "federation", or Discovery URL.


jhiemstrawisc:
@h2zh After this one is merged, we should also backport it to 7.14. Let me know if you want to chat about how to do that :)

patrickbrophy:
@jhiemstrawisc Have you tested this? I just want to make sure we can still access metrics.

jhiemstrawisc:
That's partially why I tagged you as reviewer -- I'm not super sure how to test it!

patrickbrophy (changes requested):
So checkout the `createPromMetricToken` in `web_ui/authorization.go`

This function generates a token for accessing the `/metrics` endpoint and I believe it would be relevant to this PR.

jhiemstrawisc:
@bbockelm adding you as a reviewer because neither Patrick nor I are totally sure how to test this. FWIW, I do have a separate test server in my dev environment for serving federation metadata. In the issue, you mentioned the bug only arises when `Federation.DiscoveryUrl != Federation.DirectorUrl`, which is what I'm emulating locally with the metadata server. Even when I split these up locally, I don't notice a change in behavior before/after this diff. What am I missing?

patrickbrophy (approved):
After several discussions and creating diagrams for how Prometheus uses tokens with respect to Pelican, these changes seem to be correct.

PR #2062, opened 2025-02-27
Use GetPelicanUser() to set issuer keys directory's permission group

h2zh:


jhiemstrawisc (approved):
This passes my smell check. Have we confirmed that changing the user to Pelican doesn't prevent us from exporting public keys through the `.well-known/` path at our XRootD endpoint?

h2zh:
> This passes my smell check. Have we confirmed that changing the user to Pelican doesn't prevent us from exporting public keys through the `.well-known/` path at our XRootD endpoint?

Public key can be obtained by sending a GET request to `https://localhost:8447/.well-known/issuer.jwks` (8447 is the port I run Origin), while not on `https://localhost:8443/.well-known/issuer.jwks`

h2zh:
Alarm cleared! No need to create another issue. I double checked and found out `https://localhost:8443/.well-known/issuer.jwks` **CAN** get the public key before **AND** after this PR. My previous test said "no" because it prompted me to provide a certificate when accessing this endpoint. I just found out that I could just click "cancel" to bypass this blocker and access the public key.

PR #2072, opened 2025-03-04
Use the web port for the token issuer

brianhlin:
There has been talk about removing the issuer from the xrootd
channel (e.g., port 8443)

### DIFF
--- a/docs/pages/federating-your-data/generating-tokens.mdx
+++ b/docs/pages/federating-your-data/generating-tokens.mdx
@@ -11,7 +11,7 @@ The Pelican binary comes with a command `pelican origin token create` to generat
 ```bash copy
 pelican origin token create \
     --scope "storage.read:/" \
-    --issuer https://localhost:8443 \
+    --issuer https://localhost:8444 \
     --claim "wlcg.ver=1.0" \
     --subject origin \
     --audience https://wlcg.cern.ch/jwt/v1/any


brianaydemir:
While we're here, that same documentation page also says:

> `--issuer` tells Pelican the issuer URL it should embed in the token, which will be used by other Pelican servers to verify the integrity of the token. The issuer should be the publicly accessible URL of the Origin server.

`localhost` isn't publicly accessible, regardless of the port that one might try to contact it on. Which is to say, I'm not certain that the `localhost` in the example is correct.

jhiemstrawisc (approved):
LGTM. I'll approve, but if you want to take @brianaydemir's advice and change from `localhost` to something else, go ahead. I don't feel super strongly about it because no matter what we put here, it'll have to be changed by the user. Maybe we can eventually start labeling these bash snippets to indicate "you can copy paste" vs "you have to make changes".

brianhlin:
@jhiemstrawisc before I merge, does the web issuer still work if folks disable the UI? See https://github.com/osg-htc/docs/pull/238#discussion_r1985433350

brianaydemir:
@brianhlin The built-in OA4MP issuer is not usable when `Server.EnableUI` is set to `false` because disabling the UI also disables the OAuth endpoints for authenticating the user. (Why? I don't know.)

jhiemstrawisc:
Tested locally, and setting `Server.EnableUI: false` still lets me grab the configured openid config and public key from under the path `<we host/port>/.well-known/*`. However, the suggestion linked to in those docs says opening this port is optional, at which point lots of stuff will break.

jhiemstrawisc:
Whoops, not sure what I hit on the keyboard, but some macro closed the PR unintentionally.

jhiemstrawisc:
`localhost` not withstanding, I think this is still a slight improvement so I'm going to merge it. I think there's a good chance we'll spend some time working with our token generation documentation during the March 25th docs focus, given how much this documentation has been coming up lately.

brianhlin:
> However, the suggestion linked to in those docs says opening this port is optional, at which point lots of stuff will break.

@jhiemstrawisc could you put a finer point on this? CC @matyasselmeci

jhiemstrawisc:
Do we already start origins without an externally-accessible web UI port? I'm surprised [this line](https://github.com/PelicanPlatform/pelican/blob/24b80868ae9261283d5ff22b50d03a291543f532/launchers/launcher.go#L215-L219) in our launcher sequence doesn't brick if we do 🤔

PR #2075, opened 2025-03-05
Added use of acceptable_authorization to origin scitokens files

turetske:
-- This will enforce origin capabilities for scitokens -- Adjusted the scitokens origin tests in authorization_test.go to test this functionality
-- Added test files to xrootd/resources

### DIFF
--- a/xrootd/authorization.go
+++ b/xrootd/authorization.go
@@ -61,14 +61,15 @@ type (
 
 	// Per-issuer configuration
 	Issuer struct {
-		Name            string
-		Issuer          string
-		BasePaths       []string
-		RestrictedPaths []string
-		MapSubject      bool
-		DefaultUser     string
-		UsernameClaim   string
-		NameMapfile     string
+		Name                    string
+		Issuer                  string
+		BasePaths               []string
+		RestrictedPaths         []string
+		MapSubject              bool
+		DefaultUser             string
+		UsernameClaim           string
+		NameMapfile             string
+		AcceptableAuthorization string
 	}
 
 	// Top-level configuration object for the template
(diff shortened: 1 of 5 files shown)


xrootd/resources/test-scitokens-origin-read.cfg
```diff
@@ -0,0 +1,35 @@
+#
+# Copyright (C) 2024, Pelican Project, Morgridge Institute for Research
```
jhiemstrawisc, line 2: Since we're officially changing the scitokens config template, can you update the copyright statements?

xrootd/authorization_test.go
```diff
@@ -727,37 +739,86 @@ func TestWriteCacheAuthFiles(t *testing.T) {
...
-	viper.Set("Origin.Port", 8443)
-	viper.Set("Server.WebPort", 8444)
-	viper.Set("Server.Hostname", "origin.example.com")
-	viper.Set(param.Origin_StorageType.GetName(), string(server_structs.OriginStoragePosix))
+			server_utils.ResetTestState()
+			defer server_utils.ResetTestState()
 
-	err := config.InitServer(ctx, server_structs.OriginType)
-	require.NoError(t, err)
+			viper.Set("Origin.Port", 8443)
```
jhiemstrawisc, line 782: I know a lot of this code is going to get hit with merge conflicts so I'm not sure if this comment will be relevant soon, but can you switch any of the `viper.Set`s you touch to use `param.XXX.GetName()` for the key (the way it's done in line 786 here)?

xrootd/authorization.go
```diff
@@ -529,6 +540,17 @@ func makeSciTokensCfg() (cfg ScitokensCfg, err error) {
...
 }
 
+func resolveAuthorization(auth1 string, auth2 string) string {
+	if auth1 == auth2 {
+		return auth1
+	} else {
+		// Either one is read and the other is write, in which case we want the default of
+		// all, or one is all and the other is read/write, and we'd still want it to be the default
+		// of all, which is equivalent to the empty string
+		return ""
```
jhiemstrawisc, line 550: I think I'm more in favor of using "all" in this case instead of relying on the empty string's default behavior. That should make it a bit easier for Fabio to understand the configs we generate without having to find comments in source code.

jhiemstrawisc (changes requested):

xrootd/authorization.go
```diff
@@ -479,6 +480,16 @@ func GenerateOriginIssuer(exportedPaths []string) (issuer Issuer, err error) {
...
 
+	// This will be set to true if PublicReads are enabled as well and is slightly redundant
+	// as then a token wouldn't be needed for reads, but doesn't actually change the functionality
+	// The default if the acceptable authorization isn't set is "all", so we only care if
+	// Reads is Enabled without Writes or vice versa
+	if param.Origin_EnableReads.GetBool() && !param.Origin_EnableWrites.GetBool() {
+		issuer.AcceptableAuthorization = "read"
+	} else if param.Origin_EnableWrites.GetBool() && !param.Origin_EnableReads.GetBool() {
+		issuer.AcceptableAuthorization = "write"
+	}
```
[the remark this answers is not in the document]
turetske: I think we just have to use the "all", which is what's happening now. There isn't a way to adjust this without a greater change to the xrootd code. This is already not as granular as we'd like (no difference between write/create/modify, etc.) so I think we just need to take what we can get for now.

PR #2077, opened 2025-03-05
Add RPM and Deb dependencies for the programs required to run the pre-install script

### LINKED ISSUE #2060: Pelican rpm now depends on /bin/sh but bash is not a dependency
Pelican Version:
7.13.0

Pelican Service:
- [ x] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

**Describe the bug**
This is a packaging bug more than a pelican bug. The yum/dnf install of the pelican package can fail if /bin/sh is not available. This is a corner case, but it is possible if you are creating an image from scratch. It came up in the image build process on the PSU GWave cluster.

The dependency is in [scripts/preinstall.sh](https://github.com/PelicanPlatform/pelican/blob/main/scripts/preinstall.sh).

**To Reproduce**
1. On a system without /bin/sh:
```
    rpm -ivh pelican-7.13.0-1.x86_64.rpm
```
The output will include text like:
```
 Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
error: failed to exec scriptlet interpreter /bin/sh: No such file or directory
error: %prein(pelican-7.13.0-1.x86_64) scriptlet failed, exit status 127
error: pelican-7.13.0-1.x86_64: install failed
```
    

**Expected behavior**
Yum/dnf should recognize that bash (/bin/sh) is required by pelican and install bash before the pelican RPM install that requires /bin/sh.

 - OS: Linux

**Additional context**
This came up as part of an automated build where bash and pelican are installed in the same dnf command.
This was discussed in #pelican on the OSG slack on Feb 27, 2025.

matyasselmeci:
Add the dependencies required for the pre-script (which creates the pelican user and group) to run. Note that GoReleaser does not support the `Requires(pre)` header for RPMs, so even with the dependency, there is still no guarantee that e.g. /bin/sh will exist on disk by the time the pre script is invoked.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -168,6 +168,12 @@ nfpms:
               mode: 0755
               owner: pelican
               group: pelican
+        dependencies:
+          - "/bin/sh"
+          # for useradd and groupadd:
+          - "shadow-utils"
+          # for getent:
+          - "glibc-common"
       deb:
         contents:
           - src: LICENSE
@@ -204,6 +210,13 @@ nfpms:
               mode: 0755
               owner: pelican
               group: pelican
+        dependencies:
+          # provides /bin/sh:
+          - "dash"
+          # for useradd and groupadd:
+          - "passwd"
+          # for getent:
+          - "libc-bin"
       apk:
         contents:
           - src: LICENSE


brianaydemir:
My testing, as it were:

- Confirmed that GoReleaser does indeed not seem to have support for `Requires(pre)` (nor does it look like a trivial feature to add).
- Confirmed that the RPM's newly dependencies do actually provide what the pre-script requires.
- Confirmed that the RPM installs into a basic `almalinux:9` container.
- Ditto the previous two points for the DEB and `ubuntu:24.04`.

brianaydemir (approved):
LGTM.

PR #2080, opened 2025-03-06
Add log buffering mechanism using Logrus hook

Saartank:
Implemented log buffering by creating a Logrus hook that buffers logs in memory until a flush is triggered. The flush runs only once, after which the hook is disabled.

Currently, the flush is triggered in `InitClient` and `InitServer`, sending buffered logs to the configured destination (stdout or file). For `config` commands that do not call `InitClient` or `InitServer`, logs are buffered until the program exits or terminates unexpectedly, at which point they are flushed to stdout without creating a log file.

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -112,7 +112,7 @@ func FlushLogs(pushToFile bool) {
 				DisableLevelTruncation: true,
 			})
 		} else {
-			logs.SetOutput(os.Stdout)
+			logs.SetOutput(os.Stderr)
 
 			// Restore colorized output when logging to stdout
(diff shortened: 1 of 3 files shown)


logging/logging.go
```diff
@@ -0,0 +1,188 @@
...
+		os.Exit(exitCode)
+	}()
+
+	logs.RegisterExitHandler(func() {
+		fmt.Println("\nExit handler triggered - Flushing logs...")
+		FlushLogs(false)
+	})
+}
+
+func init() {
```
bbockelm, line 174: Move this out of `init()` and into an explicit function in `cmd`.

PR #2082, opened 2025-03-06
Fix index out of bound error in attrs.Files

h2zh:
Fix the error `panic: runtime error: index out of range [3] with length 3` and add comment to explain this code snippet.


bbockelm (approved):
LGTM.

There's a failing PR validation action.  Once that's passing, feel free to merge.

Issue #2084, opened 2025-03-07
`pelican generate keygen` creates invalid public keys


jhiemstrawisc:
It looks like the `pelican generate keygen` command is creating public keys that don't have a `kid` or an `alg` field, both of which are required for token verification.

From @brianaydemir:
```
[baydemir@local Desktop]> pelican generate keygen --private-key foo --public-key bar
Successfully generated keys at: 
Private key: foo
Public Key: bar
[baydemir@local Desktop]> cat foo 
-----BEGIN PRIVATE KEY-----
<keep it secret, keep it safe>
-----END PRIVATE KEY-----
[baydemir@local Desktop]> cat bar 
{
        "keys": [
                {
                        "crv": "P-256",
                        "kty": "EC",
                        "x": "IzQKHe5_dg1J97rdyNgmjTjoEaiFtjG9eHdtctXmDsU",
                        "y": "zV1uxDLfu6AQh9oTHEsx-I14ffZaxNO1LvSBPvEH0hU"
                }
        ]
}[baydemir@local Desktop]> pelican --version
Version: 7.14.0
Build Date: 2025-03-05T17:39:03Z
Build Commit: 7bb5a5a4b1bae77fdf09b4c0ad9d1afffb22a98c
Built By: goreleaser
```

Brian also reports that testing this command with Pelican v7.9.X generates a valid public key, so the fix here should really have a regression test.

jhiemstrawisc:
I'll further note that starting an origin locally does result in a valid public key at `<origin url>/.well-known/issuer.jwks`, so there must be drift in the way we're producing these keys. Let's make sure they're all driven by the same function when we implement a fix.

brianaydemir:
7.13.1 is affected by this bug. 7.12.4 is **not** affected.

brianaydemir:
In 7.12.4 and earlier, `generate keygen` relied on the same machinery as the server initialization flow:

https://github.com/PelicanPlatform/pelican/blob/d7aea0e34573f10997325bd57ca50039655b7bc2/cmd/generate_keygen.go#L71-L75

In 7.13, the command was updated with its own key generation flow, because the server now initializes a directory of private keys. While I'm having trouble groking the code flow, I suspect something like the following got dropped in the process:

https://github.com/PelicanPlatform/pelican/blob/d7aea0e34573f10997325bd57ca50039655b7bc2/config/init_server_creds.go#L522-L533

PR #2085, opened 2025-03-07
Fix `generate keygen` so that public keys included required fields

brianaydemir:
`config.GeneratePrivateKey` is documented as generating a PEM-encoded key, so I've updated the code underlying `generate keygen` to use what seems to be the function for reading such PEM-encoded keys.

(There seem to be multiple functions for "reading" private keys, so I went with calling the function whose implementation most directly seems to address the problems of #2084.)

### DIFF
--- a/cmd/generate_keygen.go
+++ b/cmd/generate_keygen.go
@@ -19,7 +19,6 @@
 package main
 
 import (
-	"crypto"
 	"crypto/elliptic"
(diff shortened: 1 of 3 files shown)


brianaydemir:
I did confirm that my updates to the unit tests failed without the changes made in this PR.

bbockelm (approved):

PR #2086, opened 2025-03-10
Create API token listing endpoint

### LINKED ISSUE #2069: Add Listing Endpoint for API Tokens
Currently there is not an endpoint to list all tokens. This would be critical for integrating API token feature into the web UI.

patrickbrophy:
Closes #2069. This PR adds an endpoint to list all of the api tokens. This new endpoint is available with the other director web UI endpoints.

### DIFF
--- a/web_ui/ui.go
+++ b/web_ui/ui.go
@@ -507,6 +507,35 @@ func deleteApiToken(ctx *gin.Context) {
 	})
 }
 
+func listApiTokens(ctx *gin.Context) {
+	authOption := token.AuthOption{
+		Sources: []token.TokenSource{token.Cookie},
+		Issuers: []token.TokenIssuer{token.LocalIssuer},
+		Scopes:  []token_scopes.TokenScope{token_scopes.WebUi_Access},
+	}
+	status, ok, err := token.Verify(ctx, authOption)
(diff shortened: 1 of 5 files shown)


web_ui/ui.go
```diff
@@ -507,6 +507,35 @@ func deleteApiToken(ctx *gin.Context) {
...
 
+func listApiTokens(ctx *gin.Context) {
+	authOption := token.AuthOption{
+		Sources: []token.TokenSource{token.Cookie},
+		Issuers: []token.TokenIssuer{token.LocalIssuer},
+		Scopes:  []token_scopes.TokenScope{token_scopes.WebUi_Access},
+	}
+	status, ok, err := token.Verify(ctx, authOption)
+	if !ok {
+		log.Warningf("Cannot verify token: %v", err)
```
jhiemstrawisc, line 518: Can you add a bit more context to the warning to indicate more about which function is throwing the error? I worry we have so many "tokens" in Pelican, it won't be obvious what kind of token is failing to verify in these logs.

jhiemstrawisc (changes requested):
Small change request inline. The other request is for some unit test coverage of the new endpoint and the new database function.

Issue #1054, opened 2024-04-05
Support generic OIDC auth provider


haoming29:

jhiemstrawisc:
@haoming29, is this closed by the linked PR? If not, can I assign to you as a 7.9 milestone?

haoming29:
Move to 7.10 as we made a lot of changes to OIDC in 7.9 and I don't want to blow up everything with this addition

brianaydemir:
@jhiemstrawisc With the PRs for the issues mentioned above now merged into `main` (#2046, #2091, #2100), Pelican is now capable of using Microsoft AD FS and Auth0 as its authentication provider, in addition to CILogon and Globus. Whether we have completely exorcised from Pelican its assumptions on how CILogon implements OIDC is perhaps not a completely settled question.

But as I said earlier in this thread, there are no real details in this issue as to what constitutes success, so I'm substituting my own standard and closing this issue: Make it so Pelican can use something that isn't CILogon or Globus for authentication. We can always create new issues documenting specific shortcomings in Pelican's OIDC implementation.

PR #2091, opened 2025-03-11
Base64 encode the OAuth state to avoid a double URL-encoded string

brianaydemir:
Given that [double encoding attacks](https://en.wikipedia.org/wiki/Double_encoding) are a thing, in order to prevent OAuth providers from mangling the `state` parameter that Pelican constructs, the simplest fix might be to set the parameter's value to a value that does not contain any URL encoding. I've chosen to base-64 encode it.

And while not strictly related to the above issue, I'm taking this opportunity to add a line to the debug logging, specifically to record the user info received from an OAuth provider. This is related to making #1054 a reality (same as this PR), and it's otherwise "challenging" to get any insight into this part of the OAuth flow.

### DIFF
--- a/web_ui/oauth2_client.go
+++ b/web_ui/oauth2_client.go
@@ -360,6 +360,7 @@ func handleOAuthCallback(ctx *gin.Context) {
 			})
 		return
 	}
+	log.Debugf("User info from auth provider: %v", string(body))
 
 	var userInfo map[string]interface{}
 	if err := json.Unmarshal(body, &userInfo); err != nil {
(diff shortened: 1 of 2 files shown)


h2zh (approved):
LGTM. 
Note to all Pelican devs: we use lots of query parameters in Pelican APIs, though we may just perform the encoding here in the OAuth feature. Apparently, query parameters encoding is unnecessary for most straightforward cases. But it's worth to consider perform base64 encoding (just like what this PR does) when passing data to the systems in the outer world, which might trigger double-encoding. @turetske @jhiemstrawisc @Saartank  

I did some research and here is the note to myself: 
Base64 eliminates the risk of double-encoding: it produces a character set (without %) that doesn't trigger additional encoding when passed through some systems that automatically apply URL encoding (e.g. % -> %25).

PR #2097, opened 2025-03-13
Fix how paths in token scopes are calculated

brianaydemir:
Back in https://github.com/PelicanPlatform/pelican/commit/536b5427f891edb275ccf4ca2b4f6c1f0a8f29dc, we modified how we calculated the path to include in a token request to always strip off any potential filename. However, the original implementation fails in the case where the path being requested is exactly some namespace prefix. We end up treating the final component of the prefix as a filename that needs removing.

With this PR, we'll determine first what the requested path needs to be relative to, and then only after that will we strip off any potential filename. In this process, it's worth keeping in mind that [the WLCG profile](https://github.com/WLCG-AuthZ-WG/common-jwt-profile/blob/master/profile.md#authorization) requires that the path be non-empty:

> For all `storage.*` scopes, `$PATH` MUST be specified (it may be `/` to authorize the entire resource associated with the issuer)

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -80,10 +80,10 @@ func AcquireToken(issuerUrl string, entry *config.PrefixEntry, dirResp server_st
 		return nil, fmt.Errorf("issuer at %s for prefix %s does not support device flow", issuerUrl, entry.Prefix)
 	}
 
-	// Always trim the filename off the path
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
LGTM. I tested this using the same commands that generated the panic in the first place and confirmed the token scopes weren't being created with empty paths.

PR #2100, opened 2025-03-14
Add config parameter to allow preferring OIDC ID token claims

brianaydemir:
This PR introduces a new configuration parameter, `Issuer.OIDCPreferClaimsFromIDToken`, that when when set to `true`, causes Pelican's user authentication flow to first look for claims in the OIDC ID token before falling back on the OIDC UserInfo endpoint.

Pelican's current behavior is to consider only the UserInfo endpoint, hence why this new parameter's default value is `false`.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2298,9 +2298,21 @@ type: object
 default: []
 components: ["origin"]
 ---
+name: Issuer.OIDCPreferClaimsFromIDToken
+description: |+
+  If set to `true`, then the claims specified by
+  `Issuer.OIDCAuthenticationUserClaim` and `Issuer.GroupSource` will first
+  be searched for in the OIDC ID token before falling back on claims from
+  the UserInfo endpoint.
+
+  If set to `false`, then only the OIDC UserInfo endpoint will be considered.
+type: bool
+default: false  # this was the behavior prior to this parameter's introduction
+components: ["origin"]
+---
 name: Issuer.OIDCAuthenticationUserClaim
 description: |+
-  The claim in the OIDC ID token to be used as the "username" for the issuer.
+  The claim to be used as the "username" for the issuer.
 type: string
 default: sub
 components: ["origin"]
@@ -2331,7 +2343,7 @@ components: ["origin"]
 ---
 name: Issuer.OIDCGroupClaim
 description: |+
-  The claim in the OIDC ID token to be used as the group for the issuer.  If the value is a string,
+  The claim to be used as the group for the issuer.  If the value is a string,
   it is assumed that a comma is used as a group delimiter; otherwise, an array of strings is assumed.
   Check the documentation of your OIDC provider to determine the appropriate claim name.
 type: string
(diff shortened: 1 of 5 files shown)


web_ui/oauth2_client.go
```diff
@@ -179,12 +182,16 @@ func generateGroupInfo(user string) (groups []string, err error) {
 
 // Given a map from a JSON object, generate user/group information according to
 // the current policy.
-func generateUserGroupInfo(userInfo map[string]interface{}) (user string, groups []string, err error) {
+func generateUserGroupInfo(userInfo map[string]interface{}, idToken map[string]interface{}) (user string, groups []string, err error) {
```
h2zh, line 185: Could you modify the docstring of this function to incorporate the new `idToken` input and how it works in this function?

web_ui/oauth2_client.go
```diff
@@ -179,12 +182,16 @@ func generateGroupInfo(user string) (groups []string, err error) {
...
+	claimsSource := maps.Clone(userInfo)
+	if param.Issuer_OIDCPreferClaimsFromIDToken.GetBool() {
+		maps.Copy(claimsSource, idToken)
```
h2zh, line 188: Based on the code logic, if OIDC ID token is not provided, the `claimsSource` will be the same as `userInfo`. Is this what happened when you test it?

docs/parameters.yaml
```diff
@@ -2298,9 +2298,21 @@ type: object
 default: []
 components: ["origin"]
 ---
+name: Issuer.OIDCPreferClaimsFromIDToken
+description: |+
+  If set to `true`, then the claims specified by
+  `Issuer.OIDCAuthenticationUserClaim` and `Issuer.GroupSource` will first
+  be searched for in the OIDC ID token before falling back on claims from
+  the UserInfo endpoint.
```
h2zh, line 2306: I think the description here may be too long for one sentence and a little bit ambiguous for our user. Could you break it into two or more sentences?

h2zh (changes requested):
Looks good and I just have some trivial questions here and there!

docs/parameters.yaml
```diff
@@ -2298,9 +2298,21 @@ type: object
...
+
+  If set to `false`, then only the OIDC UserInfo endpoint will be considered.
+type: bool
+default: false  # this was the behavior prior to this parameter's introduction
```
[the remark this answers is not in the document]
brianaydemir: I think it's good to document, at least internally, why we chose a particular default value for a parameter, especially in a case like this, where one could the default should be `true`.

web_ui/oauth2_client.go
brianaydemir: Short answer: Yes.

Longer answer: If the ID token isn't provided, then the provider has a serious issue because it's actually [a required part of the response](https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse). That said, it's probably better if `handleOAuthCallback` initializes it to an empty map, because I don't know what `maps.Copy` will do if it gets `nil`.

docs/parameters.yaml
brianaydemir: Done.

web_ui/oauth2_client.go
brianaydemir: Done.

h2zh (approved):
LGTM🚀

PR #2102, opened 2025-03-14
Install `sssd-client` into both origin container images

brianaydemir:
Curiously, at least to me, `sssd-client` doesn't create an `sssd` UID and GID.

If I recall correctly, the reason we create an `sssd` UID and GID is so that the sockets used to communicate with the actual `sssd` daemon are owned by the daemon, which is presumably running as that UID and GID. But in at least one instance I have access to, the sockets are public read/write and owned by `root` (which is what the daemon is running as).

So, it doesn't look like we actually need to create the `sssd` UID and GID, except perhaps as a nicety for those looking at the sockets on the file system in certain specific setups.


williamnswanson (approved):
LGTM

PR #2103, opened 2025-03-14
Enable support for multiple issuers per namespace

jhiemstrawisc:
This lets origin administrators do the following in their config file:
```
Origin:
  Exports:
    - StoragePrefix: /foo
      FederationPrefix: /foo
      IssuerUrls: ["https://issuer1.com", "https://issuer2.com"]
    - StoragePrefix: /bar
      FederationPrefix: /bar
      IssuerUrls: ["https://issuer1.com", "https://issuer3.com"]
    - StoragePrefix: /baz
      FederationPrefix: /baz
     # no list provided falls back to value from `config.GetServerIssuerURL()`
```

I accomplished this by treating issuers as part of each OriginExport object, which changes a few of our assumptions about setting things up (see commit RE pre-allocating ports for XRootD). Luckily, we were forward thinking enough to design server/origin/namespace ads such that they already take a list of issuers, meaning there are no backwards compatibility issues to be concerned about there.

Otherwise, I'm punting slightly on re-defining some of the other issuer types (Origin server issuer as opposed to xrootd data issuers). Once we're sure this is working as intended, it'll let us stop registering services in Topology, which we're still doing to handle this multi-issuer case.


docs/parameters.yaml
```diff
@@ -934,6 +939,19 @@ type: string
 default: none
 components: ["origin"]
 ---
+name: Origin.AudienceUrl
```
brianaydemir, line 942: I don't think the configuration parameter's name and description need to mention "URL." Neither SciTokens nor WLCG _requires_ that the "Audience" (`aud` claim) actually be a URL.

WLCG strongly recommends using the relevant URL for services that speak HTTP(S), and URIs for other services. SciTokens is less opinionated and merely requires the same as RFC 7519 (the JWT spec), and unlike WLCG, uses a simple string for the "any" audience (literally, `ANY`).

jhiemstrawisc: Technically changing it at this point would constitute a breaking change, since I'm only really documenting a configurable parameter that was previously undocumented (and accessing it slightly differently). It's possible we've had people using this all along even though we didn't write down its existence. @turetske, I think you set up `Origin.AudienceUrl` before we had the `param` package, so how do you think we should handle this?

If we do want to change it, how do we feel about `Origin.TokenAudience` as a new name?

brianaydemir: If the parameter was previously undocumented, the only people using it would be people who went diving through the source code and people we, the developers, talked to. We can't let the former limit our ability to make changes, and the latter seems solvable?

jhiemstrawisc: In light of @bbockelm's work in[ this commit](https://github.com/PelicanPlatform/pelican/pull/2023/commits/a284b57f4e250d656e8a1c524afccf86e9871af7), it does seem like he's make a hard assumption that these audience strings _must_ be a URL. BrianB, can you weigh in?

turetske: I will also say that anything I made before we had the param package would have been in my first month of work here and I'm not attached to anything I did when I didn't know enough.

xrootd/xrootd_config.go
```diff
@@ -271,12 +271,8 @@ func CheckOriginXrootdEnv(exportPath string, server server_structs.XRootDServer,
 			" to desired daemon group %v", macaroonsSecret, groupname)
 	}
 	// If the scitokens.cfg does not exist, create one
-	if originServer, ok := server.(*origin.OriginServer); ok {
-		authedPrefixes, err := originServer.GetAuthorizedPrefixes()
```
turetske, line 275: This and the other call to GetAuthorizedPrefixes are the only two calls to that function, so I believe it should be removed for the cleanup.

turetske (changes requested):
I foresee a lot of merge conflicts between this and my PR involving scopes because we both touched a lot of the same code, but that's a problem for future me to rebase and deal with after this has been merged in.

I have a few questions and comments, otherwise, before I approve.

server_utils/origin.go
```diff
@@ -249,6 +288,33 @@ func validateFederationPrefix(prefix string) error {
...
+// handleIssuersIfNeeded checks if the export has any issuer URLs configured. If not, it attempts to
+// configure one using the server's issuer URL (which may also be the server's external web URL -- see
+// config.GetServerIssuerURL()). Not all exports require an issuer, so we skip configuration if the
+// origin doesn't require one (e.g. public reads or no writes).
+func handleIssuersIfNeeded(exports []OriginExport) error {
+	for i, export := range exports {
+		// If the export doesn't have any issuers in its list, attempt to configure one using the default
+		// server issuer. Generally we shouldn't _require_ an issuer as some exports may be truly public
+		// and not require any auth
+		if len(export.IssuerUrls) == 0 && ((export.Capabilities.Reads && !export.Capabilities.PublicReads) || export.Capabilities.Writes) {
```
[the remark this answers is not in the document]
jhiemstrawisc: Since token auth for listings is completely tied to the value of `Reads` or `PublicReads`, I don't think it's even valid to set it without one of those being true. That being said, I tried starting an origin with only listings set, and it started without any complaints 🙃

config/config.go
```diff
@@ -569,7 +569,7 @@ func GetServerIssuerURL() (issuerUrl string, err error) {
 	// Even though we prefer using this function, we'll populate the config param
 	// based on whatever we determine here.
 	defer func() {
-		if err == nil && param.Server_IssuerUrl.GetString() == "" {
+		if err == nil && param.Server_IssuerUrl.GetString() != "" {
```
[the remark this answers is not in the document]
jhiemstrawisc: Good catch, I think this survived a few early attempts I had to fix another bug related to fed tests. A new commit will correct the logic and fix the fed tests to work in this case.

brianaydemir (commented):
I've confirmed that the changes here are independent of the origin's built-in OA4MP issuer.

@jhiemstrawisc I did observe one oddity during testing, though.

Suppose we set `Origin.Exports[].IssuerUrls` to two URLs. We then attempt a `pelican object put` to that export when the client has no previously saved token. The client will attempt to send the user to two URLs: the first element of `Origin.Exports[].IssuerUrls`, followed by `Server.IssuerUrl`.

I find it odd that that the second element of `Origin.Exports[].IssuerUrls` is ignored.

Alternatively, I find it odd that any elements of `Origin.Exports[].IssuerUrls` are considered by the client given that the so-called "token generation" URL in the director UI is actually `Server.IssuerUrl`.

turetske (approved):
LGTM

Issue #1434, opened 2024-06-18
Revise Prometheus metric names to follow best practices


haoming29:
https://prometheus.io/docs/practices/naming/

PR #2105, opened 2025-03-17
Revise Prometheus Metric Names

patrickbrophy:
This PR addresses #1434. To maintain backwards compatibility, this adds metrics with new names instead of changing the existing names. Eventually we will want to use just the newly named metrics.


jhiemstrawisc (changes requested):
All the new stuff looks okay to me. Can you go through and explicitly comment/mark all of the metrics that are being replaced with a TODO to delete them along with a note about which version the new metrics were added in? Then we can file an issue to track deleting the old metrics once we see that the federations we maintain aren't running anything older than 7.16.

PR #2106, opened 2025-03-18
Enable crc32 and crc32c algorithms

bbockelm:
This helps prepare Pelican for future end-to-end checksumming support by turning on a consistent set of algorithms, including `crc32c` (which I currently plan to use as the default).

Functionality at the origin tested by hand.


h2zh (approved):
LGTM

PR #2108, opened 2025-03-18
Untie prefix from Broker communication paths

Saartank:
This PR refactors the code so that all communication paths rely solely on the origin identifier (hostname + port) and no longer depend on the prefixes exported by the origin.


bbockelm:
This covers the same ground as #2378 (which also provides functionality for the cache).  Closing this one in favor of that.

Many thanks @Saartank!

PR #2114, opened 2025-03-19
Fix signal interrupt handling in log buffering code

Saartank:
Previously, the log buffering code handled interrupt signals but prevented the main cleanup routine from executing. This update ensures that log flushing on interrupt is correctly integrated with the existing cleanup logic.

Also corrected output syncing: replaced `os.Stdout.Sync()` with `Sync()` on the configured log destination.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -36,6 +36,7 @@ import (
 	"github.com/pelicanplatform/pelican/cmd/config_printer"
 	"github.com/pelicanplatform/pelican/config"
 	"github.com/pelicanplatform/pelican/launchers"
+	"github.com/pelicanplatform/pelican/logging"
 )
 
 type uint16Value uint16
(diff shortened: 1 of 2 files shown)


Saartank:
### Reproducing the Problem and Verifying if This PR Solves It

Before this PR, if we start the cache server and then stop it by sending an interrupt,  
we will not see the log:  
`"Pelican has safely exited"` (which is logged after the cleanup completes),  
and the `xrootd` process will not be killed even after the cache stops.

After this PR, we will see the log:  
`"Pelican has safely exited"`  
and the `xrootd` process will be stopped as well after the Pelican cache stops.

jhiemstrawisc (approved):
LGTM. I tested the two situations and confirmed the failure before the diff and no failure after the diff. Thanks!

PR #2116, opened 2025-03-19
Fix client bug where schemeless pel URLs don't work with '-f' fed info

jhiemstrawisc:


brianaydemir (commented):
Right now, we have code for "determine the discovery URL (to use)" duplicated three times now, with minor differences. Is there a good reason not to take something like

https://github.com/PelicanPlatform/pelican/blob/24b80868ae9261283d5ff22b50d03a291543f532/client/main.go#L74-L94

and turn it into a standalone function?

Otherwise, I'm left wondering why some functionality falls back on the director URL as a last resort, and some doesn't. And pointing out that "info info" in `"failed to determine federation info info"` reads oddly.

brianaydemir:
@jhiemstrawisc Because "maintainers" are allowed to edit PRs, I went ahead and tweaked a few comments and one error message myself.

brianaydemir (approved):

PR #2121, opened 2025-03-20
Re-key the card list items

CannonLock:
Previously I was converting the incoming object to a string. Issue with this is the object was not deterministic so everytime the data updated React rerendered everything with new state because it thought they were all new objects.

### DIFF
--- a/web_ui/frontend/app/director/components/DirectorCardList.tsx
+++ b/web_ui/frontend/app/director/components/DirectorCardList.tsx
@@ -86,7 +86,12 @@ export function DirectorCardList({ data, cardProps }: DirectorCardListProps) {
           </Grid>
         </Grid>
       </Box>
-      <CardList data={filteredData} Card={DirectorCard} cardProps={cardProps} />
+      <CardList
+        data={filteredData}
+        Card={DirectorCard}
+        cardProps={cardProps}
+        keyGetter={(o) => o?.server?.name || 'undefined'}
+      />
     </Box>
   );
 }
--- a/web_ui/frontend/app/director/components/DirectorCardList.tsx
+++ b/web_ui/frontend/app/director/components/DirectorCardList.tsx
@@ -90,7 +90,7 @@ export function DirectorCardList({ data, cardProps }: DirectorCardListProps) {
         data={filteredData}
         Card={DirectorCard}
         cardProps={cardProps}
-        keyGetter={(o) => o?.server?.name || 'undefined'}
+        keyGetter={(o) => o.server.name}
       />
     </Box>
   );
--- a/web_ui/frontend/components/CardList.tsx
+++ b/web_ui/frontend/components/CardList.tsx
@@ -15,14 +15,13 @@ import {
   TextField,
   Typography,
 } from '@mui/material';
-import { objectToDeterministicString } from '@/helpers/util';
 
 interface CardListProps<T> {
   data?: Partial<T>[];
   Card: ComponentType<any>;
   cardProps?: Partial<T>;
   pageSize?: number;
-  keyGetter: (o: Partial<T>) => string;
+  keyGetter: (o: T) => string;
 }
 
 export function CardList<T>({
@@ -59,7 +58,7 @@ export function CardList<T>({
           } as T;
 
           return (
-            <Box pb={1} key={keyGetter(o)}>
+            <Box pb={1} key={keyGetter(props)}>
               <Card {...props} />
             </Box>
           );
--- a/web_ui/frontend/components/CardList.tsx
+++ b/web_ui/frontend/components/CardList.tsx
@@ -15,18 +15,21 @@ import {
   TextField,
   Typography,
 } from '@mui/material';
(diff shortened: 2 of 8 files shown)


web_ui/frontend/components/CardList.tsx
```diff
@@ -13,14 +13,21 @@ import {
   TextField,
   Typography,
 } from '@mui/material';
+import { objectToDeterministicString } from '@/helpers/util';
```
h2zh, line 16: Where do we use this func?

web_ui/frontend/helpers/util.ts
```diff
@@ -159,3 +159,26 @@ export const multiSort = <T>(arr: T[], ...args: ((a: T, b: T) => number)[]) => {
...
+/**
+ * Convert an object to a string key that is robust to unsorted lists
+ */
+export const objectToDeterministicString = (o: any): string => {
+  return Object.keys(o)
+    .sort()
+    .reduce((p, k) => {
+      const value = o[k];
+
+      // If the value is an array then ignore it
```
h2zh, line 172: I wonder why array is discarded..

web_ui/frontend/app/director/components/DirectorCardList.tsx
```diff
@@ -84,7 +84,12 @@ export function DirectorCardList({ data, cardProps }: DirectorCardListProps) {
           </Grid>
         </Grid>
       </Box>
-      <CardList data={filteredData} Card={DirectorCard} cardProps={cardProps} />
+      <CardList
+        data={filteredData}
+        Card={DirectorCard}
+        cardProps={cardProps}
+        keyGetter={(o) => o?.server?.name || 'undefined'}
```
h2zh, line 91: React requires a unique key for each child in a list, while calling an undefined function would throw an error I think, even it's unlikely happening. But keeping these code make me nervous... Maybe you can set a default func for keyGetter uses Math.random() to generate a unique key

web_ui/frontend/components/layout/Navigation/Navigation.tsx
```diff
@@ -25,7 +25,9 @@ const Navigation = ({
     throw new Error('Either config xor sharedPage must be defined');
   }
 
-  const { data: exports } = useSWR('getDataExport', getExportData);
+  const { data: exports } = useSWR('getDataExport', getExportData, {
+    errorRetryCount: 0,
```
h2zh, line 29: I wonder if you could explain why this is relevant in this PR and how `getExportData` consume the pass-in `errorRetryCount: 0`?

web_ui/frontend/helpers/util.ts
```diff
@@ -159,3 +159,26 @@ export const multiSort = <T>(arr: T[], ...args: ((a: T, b: T) => number)[]) => {
     return 0;
   });
 };
+
+/**
+ * Convert an object to a string key that is robust to unsorted lists
+ */
+export const objectToDeterministicString = (o: any): string => {
```
h2zh, line 166: Can't find any line calling this func

h2zh (changes requested):
A few puzzles

web_ui/frontend/helpers/util.ts
CannonLock: We don't but it is still useful so I kept it around. I couldn't use it here because the actual values in the card props do change each time put the prop that changes isn't consumed by the card so we don't want it to toggle a rerender.

web_ui/frontend/components/layout/Navigation/Navigation.tsx
CannonLock: Stops us from trying to get the exports again if they fail on a 403 for not being logged in. It is tangential, I can take it out if you would like.

web_ui/frontend/helpers/util.ts
CannonLock: I didn't want to get into the business of sorting arrays of any type. Although wouldn't be hard given I can just call this func recursively.

web_ui/frontend/app/director/components/DirectorCardList.tsx
CannonLock: Hmm, let me peak at this code. I am surprised to see that I cannot predetermine if there is a name.

CannonLock:
@h2zh Thanks for the review, I will update things with your comments and re request!

web_ui/frontend/helpers/util.ts
h2zh: Gotcha. We're not using this func now, and It's fine in most scenarios. Except the differences between two objects only lie in their array fields (this is unlikely to happen, either).

web_ui/frontend/app/director/components/DirectorCardList.tsx
CannonLock: @h2zh Fixed this! It was poor typing on my part, I was passing the keyGetter a partial so everything was possibly unknown but I knew I could provide a non Partial.

h2zh (approved):
LGTM!

PR #2125, opened 2025-03-20
Update to NextJS 15

CannonLock:
- Updated to NextJS and also patched some security issues in the npm packages
- Would prefer to also move to React 19 but have found React 18 to be a peer dependency to swagger items
- Also removed some debug lines


h2zh (approved):
LGTM

PR #2128, opened 2025-03-20
Fix `TestStatMemory` (maybe)

brianaydemir:
Fill the cache before testing for potential memory leaks.

That said, it's worth noting that Go's runtime [does not use a "stop the world" collector](https://tip.golang.org/doc/gc-guide), and thus while [`runtime.GC()` promises to stop its caller in order to perform one complete generation of garbage collection](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/runtime/mgc.go;l=467), it's not clear to me what this actually means when we ask the runtime for the heap's current metrics.


jhiemstrawisc (approved):
LGTM

PR #2132, opened 2025-03-21
Auto add the current user as the security contact in new namespace

CannonLock:
Add the current user as the default security contact when registering a namespace without a security contact. 

For example it will add their Cilogon sub. For now this is a but useless but it is a start.

### DIFF
--- a/web_ui/frontend/app/registry/components/Form.tsx
+++ b/web_ui/frontend/app/registry/components/Form.tsx
@@ -21,6 +21,7 @@ import { CustomRegistrationFieldProps } from './CustomRegistrationField';
 import { alertOnError, getErrorMessage } from '@/helpers/util';
 import { optionsNamespaceRegistrationFields } from '@/helpers/api';
 import { AlertDispatchContext } from '@/components/AlertProvider';
+import { getUser } from '@/helpers/login';
 
 interface FormProps {
   namespace?: RegistryNamespace;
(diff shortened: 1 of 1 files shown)


web_ui/frontend/dev/image/nginx.conf
```diff
@@ -27,8 +27,6 @@ http {
             # Clear the Cookie header
             proxy_set_header Cookie "";
 
-            add_header location prometheus always;
```
h2zh, line 30: I suppose we never use this custom header, and deleting it is just for keeping the response lean and avoid the confusion with the standard “Location” header, right?

web_ui/frontend/app/registry/components/Form.tsx
```diff
@@ -71,6 +72,17 @@ const Form = ({ namespace, onSubmit }: FormProps) => {
     { fallbackData: [] }
   );
 
+  // Auto-fill in the security contact if no security contact
+  const { data: user } = useSWR('getUser', getUser);
+  useEffect(() => {
+    if (
+      user !== undefined &&
+      !namespace?.admin_metadata?.security_contact_user_id
```
h2zh, line 80: On line 54-55, we store the `namespace` prop into a local state `data`, so I think here we should code like `!data?.admin_metadata?.security_contact_user_id`. Or consider rename `data` in this file to avoid confusing with other similar variable names?

web_ui/frontend/app/registry/components/Form.tsx
```diff
@@ -71,6 +72,17 @@ const Form = ({ namespace, onSubmit }: FormProps) => {
...
+    ) {
+      onChange('admin_metadata.security_contact_user_id', user?.user, setData);
+    }
+  }, [user]);
```
h2zh, line 84: It's a good practice to add `data` (or `namespace`) and `setData` to dependencies,

h2zh (changes requested):

Issue #1538, opened 2024-08-02
Server IO load metric is not available on cache servers


haoming29:
Pelican Version:
_Run `./pelican --version` to get version number._

Pelican Service:
- Origin
- Cache

**Describe the bug**
_A clear and concise description of what the bug is._

In https://github.com/PelicanPlatform/pelican/issues/873 and https://github.com/PelicanPlatform/pelican/pull/1221 We used the xrootd throttle plugin to collect server IO metrics and use that in the director to sort the serverAds returned to the client. However, based on recently observation, this metric is not available on cache servers (even if they run Pelican >7.9). I know there's an ongoing improvement to the load metric at the xrootd side (https://github.com/PelicanPlatform/pelican/issues/1372) but currently the load metric is not useful in making the sorting decision.

**To Reproduce**
_Steps to reproduce the behavior._

**Expected behavior**
_A clear and concise description of what you expected to happen._

**Screenshots**
_If applicable, add screenshots to help explain your problem._

Here is the current metric values in the OSDF:
![Screenshot 2024-08-02 at 9 53 30 AM](https://github.com/user-attachments/assets/4b6e7048-7876-4dc5-8741-3acd0629eef8)

**Desktop (please complete the following information):**
 - OS: [e.g. Linux, MacOS]
 - Browser: [e.g. chrome, safari, if the bug relates to any web user interface]

**Additional context**
_Add any other context about the problem here._

PR #2133, opened 2025-03-21
Fixed Throttle Metric integration

patrickbrophy:
This PR addresses issue #1538. Upon analyzing the UDP traffic from the XRootD monitoring stream, I discovered that packets from the throttle plugin were not being collected correctly. These packets are structured similarly to the OSS monitoring packets. To resolve this, I reused and adapted the logic previously implemented for integrating server I/O statistics.

### DIFF
--- a/metrics/xrootd_metrics.go
+++ b/metrics/xrootd_metrics.go
@@ -976,6 +976,14 @@ func handlePacket(packet []byte) error {
 			}
 			return handleOSSPacket(blobs[1:]) // Skip the header
 		}
+
+		if header.Gs.Type == "R" { // Throttle Packet
+			log.Debug("handlePacket: Received a g-stream R packet")
+			if len(blobs) < 2 {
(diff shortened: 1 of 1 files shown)


metrics/xrootd_metrics.go
```diff
@@ -1675,3 +1660,32 @@ func handleOSSPacket(blobs [][]byte) error {
 
 	return nil
 }
+
+func handleThrottlePacket(blobs [][]byte) error {
```
jhiemstrawisc, line 1664: Can you leave a docstring that explains how the function handles the packet and any assumptions that must be valid for it to do its job?

jhiemstrawisc (changes requested):
Not sure how to test this fix myself, so my review was for minor code style stuff. Otherwise it looks good to me.

jhiemstrawisc (approved):
LGTM. Thanks!

jhiemstrawisc:
The failing test here should be unrelated to the diff.

PR #2135, opened 2025-03-21
Add version to the director cards

CannonLock:


jhiemstrawisc:
@turetske, let's patch this into 7.15, it's extremely useful for this information to be obtainable through the Director's UI.

PR #2137, opened 2025-03-21
Align token endpoints with REST API best practices

CannonLock:
Token API Patches:
- Align the endpoints to best practices using the HTTP verbs to indicate the action
- Unify the json that is returned to use the same keys across all models ( expiration vs expiresAt )
- Unify the json so the object posted aligns with the one I get back (scopes is a array on POST and csv on GET)
- Update the json keys to use camelCase
- Patch the Swagger docs to work with params in the path

PR is not to be a pedant, these small things result in a messy UI and I would rather just fix them.

### DIFF
--- a/server_structs/db.go
+++ b/server_structs/db.go
@@ -18,4 +18,14 @@ type (
 		CreatedAt   time.Time `json:"createdAt"`
 		CreatedBy   string    `gorm:"column:created_by;type:text" json:"createdBy"`
 	}
+
+	ApiKeyResponse struct {
+		ID          string    `gorm:"primaryKey;column:id;type:text;not null;unique" json:"id"`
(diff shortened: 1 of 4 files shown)


h2zh (commented):
I like these new endpoints. Concise and informative. Though they seem like Director exclusive endpoints (because these data only stored in Director DB), why are they exposed to all types of Pelican servers?

CannonLock:
@h2zh I believe this API token can be used to query any services prometheus metrics. 

Also I tacked on another small tweak to align the json returns to use camelCase.

h2zh (approved):
The prerequisite PR #2147 was merged in main. This one can also rebase onto main, resolve conflicts and then be merged.

Issue #1662, opened 2024-10-17
Suffixed time values display confusingly in Pelican's printed JSON config


jhiemstrawisc:
For example, caches and origins should have a default `Xrootd.AuthRefreshInterval: 5m`, indicating 5 minutes. This value is ingested correctly by our XRootD configuration mechanism where it's turned into a suffix-less seconds value of `acc.authrefresh 300`.

However, the value printed in the config JSON Pelican dumps at startup is `300000000000` -- the same 300 second time value in nanoseconds. I think we'd be much better off displaying time values in seconds, or in their un-converted string form (e.g. `5m`). Whichever approach we pick, we should apply it to all time values printed in this JSON and make it clear to users seeing these what the unit is.

brianaydemir:
I think the root cause of this issue is that `time.Duration` has no built-in marshalling/unmarshalling support for JSON (https://github.com/golang/go/issues/10275, https://github.com/golang/go/issues/65501).

The XRootD configuration mechanism looks like it has a copy of the relevant sections of the `Config` struct, except with `AuthRefreshInterval` typed as an `int`, and then [uses a decoding hook](https://github.com/PelicanPlatform/pelican/blob/main/xrootd/xrootd_config.go#L685) to convert the `string` duration to an `int`.

One option might be to update `param_generator.go` to generate a parallel to the `Config` struct that uses `string` in place of `time.Duration`.

brianaydemir:
@jhiemstrawisc Does this output need to be in JSON format? The YAML marshaller doesn't have this issue.

jhiemstrawisc:
Not sure why we chose JSON long ago. I think @Saartank touched the logged config dump most recently -- we'd better pick his brain for details fast!

Saartank:
Yes, as @brianaydemir  mentioned, this is because there is no built-in support for `time.Duration` in JSON. We don't necessarily need our config dump at server startup to be in JSON. it could be YAML, which has better human readability.

The only argument one might have for using JSON for the config dump is that if we use structured logging in the future—where each log line is a JSON object—a JSON config dump might integrate better with it.

PR #2142, opened 2025-03-24
Use YAML for the debug-level config dumps

brianaydemir:
The YAML marshaller has support for durations. The JSON marshaller does not.

And as discussed on #1662, there isn't a particularly compelling reason for sticking with JSON.


jhiemstrawisc (approved):
I like it!

PR #2144, opened 2025-03-24
Correctly record the start time for uploads

brianaydemir:
As originally observed by @matyasselmeci, for uploads, the transfer's start time was never initialized in the `TransferResults` object.


jhiemstrawisc (approved):
Hard to believe we're just now noticing this 😆 Thanks for the fix!

PR #2147, opened 2025-03-25
Move token endpoints behind auth

CannonLock:


jhiemstrawisc (approved):
LGTM

CannonLock:
@patrickbrophy I am not familiar enough with the testing arch to test this change, mind picking this up?

h2zh:
I fixed the failed tests. This PR and the following one #2137 could be merged in the codebase. But I think this feature is not ready to go live yet.

There is something I want to confirm with @patrickbrophy next week when you’re back to office. Based on my observation on the code, when a server probes the list or delete api token endpoints, they can see everyone’s api token in the federation and delete any of them. Is that correct? If it is, I think it can be fixed in another ticket.

PR #2148, opened 2025-03-25
Clear server ads in the universal test reset function

### LINKED ISSUE #1952: `server_utils.ResetTestState()` does not clear out director's ad cache
As stated, the helper function `server_utils/server_utils.go::ResetTestState()` doesn't reset the Director's ttlcache of server ads, which is instantiated [here](https://github.com/PelicanPlatform/pelican/blob/afd33af685e277c10e4da788ed3edbd172f0d33e/director/cache_ads.go#L53).

It's not immediately clear to me whether we can just call [`serverAds.DeleteAll()`](https://pkg.go.dev/github.com/jellydator/ttlcache/v3#Cache.DeleteAll) or whether we need to do something more extensive. @bbockelm has done a lot recently with related ttlcaches and may have an opinion

h2zh:
This PR integrates server ads clearing in the universal test reset function: `server_utils.ResetTestState()`, via `config.ResetConfig()`.

It introduces a callback mechanism to avoid the import cycles: Keep the serverAds clearing logic in the `director` package and expose an interface and a callback mechanism that the `config` package can use indirectly. I believe this is going to be a good reference for other import cycles scenarios in the Pelican development.

Closes #1952

### DIFF
--- a/director/director_test.go
+++ b/director/director_test.go
@@ -1729,9 +1729,13 @@ func TestRedirectMiddleware(t *testing.T) {
 
 }
 func TestRedirects(t *testing.T) {
+	server_utils.ResetTestState()
 	ctx, cancel, egrp := test_utils.TestContext(context.Background(), t)
-	defer func() { require.NoError(t, egrp.Wait()) }()
-	defer cancel()
+	t.Cleanup(func() {
+		cancel()
+		assert.NoError(t, egrp.Wait())
+		server_utils.ResetTestState()
+	})
 
 	// Use ads generated via mock topology for generating list of caches
 	topoServer := httptest.NewServer(http.HandlerFunc(mockTopoJSONHandler))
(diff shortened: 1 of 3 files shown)


director/director_test.go
```diff
@@ -1729,19 +1729,20 @@ func TestRedirectMiddleware(t *testing.T) {
 
 }
 func TestRedirects(t *testing.T) {
+	server_utils.ResetTestState()
 	ctx, cancel, egrp := test_utils.TestContext(context.Background(), t)
-	defer func() { require.NoError(t, egrp.Wait()) }()
```
jhiemstrawisc, line 1733: It actually doesn't look like the `cancel` or `egrp` get used anywhere in this test, other than to be cleaned up at the end. Can you change this to:
```
ctx, _, _ := test_utils.TestContext(context.Background(), t)
```
to clean up some of the test logic?

director/director_test.go
```diff
@@ -1760,11 +1761,6 @@ func TestRedirects(t *testing.T) {
 	})
 
 	t.Run("redirect-link-header-length", func(t *testing.T) {
-		server_utils.ResetTestState()
-		t.Cleanup(func() {
-			server_utils.ResetTestState()
-		})
-
```
jhiemstrawisc, line 1767: Nice that we can get rid of the inter-test cleanups here and in subsequent tests, but it also means the `viper.Set("Director.CacheSortMethod", "random")` that happens in each of these must not actually be doing anything. Can you remove them from the individual `t.Run` statements and move the config to occur once somewhere before the call to `AdvertiseOSDF()` at line 1744?

jhiemstrawisc (changes requested):
Really cool approach to getting around the cyclic import, I'd never thought of this before! A few minor change requests, but otherwise this looks good to me.

PR #2151, opened 2025-03-26
Token UI

CannonLock:
Merge https://github.com/PelicanPlatform/pelican/pull/2137 first, I built this on top of the revised api endpoints.

- Add settings page for general settings
  - Plan to put config here later as well but out of scope for now
- Add api token listing and creation form
- add general page that has the server restart button
- Mark some client components as so I could SSR higher up in the tree

### DIFF
--- a/web_ui/frontend/app/settings/api/components/TokenForm.tsx
+++ b/web_ui/frontend/app/settings/api/components/TokenForm.tsx
@@ -1,0 +1,241 @@
+'use client';
+
+import {
+  useState,
+  Dispatch,
+  useContext,
+  useCallback,
+  SetStateAction,
+} from 'react';
+import {
+  Autocomplete,
+  Box,
+  Button,
+  IconButton,
+  Modal,
+  Paper,
+  TextField,
+  Typography,
+} from '@mui/material';
+import {
+  AlertDispatchContext,
+  AlertReducerAction,
+} from '@/components/AlertProvider';
+import { secureFetch } from '@/helpers/login';
+import { fetchApi } from '@/helpers/api';
+import { alertOnError } from '@/helpers/util';
+import Link from 'next/link';
+import { BaseToken } from '@/types';
+import { Check, ContentCopy, Close } from '@mui/icons-material';
+
+const TokenForm = () => {
+  const dispatch = useContext(AlertDispatchContext);
+
+  const [name, setName] = useState('');
+  const [createdBy, setCreatedBy] = useState('');
+  const [expiration, setExpiration] = useState('');
+  const [scopes, setScopes] = useState<string[]>([]);
+  const [token, setToken] = useState<string | undefined>(undefined);
+
+  const onSubmit = useCallback(async () => {
+    const tokenRequest = {
+      name,
+      createdBy,
+      expiration: new Date(expiration).toISOString(),
+      scopes,
+    };
+
+    const token = await handleSubmit(tokenRequest, dispatch);
+
+    // Remove current fields
+    setName('');
+    setCreatedBy('');
+    setExpiration('');
+    setScopes([]);
+
+    return token;
+  }, [name, createdBy, expiration, scopes]);
+
+  return (
+    <Box sx={{ mt: 3 }}>
+      <TextField
+        label='Token Name'
+        size={'small'}
+        fullWidth
+        required
+        margin={'dense'}
+        name={'name'}
+        onChange={(e) => setName(e.target.value)}
+        value={name}
+      />
+      <TextField
+        label='Creators Name'
+        size={'small'}
+        fullWidth
+        required
+        margin={'dense'}
+        name={'createdBy'}
+        onChange={(e) => setCreatedBy(e.target.value)}
+        value={createdBy}
+      />
+      <TextField
+        label='Expiration Date'
+        type='date'
+        size={'small'}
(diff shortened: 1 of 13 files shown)


h2zh (changes requested):
Could you rebase this PR to the latest main? Seems like a large amount of code here were merged already.

CannonLock:
@h2zh Rebased!

CannonLock:
@h2zh This PR is ready for review. There is a existing issue where I assume these endpoints are for all services but this is built on a system where they are only available on the Director. 

That said @patrickbrophy communicated that he is working to get things set up so that they operate on each service like I had previously thought. With that in mind we ought to just merge this so we can use it for the time being.

h2zh:
Gotcha @CannonLock. I will review this PR today or tomorrow after reviewing @patrickbrophy's #2179

web_ui/frontend/app/settings/api/components/TokenForm.tsx
```diff
@@ -0,0 +1,241 @@
...
+      />
+      <TextField
+        label='Creators Name'
+        size={'small'}
+        fullWidth
+        required
+        margin={'dense'}
+        name={'createdBy'}
+        onChange={(e) => setCreatedBy(e.target.value)}
+        value={createdBy}
```
h2zh, line 79: This should be removed since `createdBy` is now automatically filled by backend

web_ui/frontend/app/settings/api/components/TokenForm.tsx
```diff
@@ -0,0 +1,241 @@
...
+      >
+        <Box display={'flex'}>
+          <Typography variant={'subtitle1'}>
+            Copy your API Token, you cannot view it again.
+          </Typography>
+          <IconButton
+            sx={{ ml: 'auto' }}
+            size={'small'}
+            onClick={() => setToken(undefined)}
+          >
```
h2zh, line 183: Perhaps add the text "Copy" next to this icon, in case someone doesn't recognize it. Also I think the general paradigm is placing the "copy" icon on the right side of the target text.

web_ui/frontend/app/settings/api/components/TokenForm.tsx
```diff
@@ -0,0 +1,241 @@
...
+          py: 2,
+          px: 3,
+          borderRadius: 1,
+          bgcolor: 'white',
+          maxWidth: '100vw',
+        }}
+      >
+        <Box display={'flex'}>
+          <Typography variant={'subtitle1'}>
+            Copy your API Token, you cannot view it again.
```
h2zh, line 177: It would be good to explicitly ask the user paste it in a safe place after the copy. Sometimes the copy doesn't work and the user will lose this token (though it's not a big deal, they can just create another one)

web_ui/frontend/app/settings/api/components/TokenList.tsx
```diff
@@ -0,0 +1,138 @@
...
+          mx={1}
+          my={0.5}
+        >
+          <Box>
+            <Box display={'flex'}>
+              <Typography>{token.name}</Typography>
+            </Box>
+            <Box display={'flex'}>
+              <Typography variant={'subtitle2'}>
+                Expires on {expirationDate.toLocaleDateString()}
```
h2zh, line 80: User should be curious about the exact time (HH:MM) and local time/UTC

web_ui/frontend/app/settings/api/components/TokenList.tsx
```diff
@@ -0,0 +1,138 @@
...
+              </Typography>
+              <Box mx={1}>-</Box>
+              <Typography variant={'subtitle2'}>
+                Created by {token.createdBy}
```
h2zh, line 84: This may need to be removed. Because the createdBy is now fetched by backend. Its value will be a CILogon user id, which is meaningless to user.

web_ui/frontend/app/settings/api/components/TokenList.tsx
```diff
@@ -0,0 +1,138 @@
...
+              </Typography>
+            </Box>
+          </Box>
+          <Box mt={0.5}>
+            {token.scopes.map((x) => {
+              return (
+                <Tooltip key={x} title={'Token Scope'}>
+                  <Chip size={'small'} label={x} />
+                </Tooltip>
+              );
```
h2zh, line 94: It's crowded here. Could you increase the scope chip interval on x-axis? Maybe also keep them a little bit further from the bottom of this Token Box.

h2zh (changes requested):
The features of this PR needs to be tested again after #2179 merged in.

web_ui/frontend/app/settings/api/components/TokenList.tsx
CannonLock: I would keep it. You can map them, it just isn't easy.

web_ui/frontend/app/settings/api/components/TokenList.tsx
CannonLock: I will add in the time. I don't think it is a good idea to add the timezone because it is a pretty common pattern that you show local time. I don't think anyone would see this time and image it is in UTC time.

h2zh (approved):
LGTM 🚀

PR #2154, opened 2025-03-27
Add meaningful page titles

CannonLock:
- Page specific titles, so rather then only "Pelican Platform" you will see "Register Cache | Origin" for example
- Create MetaLayout for an unobtrusive layout for metadata inclusion

### DIFF
--- a/web_ui/frontend/app/(landing)/layout.tsx
+++ b/web_ui/frontend/app/(landing)/layout.tsx
@@ -21,8 +21,7 @@ import { Box } from '@mui/material';
 import { Header } from '@/components/layout/Header';
 
 export const metadata = {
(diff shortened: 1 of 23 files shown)


h2zh (commented):
This is a clever and pretty design! By leveraging the built-in Metadata `template` in NextJS, this PR renovates the webpage titles everywhere. I wonder why the title needs to reside in a separate file? In the [NextJS tutorial](https://nextjs.org/learn/dashboard-app/adding-metadata), it seems to be able to fit in the same file as the page content.

Btw, this PR also needs to rebase the latest main

h2zh (approved):
LGTM - the failed test is the lingering TestStatMemory

PR #2157, opened 2025-03-28
Enhance drop privs inter-processes test

### LINKED ISSUE #2099: Race condition in `drop_privilege_test.go`
The `xrootd/drop_privilege_test.go` tests have a race condition.  I think what's meant to happen is:

1. Main thread sends a command
2. Child thread receives command
3. Main thread closes socket pair

However, there's no synchronization between steps (2) and (3).  The main thread is free to exit and fire its `defer` function before the goroutine in (2) tries to read from its socket.  I observed this race condition at least once in the GitHub Actions although it seems to not be particularly prevalent currently.

In general, it doesn't look like this test really tests all that much -- it kinda looks half-written to me:
- It prepares a new CA bundle but doesn't seem to check anything about it.
- It has code paths for both cache and origin testing but only the origin code paths are reachable.
- It has a function named `receiveFD` which suggests it is receiving a file descriptor (the point of one of the drop privilege commands).  However, there's no receiving of file descriptors in the code!

h2zh:
[Primary] Resolve occasional race condition in the drop privs test. 
Reason: Main test thread could clean up resources before the mock XRootD goroutine finished processing commands
Solution: Adding synchronization with `sync.WaitGroup` to ensure the mock XRootD process goroutine completes its work before the main test thread proceeds with cleanup

[Second] Verify the CA file has been transferred to the expected location and contain correct content

[Third] Add Cache test, in addition to the existing Origin test

Closes #2099

### DIFF
--- a/xrootd/drop_privilege_test.go
+++ b/xrootd/drop_privilege_test.go
@@ -22,7 +22,6 @@ package xrootd
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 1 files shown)


h2zh:
@jhiemstrawisc This is the fix PR I mentioned in the meeting. Should I give it a 7.16 or 7.17 tag?

jhiemstrawisc:
Let's leave it as 7.17 -- fixing a flaky unit test isn't a high priority for 7.16 at this point.

patrickbrophy (approved):
Nice job resolving the race condition. The test of the test code look clean too.

PR #2158, opened 2025-03-28
Fix director backwards compatibility

turetske:
Fixes the backwards compatibility issue with the directors in 7.14

The problem is that while the federation director endpoint was being added to the directors that were being checked to advertisementUrls, it then was never added to the final list of director endpoints because it would fail at the director listing in doDiscovery.

This fix checks for origins/directors that if there is no director advertisement endpoints to advertise to from the federation, then advertise to the origin federation director endpoint found during federation discovery.

### DIFF
--- a/server_utils/director_discovery.go
+++ b/server_utils/director_discovery.go
@@ -44,7 +44,7 @@ var (
 )
 
 // Query all known directors & metadata, return a list of unique director ads
-// If all no director ads are found, default back to the federation director endpoint
+// If no director ads are found, default back to the federation director endpoint
 // If that endpoint doesn't exist, then all the errors encountered during the director
 // ad discovery will be returned
 func doDiscovery(ctx context.Context, isDirector bool) (endpoints []server_structs.DirectorAd, err error) {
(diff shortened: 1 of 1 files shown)


server_utils/director_discovery.go
```diff
@@ -158,7 +178,7 @@ func LaunchPeriodicDirectorDiscovery(ctx context.Context, isDirector bool) error
 	egrp := ctx.Value(config.EgrpKey).(*errgroup.Group)
 	servers, err := doDiscovery(ctx, isDirector)
 	if err != nil {
-		log.Warningln("Failed to discover available director endpoints:", err)
+		log.Warningln("Failure to discover available director endpoints:", err)
```
jhiemstrawisc, line 181: I think we should stick with "Failed" here -- it matches the phrasing in log/error statements from lines 62, 114, 129, 146 and 209.

server_utils/director_discovery.go
```diff
@@ -43,6 +44,9 @@ var (
 )
 
 // Query all known directors & metadata, return a list of unique director ads
+// If all no director ads are found, default back to the federation director endpoint
```
jhiemstrawisc, line 47: I think "if all no director ads" might be a typo.

server_utils/director_discovery.go
```diff
@@ -133,14 +140,27 @@ func doDiscovery(ctx context.Context, isDirector bool) (endpoints []server_struc
...
-	endpoints = make([]server_structs.DirectorAd, 0, len(endpointMap))
-	for _, ad := range endpointMap {
-		endpoints = append(endpoints, ad)
+	// No endpoints were found and the federation director endpoint is nil
+	if len(endpointMap) == 0 && fed.DirectorEndpoint == "" {
+		newError := errors.New("failed to find director endpoint")
+		err = errors.Join(newError, allErrors)
+	} else if len(endpointMap) == 0 { // Fall back to director endpoint
+		endpoints = make([]server_structs.DirectorAd, 0, 1)
+		if fed, err := config.GetFederation(ctx); err == nil {
```
jhiemstrawisc, line 150: You're already using `fed.DirectorEndpoint` in your conditional to enter this section, so you shouldn't need to re-invoke `config.GetFederation`. I think this also helps with some of the error handling -- it doesn't look like the error in this line gets handled or joined with the list of running errors up to this point.

jhiemstrawisc (changes requested):

jhiemstrawisc:
Other than the requested changes, I confirmed this approach solved the backwards compatibility issue by running a 7.12 Director/Registry and an Origin compiled with this diff.

server_utils/director_discovery.go
turetske: Yep, that was a copy paste issue.

jhiemstrawisc (approved):
LGTM. Thanks!

PR #2162, opened 2025-03-31
Update docs/origin startup to prevent '/view' registration

jhiemstrawisc:
This documents and prevents registration of the `/view` namespace. This prefix is disallowed because we serve web UI content under `/view`.

I also noticed we weren't treating `/pelican` registration as an error at Origin startup. We do technically register this as a namespace for federation services, but we disallow user registration of the namespace. This new check won't prevent Pelican from registering `/pelican`, but it will prevent the user from trying to use the namespace in their Origin configuration.


patrickbrophy (approved):
Look good! LGTM

PR #2167, opened 2025-04-01
Add more events to response for long history issues

CannonLock:
Increase event page size to 100 for issues with more then 30 events. This isn't a perfect update but should be good enough.


jhiemstrawisc (approved):
👍

PR #2173, opened 2025-04-01
Exclude namespace itself from its super/sub spaces & allow Registry admin to overwrite the public key of a namespace

h2zh:
It excludes the namespace itself from its super/sub spaces.
With this update, Registry admin can overwrite any namespace's public key.

Q: After this change, when a namespace registers, how do we check if there is a same namespace already registered?
A: We have a dedicated `namespaceExistsByPrefix` func to do that when a namespace registers. Note that this func won't be executed when Registry admin updates the namespace. So, it's safe to exclude the namespace itself from its super/sub spaces.

Next, @CannonLock will add a warning on the UI when Registry admin wants to edit the public key of a namespace that has super/sub spaces.

### DIFF
--- a/registry/registry_db.go
+++ b/registry/registry_db.go
@@ -133,7 +133,7 @@ func namespaceSupSubChecks(prefix string) (superspaces []string, subspaces []str
 	// Check if any registered namespaces already superspace the incoming namespace,
 	// eg if /foo is already registered, this will be true for an incoming /foo/bar because
 	// /foo is logically above /foo/bar (according to my logic, anyway)
-	superspaceQuery := `SELECT prefix FROM namespace WHERE (? || '/') LIKE (prefix || '/%')`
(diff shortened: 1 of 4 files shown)


jhiemstrawisc (approved):
I think this is okay.. The only thing that makes me a bit nervous is whether anything bad could happen by letting the admin override this check. Does anything break if public keys for `/foo` and `/foo/bar` are suddenly completely decoupled? In that case, how can we make sure the owner of `/foo` remains the logical owner of `/foo/bar` in the Registry? This is probably something for us to consider in a follow-up issue.

h2zh:
Here is what's on the top of my head (not verified yet): For the /foo and /foo/bar example above, let's say they originally have same public key and then federation (registry) admin changes the registered public of /foo/bar. If it is an addition, it doesn't need the owner of /foo do anything to keep its ownership of /foo/bar. If it is a replacement, the owner of /foo has to add the new private key to its `IssuerKeydDirectory`.

PR #2177, opened 2025-04-04
Sanity check received size and on-disk size for client GETs

jhiemstrawisc:
When the client requests an object, grab determines the expected download size by inspecting the `Content-Length` header. Previously we didn't perform any checks against this value to make sure the number of bytes we receive actually match the number of expected bytes -- we only used status codes to determine success.

With this diff, we sanity check the number of bytes witnessed by the grab client, and when the client is writing a regular file and not a directory/unpacked object, we also stat the file on disk as a second check.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -1906,6 +1906,11 @@ func parseTransferStatus(status string) (int, string) {
 // Verify that a file on disk matches the expected size. We ignore directories
 // and generic stat failures unless the file doesn't exist.
 func verifyFileSize(dest string, expectedSize int64, fields log.Fields) error {
+	if dest == "/dev/null" {
+		log.WithFields(fields).Debugf("Skipping size check because destination /dev/null")
+		return nil
+	}
+
 	fi, err := os.Stat(dest)
 	if err != nil { // Don't treat stat failure as fatal unless it indicates the file doesn't exist
 		if os.IsNotExist(err) {
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -1903,6 +1903,32 @@ func parseTransferStatus(status string) (int, string) {
 	return statusCode, strings.TrimSpace(parts[1])
 }
 
+// Verify that a file on disk matches the expected size. We ignore directories
+// and generic stat failures unless the file doesn't exist.
+func verifyFileSize(dest string, expectedSize int64, fields log.Fields) error {
+	fi, err := os.Stat(dest)
+	if err != nil { // Don't treat stat failure as fatal unless it indicates the file doesn't exist
+		if os.IsNotExist(err) {
+			return errors.Errorf("file does not exist: %s", dest)
+		}
+
+		log.WithFields(fields).Debugf("Error checking size of %s on disk: %v", dest, err)
+		return nil
+	}
+
+	if fi.IsDir() {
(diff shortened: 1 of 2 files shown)


jhiemstrawisc:
Looks like the PR Validation action is failing because GH changed the response to one of the API queries...?

client/handle_http.go
```diff
@@ -2232,6 +2232,16 @@ Loop:
...
+	// the transfer was supposed to be in the first place.
+	if totalSize > 0 && downloaded != totalSize {
+		log.WithFields(fields).Debugf("Download completed but received size '%db' does not match expected size '%db'", downloaded, totalSize)
+		err = errors.Errorf("download completed but received size '%db' does not match expected size '%db'", downloaded, totalSize)
+		return
+	}
+
 	if unpacker != nil {
 		unpacker.Close()
 		if err = unpacker.Error(); err != nil {
```
brianaydemir, line 2247: This isn't directly related to this PR, but what happens if `unpacker.Close()` is never called? There are more than a few ways this function can exit early, including via the check you just added.

client/handle_http.go
```diff
@@ -2232,6 +2232,16 @@ Loop:
 			resp.HTTPResponse.StatusCode, resp.Err().Error())}
 	}
 
+	// By now, we think the download succeeded. If we know how large the file was supposed to
+	// be based on the Content-Length header, we can check that the grab client witnessed
+	// the correct number of bytes. If totalSize is <= 0, it indicates we don't know how large
+	// the transfer was supposed to be in the first place.
+	if totalSize > 0 && downloaded != totalSize {
+		log.WithFields(fields).Debugf("Download completed but received size '%db' does not match expected size '%db'", downloaded, totalSize)
+		err = errors.Errorf("download completed but received size '%db' does not match expected size '%db'", downloaded, totalSize)
```
brianaydemir, line 2241: Why not `%d` instead of `'%db'`? Less visual noise, and I think most people will assume the size is being reported in bytes, absent a unit.

brianaydemir:
> I'd originally planned to toss in an extra stat of the bytes flushed to disk, but that won't work correctly in cases where Pelican is creating a local directory or unpacking something.

@jhiemstrawisc Is it terribly difficult to `stat` the number bytes written to storage only in the case of a single object? Although, assuming the end-to-end checksumming operates on the object on disk, maybe that's not worth worrying about here — in which case, maybe update the PR's title.

jhiemstrawisc:
> @jhiemstrawisc Is it terribly difficult to `stat` the number bytes written to storage only in the case of a single object? Although, assuming the end-to-end checksumming operates on the object on disk, maybe that's not worth worrying about here — in which case, maybe update the PR's title.

I re-added the stat check, but only for files (and not directories). Hopefully this is a satisfactory middle ground.

brianaydemir (approved):
LGTM.

But one existing and related issue that I observed:

https://github.com/PelicanPlatform/pelican/blob/016f503769593c52db1aba95da8ad87468dbf9f1/client/handle_http.go#L1878-L1882

logs the number of bytes downloaded across _all_ attempts, but the wording is intuitively read as the number of bytes in the object. (Why would anyone care to include the number of bytes from failed attempts?)

jhiemstrawisc:
@turetske, can you create a 7.15 patch out of this and let me know when you're done and all the non-github CI tests have passed?

Issue #2153, opened 2025-03-27
APIToken Enhancements


CannonLock:
List of a couple tweaks to make to the ApiTokens. 

- The db shouldn't be in the director only, if all servers have these endpoints it should be in a general db.
- `createdBy` shouldn't be determined by the user, it should be set via the user or the sub of the user that is logged in

PR #2179, opened 2025-04-07
API Token Enhancements

patrickbrophy:
This addresses issue #2153. This PR adds the API token feature to each Pelican service. It achieves this through the following:
- Create new config parameter `Server.DbLocation`. This is the new SQLite database (`pelican.sqlite`) that is shared between all Pelican services (not shared in that they share data, but shared in that each service will have this database).
- Move API token migrations from the Director to a server wide migrations folder.
- In each launcher, migrate the migrations to the  server database.

This PR also gets rid of the ability for the user to set the `createdBy` field when creating a token. This is achieved by looking at the login token and getting the `sub` field from the token and using this as the API token's creator. 

Lastly, this PR adjusts our tests to reflect the changes mentioned above.


h2zh (changes requested):
This new database hierarchy looks sleek! I hope to get this merged asap, then I will work on moving the Director's downtime table to Registry based on this, to make the Multi-Director feature possible in 7.16

h2zh (approved):
LGTM

Issue #1808, opened 2024-12-06
Truncated error message in cache's self-check feature


jhiemstrawisc:
We've observed unintended error message truncation in [this function](https://github.com/PelicanPlatform/pelican/blob/c345e676baf98a8070fd970a983e28473c2815a3/cache/self_monitor.go#L250-L268). The error message printed out was`self-test failed during download. File is cleaned up at ` without the file URL.

Let's make sure this truncation can't happen by making sure `fileUrl` can't be the empty. This probably also indicates there was an uncaught error in `generateTestFile()`.

On that note -- change all the references from "files" to "objects". Eventually we'll be able to back caches with non-posix object stores.

PR #2180, opened 2025-04-07
Fix cache self-test error message & prevent file name truncation

jhiemstrawisc:
As #1808 states, we'd observed something strange in the relevant functions where a) the file URL generation function could return an empty URL without an associated error and b) that URL file could be successfully cleaned up, but logged in a way where the URL is empty. After closer inspection of these functions, my best bet is that we're running into a case where `url.Parse()` passes on an empty/invalid string and does not produce an error. As [docs for `url.Parse()`](https://pkg.go.dev/net/url#Parse) indicate, the function may not always behave as expected due to "parsing ambiguities".

In addition to tidying up a few other error messages, I believe this makes it much harder for the conditions that caused this errant behavior to exist.

### DIFF
--- a/cache/self_monitor.go
+++ b/cache/self_monitor.go
@@ -164,6 +164,10 @@ func generateTestFile() (string, error) {
 	}
 	baseUrl.Path = extFilePath
 
+	if baseUrl.String() == "" {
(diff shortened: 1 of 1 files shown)


patrickbrophy (approved):
Looks good. LGTM

PR #2182, opened 2025-04-07
Fix log level/location issues coming from web UI config

jhiemstrawisc:
There were two bugs here:
1) The web config override function assigned all the new viper overrides but it didn't use them to set the log level. Because the logging library doesn't look at viper, it was potentially unaware it needed to update its levels.
2) Logs were flushed (and consequently their location assigned) before the web config yaml was loaded/merged.

The reviewer can test this diff by starting any Pelican server and modifying log levels and log locations via the web UI. Previously, doing this had no effect but now the values should be used correctly.

### DIFF
--- a/cmd/main_test.go
+++ b/cmd/main_test.go
@@ -26,9 +26,8 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"time"
-
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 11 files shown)


jhiemstrawisc:
Not sure what's happening here, but it doesn't look like the GHA tests ran after the last commit, and I don't see an option to run them... Any ideas?

jhiemstrawisc:
@h2zh, I think this is finally ready for review -- sorry for all the commit noise while I was trying to figure out the strange errors!

h2zh:
@jhiemstrawisc Got it! I'll do it this afternoon.

h2zh (approved):
LGTM

Issue #2183, opened 2025-04-08
Multiuser origins do not cache public keys for validating tokens


brianaydemir:
Pelican is responsible for configuring XRootD and the underlying [scitokens-cpp library](https://github.com/scitokens/scitokens-cpp) so that they are able to cache any public keys encountered during the process of validating tokens. This does not work when `Origin.Multiuser` is set to true.

The configuration is supposed to happen via the `XDG_CACHE_HOME` environment variable:

https://github.com/PelicanPlatform/pelican/blob/04e1adc88d97e232504e840067f62ba967300a3e/xrootd/xrootd_config.go#L437-L452

However, when launching a multiuser origin

https://github.com/PelicanPlatform/pelican/blob/04e1adc88d97e232504e840067f62ba967300a3e/launchers/origin_serve.go#L115-L116

instead of allowing it to inherit its environment from the invoking process, we build the environment from scratch:

https://github.com/PelicanPlatform/pelican/blob/04e1adc88d97e232504e840067f62ba967300a3e/xrootd/launch_linux.go#L94-L99

PR #2184, opened 2025-04-08
Enable multiuser origins to cache public keys for validating tokens

brianaydemir:
The rationale for this fix is laid out in #2183.

When testing this PR, I checked for the presence of `scitokens_cpp.sqllite` in the expected locations, both for origins and caches, with `Origin.Multiuser` set to the default (`false`) and `true`. (When the bug is present, this file is never created.) I deleted `/run/pelican` before each test.

### DIFF
--- a/xrootd/launch_linux.go
+++ b/xrootd/launch_linux.go
@@ -96,8 +96,8 @@ func (plauncher PrivilegedXrootdLauncher) Launch(ctx context.Context) (context.C
 		env = []string{"XRDHTTP_PELICAN_INFO_FD=3"}
 	}
 	// Set the directory where the SciTokens library should cache public keys.
-	if xdg_cache_home, ok := os.LookupEnv("XDG_CACHE_HOME"); ok {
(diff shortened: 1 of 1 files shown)


xrootd/launch_linux.go
```diff
@@ -95,6 +95,10 @@ func (plauncher PrivilegedXrootdLauncher) Launch(ctx context.Context) (context.C
 	if plauncher.fds[1] != -1 {
 		env = []string{"XRDHTTP_PELICAN_INFO_FD=3"}
 	}
+	// Set the directory where the SciTokens library should cache public keys.
+	if xdg_cache_home, ok := os.LookupEnv("XDG_CACHE_HOME"); ok {
```
jhiemstrawisc, line 99: Can you switch `xdg_cache_home` to `xdgCacheHome` to stay consistent with our use of camel case?

brianaydemir:
@turetske This PR needs to ported back to 7.15.

PR #2185, opened 2025-04-08
Update 'componenets' for Monitoring.XXX configs

jhiemstrawisc:
The reviewer should double check that I got these correct -- I tried to determine where each one of these was used, but may have misunderstood some of it.


patrickbrophy (approved):
LGTM

PR #2186, opened 2025-04-08
Make Origin throttling configurable

jhiemstrawisc:
This lets an Origin admin set:
```
Origin:
  Concurrency: <int>
```
to limit the number of accepted concurrent requests to the Origin. You can test the diff by adding the config and observing the line `throttle.throttle concurrency <int>` in the generated xrootd config.

I also got annoyed enough by all the errant whitespace in the generated xrootd config that I tidied some of it up.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1196,6 +1196,15 @@ func SetServerDefaults(v *viper.Viper) error {
 		v.Set(param.Server_ExternalWebUrl.GetName(), parsedExtAdd.String())
 	}
 
+	if originConcurrency := v.GetInt(param.Origin_Concurrency.GetName()); originConcurrency < 0 {
+		return errors.Errorf("invalid value of '%d' for config param %s; must be greater than or equal to 0, where 0 disables the feature",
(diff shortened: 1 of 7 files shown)


h2zh:
It would be nice to also sanitize the input number to avoid the following situation. @jhiemstrawisc 

In `pelican.yaml`:
```
##########################
#   Origin Directives    #
##########################
Origin:
  Concurrency: 015
```

In the printed config:
```
Origin:
    Concurrency: 13
...
throttle.throttle concurrency 13
```

jhiemstrawisc:
Huh, I was surprised by that, but after a bit of googling, it looks like the conversion of `015` to `13` is intentional and comes from the YAML specification itself. Apparently yaml treats numbers with a leading `0` as octal. I don't think we should change the way we handle it because it's part of the YAML specification.

jhiemstrawisc:
@turetske, here's another one to patch into 7.15. Can you let me know when it's an official release so I can pass it along to our friends at NRAO?

PR #2189, opened 2025-04-08
Disable topology flags

turetske:
Added flags to disable getting downtime from Topology, getting X509 authentication from Topology (caches and origins), getting caches themselves, and getting origins.

These are meant to be modular, so rather than gutting AdvertiseOSDF or playing with the "isTopology" settings in the ads, this just selectively disables what information it creates/retrieves from Topology as needed.

I recommend testing by using the osdf binary and selectively applying various versions of the TopologyFlags and seeing that the various features are turned off.

### DIFF
--- a/config/resources/osdf.yaml
+++ b/config/resources/osdf.yaml
@@ -27,3 +27,5 @@ Federation:
 Registry:
   RequireCacheApproval: true
   RequireOriginApproval: true
+Topology:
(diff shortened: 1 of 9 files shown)


patrickbrophy (approved):
Looks good. This will be a nice feature to hopefully phase out topology in the future.

PR #2196, opened 2025-04-09
Fix metrics scraping by accepting fed tokens for Director service discovery

jhiemstrawisc:
The bug was that we switched the Director to issue service discovery tokens using the Federation root as an issuer, but neglected to update the auth handler in the service discovery endpoint to accept federation tokens.

This adds the extra auth handler and gives the test a face lift that I think would have caught the bug the first go around had they existed.

I also added a mock federation root server, which I've wound up re-implementing too many times. Some other time I'll go back through some old code and switch it to use this standard test utility.

### DIFF
--- a/test_utils/utils.go
+++ b/test_utils/utils.go
@@ -285,15 +285,17 @@ func MockFederationRoot(t *testing.T, fInfo *pelican_url.FederationDiscovery, kS
 			w.Header().Set("Content-Type", "application/json")
 			w.WriteHeader(http.StatusOK)
 
-			_, err := w.Write([]byte(fmt.Sprintf(`{
-				"director_endpoint": "%s",
-				"director_advertise_endpoints": %s,
-				"namespace_registration_endpoint": "%s",
-				"broker_endpoint": "%s",
-				"jwks_uri": "%s"
-			}`, getInternalFInfo().DirectorEndpoint, getInternalFInfo().DirectorAdvertiseEndpoints, getInternalFInfo().RegistryEndpoint,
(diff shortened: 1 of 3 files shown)


test_utils/utils.go
```diff
@@ -246,3 +250,100 @@ func GetUniqueAvailablePorts(count int) ([]int, error) {
...
+			w.WriteHeader(http.StatusOK)
+
+			_, err := w.Write([]byte(fmt.Sprintf(`{
+				"director_endpoint": "%s",
+				"director_advertise_endpoints": %s,
+				"namespace_registration_endpoint": "%s",
+				"broker_endpoint": "%s",
+				"jwks_uri": "%s"
+			}`, getInternalFInfo().DirectorEndpoint, getInternalFInfo().DirectorAdvertiseEndpoints, getInternalFInfo().RegistryEndpoint,
+				getInternalFInfo().BrokerEndpoint, getInternalFInfo().JwksUri)))
```
h2zh, line 295: This is all good, though I prefer to construct a Go struct or map and then call json.Marshal. It would help catch typos or formatting issues at compile time and handle proper JSON escaping automatically.

h2zh (changes requested):
Very trivial things..

h2zh (approved):
🚀

jhiemstrawisc:
Failing tests look to be unrelated

jhiemstrawisc:
@turetske, this should be patched into 7.15. Can you let me know when you start the process?

Issue #2073, opened 2025-03-04
`director_ui/servers` endpoint produces an incomplete server ad


jhiemstrawisc:
I recalled we added version info to server ads in a previous release and tried to use that to grab the version of a cache from the `/api/v1.0/director_ui/servers` API, which should return the Director's list of known server ads. However, it looks like the version info didn't get plumbed into that because the API function to regenerate the server ad does so manually. See https://github.com/PelicanPlatform/pelican/blob/b539ed596086bb22e94a8a9811e6220db8362e4c/director/director_ui.go#L232

The fix for this should get as close as possible to automatically generating the server ad rather than manually enumerating its fields. Otherwise, any time we make an addition to server ads, we'd also need to remember to update this function.

h2zh:
I think this issue collides with #1824 and was done in another PR #2083. Could you take a look at that PR and close this issue if you think what's done there is sufficient.

jhiemstrawisc:
My issue with that PR is it still manually syncs the two structs and requires we always remember to do that any time we update one of them. I'd like to keep this issue open for tracking the process of guaranteeing the two update in lockstep without touching the code in both places.

PR #2203, opened 2025-04-11
Add Warning Comment for Misaligned ServerAd and serverResponse structs

patrickbrophy:
This PR addresses issue #2073. After a series of trials and conversations with @jhiemstrawisc. We concluded that it would be difficult to align these structures together without breaking compatibility. This PR adds a warning message comment so that hopefully to prevent misalignments in the future.


jhiemstrawisc (approved):
LGTM

Issue #2199, opened 2025-04-10
Refactor the downtime management logics to embrace Multi Directors


h2zh:
With the introduction of multiple directors, we are no longer able to store the downtimes in the sqlite database residing in the Director, because there is no way to sync the data in the database among multiple directors.

Now, the only data in the Director db is the downtimes set by federation admin. The plan is moving this downtimes table to the Registry database.

In this new world, the Director only display the downtimes, which means you can't toggle downtime there. Instead, federation admin needs to schedule downtime for an origin or cache server in the Registry database, then Director will get them via a recurring API call Go routine and keep them in an in-memory variable `fedDowntimes` with the data structure [serverName: a list of Downtime]. Origin or cache server admin will set their scheduled downtimes in their own database, then propagate to the Director via server ads. Besides, there will be another variable `topoDowntimes` to save the downtimes set by topology.

![Image](https://github.com/user-attachments/assets/8522b09d-deb6-4e75-bf6b-e159e003ea0d)
(Highlights in yellow represent the TODO in this ticket. I'll replace this draft diagram later) 

This ticket will build upon #2153. All Pelican services will share the same Downtime table schema, while they have different duties:
- Registry: store the downtimes set by Federation admin for all Origin/Cache servers
- Origin: store its own downtimes set by Origin server admin
- Cache: store its own downtimes set by Cache server admin
- Director: It just displays the downtimes fetched from the above databases. However, in this ticket, it also gets a Downtime table in its database so that it can store its own downtimes in the future (not in the scope of this PR). In the world of high availability federation (multiple Directors), a Director finally can take a rest!

PR #2218, opened 2025-04-18
Downtime workflow for stateless Directors

h2zh:
With the introduction of multiple directors, we are no longer able to store the downtimes in the sqlite database residing in the Director, because there is no way to sync the data in the database among multiple directors.

The only data in the Director db is the downtimes set by federation admin. The PR moved it to the Registry database.

In this new world, the Director only display the downtimes, which means you can't toggle downtime there. Instead:

1. Federation admin needs to schedule downtime for an origin or cache server in the Registry database, then Director will get them via a recurring API call Go routine every `Director_RegistryQueryInterval` (default: 1m), and keep them in an in-memory variable `federationDowntimes` with the data structure [serverName: a list of Downtime]. This is the data structure for `serverDowntimes` and `topologyDowntimes` too.
2. Origin or cache server admin will set their scheduled downtimes in their own database, then propagate to the Director via server ads on the basis of every `Server_AdvertisementInterval` (default: 1m). `serverDowntimes` will record these downtimes.
3. Besides, there is another variable`topologyDowntimes` to save the downtimes polled from topology every `Federation_TopologyReloadInterval` (default: 4.5m).

**IMPORTANT**: For 1,2,3, whenever the Director receives downtimes, it traverses them to identify any active ones and adjusts the `filteredServers` (the in-memory tracker for active downtimes) accordingly. 
Behind the scenes, it removes all existing filters set by this source (fed admin→tempFiltered, server admin→serverFiltered, topology→topoFiltered) in `filteredServers` and reapplies filters only to servers that are currently in downtime.

The general idea is that in a federation, each server maintains their own downtimes locally, and federation admin uses Registry to set the downtime to any server. Then all of these downtimes are propagated to the Director.
If a server is already marked in downtime in the Director, a new overlapped downtime can be accepted by the Director and stored in `<type>Downtimes`, but it will not change `filteredServers`.


patrickbrophy:
Could you update the swagger docs for the downtime APIs?

patrickbrophy (commented):
Can you create a design document for this? A diagram would be nice too

h2zh:
hey @patrickbrophy, would the draft illustration in the linked issue be helpful?

patrickbrophy:
Just testing the functionality I believe I may have found a bug. Here are the steps to recreate the bug:
1. Create a downtime to be scheduled for 5 minutes:
```bash
curl -X 'POST' \
  'https://localhost:8444/api/v1.0/downtime' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "class": "SCHEDULED",
  "description": "Test",
  "severity": "Outage (completely inaccessible)",
  "startTime": 1745246580,
  "endTime": 1745246880,
"source": "cache"
}'
```

2. Attempt to retrieve the active downtime:
 ```bash
curl -X 'GET' \
  'https://localhost:8444/api/v1.0/downtime?status=incomplete&source=cache' \
  -H 'accept: application/json'
```
The result of this request is 
```json
[]
```

However I'd expect it to be something like this
```json
[
  {
    "id": "019658cd-cc01-7323-a402-ee839c62c486",
    "serverName": "",
    "createdBy": "admin",
    "updatedBy": "admin",
    "source": "cache",
    "class": "SCHEDULED",
    "description": "Test",
    "severity": "Outage (completely inaccessible)",
    "startTime": 1745246580,
    "endTime": 1745246880,
    "createdAt": 1745246604289,
    "updatedAt": 1745246604289
  }
]
```

CannonLock:
@patrickbrophy This could because your post is in seconds and the backend is looking for milliseconds.

patrickbrophy:
'doh...
I didn't even notice that. I'll try again.

patrickbrophy:
After trying again things seem to working as expected. Sorry @h2zh for the confusion.

h2zh:
Good catch though! I will add a time format validation in `validateDowntimeInput`.
Btw, the draft illustration is in #2199. You can also find it in "development" section on the sidebar in this PR. @patrickbrophy

CannonLock:
@h2zh Couple high level suggestions:

- Clean up the old director downtime cruft
  - Do we need the filteredServers map?
  - Can we get rid of the two director endpoints used for filtering?

Director API's:
- Is there any reason we can't use the same endpoint to GET downtimes as we do with the cache and the origin? ( GET `/downtimes` )
- The path for getting a servers downtimes should be `director_ui/servers/{name}/downtimes` I would expect the current one to return a single downtime by the id.

h2zh:
Thanks for these feedbacks! @CannonLock 

1. I did remove the old cruft in the codebase as much as possible, while I think `filteredServers` is still necessary. It helps us track the latest downtime info for each server, otherwise for each server we need to loop through all places (federationDowntimes, serverDowntimes and topologyDowntimes) to compute if it is currently in downtime, which will drastically slow down APIs like `/api/v1.0/director_ui/servers`.

2. Thanks for the reminder - I deleted Director's two downtime filter toggling APIs but forgot to delete the corresponding Swagger docs. I just fixed this problem and will push the commit soon.

3. Because the general downtime database and APIs are applied in each Pelican server, including Director. Though we say the Director will be "stateless", it still gets a downtime table in its database, to record its own downtime in the world of multi Directors. It is an empty table now as a reserved design for future functionality. While the downtimes Director routinely scrapped from other Pelican services are stored in the in-memory maps, I decided to assigned it an unique endpoint in director_ui.

4. That makes sense! I will make the adjustments and push a new commit.

CannonLock:
@h2zh Ah I see. 

My only other thought is that that backend should be setting a default servername that is the name of the current services server. Only on the registry should I be allowed to set the servername to something. Although if I don't have one we assume I am marking downtime for the registry itself.

h2zh:
1. Done. Registry can't put itself to downtime yet. 

> In theory, Registry can do it using the downtime table in Registry. It will be passed to Director in `federationDowntimes` and recorded in `filteredServers`. But right now, Director doesn't have a way to handle it. Besides, I suppose it will never happen in production? 

2. Sure, I am in favor of client-side grouping to allow dynamic grouping strategies, so a flat list will be returned for `/director_ui/downtimes` and `/director_ui/server/{name}/downtimes`.

@CannonLock

patrickbrophy (approved):
Look good to me.

Issue #2131, opened 2025-03-21
Client token selection should take issuers into account


brianhlin:
Pelican Version:
7.14.1

Pelican Service:
- [X] Client
- [X] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

**Describe the bug**
When presented with multiple tokens in a `_CONDOR_CREDS` dir, Pelican only compares the paths in the token's scopes vs the `targetResource` to determine whether or not a token is likely to pass federation authZ checks. This means that a token with the correct issuer / scopes may get ignored if another token 

1. Comes before the appropriate token in `filepath.Walk` (alphabetical order?)
2. Has scopes that look like they would work for the relevant action + scope path

**To Reproduce**
Generate a WLCG token (see #2130 ) named `a.use` with `storage.modify:/ storage.read:/` with issuer A and another token named `b.use` with `storage.modify:/chtc/ storage.read:/chtc` from `iss: https://chtc.cs.wisc.edu`, then run:

```
export _CONDOR_CREDS=<path to dir containing .use JWTs>
pelican object get -d osdf://chtc/staging/blin/test /tmp/foo
```

The command should fail but the most interesting bit is that Pelican will pick the token `a.use`.

**Expected behavior**
I would expect Pelican to also compare the issuer in the token to the issuer in the director response asociated with the `targetResource`.

PR #2219, opened 2025-04-18
Client uses token issuer as selection criteria during token selection process

patrickbrophy:
This PR addresses issue #2131 by adding an additional selection criterion to the token selection process. Specifically, it introduces a check to compare the issuer provided in the Director’s response with the issuer in each token. Tokens are only considered further if the issuers match.


jhiemstrawisc (approved):
LGTM. Thanks!

PR #2223, opened 2025-04-21
Accept SciTokens-profile JWT

### LINKED ISSUE #2130: Client token selection logic ignores Scitokens JWTs
Pelican Version:

7.14.1

Pelican Service:
- [X] Client
- [X] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

**Describe the bug**
When Pelican is presented with multiple SciTokens, it cannot find the correct one because it deems non-WLCG tokens as unacceptable (https://github.com/PelicanPlatform/pelican/blob/main/client/acquire_token.go#L441-L444). The client may get lucky picking the correct token if it's named `scitokens.use` or in the absence of a `scitokens.use`, whatever token comes first in the `filepath.Walk` (https://github.com/PelicanPlatform/pelican/blob/main/client/acquire_token.go#L201).

**To Reproduce**
Generate a JWT with `ver: "scitokens:2.0"` and pass it to `pelican object get -d -t <path to token>`. You should always get a warning:

```
Using provided token even though it does not appear to be acceptable to perform transfer
```

**Expected behavior**
I expect Pelican to be able to perform the same acceptability checks (e.g., scope + path comparisons) for SciTokens as it does for WLCG tokens

h2zh:
Pelican’s client today only recognizes WLCG tokens (i.e. JWTs with a wlcg.ver claim). When a user supplies a SciTokens‑profile JWT (with ver: "scitokens:2.0"), Pelican will reject it as “unacceptable”. This PR extends the token profiles Pelican support from WLCG to SciTokens. Check out the linked issue for more details.

Close #2130

### DIFF
--- a/client/main_test.go
+++ b/client/main_test.go
@@ -506,22 +506,18 @@ func TestParseRemoteAsPUrl(t *testing.T) {
 	}
 }
 
-// mustURL parses a string into *url.URL or fails the test.
-func mustURL(t *testing.T, s string) *url.URL {
-	u, err := url.Parse(s)
-	require.NoError(t, err)
-	return u
-}
-
 // Verify if a scitoken‐profile JWT is acceptable for a given namespace
 func TestTokenIsAcceptableForSciTokens(t *testing.T) {
+	issuerURL, err := url.Parse("https://issuer.example")
+	require.NoError(t, err)
+
 	// 1) Build a minimal DirectorResponse whose namespace is "/foo"
 	dirResp := server_structs.DirectorResponse{
 		XPelNsHdr: server_structs.XPelNs{
 			Namespace: "/foo",
 		},
 		XPelTokGenHdr: server_structs.XPelTokGen{
-			Issuers:   []*url.URL{mustURL(t, "https://issuer.example")},
+			Issuers:   []*url.URL{issuerURL},
 			BasePaths: []string{"/foo"},
 		},
 	}
--- a/client/main_test.go
+++ b/client/main_test.go
@@ -20,6 +20,9 @@ package client
 
 import (
 	"context"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


client/main_test.go
```diff
@@ -497,3 +505,61 @@ func TestParseRemoteAsPUrl(t *testing.T) {
 		})
 	}
 }
+
+// mustURL parses a string into *url.URL or fails the test.
+func mustURL(t *testing.T, s string) *url.URL {
```
turetske, line 510: It seems very odd to have this as a separate function for something that is only used once. Why isn't this check done before the dirResp assignment and then have the result places in the slice rather than doing a separate function.

turetske (changes requested):

client/main_test.go
h2zh: My bad. That's an overthinking. Just reorganized the code.

turetske (approved):

josh-willis:
I realize I am coming to this months after it was merged, because I just saw it in recent release notes. I am happy to open an issue if asked.

This fix checks for the `ver` claim to start with `scitokens:` in order to conclude that the JWT is a scitoken. I want to raise here the analogous bug in HTCondor that we in LIGO opened in [HTCONDOR-2285](https://opensciencegrid.atlassian.net/browse/HTCONDOR-2285). Like this fix, the local HTCondor-issued scitokens by default had `ver` claim `scitokens:2.0` (though this could be overriden with a config knob). However, the reference Python scitokens library specifies that the claim must be `scitoken:2.0` (note the singular!) as you can see for instance [here](https://github.com/scitokens/scitokens/blob/v1.8.1/src/scitokens/scitokens.py#L488).

This is also what I understand the technical documentation [here](https://scitokens.org/technical_docs/Verification) to say (you'll have to scroll down to the version discussion under *SciToken-specific Claim Processing*, I couldn't see an anchor for that section heading). 

So, both the reference implementation and those docs seem to imply that the correct behavior is for the `ver` claim to be `scitoken:2.0` singular, thus I am concerned that tokens following this specification will not be accepted by the pelican client and plugin.

PR #2224, opened 2025-04-21
Create cache socket as temporary and overwrite

### LINKED ISSUE #2216: Pelican local cache should clean up socket upon startup
Pelican Version:

```
$ pelican --version
Version: 7.15.1
Build Date: 2025-04-14T21:45:02Z
Build Commit: b96908de5b586bf9227b25ff83aca793911d6d65
Built By: goreleaser
```

Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [X] Cache

**Describe the bug**
If the Pelican local cache is unexpectedly killed, it will leave behind a `cache.sock` and subsequent attempts to restart the cache will fail:

```
time="2025-04-04T06:20:53-05:00" level=error msg="Fatal error occurred at the start of the program. Cleanup started: listen unix /var/cache/pelican/localcache/cache.sock: bind: address already in use"
```

**To Reproduce**
Start the Pelican local cache, sigkill the process, then try to restart it.

**Expected behavior**
I would expect Pelican to clear out the socket and any "remains" from previous invocations of the service. Perhaps making use of a lockfile that records the PID and adding a startup check for said PID before cleaning up.

**Additional context**
Additionally, the paths in the local cache seem a bit repetitive:

```
$ ls -l /var/cache/pelican/localcache/
total 4
drwx------ 2 root root 4096 Apr 16 13:03 cache
srwxr-xr-x 1 root root    0 Apr 16 13:03 cache.sock
```

This may be coming from HTCondor config setting:

```
$ condor_config_val -v pelicancache_runlocation
PELICANCACHE_RUNLOCATION = /var/cache/pelican/localcache
 # at: /etc/condor/include/pelican_cache.conf, line 3, use FEATURE:PelicanCache+5
 # raw: PELICANCACHE_RUNLOCATION = /var/cache/pelican/localcache
```

Should we just set that to `/var/cache/pelican`?

bbockelm:
Create a temporary file for the local cache unix socket and rename it into place.  This prevents any pre-existing socket (say, from a prior crash of the local cache) from causing startup to fail.

Fixes #2216

### DIFF
--- a/fed_test_utils/fed.go
+++ b/fed_test_utils/fed.go
@@ -146,7 +146,10 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
 	viper.Set(param.Cache_DbLocation.GetName(), filepath.Join(t.TempDir(), "cache.sqlite"))
 	viper.Set(param.Server_EnableUI.GetName(), false)
 	viper.Set(param.Server_WebPort.GetName(), ports[2])
-	viper.Set(param.LocalCache_RunLocation.GetName(), filepath.Join(tmpPath, "local-cache"))
+	// Unix domain sockets have a maximum length of 108 bytes, so we need to make sure our
+	// socket path is short enough to fit within that limit. Mac OS X has long temporary path
+	// names, so we need to make sure our socket path is short enough to fit within that limit.
+	viper.Set(param.LocalCache_RunLocation.GetName(), filepath.Join(tmpPath, "lc"))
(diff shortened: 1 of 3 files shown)


bbockelm:
Given there's active suffering in CHTC over this one (and there's a unit test which is passing), I'm going to pull PI privilege and merge for v7.16.

Note the failing test is due to the fact I used the "Fixes" semantics which breaks our PR validation action.

PR #2228, opened 2025-04-22
Warn user if they still use the key file at `IssuerKey`

h2zh:
This PR:
~~1. Update the `--private-key` flag to work with IssuerKeysDirectory~~
2. Warn the user if they still use the deprecated `IssuerKey` (but the private key it points to is still incorporated)

I did a thorough check on the usage of `--private-key` flag in the Pelican CLI. 
1. `pelican origin token create --private-key=...` It specifies Pelican's valid private keys directory. This is updated in this PR.
2. `pelican generate keygen --private-key=...` It sets the location for the private key file you're going to generate. It doesn't needed to be updated to accommodate `IssuerKeysDirectory`
3. `pelican object {get, put, ls}` @brianaydemir mentions these commands use `--private-key=...` flag, but I can't find it in the [code](https://github.com/PelicanPlatform/pelican/blob/0443c59eeecc5468af413abcc39a66511231e73c/cmd/object_ls.go#L47)? Instead, I find the `--token` flag.

### DIFF
--- a/cmd/origin.go
+++ b/cmd/origin.go
@@ -24,6 +24,8 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+
+	"github.com/pelicanplatform/pelican/param"
 )
 
 var (
@@ -185,8 +187,8 @@ instead.
 	if err := viper.BindPFlag("Server.IssuerUrl", originTokenCreateCmd.Flags().Lookup("issuer")); err != nil {
 		panic(err)
 	}
-	originTokenCreateCmd.Flags().String("private-key", "", "Filepath designating the location of the private key in PEM format to be used for signing, if different from the origin's default.")
-	if err := viper.BindPFlag("IssuerKey", originTokenCreateCmd.Flags().Lookup("private-key")); err != nil {
+	originTokenCreateCmd.Flags().String("private-key", "", "Directory designating the location of the private key file(s) in PEM format to be used for signing, if different from the origin's default.")
+	if err := viper.BindPFlag(param.IssuerKeysDirectory.GetName(), originTokenCreateCmd.Flags().Lookup("private-key")); err != nil {
 		panic(err)
 	}
 	originTokenCmd.AddCommand(originTokenVerifyCmd)
--- a/cmd/origin.go
+++ b/cmd/origin.go
@@ -24,8 +24,6 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-
-	"github.com/pelicanplatform/pelican/param"
 )
 
 var (
@@ -187,8 +185,8 @@ instead.
 	if err := viper.BindPFlag("Server.IssuerUrl", originTokenCreateCmd.Flags().Lookup("issuer")); err != nil {
 		panic(err)
 	}
-	originTokenCreateCmd.Flags().String("private-key", "", "Directory designating the location of the private key file(s) in PEM format to be used for signing, if different from the origin's default.")
-	if err := viper.BindPFlag(param.IssuerKeysDirectory.GetName(), originTokenCreateCmd.Flags().Lookup("private-key")); err != nil {
+	originTokenCreateCmd.Flags().String("private-key", "", "Filepath designating the location of the private key in PEM format to be used for signing, if different from the origin's default.")
+	if err := viper.BindPFlag("IssuerKey", originTokenCreateCmd.Flags().Lookup("private-key")); err != nil {
 		panic(err)
 	}
 	originTokenCmd.AddCommand(originTokenVerifyCmd)
--- a/config/config.go
+++ b/config/config.go
@@ -1019,6 +1019,22 @@ func PrintClientConfig() error {
 	return nil
 }
 
+// Helper function to emit a warning telling the user to switch to IssuerKeysDirectory,
+// if the IssuerKey config param still points to a file on disk.
+func warnDeprecatedConfigs(v *viper.Viper) {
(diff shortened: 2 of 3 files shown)


brianaydemir:
@h2zh There is some ideal world in which any Pelican operation that requires a token would attempt to fabricate such a token if it had access to a private key. I thought that this is how the Pelican CLI was intended to behave, but I'm not actually sure I tested that for myself.

config/config.go
```diff
@@ -1019,6 +1019,22 @@ func PrintClientConfig() error {
...
+		if _, err := os.Stat(legacyKey); err == nil {
+			log.Warnf(
+				"Deprecated configuration %q detected — this will be removed in a future release. "+
+					"Please switch to %q, pointing at a directory of private key files with .pem extension.",
+				param.IssuerKey.GetName(),
+				param.IssuerKeysDirectory.GetName(),
+			)
+		}
+	}
+}
```
jhiemstrawisc, line 1036: I think this should already be handled by `handleDeprecatedConfig()`, no? If we do need an extra handler, it should use the parsed `parameters.yaml`, which marks configs as deprecated instead of requiring duplication (and ongoing maintenance) of a list here.

cmd/origin.go
```diff
@@ -185,8 +187,8 @@ instead.
 	if err := viper.BindPFlag("Server.IssuerUrl", originTokenCreateCmd.Flags().Lookup("issuer")); err != nil {
 		panic(err)
 	}
-	originTokenCreateCmd.Flags().String("private-key", "", "Filepath designating the location of the private key in PEM format to be used for signing, if different from the origin's default.")
-	if err := viper.BindPFlag("IssuerKey", originTokenCreateCmd.Flags().Lookup("private-key")); err != nil {
+	originTokenCreateCmd.Flags().String("private-key", "", "Directory designating the location of the private key file(s) in PEM format to be used for signing, if different from the origin's default.")
+	if err := viper.BindPFlag(param.IssuerKeysDirectory.GetName(), originTokenCreateCmd.Flags().Lookup("private-key")); err != nil {
```
jhiemstrawisc, line 191: Can you confirm that `IssuerKey` still works here? The bigger issue in my mind is that the previous changes generally broke the ability to use `IssuerKey` in configuration, and while I agree we probably shouldn't bind to a deprecated param, we still want to make sure this works.

docs/pages/federating-your-data/generating-tokens.mdx
```diff
@@ -37,5 +37,5 @@ where:
 
 ### Additional Command Line Arguments
 
-- `--private-key` encodes the path to the private key used to sign the token. By default it uses the private key defined by the Origin server that the command runs on.
+- `--private-key` encodes the directory where you saved the private key(s) used to sign the token. By default it uses the private key defined by the Origin server that the command runs on.
```
jhiemstrawisc, line 40: We shouldn't change the behavior here -- the `--private-key` flag should still point to an explicit private key and not a directory.

config/config.go
```diff
@@ -1259,6 +1275,9 @@ func SetServerDefaults(v *viper.Viper) error {
 		v.SetDefault("Federation_DirectorUrl", v.GetString(param.Server_ExternalWebUrl.GetName()))
 	}
 
+	// Warn the user if they still use IssuerKey
+	warnDeprecatedConfigs(v)
+
```
jhiemstrawisc, line 1280: See previous comment -- we should already be printing out a warning when `IssuerKey` is set in configuration.

jhiemstrawisc (changes requested):

brianaydemir:
I thought I was told by someone that `object {get,put,ls} --private-key {key}` was or would be a thing. Or maybe I misunderstood the scope of #1591.

docs/pages/federating-your-data/generating-tokens.mdx
h2zh: I think there's nothing need to change on the current `private-key` flag for now. Since it should still point to an explicit private key, which is a file location thus can't bind to `IssuerKeysDirectory`. It is currently binding to `IssuerKey`, and Pelican has always incorporated the key there.

cmd/origin.go
h2zh: See below.

config/config.go
brianaydemir: Empirically, @h2zh is correct here. That said, I find the wording of the warning to be misleading. If I run Pelican with an empty config file and set the log level to `warn` (the default is `error`), I am told:

```
$ PELICAN_LOGGING_LEVEL=warn ./.pelican.sh --config empty.yaml object ls osdf://asdfsdf
WARNING[2025-04-23T20:14:45Z] Deprecated configuration "IssuerKey" detected — this will be removed in a future release. Please switch to "IssuerKeysDirectory", pointing at a directory of private key files with .pem extension.
```

Telling me to switch away from a configuration parameter that I never set is confusing.

I'm not sure exactly how to word this, but I'd try to convey the following, in order:

1. The action the user needs to take: File X  should be moved into directory Y.
2. The reason that action is necessary: `IssuerKey` (which is currently set to X by {default or explicit config}) is being deprecated in favor of `IssuerKeyDirs` (which is currently set to Y by {default or explicit config}).

h2zh: Thanks Brian! I improved the wording and set this message to "error" level so that user can see it by default.

brianaydemir:
@jhiemstrawisc @h2zh At this point, do we still need the "create patch" label on this PR? Emitting a new warning doesn't strike me as that important.

brianaydemir (approved):

PR #2229, opened 2025-04-23
Add New Cache Metrics

patrickbrophy:
This PR adds new cache metrics. It includes new statistics for the bytes prefetched and the bytes to disk. These new statistics can be used to accurately reflect the hits and misses for a cache.


jhiemstrawisc (approved):
Looks okay to me

Issue #2232, opened 2025-04-23
Docker images may silently overwrite XRootD with incorrect version


jhiemstrawisc:
While working on #2221, I encountered an issue that left me scratching my head for a while. The issue arose from this Dockerfile configuration:
```
ARG XRDHTTP_PELICAN_SRC_BUILD=false
ARG XRDHTTP_PELICAN_VER=0.0.4

...

ARG XROOTD_VER="5.8.1"
```

Due to the Dockerfile's order of operations, we first install `xrootd-5.8.1`. However, the spec file for `xrdhttp-pelican-0.0.4` has an explicit requirement for xrootd < 5.8.0, so when it gets installed it overwrites xrootd with `xrootd-5.7.3`.

I discovered this after running `rpm -q xrootd` in the container after it was built, and slowly working through each of the rpm installs to see which was causing the issue.

It should be possible to run something somewhere in the build process to make sure we don't overwrite software like this.

PR #2241, opened 2025-04-23
Prevent package dependencies from changing the XRootD install

brianaydemir:
From #2232:

> It should be possible to run something somewhere in the build process to make sure we don't overwrite software like this.

This is exactly what [DNF's versionlock plugin](https://dnf-plugins-core.readthedocs.io/en/latest/versionlock.html) is for.


jhiemstrawisc (approved):
I confirmed this would've prevented the issue that caused me to open the issue. Thanks!

PR #2244, opened 2025-04-23
Security update as much as possible

CannonLock:


CannonLock:
@jhiemstrawisc This is the best we can do right now.

https://github.com/react-syntax-highlighter/react-syntax-highlighter/issues/591

jhiemstrawisc:
@CannonLock can you try rebasing this against main again? I might have temporarily broken something when you created this branch.

jhiemstrawisc (approved):
I think the test failures are noise from another commit. I'm going to merge and keep a close eye on subsequent action runs to make sure.

PR #2247, opened 2025-04-24
Update packages to prevent dependabot security warnings

jhiemstrawisc:


images/Dockerfile
```diff
@@ -35,7 +35,7 @@ ARG OSG_SERIES=24
 ARG OSG_REPO=release
 
 # For compiling Pelican.
-ARG GO_VER=1.21.6
+ARG GO_VER=1.23.0
```
brianaydemir, line 38: Should this be `1.23.8`, to match the `toolchain` in `go.mod`?

brianaydemir (approved):
@jhiemstrawisc I left one comment for you, that you can address as you see fit before merging.

images/Dockerfile
jhiemstrawisc: I was wondering that myself. After a bit more research, I think the answer is "probably yes".

jhiemstrawisc:
The remaining test failures are known flakes (director memory stat and macos filesize length limits). Not sure what's going on with the linter, but I'm suspecting there's an order of operations issue here where these linter failures will go away once this is merged to main. I'm going to do that and then keep a close on subsequent action runs to determine whether there's more to fix.

PR #2249, opened 2025-04-24
Update the version of golangci-lint used by pre-commit

brianaydemir:
This seems to resolve the failure in the GitHub Action.

In the name of expediency, I'm not attempting to upgrade everything that's fallen behind.


jhiemstrawisc (approved):
LGTM. I'll approve and let you rebase as you see fit to see if we can tidy up the test failures.

brianaydemir:
The test failures are due to errors in building the web assets or some such.

PR #2259, opened 2025-04-26
Bugfixes, tests, and tools around checksums

bbockelm:
This PR improves the checksum functionality:
- Adds a bugfix that prevents checksum failures from not being reported as checksum mismatch errors (@jhiemstrawisc - I think this is probably important enough to put into a 7.16 RC).
- Records the client and server-side checksums into the plugin outputs.  This will allow ElasticSearch to detect the frequency of errors.
- Adds a flag to `pelican object stat` that allows CLI users to get a checksum back.

Example:
```
pelican object stat --json --checksums md5 pelican://`hostname`:8444/tmp/test/hello_world.txt
{"Name":"/tmp/test/hello_world.txt","Size":12,"ModTime":"2025-04-26T15:16:26Z","IsCollection":false,"checksums":{"md5":"6f5902ac237024bdd0c176cb93063dc4"}}
```

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -350,14 +350,21 @@ const (
 	attrProjectName classAdAttr = "ProjectName"
 	attrJobId       classAdAttr = "GlobalJobId"
 
+	// The checksum algorithms supported by the client
+	//
+	// Note we have a helper function, KnownChecksumTypes, that returns a list
+	// of all the elements enumerated below; do not skip integers in this list
+	// or that functionality will break.
+	//
 	AlgMD5     ChecksumType = iota // Checksum is using the MD5 algorithm
 	AlgCRC32C                      // Checksum is using the CRC32C algorithm
 	AlgCRC32                       // Checksum is using the CRC32 algorithm
 	AlgSHA1                        // Checksum is using the SHA-1 algorithm
-	AlgUnknown                     // Unknown checksum algorithm
+	AlgUnknown                     // Unknown checksum algorithm.  Always a "trailer" indicating the last known algorithm.
 
 	AlgDefault = AlgCRC32C // Default checksum algorithm is CRC32C if the client doesn't specify one.
 	algFirst   = AlgMD5
+	algLast    = AlgUnknown
 
 	SyncNone  = iota // When synchronizing, always re-transfer, regardless of existence at destination.
 	SyncExist        // Skip synchronization transfer if the destination exists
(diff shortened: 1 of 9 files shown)


client/handle_http.go
```diff
@@ -386,7 +396,26 @@ func checksumFromHttpDigest(httpDigest string) ChecksumType {
...
 
-func httpDigestFromChecksum(checksumType ChecksumType) string {
+// List all the checksum types known to the client
+func KnownChecksumTypes() (result []ChecksumType) {
+	result = make([]ChecksumType, AlgUnknown-algFirst)
+	for idx := algFirst; idx < AlgUnknown; idx++ {
+		result[idx-algFirst] = idx
+	}
+	return
+}
```
jhiemstrawisc, line 406: This feels like a complex way to declare what could be a static list, and while I think your goal here was to prevent creating a list that needs to be maintained, this approach comes with its own assumptions. If you really think we need the iota math here, can you:
1. Leave a comment near the iota declaration that indicates the iota must not skip any integers
2. Declare and use an `algLast` to be more explicit about which algorithms must be first and last in the list (plus it rounds out the symmetry).

client/handle_http_test.go
```diff
@@ -763,6 +764,237 @@ func TestGatewayTimeout(t *testing.T) {
...
+	assert.Equal(t, ChecksumType(AlgCRC32C), info.Algorithm)
+
+	assert.Equal(t, 1, len(transferResult.ClientChecksums), "Checksum count is %d but should be 1", len(transferResult.ClientChecksums))
+	info = transferResult.ClientChecksums[0]
+	assert.Equal(t, "977b8112", hex.EncodeToString(info.Value))
+	assert.Equal(t, ChecksumType(AlgCRC32C), info.Algorithm)
+}
+
+// Test behavior when checksum is missing
+func TestChecksumMissing(t *testing.T) {
```
jhiemstrawisc, line 872: Each of these new checksum tests is largely a copy-paste-tweak of the others. Instead of using three separate tests, can you create one test that handles a slice of test cases?

jhiemstrawisc (changes requested):

client/handle_http.go
bbockelm: Done.

jhiemstrawisc (approved):
LGTM. I'll add this as a 7.16 patch as well.

Issue #2266, opened 2025-04-29
Windows client cannot save tokens it obtains


brianaydemir:
Supposing that #2264 is fixed, the next issue one will encounter with the Windows client is that it cannot actually save the token it has obtained:

```
WARNING[2025-04-29T12:20:23-05:00] Failed to save new token to configuration file: rename C:\Users\baydemir\.config\pelican\credentials\oauth2-client.pem2906852323 C:\Users\baydemir\.config\pelican\credentials\client-credentials.pem: The process cannot access the file because it is being used by another process.
```

This leads to a poor user experience, as every operation requiring a token will require the user to authorize the operation.

Issue #2264, opened 2025-04-29
Windows client cannot be used to obtain tokens


brianaydemir:
When an `object` operation requires a token and the client does not already have one, the client goes through the OAuth2 device code flow to request one. As part of this flow, the user is supposed to be prompted to visit a URL and authenticate. On Windows, this URL is never displayed:

```
DEBUG[2025-04-29T10:46:12-05:00] Requesting a credential with the following scope: storage.create:/
DEBUG[2025-04-29T10:46:17-05:00] After sleep of 5 seconds, attempting to retrieve token.
```

On Linux and macOS, text to the effect of "To approve credentials for this operation..." is displayed between those two lines of debug output.

PR #2267, opened 2025-04-29
Fix the Windows client's ability to request and save tokens

brianaydemir:
#2264 is caused by trying to write the authorization URL to standard input instead of standard output.

#2266 is seemingly caused by keeping the file open when calling `os.Rename` (the process that's still using the file is apparently the process itself). As I understand it, calling `fp.Close()` multiple times is not well-defined behavior, hence the awkward scoping of `defer fp.Close()`.


jhiemstrawisc (approved):
LGTM

PR #2270, opened 2025-04-29
Properly handle the trailing`/` in namespace registration and advertisement

### LINKED ISSUE #1806: Better handling of trailing `/` in federation prefixes
A user tried registering a namespace like `/foo/bar/`, which generated an error. The correct registration was `/foo/bar` without a trailing `/`.

I _do_ think it's important that our internal representations of prefixes contain a trailing `/` so that it's much harder to accidentally match `/foo/barbaz` to `/foo/bar`, but we can make this logic transparent to users by permitting registration of both `/foo/bar` and `/foo/bar/` (and treating them the same under the hood).

h2zh:
We should allow user to set either `/foo/bar` or `/foo/bar/` as the federationprefix, but fundamentally treat them as the same (without trailing slash).

This patch switches the order of prefix validation and prefix existence checking. It removes the trailing slash for the user input prefix, which prevents the error [like this](https://github.com/PelicanPlatform/pelican/issues/1806#issuecomment-2840109314) in the future.

Closes #1806

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -993,6 +993,11 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
 		adV2 = server_structs.ConvertOriginAdV1ToV2(ad)
 	}
 
+	// Check every namespace path to strip the trailing slash
+	for i := range adV2.Namespaces {
+		adV2.Namespaces[i].Path = server_utils.RemoveTrailingSlash(adV2.Namespaces[i].Path)
+	}
+
 	// Filter the advertised prefixes in the cache server ad
 	// based on the allowed prefixes for caches data.
 	if sType == server_structs.CacheType {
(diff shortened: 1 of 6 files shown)


h2zh:
For 1), I create a new function to remove the trailing '/'  in Origin exports. Because `prefix` is just a local copy of the string passed in, reassigning it inside `validateFederationPrefix` won’t touch the original `b.Exports[i].FederationPrefix`.

h2zh:
For 2), after some digging, I think the origin test framework would be sufficient for this modification. While spinning up a new fed test is unnecessary - we can't rely on fed tests too much because each fed test increase the test time significantly (Tests within a package are running sequentially).

For 3), I'm happy to clear an eyesore 🧹🤪

jhiemstrawisc (approved):
LGTM. Thanks!!

PR #2274, opened 2025-04-30
Remove downtime toggles on Director

CannonLock:
@jhiemstrawisc This will need to make it into a rc. 

@h2zh Quick one that removes the downtime toggles on the director that I missed in the first PR.

### DIFF
--- a/web_ui/frontend/app/director/components/DirectorCard.tsx
+++ b/web_ui/frontend/app/director/components/DirectorCard.tsx
@@ -16,7 +16,7 @@ import {
 } from '@mui/material';
 import { red, grey } from '@mui/material/colors';
 import { Server } from '@/index';
-import { Equalizer, Language } from '@mui/icons-material';
+import { Equalizer, Language, FolderOff } from '@mui/icons-material';
 import { NamespaceIcon } from '@/components/Namespace/index';
 import useSWR from 'swr';
 import Link from 'next/link';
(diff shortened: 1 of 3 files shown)


web_ui/frontend/app/director/components/DirectorCard.tsx
```diff
@@ -87,47 +87,6 @@ export const DirectorCard = ({ server, authenticated }: DirectorCardProps) => {
...
-                              dispatch
-                            );
-
-                            mutate();
-
-                            setDisabled(false);
-                          }}
-                        />
-                      }
-                      label={server.filtered ? 'Disabled' : 'Active'}
```
h2zh, line 126: While the toggle is removed, I'm wondering if it would be better to keep the text to also display if a server is in the downtime?

h2zh (changes requested):

web_ui/frontend/app/director/components/DirectorCard.tsx
CannonLock: Good call. 

I didn't want to make it super first class so as to compete with the other items in the card. 

I did add it to the dropdown data as well as showing itself as a grey icon on the card rather then the blue.

jhiemstrawisc: In addition to the gray box, can we have some explicit text? So something like:
```
<origin icon> Origin * 7.8.0 (down)          <globe icon> <stats icon>
```

h2zh (approved):
LGTM!

PR #2275, opened 2025-04-30
Correctly compare the total local size with the total downloaded size

bbockelm:
Previously, only the bytes from the latest attempt (which doesn't necessarily start at the beginning of the file if done during a retry) were used, meaning the comparison with total size always failed.

Additionally, avoid doing a transfer retry when there are zero bytes left to transfer.  That will always fail.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -2165,6 +2165,12 @@ func downloadObject(transfer *transferFile) (transferResults TransferResults, er
 			log.WithFields(fields).Debugln("Downloaded bytes:", downloaded)
 			success = true
 			break
+		} else if size > 0 && downloaded == size {
+			// We have downloaded all the data but we still have an error.  If we retry again,
+			// we will read past the end of the file and generate yet another error.  So, break
+			// and cause a permanent failure.
+			// In the future, we can consider deleting the file and trying again.
+			break
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


client/handle_http.go
```diff
@@ -2856,9 +2862,9 @@ Loop:
 	// <= 0, it indicates we don't know how large the transfer was supposed to be in the first
 	// place.
 	if totalSize > 0 {
-		if downloaded != totalSize {
-			log.WithFields(fields).Debugf("Download completed but received size %db does not match expected size '%db'", downloaded, totalSize)
-			err = errors.Errorf("download completed but received size %db does not match expected size '%db'", downloaded, totalSize)
+		if bytesSoFar+downloaded != totalSize {
+			log.WithFields(fields).Debugf("Download completed but received size %db does not match expected size '%db'", bytesSoFar+downloaded, totalSize)
+			err = errors.Errorf("download completed but received size %db does not match expected size '%db'", bytesSoFar+downloaded, totalSize)
```
jhiemstrawisc, line 2867: Can you fix a mistake I made that's now bothering me? One of the format specifiers is in single quotes and the other isn't. I don't have a strong preference either way, but it'd be better to be consistent.

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM. Feel free to merge when you're ready and tests are passing.

PR #2280, opened 2025-04-30
Toggle origin/cache downtimes in director correctly

### LINKED ISSUE #2279: Director doesn't toggle origin/cache downtimes correctly
Pelican Version:
7.16.0-rc.0

Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [X] Director
- [ ] Origin
- [ ] Cache

**Describe the bug**
When the Director receives the downtimes set by origin/cache along with the server ads, it doesn't update the server entry in `filteredServers` that were previously set into downtime. As a result:

- New active downtimes aren’t applied when a server is already filtered
- Expired downtimes aren’t cleared
- Servers can’t flip between “active” and “future” states

Downtimes created/modified by the Registry don't have this issue.

**To Reproduce**
Create a downtime in Origin, check it out in the Director. Then modify this downtime in Origin (e.g. change the start/end time, active->future), check the Director again.

h2zh:
Fix the bug that when the Director toggles the downtimes set by origin/cache, it skips re-evaluating any server that already has a filter (downtime).

Also simplify the code logics.

For example, let's say there is a downtime entry set by the Origin and propagated to the Director with server ads every minute:

> Note: "existing" means "existingFilterType", "had" means "isServerFiltered" in the code
> First ever “active” → had==false, so it sets serverFiltered.
> 
> Subsequent still-active → had==true && existing==serverFiltered, it doesn’t enter the `!had` branch, but we also don’t delete, because active==true skips the delete block.
> 
> Subsequent no longer active → active == false && had == true && existing == serverFiltered, so it falls into the delete block
> 
> Change time: Active to future downtime → active==false, so it hits the delete block (because existing==serverFiltered) and remove it.
> 
> Change time: Future to active downtime → had==false again, so it’ll set it back in filteredServers.

Closes #2279

### DIFF
--- a/director/director.go
+++ b/director/director.go
@@ -1169,33 +1169,36 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
 	// Process received server(origin/cache) downtimes and toggle the director's in-memory downtime tracker when necessary
 	if adV2.Downtimes != nil {
 		filteredServersMutex.Lock()
+		defer filteredServersMutex.Unlock()
+
+		// Update the cached server downtime list
 		serverDowntimes[adV2.Name] = adV2.Downtimes
-		currentTime := time.Now().UTC().UnixMilli()
-		bringItDown := false // Flag to indicate if this server is put in downtime during the traversal of the downtimes in this server ad
+
+		now := time.Now().UTC().UnixMilli()
 		sn := adV2.Name
-		// Check existing downtime filter
-		originalFilterType, hasOriginalFilter := filteredServers[sn]
-		// If this server is already put in downtime, we don't need to do anything
-		if !(hasOriginalFilter && originalFilterType != tempAllowed) {
-			// Check if the server is currently in downtime
-			for _, downtime := range adV2.Downtimes {
-				if downtime.StartTime < currentTime && (downtime.EndTime > currentTime || downtime.EndTime == server_structs.IndefiniteEndTime) {
-					// Server is currently in downtime
-					filteredServers[sn] = serverFiltered
-					bringItDown = true
-					break
-				}
+		active := false // Flag to indicate if this server has active downtime in this server ad
+		for _, dt := range adV2.Downtimes {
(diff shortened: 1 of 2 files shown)


director/director.go
```diff
@@ -1169,33 +1169,36 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
...
 			}
 		}
-		// If the server doesn't have an active downtime, it means Director's previously
-		// recorded downtime set by the server admin is stale and should be removed.
-		// It only removes the downtime set by the server admin, not the downtime set by others.
-		if !bringItDown {
-			if hasOriginalFilter && originalFilterType == serverFiltered {
+
+		// Inspect the existing downtime status for this server
+		existing, had := filteredServers[sn]
```
CannonLock, line 1189: Can you update the variable names to be more informative? I had to look in the code to figure out what values to expect here.

existingFilterType, serverIsFiltered would make it so you wouldn't need the comments.

CannonLock (changes requested):
Can you note for me the precedence of downtimes?

I assume Topo/Registry > Topo/Registry > ServerAds where ServerAds has the lowest precedence?

I want to double check that this precedence is enforced in the other locations.

h2zh:
@CannonLock At this moment the precedence is ... whoever set the downtime first. For example, if server A is set to downtime by the server admin, and later the Registry also sets this server to downtime, the Director will show both downtime details, but the reason (`filteredType` in json) will stick to whoever sets the downtime first ("Temporarily disabled by the server admin"). 

If the downtime period set by the server admin ends, but the time is still in the downtime period set by the Registry, the Director will still consider this server is in downtime (`filtered: true` in json) and the reason will switch to `filteredType: Temporarily disabled via the admin website`.

director/director.go
h2zh: Done. Thanks for the advice!

CannonLock (approved):

PR #2286, opened 2025-05-01
RPM: Specify pre-install script requirements by binary instead of by containing package

### LINKED ISSUE #2284: The pelican RPM cannot be installed on openSUSE  Leap 15
Pelican Version:
7.15.0+

Pelican Service:
- [X] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

**Describe the bug**
Cannot install on openSUSE Leap 15

**To Reproduce**
zypper install pelican

**Expected behavior**
Pelican gets installed

 - OS: OpenSUSE Leap 15
 

**Additional context**
The pelican RPM depends on glibc-common, which does not exist on openSUSE Leap 15.

    # rpm -qf /usr/bin/getent 
    glibc-2.38-150600.14.23.1.x86_64

matyasselmeci:
Closes #2284 . The packages providing `/usr/bin/getent`, `/usr/sbin/useradd`, and `/usr/sbin/groupadd` (used in the pre-install script to create the `pelican` user and group) are not named the same on openSUSE Leap vs RHEL-based distributions. We can get around the issue by requiring the specific files instead.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -170,10 +170,9 @@ nfpms:
               group: pelican
         dependencies:
           - "/bin/sh"
+          - "/usr/bin/getent"
           # for useradd and groupadd:
           - "shadow-utils"
-          # for getent:
-          - "glibc-common"
       deb:
         contents:
           - src: LICENSE
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -171,8 +171,8 @@ nfpms:
         dependencies:
           - "/bin/sh"
           - "/usr/bin/getent"
-          # for useradd and groupadd:
-          - "shadow-utils"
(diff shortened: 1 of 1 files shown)


matyasselmeci:
This might be worth putting into 7.16 and 7.15 as well...

jhiemstrawisc (approved):
👍

jhiemstrawisc:
I'll add this in the next patch I cut for 7.16, but I don't plan to create a patch solely for this (unless you think I should).

I'll leave it to you and @turetske to decide whether or not this goes into 7.15. It might be worth it since we still don't have an official 7.16 release.

timtheisen:
I can wait for the 7.16.1 release for the RPM patch.

PR #2290, opened 2025-05-02
Don't refresh expiration on serverAds.Get()

jhiemstrawisc:
The bug here was nuanced -- if a server had previously succeeded at advertising but was restarted quickly in a way that caused advertisement failure, the IO Stats routine would refresh the ad every 15 seconds, completely preventing expiration. This adds the disable on touch option so that these ads can expire.

I've also added it as the default behavior on ttlcache instantiation as well. Why do it in both places, you may ask? I want to future proof the bug against anyone purposefully/accidentally removing the default from the ttlcache instantiation, while also adding some protection against this type of bug for anyone in the future who might write a routine that calls `serverAds.Get()` without remembering to include the option there as well. Including the option in both places covers both of these bases.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1559,9 +1559,13 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
 
 			viper.Set(param.Director_RegistryQueryInterval.GetName(), "1m")
 		}
-	}
 
-	if currentServers.IsEnabled(server_structs.DirectorType) {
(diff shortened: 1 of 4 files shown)


jhiemstrawisc:
@brianaydemir testing instructions for this are nuanced -- let's talk in person about how to make sure this fix works.

bbockelm (approved):
LGTM.

Given this is the Nth issue in correct use of the TTLCache API (other common problem has been forgetting to start or stop the eviction routine), I do wonder if we should have some sort of helper class to polish off the rough edges.

PR #2308, opened 2025-05-13
A wider test downtime window to avoid flaky test

h2zh:
There is a flaky test on GitHub Action (actually, almost always failing) because the test downtime window is too tight, which results in the test downtime expires before the server ad gets to the Director.


jhiemstrawisc (approved):
🚀

PR #2316, opened 2025-05-14
Stop restarting Pelican if it crashes too frequently

matyasselmeci:
Disable the SystemD auto-restart of Pelican services if they've exited after less than 60 seconds, 3 times in a row.  After that, the admin will need to `systemctl start SERVICE` by hand.


brianhlin (approved):
LGTM

bbockelm:
@matyasselmeci - what do you want to do with this PR in the end?  Per the last comment, should we just close it out?

matyasselmeci:
Yeah, I'll close it.

PR #2334, opened 2025-05-20
Explicitly install make in pelican-build-base image

jhiemstrawisc:
This dependency was previously picked up implicitly by goreleaser, but stopped being installed automatically. However we still need it to run goreleaser's `make web-build` hook.


brianaydemir (approved):
LGTM. (I'm assuming the build tests will pass.)

PR #2338, opened 2025-05-21
Add health status to server ads and use to report high concurrency in origins/caches

jhiemstrawisc:
This PR does two primary things:
1. It adds information from an xrootd server's health status map to its outgoing advertisement so that the Director may eventually use this information in server filtering.
2. It spins off a routine that lets caches/origins scrape their own prometheus endpoint for information reported from the throttle plugin. When the result of that scrape indicates the server has more active IO operations than the configured limit, the server puts itself into a "degraded" health state.

This does _not_ implement any special logic Director side to use that information.


jhiemstrawisc:
@patrickbrophy I'm hoping you can take a look at the prometheus bits in this. In particular, I wasn't totally sure which metric I actually needed. I know the config I compare the scraped value against is fed to the throttle plugin, which helped me narrow things down to those metrics. I couldn't find documentation to understand the differences between the options I had there, however.

director/director.go
```diff
@@ -1277,6 +1277,7 @@ func finishRegisterServeAd(engineCtx context.Context, ctx *gin.Context, adV2 *se
 		RequiredFeatures:    adV2.RequiredFeatures,
 		IOLoad:              0.0, // Explicitly set to 0. The sort algorithm takes 0.0 as unknown load
 		Downtimes:           adV2.Downtimes,
+		Status:              adV2.Status,
```
h2zh, line 1280: Now the new `Status` field is stored in Director's in-memory `serverAds`. Next, modify the functions in `listServers` to pass `Status` all the way along to the json in `/api/v1.0/director_ui/servers` endpoint. 

Note that there is another similar field named `healthStatus` in that endpoint, which keeps track of director-based health tests it created. It serves a different purpose comparing to the new `Status` field here, so I think they should co-exist.

You could look at the logics of `healthTestUtils`. It has a different set of enumerations.
```
const (
	HealthStatusDisabled HealthTestStatus = "Health Test Disabled"
	HealthStatusUnknown  HealthTestStatus = "Unknown"
	HealthStatusInit     HealthTestStatus = "Initializing"
	HealthStatusOK       HealthTestStatus = "OK"
	HealthStatusError    HealthTestStatus = "Error"
)
```

server_utils/server_utils.go
```diff
@@ -511,3 +512,74 @@ func SetFedTok(ctx context.Context, server server_structs.XRootDServer, tok stri
...
+		}
+
+		if concLimit <= 0 {
+			return errors.Errorf("invalid config value: %s is %d. Must be greater than 0.", paramName, concLimit)
+		}
+
+		// TODO: Do we need to make this configurable?
+		concQuery := `avg_over_time(xrootd_server_io_active[1m])`
+
+		ticker := time.NewTicker(15 * time.Second)
```
h2zh, line 542: Why 1m and 15s respectively?

server_utils/server_utils.go
```diff
@@ -511,3 +512,74 @@ func SetFedTok(ctx context.Context, server server_structs.XRootDServer, tok stri
...
+		if sType == server_structs.CacheType {
+			concLimit = param.Cache_Concurrency.GetInt()
+			paramName = param.Cache_Concurrency.GetName()
+		} else if sType == server_structs.OriginType {
+			concLimit = param.Origin_Concurrency.GetInt()
+			paramName = param.Origin_Concurrency.GetName()
+		}
+
+		if concLimit <= 0 {
+			return errors.Errorf("invalid config value: %s is %d. Must be greater than 0.", paramName, concLimit)
```
h2zh, line 536: This is duplicated with the check before calling LaunchConcurrencyMonitoring

server_utils/server_utils.go
```diff
@@ -511,3 +512,74 @@ func SetFedTok(ctx context.Context, server server_structs.XRootDServer, tok stri
...
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-ticker.C:
+				log.Debug("Starting a new concurrency monitoring cycle")
+
+				qResult, err := QueryMyPrometheus(ctx, concQuery, true, sType.String())
+				if err != nil {
+					log.Errorf("Failed to query prometheus for current xrootd concurrency value; query: %s; error: %v", concQuery, err)
```
h2zh, line 553: When err is not null here, qResult is likely to be `null`. `qResult.ResultType` below will cause a panic.

Maybe add a `continue` to skip this when error occurs?

server_utils/server_utils.go
```diff
@@ -511,3 +512,74 @@ func SetFedTok(ctx context.Context, server server_structs.XRootDServer, tok stri
...
+							log.Errorf("Could not parse Prometheus-supplied xrootd concurrency value %q as float: %v\n", pair.Value, err)
+							continue
+						}
+
+						log.Tracef("Average 1m IO concurrency value from Prometheus is %.2f", avgConc)
+						if avgConc >= float64(concLimit) {
+							log.Debugf("Putting IO Concurrency health status into degraded state; the average 1m concurrency value of %f is higher than the configured limit %d", avgConc, concLimit)
+							metrics.SetComponentHealthStatus(metrics.OriginCache_IOConcurrency, metrics.StatusDegraded, "The server is currently experiencing more than its configured IO concurrency limit, performance may degraded")
+						} else {
+							log.Debugln("Putting IO Concurrency health status into OK state")
```
h2zh, line 573: It'd be nice to show the value of `avgConc` in the debug log here, in case it is close to `concLimit`

h2zh (changes requested):

server_utils/prom_query.go
```diff
@@ -74,7 +74,7 @@ type (
 // where the only arg is the query to execute, without "?query="
 //
 // Example: queryPromtheus("up") // Get metric of the running Prometheus instances
-func queryPromtheus(ctx context.Context, query string, withToken bool) (promParsed promQLParsed, err error) {
+func QueryMyPrometheus(ctx context.Context, query string, withToken bool, identity string) (promParsed promQLParsed, err error) {
```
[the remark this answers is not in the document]
patrickbrophy: This is not entirely true. You could disable `Monitoring.PromQLAuthorization`. Also it should be noted that the doc that was linked refers to how scraping is done in Prometheus, not querying.

director/director_api.go
```diff
@@ -307,7 +308,7 @@ func LaunchServerIOQuery(ctx context.Context, egrp *errgroup.Group) {
 				// Query all the servers and filter them out later
 				// We are interested in the derivative/rate of the total server IO over the past 5min
 				query := `rate(xrootd_server_io_total{job="origin_cache_servers"}[5m])`
-				queryResult, err := queryPromtheus(ddlCtx, query, true)
+				queryResult, err := server_utils.QueryMyPrometheus(ddlCtx, query, true, "director")
```
patrickbrophy, line 311: I'm not a huge fan of hardcoding strings, maybe try using the `server_structs.ServerType.String()`

h2zh: Just a reminder that `server_structs.ServerType.String()` get "**D**irector" with capitalized first character, which may need to pay attention to.

server_utils/prom_query.go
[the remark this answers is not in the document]
jhiemstrawisc: I'll change this to remove the `withToken` bool in the signature, and instead grab the value of `Monitoring.PromQLAuthorization` inside the function to determine whether a token needs to be paired with the request. Thank you both for the suggestions!

server_utils/server_utils.go
jhiemstrawisc: This was intentional -- since `LaunchConcurrencyMonitoring` is an exported function, I'm operating under the assumption that it _could_ be called elsewhere in the future. The outer conditional triggers whether the function is called in the first place and does not generate an error based on the configured value, while the inner conditional represents an error condition.

server_utils/server_utils.go
patrickbrophy: Your reasoning makes sense to me. I also don’t think making these timings configurable is the right approach, admins likely won’t be interested in tweaking them, especially since this query doesn’t have a significant performance impact.

h2zh (approved):
LGTM

PR #2342, opened 2025-05-22
Bump containerized xrdcl-pelican to v1.2.3

jhiemstrawisc:
This patch comes with additional use-after-free fixes for xrdcl-pelican in caches. Note that 1.2.2 is skipped because it was DOA.


turetske (approved):
Tentatively approving assuming the tests and builds pass

PR #2344, opened 2025-05-22
Transfer response errors

turetske:
First attempt at 404 transfer response error

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -482,6 +482,7 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
 			resultAd := classads.NewClassAd()
 			// Set our DeveloperData:
 			developerData := make(map[string]interface{})
+			var transferErrorData []interface{}
 			developerData["PelicanClientVersion"] = config.GetVersion()
 			developerData["Attempts"] = len(result.Attempts)
 			for _, attempt := range result.Attempts {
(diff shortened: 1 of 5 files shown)


bbockelm:
I'm awfully reluctant to change the `Link` header given that's a RFC standard and this seems to be a non-standard extension... what's it used for?

cmd/plugin.go
```diff
@@ -770,6 +774,39 @@ func writeTransferErrorMessage(currentError string, transferUrl string) (errMsg
...
+	developerData := make(map[string]interface{})
+	errMsg := err.Error()
+
+	if errors.Is(err, &client.SlowTransferError{}) {
+		developerData["PelicanErrorCode"] = 6002 // From error_codes.yaml
+		developerData["Retryable"] = true        // SlowTransferError is always retryable
+		developerData["ErrorMessage"] = "Slow transfer"
+		developerData["ErrorType"] = "Transfer.SlowTransfer"
+		transferError["ErrorType"] = "Transfer"
+	} else if strings.Contains(errMsg, "server returned 404 Not Found") {
```
patrickbrophy, line 788: From recent discussions about 404 errors, it appears that there’s enough ambiguity surrounding them that assigning a distinct error type, like we do in other cases, isn’t sufficient.

cmd/plugin_test.go
```diff
@@ -1276,3 +1277,328 @@ func TestWriteTransferErrorMessage(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Second server that returns 404
+	secondServer := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.NotFound(w, r)
+	}))
+	defer secondServer.Close()
+
+	// First server returns Link header pointing to second server
+	firstServer := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
patrickbrophy, line 1298: For clarity, can you rename this server to `directorServer` like the other tests?

PR #2348, opened 2025-05-23
Inform client which token it selected

h2zh:
Client should know about which token it selects for a request. This would be helpful for debugging.


patrickbrophy (approved):
LGTM

Issue #2346, opened 2025-05-22
Shorten message when attempting to "get" without a destination


aowen-uwmad:
Pelican Service:
- [x] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

**Is your feature request related to a problem? Please describe.**
For whatever reason, my muscle memory tries to use `pelican object get` the same way as a `wget` command. That means I don't automatically include the destination at the end of the command like I'm supposed to. I'm fine with that behavior, but the error message that is generated includes the usage help text. It takes time to parse that and realize that there was an error and what the error was.

For example,

```terminal
$ pelican object get osdf:///pelicanplatform/test/hello-world.txt
ERROR[2025-05-22T20:45:21Z] No Source or Destination                     
Get a file from a Pelican federation

Usage:
  pelican object get {source ...} {destination} [flags]

Flags:
  -c, --cache string    Cache to use
      --caches string   A JSON file containing the list of caches
  -h, --help            help for get
  -r, --recursive       Recursively download a collection.  Forces methods to only be http to get the freshest collection contents
  -t, --token string    Token file to use for transfer

Global Flags:
      --config string       config file (default is $HOME/.config/pelican/pelican.yaml)
  -d, --debug               Enable debug logs
  -f, --federation string   Pelican federation to utilize
      --json                output results in JSON format
  -L, --log string          Specified log output file
      --version             Print the version and exit
```

**Describe the solution you'd like**
Shorten the output for this error type to only a couple of lines.

Propose something like the following:

```terminal
$ pelican object get osdf:///pelicanplatform/test/hello-world.txt
ERROR[2025-05-22T20:45:21Z] No Source or Destination  
For usage, run 'pelican object get --help'.
```

**Additional context**
In principle, could change it so that `destination` is optional. 
I don't know that it's required though. 
Perhaps as we get more users encountering this issue we could revisit.

PR #2351, opened 2025-05-27
Shorten message when attempting using `get` without a destination

patrickbrophy:
This PR address issue #2346. This PR removes the large help message that obscures that an error occurred.


jhiemstrawisc (approved):
Works as advertised. Thanks, @patrickbrophy!

PR #2355, opened 2025-05-27
Implement new logging level configuration parameters.

patrickbrophy:
This PR introduces two new logging configuration parameters, `Logging.Client.Level` and `Logging.Server.Level`. These new parameters are meant to separate the existing logging level into separate parameters. However if either of these are unset, it will fallback to `Logging.Level`. This PR also sets the default server logging level (Logging.Server.Level) to `info` and sets the default client logging level to `error`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -783,14 +783,35 @@ func setLoggingInternal() error {
 	if param.Debug.GetBool() {
 		SetLogging(log.DebugLevel)
 		log.Warnf("Debug is set as a flag or in config, this will override anything set for '%s' within your configuration", param.Logging_Level.GetName())
-	} else {
-		logLevel := param.Logging_Level.GetString()
-		level, err := log.ParseLevel(logLevel)
-		if err != nil {
-			return errors.Wrapf(err, "failed to parse value of config param %s", param.Logging_Level.GetString())
-		}
-		SetLogging(level)
+		return nil
 	}
+
+	var logLevelParam param.StringParam
+	var logLevel string
+
+	if len(GetEnabledServerString(true)) > 0 { // Server context
+		if viper.IsSet(param.Logging_Server_Level.GetName()) {
+			logLevelParam = param.Logging_Server_Level
+			logLevel = logLevelParam.GetString()
+		} else {
+			logLevelParam = param.Logging_Level
+			logLevel = logLevelParam.GetString()
+		}
+	} else { // Client context
+		if viper.IsSet(param.Logging_Client_Level.GetName()) {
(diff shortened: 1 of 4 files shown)


brianhlin:
Does this mean that we have the following (ignoring the `Debug` config) precedence order for configuration?

1. `Logging.<service>.<subsystem>`, e.g. `Logging.Origin.Scitokens`
2. `Logging.Server.Level`
3. `Logging.Level`

I spoke with Mat and William and we agreed that it'd be more intuitive if the config aligned on the specific services like so:

1. `Logging.<service>.<subsystem>`, e.g. `Logging.Origin.Scitokens`
2. `Logging.<service>.Level`, e.g. `Logging.Origin.Level`
3. Logging.Level

patrickbrophy:
Ok sounds good. I just implemented the following logging hierarchy (from most fine grained to least):
1. Logging.{Origin,Cache}.Scitokens
2. Logging.{Origin,Cache,Director,Registry}.Level
3. Logging.Level

To be clear on the second level, the enabled service only takes precedence, so Logging.Registry.Level won't override Logging.Origin.Level.

Essentially each level will fallback to the next level if it is unset.

brianhlin:
@jhiemstrawisc I see your point, especially about the rabbit hole, but I think it's the right direction unless you want to somehow differentiate between operator vs end-user client commands (and that seems really hard). One of the big annoyances in my day-to-day operation of services is stuff like this:

```
[blin@ap40 ~]$ pelican --config /etc/pelican/osdf-origin.yaml config get logging.loglocation
logging.loglocation: "/var/log/pelican/osdf-origin.log"
INFO[2025-06-12T16:34:36-05:00] Configuration constructed according to directory-scoped lexicographical file order from the following directories: /usr/share/pelican/config.d, /etc/pelican/config.d 
WARNING[2025-06-12T16:34:36-05:00] The configuration key 'Server.TLSCertificate' is deprecated. The value from its replacement 'Server.TLSCertificateChain' will be used instead, and the value of the deprecated configuration key 'Server.TLSCertificate' will be ignored. 
```

That's a lot of noise that makes it hard to find the one line of information that I'm looking for. And if you separate out the config files to solve this, the operator will *always* have to specify `pelican --config <configpath> config` to make that verb even remotely useful.

> Maybe instead the solution is to provide separate default/fallback paths for servers vs clients? I don't know.

This is what HTCondor does: `SEC_CLIENT_AUTHENTICATION_METHODS` specifies the authN methods for the client but if that's not set, it just uses what's set in `SEC_DEFAULT_AUTHENTICATION_METHODS`

PR #2370, opened 2025-05-30
Filter OSS Packets

patrickbrophy:
This PR adds an allow list to the OSS packet handler so that we can filter packets that are disallowed from being processed into metrics. If we want to add another event we can just update the allow list.

### DIFF
--- a/metrics/xrootd_metrics_test.go
+++ b/metrics/xrootd_metrics_test.go
@@ -805,8 +805,14 @@ func TestHandlePacket(t *testing.T) {
 	})
 }
 
-// setupTestRegistry creates a new registry and registers the OSS metrics for testing
-func setupTestRegistry() prometheus.TransactionalGatherer {
+// setupTestPrometheusRegistry creates a new registry and registers the OSS metrics for testing
+func setupTestPrometheusRegistry(t testing.TB) prometheus.TransactionalGatherer {
+	oldCounter := OssReadsCounter
+
+	t.Cleanup(func() {
+		OssReadsCounter = oldCounter
+		lastOssStats = OSSStatsGs{}
+	})
 	reg := prometheus.NewRegistry()
 
 	// Create a new counter for each test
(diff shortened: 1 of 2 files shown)


metrics/xrootd_metrics_test.go
```diff
@@ -804,6 +804,20 @@ func TestHandlePacket(t *testing.T) {
...
+
+		badPacket := []byte(`{"event":"bad_packet","reads":151577,"writes":0,"stats":152346,"pgreads":0,"pgwrites":0,"readvs":0,"readv_segs":0,"dirlists":0,"dirlist_ents":0,"truncates":0,"unlinks":0,"chmods":0,"opens":10754,"renames":0,"slow_reads":521,"slow_writes":0,"slow_stats":0,"slow_pgreads":0,"slow_pgwrites":0,"slow_readvs":0,"slow_readv_segs":0,"slow_dirlists":0,"slow_dirlist_ents":0,"slow_truncates":0,"slow_unlinks":0,"slow_chmods":0,"slow_opens":0,"slow_renames":0,"open_t":854.9008,"read_t":20701.8367,"readv_t":0.0000,"pgread_t":0.0000,"write_t":0.0000,"pgwrite_t":0.0000,"dirlist_t":0.0000,"stat_t":68.3124,"truncate_t":0.0000,"unlink_t":0.0000,"rename_t":0.0000,"chmod_t":0.0000,"slow_open_t":0.0000,"slow_read_t":810.3797,"slow_readv_t":0.0000,"slow_pgread_t":0.0000,"slow_write_t":0.0000,"slow_pgwrite_t":0.0000,"slow_dirlist_t":0.0000,"slow_stat_t":0.0000,"slow_truncate_t":0.0000,"slow_unlink_t":0.0000,"slow_rename_t":0.0000,"slow_chmod_t":0.0000}`)
```
bbockelm, line 811: Make sure you have a case where:
- multiple events are in the same packet.
- there are no parseable events in a packet.
- only the first event is parseable.

metrics/xrootd_metrics_test.go
```diff
@@ -804,6 +804,20 @@ func TestHandlePacket(t *testing.T) {
...
+
+		err := handleOSSPacket([][]byte{goodPacket})
+		assert.NoError(t, err, "Error handling a good OSS packet")
+
+		err = handleOSSPacket([][]byte{badPacket})
+		assert.Error(t, err, "Error handling a good OSS packet")
```
bbockelm, line 817: Copy/paste error?

jhiemstrawisc:
Using super powers to merge this -- I believe Patrick has addressed BrianB's comments.

Issue #2371, opened 2025-06-04
disableHTTPProxy isn't really being considered


turetske:
It seems we are always turning off the HTTP or HTTPS Proxy regardless of whether disableHTTPProxy is set.

This should be fixed to only be disabled when the parameter is set.

We should also determined if we want a separate disableHTTPSProxy parameter or if the same parameter should handle all proxies (and then potentially be renamed?)

Issue #2248, opened 2025-04-24
"pelican object put" fails if the HTTPS_PROXY env var is set


brianhlin:
Pelican Version:
7.15.1

Pelican Service:
- [X] Client
- [X] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

**Describe the bug**
Pelican fails to `pelican object put` if `HTTPS_PROXY` is set:

```
[ERRORESC[0m[2025-04-23T17:34:40-04:00] Error with PUT: Put "https://ap21.uc.osg-htc.org:8443/ospool/ap21/data/blin/hello-writeback.txt": Forbidden 
ESC[37mDEBUGESC[0m[2025-04-23T17:34:40-04:00] File closed                                  
ESC[31mERRORESC[0m[2025-04-23T17:34:40-04:00] Unexpected error when performing upload: Put "https://ap21.uc.osg-htc.org:8443/ospool/ap21/data/blin/hello-writeback.txt": Forbidden 
```

On the other hand, `pelican object get` appears to work fine

brianhlin:
Setting `PELICAN_CLIENT_DISABLEHTTPPROXY=true` does not appear to help

brianaydemir:
> Setting `PELICAN_CLIENT_DISABLEHTTPPROXY=true` does not appear to help

@brianhlin This setting appears to apply only when `http_proxy` is set in the environment:

https://github.com/PelicanPlatform/pelican/blob/81c64af59635a16272126100aa88564797171bea/client/handle_http.go#L426-L435

There's probably some historical context that I'm missing here.

PR #2372, opened 2025-06-04
Updated runPut to mimic downloadHttp's proxy logic

turetske:
This is fixing the issue described here: #2248 

It mimics how downloadHTTP deals with the proxy.
 
This is just meant to be a fix to make sure both functions are behaving the same way, but I think there is a larger issue involving the correctness of the code that I created a different issue for in #2371 

Test this by running a put using the built binary on the ospool with 

Requirements = GLIDEIN_Site == "UA Little Rock ITS"


patrickbrophy (approved):
LGTM

PR #2380, opened 2025-06-09
Add `pelican key create` command

### LINKED ISSUE #2192: New Pelican CLI command: pelican key create
Brought by @aowen-uwmad in the Pelican-RCF meeting 04/08, we need a new `pelican key create` command to replace `pelican generate keygen`.

h2zh:
What this new command does is identical to `pelican generate keygen` command - create a pair of private/public keys. The only difference is the new private key file will be named as "issuer.pem" instead of "issuer.jwk". The new extension reflects the real format (PEM, Privacy-Enhanced Mail) of this private key file and is required by Pelican to automatically detect it in the `IssuerKeysDirectory`. 

There is no backward compatibility issue for this change because it won't affect the existing "issuer.jwk" file (which still work via the `IssuerKey` config param). I also updated the Pelican docs to let facilitators and users know this change.

`pelican generate keygen` is planned to retire it in the future. For now, it remains available for backward compatibility.

Closes #2192

### DIFF
--- a/cmd/key.go
+++ b/cmd/key.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	keyCmd = &cobra.Command{
+		Use:   "key",
+		Short: "Manage Pelican issuer keys",
+	}
+
+	// `pelican key create` command aims to replace `pelican generate keygen` command in the future.
+	// For now, they are both available and do the same thing to maintain backward compatibility.
+	keyCreateCmd = &cobra.Command{
+		Use:   "create",
+		Short: "Generate a public-private key-pair for Pelican OIDC issuer",
+		Long: `Generate a public-private key-pair for a Pelican server.
+The private key is a ECDSA key with P256 curve. The corresponding public key
+is a JWKS in JSON. The public key follows OIDC protocol and can be used
+for JWT signature verification.`,
+		RunE: keygenMain,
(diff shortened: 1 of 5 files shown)


cmd/key.go
```diff
@@ -0,0 +1,35 @@
...
+		// which is already tested in `pelican/cmd/generate_keygen_test.go`.
+		SilenceUsage: true,
+	}
+)
+
+func init() {
+	keyCmd.AddCommand(keyCreateCmd)
+
+	// Attach flags to the `create` sub-command
+	keyCreateCmd.Flags().StringVar(&privateKeyPath, "private-key", "", "The path to the generate private key file. Default: ./issuer.pem")
```
jhiemstrawisc, line 33: Three notes here:
1. I'd say "The file path where the generated private key will be saved"
2. Instead of `issuer.pem`, may I propose `private-key.pem`? My reasons for this are that a) I'd prefer the default name contains the word "private" to be super explicit to the user that this is a secret and b) this is just a private pem and it could be used for stuff other than setting up an Origin's issuer (which we do a poor job of defining anyway).
3. Can you also move the "Default" out of the help string and instead using the command's default value field instead? For example, this:
```
keyCreateCmd.Flags().StringVar(&privateKeyPath, "private-key", "./private-key.pem", "The file path where the generated private key will be saved")
```
will already generate this help text _and_ set the default:
```
--private-key string   The file path where the generated private key will be saved (default "./issuer.pem")
```

PR #2381, opened 2025-06-09
Relax log directory permissions

h2zh:
Relax log directory permissions.

This can unblock unpriv users to use Pelican config tool.


matyasselmeci (approved):
LGTM.

PR #2383, opened 2025-06-09
Memory jitter cleanup

bbockelm:
This PR has "various and sundry" cleanups around fixing up log messages, memory churn, and VS Code linter (which appears to be a bit more picky than the GHA?).

The unifying theme is various small changes that are prerequisites to a larger set of fixes for the "memory stress test" (which I'd like to make more reliable!).


jhiemstrawisc (approved):
LGTM. Will merge once post-conflict tests have passed.

PR #2385, opened 2025-06-09
Replace esoteric error message with one more meaningful for users

bbockelm:
The sTREAM timeout error message means the upstream origin was unresponsive.


patrickbrophy (approved):
Looks good, happy to have better error messages.

LGTM

patrickbrophy (changes requested):
Looks like there is a client test failing, appears to be caused by strings not lining up in the failing test.

PR #2387, opened 2025-06-10
Fix debug logging of the xrootd framework itself

bbockelm:
It looks like the original author was under the impression `debug` implied the other message levels. It doesn't - you have to specify all the ones at the higher level _plus_ the verbose debug messages.


patrickbrophy (approved):
LGTM

PR #2395, opened 2025-06-11
Update detailed server on server update

CannonLock:
- Only do so if the dropdown is open and the detailedServer has been previously fetched

@h2zh There is two levels of detail for a server, detailed and not. When you click the dropdown I fetch the detailed view and use that to display more granular details in the dropdown. 

The issue was that I preferred that detailed data over the other data in the dropdown, but unlike the normal data that updates itself the detailed never gets refetched. 

To fix this I added an effect that watches for server change and if it happens well the dropdown is populated and open, I refetch the detailed data.

### DIFF
--- a/web_ui/frontend/app/director/components/DirectorCard.tsx
+++ b/web_ui/frontend/app/director/components/DirectorCard.tsx
@@ -32,7 +32,7 @@ export const DirectorCard = ({ server, authenticated }: DirectorCardProps) => {
     if (detailedServer !== undefined && dropdownOpen) {
       updateDetailedServer(server.name, setDetailedServer, dispatch);
     }
-  }, [server, setDetailedServer, dispatch]);
+  }, [server, setDetailedServer, dispatch, dropdownOpen, detailedServer]);
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


web_ui/frontend/app/director/components/DirectorCard.tsx
```diff
@@ -46,15 +56,10 @@ export const DirectorCard = ({ server, authenticated }: DirectorCardProps) => {
...
-                async () => {
-                  const response = await getDirectorServer(server.name);
-                  setDetailedServer(await response.json());
-                },
-                'Failed to fetch server details',
-                dispatch
-              );
+
+            // Update detailed server when the dropdown is opened
+            if (dropdownOpen) {
```
h2zh, line 61: The state update `setDropdownOpen(!dropdownOpen);` is asynchronous and doesn't happen immediately. The `if (dropdownOpen)` condition does check the old value before the state update takes effect.

h2zh (changes requested):
Clever design to minimize the rendering speed 💡 One potential problem as below

Issue #2398, opened 2025-06-11
Director's downtime calendar is difficult to read


brianaydemir:
As of late, every time I visit https://osdf-director.osg-htc.org/view/director/downtime/, I'm greeted by alternating rows of blue and red:

<img width="816" alt="Image" src="https://github.com/user-attachments/assets/ed149ebc-7245-4d46-86ac-b3ea6ade2d51" />

The first time I visited this page, I actually didn't realize that this is supposed to be a calendar.

Some suggestions to consider:

- Don't use a background color anywhere in the "square" for each day, but do give the "squares" a defined border.
- Label any number that isn't the calendar day.
- Obey my locale setting (the first day of the week is Sunday, not Monday).
- Don't use any special colors for Saturday and Sunday (save the coloring for values of interest).

PR #2409, opened 2025-06-16
Downtime Upgrades

### LINKED ISSUE #2254: Downtime Improvements
- [ ] Remove the '/origins' and '/caches' prefix
- [ ] Convert the Registry ServerName box into a autocomplete field
- [ ] Filter out nonsense namespaces in the Registry ServerName field ( For instance ones that have been denied )

### LINKED ISSUE #2373: Various downtime web UI nitpicks
Pelican Service:
- [ ] Client
- [ ] Plugin
- [X] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

1. When creating a downtime via the `Service Downtime` page, it's very difficult to find the service that you want to put into downtime. Filtering and/or alphabetical ordering would be helpful
2. Start / end time time zones are unclear
3. When creating a downtime via the `Service Downtime` page, the start time seems to autopopulate to the beginning of the month. OTOH, when creating a downtime for a specific service via the main page, the start time autopopulates to the current time. I think it would make sense to unify on the latter.

CannonLock:
- Sort, add icons, clean prefix in server options on registry
- Add note about current timezone on time entry
- Always have start and end time autopopulate to current time
- Filter out not approved servers
- Update styles to prevent number overload and add more meaning to Tiles

Closes #2373
Closes #2254
Closes https://github.com/PelicanPlatform/pelican/issues/2398

### DIFF
--- a/web_ui/frontend/components/Downtime/Registry/ServerUnknownDowntimeForm.tsx
+++ b/web_ui/frontend/components/Downtime/Registry/ServerUnknownDowntimeForm.tsx
@@ -73,7 +73,7 @@ const ServerUnknownDowntimeForm = ({
   // Get the cache and origin prefixes to key the downtimes for the director
   const { data: namespaces } = useApiSWR<RegistryNamespace[]>(
     'Could not fetch Origins and Caches to populate downtime form',
-    'getNamespaces',
+    'getNamespaces-TODO-update-this-key-to-share-cache',
     getNamespaces
   );
   const servers = useMemo(() => {
--- a/web_ui/frontend/components/Downtime/Registry/ServerUnknownDowntimeForm.tsx
+++ b/web_ui/frontend/components/Downtime/Registry/ServerUnknownDowntimeForm.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import {
+  Autocomplete,
   Box,
   Button,
   Checkbox,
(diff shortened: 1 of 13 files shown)


web_ui/frontend/components/Downtime/Registry/ServerUnknownDowntimeForm.tsx
```diff
@@ -71,48 +76,69 @@ const ServerUnknownDowntimeForm = ({
     'getNamespaces',
     getNamespaces
   );
-  const prefixes = useMemo(
-    () => namespacesToOriginAndCachePrefixes(namespaces || []),
-    [namespaces]
-  );
+  const servers = useMemo(() => {
```
h2zh, line 79: Don't process until data is actually loaded
```suggestion
  const servers = useMemo(() => {
  if (isLoading || !namespaces || namespaces.length === 0) {
      return [];
    }
```

h2zh (changes requested):
Great enhancement! This would save admin's time and energy a lot.

When I click on "create downtime" button on the Registry, the web page crashes with error: `Application error: a client-side exception has occurred while loading localhost (see the browser console for more information).`
Console: `Uncaught TypeError: Cannot read properties of undefined (reading 'startsWith')`
I spent some time to investigate and have two solutions (short-circuit evaluation OR useSWR's `isLoading`) as below.

I'll allocate other comments in another review so this review can just focus on one problem.

CannonLock:
@h2zh I temp patched the bug and created a new ticket to give the issue full consideration.

https://github.com/PelicanPlatform/pelican/issues/2449

h2zh (approved):
LGTM after a rebase!

Issue #2359, opened 2025-05-29
Provide tracing option for S3 origins


bbockelm:
The `s3.trace` XRootD configuration directive is essential to debugging what's going on with the S3 backend.

We should have this exposed as part of the origin logging configuration.  Since S3 is not going to be used everywhere, instead of having a new configuration option, `Logging.Origin.S3`, let's use the fact the S3 is the OSS plugin and hook it up to `Logging.Origin.Oss`.

The translation between the log level names should be straightforward.  Note there's currently a bug where the log level is quite literal -- e.g., turning on `debug` only enables `debug` messages but not `info` or `warning` or `error`.  Let's workaround that in the Pelican layer (it's also fixed in `main` for upstream).

Example xrootd config:

```
s3.trace debug info warning error
```

Note there is a `dump` level which is incredibly verbose - every byte of traffic is recorded in the log multiple times.  I suppose that can map to `trace`.

PR #2410, opened 2025-06-16
Enable S3 tracing option via Logging.Origin.Oss configuration

patrickbrophy:
This PR addresses #2359. This PR enables the `s3.trace` configuration when the S3 backend is enabled. This doesn't introduce any new Pelican configuration parameters and the verbosity of the logging is handled through the `Logging.Origin.Oss`.


bbockelm (approved):
LGTM.

PR #2411, opened 2025-06-16
Support multiple keys in string encrypt/decrypt functions

h2zh:
- Update string encrypt/decrypt functions with NaCl's box pkg
- Record the id of the key used for encryption in the output string
- Iterate all valid issuer key to decrypt

  - Enable support for key rotation and decryption of data encrypted with previous keys.
  - To achieve this goal, GetSecret and key pair derivation functions now accept a specific issuer key as input

- No backward compatibility issue because the `EncryptString` and `DecryptString` are used for globus collection only, which is never in the prod. cc @whwjiang: This PR may have impact on your globus development.
- Replace the standard errors package with github.com/pkg/errors

### DIFF
--- a/config/encrypted.go
+++ b/config/encrypted.go
@@ -32,6 +32,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
@@ -382,26 +383,19 @@ func SaveConfigContents_internal(config *OSDFConfig, forcePassword bool) error {
 // How we generate the secret:
 // Concatenate the byte array pelican with the DER form of the service's private key,
 // Take a hash, and use the hash's bytes as the secret.
-func GetSecret() (string, error) {
-	// Use issuer private key as the source to generate the secret
-	privateKey, err := GetIssuerPrivateJWK()
-	if err != nil {
-		return "", err
-	}
-
+func GetSecret(issuerKey jwk.Key) (string, error) {
 	// Extract the underlying ECDSA private key in native Go crypto key type
 	var rawKey interface{}
-	if err := privateKey.Raw(&rawKey); err != nil {
+	if err := issuerKey.Raw(&rawKey); err != nil {
 		return "", err
 	}
 
 	derPrivateKey, err := x509.MarshalPKCS8PrivateKey(rawKey)
-
 	if err != nil {
 		return "", err
 	}
-	byteArray := []byte("pelican")
 
+	byteArray := []byte("pelican")
 	concatenated := append(byteArray, derPrivateKey...)
 
 	hash := sha256.Sum256(concatenated)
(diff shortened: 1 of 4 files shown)


config/encrypted.go
```diff
@@ -411,55 +404,123 @@ func GetSecret() (string, error) {
 	return secret, nil
 }
 
+func getEncryptionKeyPair(issuerKey jwk.Key) (privateKey, publicKey *[32]byte, err error) {
```
bbockelm, line 407: Please add a doc string for the function.

config/encrypted.go
```diff
@@ -411,55 +404,123 @@ func GetSecret() (string, error) {
...
+	if _, err := io.ReadFull(rand.Reader, nonce[:]); err != nil {
+		return "", err
+	}
+
+	// Encrypt the message
+	message := []byte(stringToEncrypt)
+	encrypted := box.Seal(nil, message, &nonce, publicKey, privateKey)
+
+	// Get current issuer key
+	currentIssuerKey, err := GetIssuerPrivateJWK()
```
bbockelm, line 449: Don't we already get the issuer key above?

config/encrypted.go
```diff
@@ -411,55 +404,123 @@ func GetSecret() (string, error) {
...
+
+	if len(nonceBytes) != 24 {
+		return "", errors.New("invalid nonce length")
 	}
+	var nonce [24]byte
+	copy(nonce[:], nonceBytes)
+
+	var decryptedMsg []byte
+	var decrypted bool
+	for _, issuerKey := range allIssuerKeys {
```
bbockelm, line 501: Write a helper function to try decrypting for a private key.

Try the current key first (most likely to succeed) then iterate through all others.

config/encrypted.go
```diff
@@ -411,55 +404,123 @@ func GetSecret() (string, error) {
...
 
-	if len(ciphertext) < aes.BlockSize {
-		return "", fmt.Errorf("ciphertext too short")
+		// Decrypt the message
+		msg, ok := box.Open(nil, encryptedMsg, &nonce, publicKey, privateKey)
+		if ok {
+			decryptedMsg = msg
+			decrypted = true
+			// Inform the user when the key used in encryption does not match the current issuer key
+			if keyID != currentIssuerKey.KeyID() {
```
bbockelm, line 514: This seems strange.  Test to see if you have the correct key ID first (that's the point of storing it) then try the decryption.

bbockelm (commented):
A few hopefully helpful comments while it's still in draft.

config/encrypted.go
h2zh: Thanks for the suggestion! I feel the same when revisiting this snippet after a few days.

config/encrypted.go
h2zh: Since we record keyID in the string as `$KEYID.$NONCE.$MESSAGE`, we can directly find the key that did the encryption before in all valid keys.
```
// Find the right key to decrypt the message
	allIssuerKeys := GetIssuerPrivateKeys()

	encryptionKey, found := allIssuerKeys[keyID]
	if !found {
		return "", "", errors.Errorf("the key used for encryption (with ID %s) is not found", keyID)
	}
```

jhiemstrawisc:
@h2zh, can you either re-request a review from @bbockelm or ask another Pelican dev to take it over? Since this PR's milestone has already been bumped twice, let's focus on finishing it up for `v7.20`.

PR #2417, opened 2025-06-18
Correctly scope permission modification of LotHome directory

jhiemstrawisc:


jhiemstrawisc:
Given the current time crunch, I'm using super powers to merge this while I work through a few Lotman items.

PR #2428, opened 2025-06-19
Improvements for HTTP transports and stat tests

bbockelm:
This PR overhauls the stat stress test to make it more reliable.  This includes:
- Using a common client throughout the code base.  As a side-effect, I noticed we were editing the global transport object from multiple threads simultaneously when proxies are present which is probably ... bad.
- Ensure we correctly handle HTTP responses.  For a TCP connection to be reused, we *must* fully read the response and close it.  One issue with the stat test was its churning through TCP connections (and associated data structures) at an incredible rate.
- Having some environment-variable based tuning of the test, allowing one to run it for longer and dump heap profiles.

Honestly, if we're still seeing weirdness after this PR, I think I might just mark the test skippable and move on with life.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -806,7 +806,7 @@ func NewTransferEngine(ctx context.Context) (te *TransferEngine, err error) {
 	results := make(chan *clientTransferResults, 5)
 
 	// Start the URL cache to avoid repeated metadata queries
-	pelicanUrlCache := pelican_url.StartCache()
+	pelicanUrlCache := pelican_url.StartCache(ctx, egrp)
 
 	te = &TransferEngine{
 		ctx:             ctx,
@@ -956,7 +956,6 @@ func (te *TransferEngine) Shutdown() error {
 	te.Close()
 	<-te.closeDoneChan
 	te.ewmaTick.Stop()
-	te.pelicanUrlCache.Stop()
 	te.cancel()
 
 	err := te.egrp.Wait()
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -1577,7 +1577,7 @@ func (tc *TransferClient) CacheInfo(ctx context.Context, remoteUrl *url.URL, opt
 	cacheUrl := *sortedServers[0]
 	cacheUrl.Path = remoteUrl.Path
 
-	return objectCached(ctx, &cacheUrl, token, config.GetTransport())
+	return objectCached(ctx, &cacheUrl, token)
 }
 
 // Close the transfer client object
@@ -1924,7 +1924,6 @@ func sortAttempts(ctx context.Context, path string, attempts []transferAttemptDe
 		results = attempts
 		return
 	}
-	transport := config.GetTransport()
 	type checkResults struct {
 		idx  int
 		size uint64
@@ -1955,7 +1954,7 @@ func sortAttempts(ctx context.Context, path string, attempts []transferAttemptDe
 				return
 			}
 
-			if age, size, err := objectCached(ctx, tUrl, token, transport); err != nil {
+			if age, size, err := objectCached(ctx, tUrl, token); err != nil {
 				headChan <- checkResults{idx, 0, -1, err}
 				return
 			} else {
@@ -2408,8 +2407,7 @@ func fetchChecksum(ctx context.Context, types []ChecksumType, url *url.URL, toke
 		}
 		request.Header.Set("Want-Digest", val)
 	}
-	transport := config.GetTransport()
-	client := &http.Client{Transport: transport}
+	client := config.GetClient()
 	response, err := client.Do(request)
 	if err != nil {
 		return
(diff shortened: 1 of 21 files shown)


jhiemstrawisc:
Aha! Your first bullet point rang a bell for me -- I observed a panic a month or two ago related to two threads writing to some map associated with the `transport` object at the same time, but I never had time to track down where that was happening. I bet you're fixing that in this PR.

director/stat_stress_test.go
```diff
@@ -54,61 +114,149 @@ var (
...
+	//
+	// We make sure the warmup is for at least the size of the cache and 10%
+	// the length of the entire test duration.
+	isCanceled := false
+	warmupDuration := testDuration / 10
+	for idx < cacheSize || time.Since(start) < warmupDuration {
 		downloadURL := fmt.Sprintf("pelican://%s%s/stress/%v.txt", discoveryUrl.Host, fed.Exports[0].FederationPrefix, idx)
+		destName := filepath.Join(t.TempDir(), fmt.Sprintf("dest.%v.txt", idx))
+		src := filepath.Join(fed.Exports[0].StoragePrefix, fmt.Sprintf("stress/%v.txt", idx))
+		fmt.Println(src)
```
jhiemstrawisc, line 175: Do you mean to have this? If so, can we change it to a `test.Log` line?

director/stat_stress_test.go
```diff
@@ -54,61 +114,149 @@ var (
...
 
 	// Now, do enough work to fully evict and replace the cache's
 	// contents from the "warm up" stage. If we're on an unusually
 	// fast host, keep going until "enough" time has elapsed.
-	for idx < 2*cacheSize || time.Since(start) < 10*time.Second {
+	for idx < 2*cacheSize || time.Since(start) < testDuration {
 		downloadURL := fmt.Sprintf("pelican://%s%s/stress/%v.txt", discoveryUrl.Host, fed.Exports[0].FederationPrefix, idx)
+		destName := filepath.Join(t.TempDir(), fmt.Sprintf("dest.%v.txt", idx))
+		src := filepath.Join(fed.Exports[0].StoragePrefix, fmt.Sprintf("stress/%v.txt", idx))
+		fmt.Println(src)
```
jhiemstrawisc, line 221: Same question here.

director/stat_stress_test.go
```diff
@@ -54,61 +114,149 @@ var (
...
+		require.NoError(t, os.WriteFile(src, []byte("foo"), os.FileMode(0600)))
+		require.NoError(t, os.Chown(src, ui.Uid, ui.Gid))
 		grp.Go(func() error {
-			_, err := client.DoGet(fed.Ctx, downloadURL, dest, false)
-			assert.Error(t, err)
+			fmt.Println("Launched download URL", downloadURL, err)
```
jhiemstrawisc, line 225: And here.

launcher_utils/advertise.go
```diff
@@ -54,11 +54,14 @@ type directorResponse struct {
 
 func doAdvertise(ctx context.Context, servers []server_structs.XRootDServer) {
 	log.Debugf("About to advertise %d XRootD servers", len(servers))
+	start := time.Now()
 	err := Advertise(ctx, servers)
+	duration := time.Since(start)
 	if err != nil {
-		log.Warningln("XRootD server advertise failed:", err)
+		log.Warningf("XRootD server advertise failed (duration %v): %s", duration.String(), err)
```
jhiemstrawisc, line 61: I think the format specifiers might be backwards here (although I think this is probably fine, just not as idiomatic).

director/stat_stress_test.go
```diff
@@ -54,61 +114,149 @@ var (
 func TestStatMemory(t *testing.T) {
 	server_utils.ResetTestState()
 
+	// To allow for developer control over testing, we add two environment variables
+	// specific to this unit test:
+	// - PELICAN_STRESS_DURATION: Parsed as a time duration, this is the time of the
```
jhiemstrawisc, line 119: Not a change request, but a note for posterity -- these env vars will trigger a warning related to unrecognized Pelican configuration because the client init a few lines down matches all `PELICAN_` env vars and tries to marshal them into the config structs. I think this is fine for an internal unit test.

client/handle_http.go
```diff
@@ -2597,9 +2598,9 @@ func downloadHTTP(ctx context.Context, te *TransferEngine, callback TransferCall
 		// The host is ignored since we override the dial function; however, I find it useful
 		// in debug messages to see that this went to the local cache.
 		transferUrl.Host = "localhost"
+		client = &http.Client{Transport: transport}
```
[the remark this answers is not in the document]
bbockelm: This is a really unique situation - we replace the TCP-based transport with one that reads from a Unix socket.  I decided that there wasn't going to be much reuse so I didn't centralize it.

director/stat_stress_test.go
bbockelm: I think we can do without - I'll delete.

jhiemstrawisc:
What's your thought on the tests suddenly failing again?

PR #2431, opened 2025-06-20
Upgrade to xrdcl-pelican v1.3.1

bbockelm:
xrdcl-pelican splits pelican:// and https:// into two different plugins (the former relies on the latter); this includes corresponding changes to the plugin configuration generation.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -53,8 +53,11 @@ ARG NODEJS_VER=20
 #
 ARG LOTMAN_SRC_BUILD=false
 ARG LOTMAN_VER=0.0.4
-ARG XRDCL_PELICAN_SRC_BUILD=false
-ARG XRDCL_PELICAN_VER=1.2.3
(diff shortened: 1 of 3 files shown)


images/Dockerfile
```diff
@@ -53,8 +53,8 @@ ARG NODEJS_VER=20
 #
 ARG LOTMAN_SRC_BUILD=false
 ARG LOTMAN_VER=0.0.4
-ARG XRDCL_PELICAN_SRC_BUILD=false
-ARG XRDCL_PELICAN_VER=1.2.3
+ARG XRDCL_PELICAN_SRC_BUILD=true
+ARG XRDCL_PELICAN_VER=1.3.1
```
jhiemstrawisc, line 57: This version should also be updated in `github_scripts/osx_install.sh`. When you do that, can you add a note/warning comment above this section so developers know there's a dependency update required in both places?

jhiemstrawisc (approved):
LGTM

Issue #2360, opened 2025-05-29
Parse S3 file cache statistics and create Prometheus metrics


bbockelm:
As of b7518b8b322ce5fdf514d4940773df1e7a953dce in `xrootd-s3-file` main (I'll turn that into v0.3.0 momentarily), the S3 OSS backend will use the OSS's g-stream integration to send out monitoring statistics.  These, like the other OSS stats, are designed to easily integrate with Prometheus.

Here's a sample event run through `jq`:

```
{
  "event": "s3file_stats",
  "hit_b": 0,
  "miss_b": 0,
  "full_hit": 0,
  "part_hit": 0,
  "miss": 0,
  "bypass_b": 12265,
  "bypass": 6,
  "fetch_b": 0,
  "fetch": 0,
  "unused_b": 0,
  "prefetch_b": 0,
  "prefetch": 0,
  "errors": 0,
  "bypass_s": 1.27,
  "fetch_s": 0
}
```

A bit of explanation of the keys:
- `hit_b`: Bytes that were served to the client directly from the cache (a "cache hit")
- `miss_b`: Bytes that were served to the client that were a cache miss (not immediately available).
- `full_hit`: Count of read requests from the client that were completely served from the cache.
- `part_hit`: Count of read requests from the client that were partially served from the cache.
- `miss`: Count of read requests that were entirely a miss.
- `bypass_b`: Bytes that were read in "bypass mode", skipping the cache.  Typically, this is from read requests that are larger than the cache size.
- `bypass`: Count of read requests that were served by bypassing the cache.
- `fetch_b`: Bytes fetched from S3 triggered by cache misses.  Note the cache may read in more bytes than requested for the miss.
- `fetch`: Count of GET requests sent to S3.
- `unused_b`: Count of bytes that were fetched from S3 but not sent to the client.
- `prefetch_b`: Bytes prefetched from S3 independent of a client request.
- `prefetch`: Count of prefetch requests sent to S3.
- `bypass_s`: Seconds spent in GET requests serving bypass requests.  Floating point number.
- `fetch_s`: Seconds spent in GET requests serving cache misses.  Floating point number.

PR #2433, opened 2025-06-23
S3 file cache metrics

patrickbrophy:
This PR addresses issue #2360. This PR consumes the new statistics from the xrootd-s3-http plugin and processes them into Prometheus metrics. This PR introduces the following Prometheus metrics:
- `xrootd_s3_cache_bytes_total`: Bytes transferred by the S3 cache plugin.
- `xrootd_s3_cache_hits_total`: Number of cache hits, partial hits, or misses.
- `xrootd_s3_cache_requests_total`: Number of cache requests.
- `xrootd_s3_cache_errors_total`: Number of errors encountered by the S3 cache plugin.
- `xrootd_s3_cache_request_seconds_total`: Total time spent in S3 requests.

This PR also cleans up how we handle events from the OSS g-stream and introduces a dispatcher to the correct handler.

### DIFF
--- a/metrics/xrootd_metrics.go
+++ b/metrics/xrootd_metrics.go
@@ -203,6 +203,27 @@ type (
 		IOTotal    int     `json:"io_total"`
 	}
 
+	OssGStreamEvent struct {
+		Event string `json:"event"`
+	}
+	OssS3CacheGs struct {
+		Event     string  `json:"event"`
+		HitB      int     `json:"hit_b"`      // Bytes that were served to the client directly from the cache (a "cache hit")
+		MissB     int     `json:"miss_b"`     // Bytes that were served to the client that were a cache miss (not immediately available).
+		FullHit   int     `json:"full_hit"`   // Count of read requests from the client that were completely served from the cache.
+		PartHit   int     `json:"part_hit"`   // Count of read requests from the client that were partially served from the cache.
+		Miss      int     `json:"miss"`       // Count of read requests that were entirely a miss.
+		BypassB   int     `json:"bypass_b"`   // Bytes that were read in "bypass mode", skipping the cache. Typically, this is from read requests that are larger than the cache size.
+		Bypass    int     `json:"bypass"`     // Count of read requests that were served by bypassing the cache.
+		FetchB    int     `json:"fetch_b"`    // Bytes fetched from S3 triggered by cache misses. Note the cache may read in more bytes than requested for the miss.
+		Fetch     int     `json:"fetch"`      // Count of GET requests sent to S3.
+		UnusedB   int     `json:"unused_b"`   // Count of bytes that were fetched from S3 but not sent to the client.
+		PrefetchB int     `json:"prefetch_b"` // Bytes prefetched from S3 independent of a client request.
+		Prefetch  int     `json:"prefetch"`   // Count of prefetch requests sent to S3.
+		Errors    int     `json:"errors"`     // Count of errors encountered.
+		BypassS   float64 `json:"bypass_s"`   // Seconds spent in GET requests serving bypass requests.
+		FetchS    float64 `json:"fetch_s"`    // Seconds spent in GET requests serving cache misses.
+	}
 	OSSStatsGs struct {
 		Event           string  `json:"event"`
 		Reads           int     `json:"reads"`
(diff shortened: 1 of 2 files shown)


metrics/xrootd_metrics.go
```diff
@@ -364,6 +385,11 @@ const (
 	ProcStat  SummaryStatType = "proc"  // https://xrootd.web.cern.ch/doc/dev57/xrd_monitoring.htm#_Toc138968507
 )
 
+const (
+	OssStatsEvent    = "oss_stats"
```
jhiemstrawisc, line 389: Can you leave a comment explaining what these are?

metrics/xrootd_metrics.go
```diff
@@ -1740,7 +1792,90 @@ func HandleSummaryPacket(packet []byte) error {
...
+// The s3file_stats event schema is as follows:
+//
+//	{
+//	  "event": "s3file_stats",
+//	  "hit_b": 0,
+//	  "miss_b": 0,
+//	  "full_hit": 0,
+//	  "part_hit": 0,
+//	  "miss": 0,
+//	  "bypass_b": 12265,
```
jhiemstrawisc, line 1830: Since you call this a schema, I think it'd be better to indicate valid value types instead of mostly 0s, e.g.:
```
{
    "event": "s3file_stats",
    "hit_b": uint,
    "miss_b": uint,
    "bypass_s": float,
```
(I don't know if these are correct, just giving an example).

metrics/xrootd_metrics.go
```diff
@@ -1740,7 +1792,90 @@ func HandleSummaryPacket(packet []byte) error {
...
+
+	// we need to process the blobs backwards to ensure that we are processing the last valid event
+	// the most relevant data is the last valid event in the sequence of blobs
+	for i := len(blobs) - 1; i >= 0; i-- {
+		blob := blobs[i]
+		s3fileStats := OssS3CacheGs{}
+		if err := json.Unmarshal(blob, &s3fileStats); err != nil {
+			return errors.Wrap(err, "failed to parse S3 file stats json")
+		}
+		if !allowedEvents[s3fileStats.Event] {
```
jhiemstrawisc, line 1851: This is already checked in `handleOSSPacket`, isn't it?

metrics/xrootd_metrics.go
```diff
@@ -1740,7 +1792,90 @@ func HandleSummaryPacket(packet []byte) error {
...
+			log.Debugf("handleS3CacheStats received an S3 file stats packet with an unrecognized event type (%s)", s3fileStats.Event)
```
jhiemstrawisc, line 1852: If this is something we really don't expect to happen, it may be better to set this at the warning or error level so it's a bit more obvious. I'll defer to you.

metrics/xrootd_metrics.go
```diff
@@ -1815,18 +1950,7 @@ func HandleSummaryPacket(packet []byte) error {
...
-		incBy := math.Max(0, float64(new-old))
-		counter.Add(incBy)
-
-		if new < old {
-			return old
-		}
-		return new
-	}
-
+func handleOSSStats(blobs [][]byte) error {
```
jhiemstrawisc, line 1953: Git doesn't let you comment on lines that are far from changes, but with this function name change I think you'll also need to change the debug line that says "handleOSSPacket received an OSS packet with an unrecognized event type (%s)" (line 1979 in my current view).

metrics/xrootd_metrics.go
```diff
@@ -1740,7 +1792,90 @@ func HandleSummaryPacket(packet []byte) error {
...
+	incBy := new - old
+	if incBy < 0 {
+		incBy = 0
+	}
+	counter.Add(incBy)
+	if new < old {
+		return old
+	}
+	return new
+}
```
jhiemstrawisc, line 1813: If you haven't played around with go's "generics", this smells like exactly the type of thing they're meant for. You should take a look and see whether this could be written as one function on both ints and floats 🤓

jhiemstrawisc (changes requested):

metrics/xrootd_metrics.go
```diff
@@ -1740,7 +1792,90 @@ func HandleSummaryPacket(packet []byte) error {
...
+func handleS3CacheStats(blobs [][]byte) error {
+
+	// we need to process the blobs backwards to ensure that we are processing the last valid event
+	// the most relevant data is the last valid event in the sequence of blobs
+	for i := len(blobs) - 1; i >= 0; i-- {
+		blob := blobs[i]
+		s3fileStats := OssS3CacheGs{}
+		if err := json.Unmarshal(blob, &s3fileStats); err != nil {
+			return errors.Wrap(err, "failed to parse S3 file stats json")
+		}
```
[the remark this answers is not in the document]
patrickbrophy: An invalid event could have the following properties:
- It could be malformed (invalid JSON)
- It could have an unrecognized event

I agree the logic is incorrect and we should just skip over the blob and move on to find a valid one.

PR #2435, opened 2025-06-23
Trim trailing '/' from federation-derived lot paths

jhiemstrawisc:
While trying to set up Lotman in a production setting, I noticed a regression that caused all lot usage to be attributed to the default lot. After some debugging, I discovered a comment in the Lotman source that handles "directory-based usage updates" that indicated Lotman does not expect to have a trailing '/', but this is exactly how the federation-supplied namespace paths were being initialized in lotman.

I view this more as a Lotman bug, but in the interest of getting this moving I'm trimming the '/' in Pelican for now.


jhiemstrawisc:
Given the time crunch and this diff's isolation to Lotman stuff, I'm using superpowers to merge it.

PR #2437, opened 2025-06-23
Fix link depth header for listings

turetske:
Fixes an error where the link header depth when querying the directory for collections was incorrect at the top level of a namespace because the prefix would have an "/" appended to it which would cause the prefix matching to fail.

Simple solution, remove the potential extra "/" after doing the matching.

Also ensures that a full match has a link depth of 0.


jhiemstrawisc (changes requested):
Since this was a previously uncaught bug and there's already a set of unit tests for this function, can you add a new test that would've caught the bug this solves?

jhiemstrawisc (approved):
LGTM

PR #2444, opened 2025-06-26
Update OA4MP

brianaydemir:
This PR preserves the approach of taking the upstream [scitokens-oauth2-server](https://github.com/scitokens/scitokens-oauth2-server) container image and copying the OA4MP install from it and into the Pelican origin container image.

Hidden away in our Dockerfile are two important things:

* A patch to one of `.jsp`'s, to ensure that the web page's form action submits to the correct (i.e. default) URL.
* A replacement for the template used for dynamically registered clients.

The changes to which QDL scripts are run in which phases is necessary to ensure that the OAuth flow correctly records the user and group information sent by Pelican to OA4MP ([reference on what the phases are](https://oa4mp.org/pdf/oa4mp_running_qdl_scripts.pdf)).

The changes to `server-config.xml` are me making a pass at cleaning it up a bit. The only functional change, which is required for this update, is to the class name for `OA2QDLLoader`.

The additional handling for `consent_required` in `device_auth.go` is motivated (required) by one of the relevant OAuth and/or OIDC specifications, but I can't recall exactly which one.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -64,7 +64,7 @@ ARG XROOTD_S3_HTTP_VER=0.4.1
 
 # Set scitokens-oauth2-server image as a build stage so that we can copy its
 # OA4MP installation into the images being built here.
-FROM --platform=linux/amd64 hub.opensciencegrid.org/sciauth/scitokens-oauth2-server:release-20250507-1732 AS scitokens-oauth2-server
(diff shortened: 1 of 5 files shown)


CannonLock:
LGTM.

Can confirm that this fixes the token flow and actually fixes the Auth Code flow as well as a bit of a plus. Additionally sorts out the issues revolving around the OA4MP action targets pointing at the proxy endpoint.

I have tested by running the federation and pulling down a protected file via the cli.

PR #2447, opened 2025-06-26
Actually make use of Origin.ScitokensNameMapFile

brianaydemir:
We just forgot to plumb the parameter's value through the process of configuring XRootD.


jhiemstrawisc (approved):
LGTM

Issue #2441, opened 2025-06-25
Client will attempt to reuse a token from the local credential store when it lacks an appropriate scope


brianaydemir:
What the title says. I tested this using Pelican 7.17.0 and 7.12.0.

Let's start by removing the local credential store (I'm in a Docker Compose setup, hence why the store is the one `root` uses):

```
$ rm /etc/pelican/credentials/client-credentials.pem
```

Now, let's try a simple `ls`, so that we acquire a read token:

```
$ ./pelican --version
Version: 7.17.0
Build Date: 2025-06-20T18:26:03Z
Build Commit: c6c4d24e1f055fbed89aeeefcbc28227a4f814db
Built By: goreleaser
$ ./pelican object ls --debug pelican://director:8444/testing-1/
...
DEBUG[2025-06-25T18:56:44Z] Requesting a credential with the following scope: storage.read:/
...
[...]

PR #2455, opened 2025-06-27
Added per operation scope checking

patrickbrophy:
This PR addresses issue #2441. This PR fixes a bug where the token is marked acceptable when it shouldn't as described in the issue. This PR adds the appropriate logic to mark the token as acceptable based on the operation that the client is trying to achieve.


brianaydemir:
I added the `create-patch` label to this because of [this request from a user](https://opensciencegrid.freshdesk.com/a/tickets/80173).

brianaydemir:
I'm going to let @turetske be the official reviewer for this PR because I don't have time to test the changes myself. But that said, the proposed code looks like what I'd expect.

turetske (changes requested):
This looks good, but is there any way you can write a test that would have failed before and now passes?

PR #2457, opened 2025-06-27
When creating custom TLS certs, add IPs to SANs as IPs and not DNS names

jhiemstrawisc:
The issue being fixed here is that some of our internal tests were producing bad certificates by adding IP addresses as DNS names, which was exposed recently when we started checking the configured certs against the configured server hostnames

This diff does two things -- it only toggles the cert<-->hostname validation when TLS is enabled in Pelican, and it adds the IP addresses to the cert correctly regardless.

### DIFF
--- a/config/init_server_creds.go
+++ b/config/init_server_creds.go
@@ -30,6 +30,7 @@ import (
 	"fmt"
 	"io/fs"
 	"math/big"
+	"net"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


jhiemstrawisc:
Indeed -- the offending test comes from using an `httptest` server for Pelican config:
https://github.com/PelicanPlatform/pelican/blob/c2068a1eea9eb8e2ee9893b8a7ff5ca23832eddf/director/stat_test.go#L942

jhiemstrawisc:
@bbockelm Since fixing these underlying tests can be done in a subsequent PR, and since I believe this PR does fix an issue even if it's not the root issue, I'm going to merge. In the short term this unblocks #2035.

Issue #2443, opened 2025-06-25
Implement Regex-Based Group Authorization in Issuer


patrickbrophy:
Currently, the Pelican issuer requires administrators to explicitly list every group name within an AuthorizationTemplates entry to grant token scopes.  For collaborations with hundreds of similarly-named groups this is not scalable. 

To remedy this, I suggest introducing a new `group_regexes` key within the Issuer.AuthorizationTemplates objects in pelican.yaml. This will allow administrators to use regular expressions to define which groups an authorization rule applies to.

brianaydemir:
@patrickbrophy This issue seems to be related to #2169.

Issue #2207, opened 2025-04-15
Add comments to the generated `policies.qdl` file and the Go template that generates it


brianaydemir:
https://github.com/PelicanPlatform/pelican/blob/main/oa4mp/resources/policies.qdl is an imposing Go template that generates QDL from Pelican's configuration for the built-in issuer (OA4MP).

The logic in the Go template and the logic in the generated QDL would both benefit from some commentary explaining the logic, and in the case of the generated QDL, maybe even a pointer to the Pelican configuration that lead to it.

brianaydemir:
It's possible that the changes in #2462 will make much of the logic in `policies.qdl` unnecessary.

PR #2462, opened 2025-07-01
Implement regex based group authorization

patrickbrophy:
Add regex-based group authorization, addressing #2443. This change introduces a `group_regexes` field to `Issuer.AuthorizationTemplates`, enabling deployments to map scopes with regular-expression patterns instead of only exact group names. `compileAuthzRules` now pre-compiles both literal and regex matchers at start-up to eliminate per-request overhead, and `calculateAllowedScopes` matches each incoming group against these matchers, deduplicates the resulting scopes, and injects them into the OA4MP proxy header. QDL policy files (id_token_policies.qdl, policies.qdl) are updated so the computed scopes flow into claims.scopes and merge with default_scopes.

### DIFF
--- a/oa4mp/proxy.go
+++ b/oa4mp/proxy.go
@@ -63,6 +63,50 @@ func getTransport() *http.Transport {
 	return transport
 }
 
+func calculateAllowedScopes(user string, groupsList []string) []string {
+	if len(compiledAuthzRules) == 0 {
+		return nil
+	}
+
+	scopeSet := make(map[string]struct{})
+	for _, rule := range compiledAuthzRules {
+		if len(rule.UserSet) > 0 {
+			if _, ok := rule.UserSet[user]; !ok {
+				continue
+			}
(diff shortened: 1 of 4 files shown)


brianaydemir:
@patrickbrophy Can you update the documentation in [parameters.yaml](https://github.com/PelicanPlatform/pelican/blob/89634f228657daa88cbade4da1266657cba03c62/docs/parameters.yaml#L2529-L2533)? Unless this is supposed to be a hidden feature, in which case, a comment in this PR might be helpful for posterity.

brianaydemir (changes requested):

brianaydemir:
@patrickbrophy Out of curiosity: Does this PR eliminate some or all of the need for `policies.qdl` to do the scope calculations, including subsituting for `$USER` and `$GROUP`? If so, maybe I'll clean that up as a part of #2207.

PR #2268, opened 2025-04-29
Add new Docker setups to run as federation and web service

CannonLock:
Create new docker based pelican testing environments for development

PR #2463, opened 2025-07-01
Avoid duplicated server sitename in registration and update via CLI/webUI

h2zh:
In this PR, we check if a server's sitename has already been taken by any same-type server in Registry every time when admin wants to register or update a server via CLI or webUI. If so, Pelican will terminate the registration/update process and prompt the admin to provide another sitename.

### Design Thinking
This feature looks simple at the first glance, but requires much more consideration. We cannot just add a new "unique" constraint to DB because a) sitename is not a DB column, but a field in a json string column `admin_metadata` b) sitename could be duplicate for a server namespace and the namespaces it host c) there are lots of duplicate namespaces in DB because of b)

So this PR focuses on the prevention of duplicate sitename for server (Origin/Cache) namespace _from now on_. If the a server's sitename was taken by another, admin will not able to register or update that server namespace _without_ changing the sitename. 

To clean up the existing duplicate registrations in OSDF Registry DB, there is a counterpart [Jira ticket](https://opensciencegrid.atlassian.net/browse/OPS-416). 

### Impact on dev env
With the introduction of no duplicate sitename for Pelican servers, running all Pelican services in a local dev environment becomes a little bit tricky - because by default Origin and Cache both use os.Hostname as `sitename`. If you spin up Cache after Origin in a fresh federation (meaning no DB entries in Registry DB), you will be hit by the following error message:

```
ERROR[2025-07-02T21:08:01Z] Server responded with an error: Bad request for key-sign challenge: The SiteName '88386cd92f0a' is already in use by another registered origin or cache namespace. . The POST attempt to https://88386cd92f0a:8446/api/v1.0/registry resulted in status code 400 
```

To avoid this error, you need to either:

1. After spinning up an Origin, change the sitename of the Origin in the Registry Web UI. For example, change the sitename from "88386cd92f0a" to "88386cd92f0a-Origin".
2. Use the new `docker compose` Pelican development environment. https://github.com/PelicanPlatform/pelican/pull/2268 It provides different config yaml file for Origin and Cache respectively, so that you could set different `Xrootd.Sitename` for them.

Note: this only impacts on a fresh federation. Once the registrations of both Origin and Cache are done, you don't need to do this trick when spinning up Origin and Cache in the future, as long as you don't nuke the registrations in Registry DB.

### DIFF
--- a/registry/client_commands_test.go
+++ b/registry/client_commands_test.go
@@ -403,20 +403,20 @@ func TestRegistryKeyChainingOSDF(t *testing.T) {
 	require.NoError(t, err)
 
 	// Start by registering /foo/bar with the default key
-	err = NamespaceRegister(privKey, registrySvr.URL+"/api/v1.0/registry", "", "/foo/bar", "")
+	err = NamespaceRegister(privKey, registrySvr.URL+"/api/v1.0/registry", "", "/foo/bar", "test-site-foobar")
 	require.NoError(t, err)
 
 	// Perform one test with a subspace and the same key -- should succeed
-	err = NamespaceRegister(privKey, registrySvr.URL+"/api/v1.0/registry", "", "/foo/bar/test", "")
+	err = NamespaceRegister(privKey, registrySvr.URL+"/api/v1.0/registry", "", "/foo/bar/test", "test-site-foobar-test")
 	require.NoError(t, err)
 
 	// If the namespace is a subspace from the topology and is registered without the identity
 	// we deny it
-	err = NamespaceRegister(privKey, registrySvr.URL+"/api/v1.0/registry", "", "/topo/foo/bar", "")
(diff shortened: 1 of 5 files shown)


h2zh:
I think this PR would be better put on hold, because there is going to be some major changes to Registry DB ("sitename" will be replaced by "name" in a new table). We can just put a unique constraint on the new "name" column.

Besides, if the sitename uniqueness is required among all servers, it would be a problem for a server that has both Origin and Cache services. They have the same hostname, thus the same sitename. There's no way to set different sitename for Origin and Cache **within a server**. The local dev environment and some unit tests (e.g. `TestPrestage` as in the GitHub actions) will also be blocked, because their Origin and Cache have the same sitename. 

We should allow this:
server-service-relation drawio

P.S. Here are some concept definitions by BrianB https://github.com/orgs/PelicanPlatform/discussions/2445#discussioncomment-13621390

Issue #2458, opened 2025-06-30
Origin.Concurrency health check triggers Prometheus query errors when Web UI is disabled


patrickbrophy:
An origin operator from LIGO reported that they were seeing a large amount of log messages like the following:
```
time="2025-06-30T11:03:24-07:00" level=error msg="Failed to query prometheus for current xrootd concurrency value; query: avg_over_time(xrootd_server_io_active[1m]); error: Prometheus responded 404 for the query \"avg_over_time(xrootd_server_io_active[1m])\": 404 page not found"
```

The operator noted that `Origin.Concurrency: 80` was set and that the web UI was disabled. 

I believe that the issue is coming from PR #2338 by @jhiemstrawisc. This PR makes the assumption that the Origin's Prometheus server is reachable. 

Proposed solutions:
- Verify that the internal Prometheus endpoint is available before attempting the query.
- Make the health check a no-op if Server.EnableUI is false.
- Introduce a new configuration parameter like `Server.EnablePrometheus` and make the Prometheus server and check rely on checking this parameter is enabled.

PR #2465, opened 2025-07-02
Add a `Monitoring.EnablePrometheus` configuration parameter to disable the embedded Prometheus server and related checks

patrickbrophy:
This PR address issue #2458. This PR introduces a new configuration parameter, `Monitoring.EnablePrometheus`, which defaults to true.  When set to false, the embedded Prometheus server will not be started and will skip the IO-concurrency monitoring introduced in PR #2338. Also when Prometheus is disabled and the Web UI is enabled, it will put a warning message in the logs to warn that metrics will not be available in the Web UI.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2880,6 +2880,13 @@ components: ["origin", "cache"]
 ############################
 # Monitoring-level configs #
 ############################
+name: Monitoring.EnablePrometheus
+description: |+
+  Enable the internal Prometheus server.
+type: bool
+default: true
+components: ["origin", "cache", "director", "registry"]
+---
 name: Monitoring.DataLocation
 description: |+
   A filepath where Prometheus should host its monitoring data.
--- a/launchers/launcher.go
+++ b/launchers/launcher.go
@@ -89,6 +89,10 @@ func LaunchModules(ctx context.Context, modules server_structs.ServerType) (serv
 
 	// Register OIDC endpoint
 	if param.Server_EnableUI.GetBool() {
+		// Warn if Prometheus is disabled, but Web UI is enabled. Metrics via Web UI will not be available.
+		if !param.Monitoring_EnablePrometheus.GetBool() {
+			log.Warn("Prometheus is disabled, but Web UI is enabled. Metrics via Web UI will not be available.")
+		}
 		if modules.IsEnabled(server_structs.RegistryType) ||
 			(modules.IsEnabled(server_structs.OriginType) && param.Origin_EnableOIDC.GetBool()) ||
 			(modules.IsEnabled(server_structs.CacheType) && param.Cache_EnableOIDC.GetBool()) ||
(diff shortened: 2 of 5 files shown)


docs/parameters.yaml
```diff
@@ -2880,6 +2880,13 @@ components: ["origin", "cache"]
 ############################
 # Monitoring-level configs #
 ############################
+name: Monitoring.EnablePrometheus
+description: |+
+  Enable the internal Prometheus server. This server is binded to the Server.WebPort
```
jhiemstrawisc, line 2885: Can you do "... bound to the ${Server.WebPort}."?

launchers/launcher.go
```diff
@@ -370,9 +378,14 @@ func LaunchModules(ctx context.Context, modules server_structs.ServerType) (serv
...
-	if param.Server_EnableUI.GetBool() {
+	if param.Monitoring_EnablePrometheus.GetBool() {
 		metrics.SetComponentHealthStatus(metrics.Prometheus, metrics.StatusWarning, "Prometheus not started")
-		if err = web_ui.ConfigureEmbeddedPrometheus(ctx, engine); err != nil {
+		// Due to federation tests / fed-in-a-box, we need to configure the embedded prometheus instance only once
+		// and not for each server. This is why we use a sync.Once here.
+		oncePrometheus.Do(func() {
+			err = web_ui.ConfigureEmbeddedPrometheus(ctx, engine)
+		})
+		if err != nil {
```
jhiemstrawisc, line 388: Two comments here:
- The error handling and other metrics settings should also be moved into the `sync.Once` so that as modules run through this, the status updates and error handling are only triggered by the first invocation. You'll also want to have an externally-scoped error value to set `err` on the outer functions return, e.g.
```
var prometheusInitErr error
		oncePrometheus.Do(func() {
			prometheusInitErr = web_ui.ConfigureEmbeddedPrometheus(ctx, engine)
			if prometheusInitErr != nil {
				prometheusInitErr = errors.Wrap(prometheusInitErr, "failed to configure embedded Prometheus instance")
				metrics.SetComponentHealthStatus(metrics.Prometheus, metrics.StatusCritical, prometheusInitErr.Error())
				return
			}
			metrics.SetComponentHealthStatus(metrics.Prometheus, metrics.StatusOK, "Prometheus started")
		})

		if prometheusInitErr != nil {
			err = prometheusInitErr
			return
		}
```
- I think there's a bit more webUI/Prometheus disentanglement that needs to happen here. If I disable Prometheus but enable the Web UI in a fresh environment that hasn't enabled Web UI in the past, I never get the initialization prompt with a code because that's triggered by the return in the error group here. You should move that to its own `if param.Server_EnableUI.GetBool()` block the way it used to be.

jhiemstrawisc (changes requested):

launchers/launcher.go
patrickbrophy: Ah that makes a lot of sense. I just pushed some changes to align with this.

jhiemstrawisc (approved):
LGTM -- I tested the grid of [webUI, Prometheus] x [on, off] and it looks like all four worked as expected for me.

Thanks for this one!

PR #2468, opened 2025-07-02
Web client - All Changes

CannonLock:
Intended to be merged after #2467

**Add redirect=false to the director so that the web client can read the directors headers before being redirected**

Used by the web client which cannot read headers from a url that has redirected itself. Allows the web client to consume the `X-Pelican-Authorization` header, which in turn points it at the issuer to begin the token exchange.

**Add CORs configuration on the Director and the Origin Issuer**

Enables browser requests to the Director and the Origin. These are used on PUT, POST, OPTION, GET methods depending on the service. 

The director needs to be able to handle the PUT, POST, GET and OPTION which are typically forwarded to the cache/origin.

The Origin needs to be able to respond to the Issuer requests such as the code exchange and the Client registration.

**Configure Issuer to only respond to configured Web Clients**

Added a new configuration knob called `Issuer.RedirectUris`. This knob makes it so that only RedirectUris that the Origin admin has approved can be used for dynamic clients. The hosts from these clients are also the only hosts that can make browser requests to the Origin issuer.

This is to prevent anyone from setting up a web client for a Origin. If that was possible it would not be challenging for someone to accidentally use a fake client and expose their object tokens. 

**Add new configuration to `register` web client uris**

Used for the purpose above. Is a slice of strings.

### DIFF
--- a/client/acquire_token.go
+++ b/client/acquire_token.go
@@ -534,7 +534,6 @@ func registerClient(dirResp server_structs.DirectorResponse) (*config.PrefixEntr
 	}
 
 	drcp := oauth2.DCRPConfig{ClientRegistrationEndpointURL: issuer.RegistrationURL, Transport: config.GetTransport(), Metadata: oauth2.Metadata{
-		RedirectURIs:            []string{"https://localhost/osdf-client"},
 		TokenEndpointAuthMethod: "client_secret_basic",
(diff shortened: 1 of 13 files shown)


h2zh (changes requested):
One small thing - the linter you used for param/parameters.go and param/parameters_struct.go may be different from the standard Pelican linter setting

PR #2474, opened 2025-07-08
Bump xrootd release to 5.8.3-1.3

matyasselmeci:


jhiemstrawisc:
LGTM. We discussed offline, and this does not currently need to be patched into 7.17.

PR #2476, opened 2025-07-08
Upgrade xrdhttp-pelican to v0.0.7

h2zh:
This xrdhttp-pelican release should enable Pelican's 'Drop Privileges' mode. The Authfile and SciTokens config file are now moved into the directories owned by the xrootd user.

### How to test
Checkout to this branch, then rebuild the container.
Run `rpm -q xrdhttp-pelican` in the new container, you should see `xrdhttp-pelican-0.0.7-........`

If you want to further check, spin up the federation with the following configs for Cache only. Make sure there is no obvious error like `Unable to find /run/pelican/xrootd/cache/authfile-cache-generated; no such file or directory` in the running logs.
```
Server:
  DropPrivileges: true
  UnprivilegedUser: pelican
```


h2zh:
Could you try to run `goreleaser --clean --snapshot` to rebuild the binary on this branch? The error logs show a cgo problem but Pelican is built with without cgo by default (CGO_ENABLED=0 in line 34, .goreleaser.yml) @patrickbrophy

patrickbrophy (approved):
After lots of testing, things are looking good!

LGTM

PR #2478, opened 2025-07-09
OSDF systemd services: use `pelican-server` everywhere

### LINKED ISSUE #2464: OSDF systemd service files should stop using /usr/bin/osdf name
Pelican Service:
- [x] Registry
- [x] Director

The systemd service files for the director and the registry currently rely on the `/usr/bin/osdf` binary, which makes it hard to discern the correct config (see #2451 ).  The origin and the cache have already been adjusted to use `pelican-server`; let's do the same for the director and registry.

matyasselmeci:
This avoids reliance on the magic config from the `osdf` binary. Fixes #2464.

### DIFF
--- a/systemd/osdf-director.service
+++ b/systemd/osdf-director.service
@@ -4,7 +4,7 @@ After = network.target nss-lookup.target
 
 [Service]
 EnvironmentFile = -/etc/sysconfig/osdf-director
-ExecStart = /usr/bin/osdf --config /etc/pelican/osdf-director.yaml director serve
+ExecStart = /usr/bin/pelican-server --config /etc/pelican/osdf-director.yaml director serve
 Restart = on-failure
 RestartSec = 20s
 
--- a/systemd/osdf-registry.service
+++ b/systemd/osdf-registry.service
@@ -4,7 +4,7 @@ After = network.target nss-lookup.target
 
 [Service]
 EnvironmentFile = -/etc/sysconfig/osdf-registry
-ExecStart = /usr/bin/osdf --config /etc/pelican/osdf-registry.yaml registry serve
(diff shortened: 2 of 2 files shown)


brianhlin (approved):
LGTM

brianhlin:
@jhiemstrawisc we're already doing so for the origin:

1. https://github.com/PelicanPlatform/pelican/blob/main/systemd/pelican-origin.service#L7
2. https://github.com/PelicanPlatform/pelican/blob/main/systemd/osdf-origin.service#L7

Is there a reason we shouldn't?

jhiemstrawisc:
Roping in @h2zh on this -- we just merged a change that enables CGO in the `pelican-server` binary, a decision that was predicated on our belief the `pelican-server` binary was not in use outside the controlled environment of our cache images (see issue #2485 and corresponding PR #2488).

I don't have a full grasp of potential consequences here, but I think enabling CGO means we lose having a static binary (haven't confirmed this) and need to worry more about distribution.

h2zh:
One concern I have for this change is - now with `CGO_ENABLED=1`, `pelican-server` depends on `libc` and maybe more other c libraries. In deployment, though most Linux systems have `libc` by default, we may need to check that.

brianhlin:
@jhiemstrawisc @h2zh can we just change the origin + director + registry services to use the pelican binary until you guys are ready?

I checked the Dockerfile and that looked fine. Separately, we should probably move the `osdf-*` containers to the `pelican` and `pelican-server` binaries

jhiemstrawisc:
That sounds like a good plan to me. I'm guessing we should hold the 7.18 release for this?

Issue #2470, opened 2025-07-07
When uploading an object, the client should prevent overwriting existing objects.


patrickbrophy:
The client currently is able to overwrite objects when given a token with the scope `storage:modify`. However within Pelican objects are immutable. The client should respect this principle. 

Instead of messing with the tokens, we should just simply check if the object exists but uploading the object. If the object does exist it should return an error.

PR #2480, opened 2025-07-09
Prevent overwriting objects at upload

patrickbrophy:
This PR addresses issue #2470. This PR adds a simple check to see if the object exists at the origin before proceeding with the upload. This check doesn't take into account any tokens and will simply block the upload if the object exists. 

I added test cases to ensure this behavior works as expected.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -3108,7 +3108,9 @@ func uploadObject(transfer *transferFile) (transferResult TransferResults, err e
 	transferResult.job = transfer.job
 
 	// Check if the remote object already exists using statHttp
-	if transfer.remoteURL != nil && transfer.job != nil {
+	// If the job is recursive, we skip this check as the check is already performed in walkDirUpload
+	// If the job is not recursive, we check if the object exists at the origin
+	if transfer.remoteURL != nil && transfer.job != nil && (transfer.job.syncLevel == SyncNone || !transfer.job.recursive) {
 		remoteUrl, dirResp, token := transfer.job.remoteURL, transfer.job.dirResp, transfer.job.token
 		_, statErr := statHttp(remoteUrl, dirResp, token)
 		if statErr == nil {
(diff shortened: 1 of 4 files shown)


client/handle_http.go
```diff
@@ -3084,6 +3084,22 @@ func uploadObject(transfer *transferFile) (transferResult TransferResults, err e
...
+	// If the job is recursive, we skip this check as the check is already performed in walkDirUpload
+	// If the job is not recursive, we check if the object exists at the origin
+	if transfer.remoteURL != nil && transfer.job != nil && transfer.job.syncLevel != SyncNone && !transfer.job.recursive {
+		remoteUrl, dirResp, token := transfer.job.remoteURL, transfer.job.dirResp, transfer.job.token
+		_, statErr := statHttp(remoteUrl, dirResp, token)
+		if statErr == nil {
+			// Object exists, abort upload
+			transferResult.Error = errors.New("remote object already exists, upload aborted")
+			return transferResult, transferResult.Error
+		} else if !strings.Contains(statErr.Error(), "not found") && !strings.Contains(statErr.Error(), "no such object") && !strings.Contains(statErr.Error(), "cannot remove remote path") {
```
turetske, line 3097: I'm a little concerned about this, solely because if the error ever changes, then this will fail, but I'm unsure if there is a better solution. As long as there is a test which will check for these, then when things change and that test fails, then the developer will know to update this check.

client/handle_http_test.go
```diff
@@ -1797,3 +1880,197 @@ func TestTLSCertificateError(t *testing.T) {
...
+			attempts: []transferAttemptDetails{
+				{
+					Url: svrURL,
+				},
+			},
+		}
+
+		result, err := uploadObject(transfer)
+		require.NoError(t, err) // We should not get an error from the uploadObject call
+		require.NoError(t, result.Error)
```
turetske, line 2074: Can you also capture the log output from this line?

```log.Warningln("Failed to check if object exists at the origin, proceeding with upload")```

and ensure it generates the correct warning? Otherwise it looks good.

turetske (changes requested):

PR #2481, opened 2025-07-10
Add a default /etc/pelican/pelican.yaml file to the image

matyasselmeci:
which just loads all the config from `/usr/share/pelican/config.d` and `/etc/pelican/config.d`. Create those directories when building the image.


brianhlin (approved):
This LGTM but was there really no default `pelican.yaml` config before this? Interesting

matyasselmeci:
> This LGTM but was there really no default `pelican.yaml` config before this? Interesting

I guess strictly speaking you don't need one if you're doing all the config by env var.

PR #2488, opened 2025-07-14
Enable CGO for the binary build that includes purego

h2zh:
Enable CGO for the build that includes lotman. Lotman uses purego, which implements a fakecgo and persuades runtime that CGO is enabled. Then in the drop-privs feature, CGO_ENABLED=0 would make syscall.Setgid and syscall.Setuid call AllThreadsSyscall(), while this function is incompatible with binaries that use cgo (and fakecgo also counts). Set CGO_ENABLED=1 to avoid calling AllThreadsSyscall(), which causes this panic.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -61,6 +61,9 @@ jobs:
         key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
         restore-keys: |
           ${{ runner.os }}-go-
+    - name: Install build tools
+      run: |
+        dnf -y install gcc gcc-c++ || (apt-get -y update && apt-get -y install gcc g++)
     - name: Test
       run: |
         make web-build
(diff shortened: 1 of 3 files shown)


jhiemstrawisc:
> Seems like enabling CGO crashes the lotman tests... @jhiemstrawisc I wonder does this error ring any bell?
> 
> ```
> time="2025-07-15T04:40:58Z" level=error msg="Error opening LotMan library: /usr/lib64/libLotMan.so: undefined symbol: _ZNK8nlohmann11json_schema14json_validator8validateERKNS_16json_abi_v3_11_210basic_jsonISt3mapSt6vectorNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEblmdSaNS2_14adl_serializerES5_IhSaIhEEEE"
> ```

I think this means that enabling `CGO` breaks `purego` and that we can have one or the other, but not both.

h2zh:
To sum up, it turned out that the failed lotman tests on pelican-server GitHub Action were caused by the ABI mismatch when building `xrootd-lotman` plugin from source. See the linked PR above for more info.

PR #2499, opened 2025-07-15
Fix token acquisition from token cache

patrickbrophy:
This issue addresses issue #2441. Changes made in PR #2455 were necessary but insufficient for fix the issue. This PR introduces the following:
- Tokens are now checked to be valid before being pulled from the cache
- Tokens are now correctly selected if they are both not expired and have the correct scope. 

The token selection logic failed to ensure that a cached token was both unexpired and had the appropriate scopes, sometimes resulting in the use of a valid token with the wrong permissions. The logic has been corrected to first find a token that satisfies both conditions, and the underlying tokenIsValid check has been fixed to properly validate a token's expiration.

### DIFF
--- a/client/acquire_token.go
+++ b/client/acquire_token.go
@@ -377,8 +377,13 @@ func (tg *tokenGenerator) get() (token string, err error) {
 	// First, see if the existing token is valid
 	info := tg.Token.Load()
 	if info != nil && time.Until(info.Expiry) > 0 && info.Contents != "" {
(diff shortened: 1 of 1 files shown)


jhiemstrawisc:
I got a security notification that merging this PR introduced some code that may log tokens in clear text:
https://github.com/PelicanPlatform/pelican/security/code-scanning/69

Can someone look into it?

PR #2501, opened 2025-07-15
Install xrootd-lotman and lotman from osg repos

jhiemstrawisc:
Now that `xrootd-lotman-0.0.5` is in the OSG repos, we should install from there instead of building from source.

This should also fix the currently-failing tests that point to an un-located symbol when dlopening Lotman.

Since this passed the GHA tests last time, I'd recommend the following testing procedure:
1. Checkout this diff
2. Build the image, launch tests in the container, e.g.:
```
# from repo root
docker build -t pelican-test:latest -f images/Dockerfile
docker run --rm --volume $(pwd):/pelican --entrypoint go -w /pelican/lotman pelican-test:latest test -tags=lotman
```

This avoids any container/repo mangling that happens in the GHA tests.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -51,13 +51,13 @@ ARG NODEJS_VER=20
 #     claims to be will *always* come from the .spec file. The upstream Git
 #     repository must have a correct .spec file, or confusion will ensue.
 #
-ARG LOTMAN_SRC_BUILD=true
+ARG LOTMAN_SRC_BUILD=false
 ARG LOTMAN_VER=0.0.4
 ARG XRDCL_PELICAN_SRC_BUILD=false
 ARG XRDCL_PELICAN_VER=1.2.3
 ARG XRDHTTP_PELICAN_SRC_BUILD=false
 ARG XRDHTTP_PELICAN_VER=0.0.7
-ARG XROOTD_LOTMAN_SRC_BUILD=true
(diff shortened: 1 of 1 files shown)


h2zh:
Seems like this PR is not able to fix the failed lotman tests. See #2499 (I re-run that failed github action after this PR merged in main, https://github.com/PelicanPlatform/pelican/actions/runs/16301662289/job/46048736669?pr=2499)

jhiemstrawisc:
I'm not totally sure why it didn't pick up the changes -- I thought your rebase should have worked. I don't think it's something specific to Harbor, but is probably related to some Github Actions build cache not realizing the `latest-itb` tag pointed at a new container or something like that.

Maybe @brianaydemir has more insight.

brianaydemir:
@h2zh @jhiemstrawisc From the action run:

```
2025-07-15T21:41:34.9070602Z Digest: sha256:88ef4a36594d3b60c230e5d44b946f22685e2675d1a8ed675bea7db659f7e26e
2025-07-15T21:41:34.9086005Z Status: Downloaded newer image for hub.opensciencegrid.org/pelican_platform/pelican-test:latest-itb
```

That is indeed the second-most recent `pelican-test:latest-itb` image at the moment.

However, the most recent image was, according to Harbor, pushed at `2025-07-15T21:42Z`.

TL;DR The action ran before the image was pushed, by ~30 seconds or so.

Issue #2485, opened 2025-07-11
Failed to drop privileges


h2zh:
Pelican Version:
v7.17.3-rc.0

Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [X] Cache
- [ ] Other (please give the detail)

**Describe the bug**
Pelican Cache process is still owned by `root` after enable `DropPrivileges` in the configs.

**To Reproduce**
Set `DropPrivileges: true` in the configs, then spin up a Pelican Cache.

**Expected behavior**
Pelican Cache process should be owned by `pelican` after enable `DropPrivileges` in the configs.

**Reason**
Go runtime creates multiple OS threads (goroutines)
syscall.RawSyscall() only affects the current thread that makes the syscall
Other threads remain as root - So the process is still considered "root"
We need to go back to syscall.Setuid(), syscall.Setgid() to change the owner of **all threads**.

bbockelm:
Not 100% sure how to solve it but part of the issue is here:

https://github.com/golang/go/blob/master/src/syscall/syscall_linux.go#L1178-L1187

Looks like `syscall.Setgid` uses libc if available; if not, it uses `syscall.runtime_doAllThreadsSyscall`.  However, if `iscgo` is set (which it looks like purego does set), then the `syscall.runtime_doAllThreadsSyscall` will panic.

@jhiemstrawisc - we build `pelican-server` in a very limited context, right?  Can we just enable "real" cgo there?  That should link to `glibc` (we already build on the correct platform) and allow `syscall.Setgid` use the "real" glibc function definition.

jhiemstrawisc:
We _currently_ build pelican-server only for cache containers, but the plan is to be using pelican-server for everything other than clients. I don't know enough about the implications of enabling CGO, but I think avoiding it was the primary reason we opted for the `purego` library, wasn't it?

h2zh:
A tentative workaround for this problem would be using `pelican` instead of `pelican-server` command in testing the drop-privs Cache. Though it would temporarily violate the "using pelican-server for everything other than clients" principle until this `syscall.Setuid`/`purego` incompatibility problem is solved.

PR #2504, opened 2025-07-16
Disable CGO again

h2zh:
Turning on CGO would break our infra pipeline (see the [failure logs](https://github.com/PelicanPlatform/pelican/actions/runs/16327938614/job/46122841447#step:5:77) in new release GitHub Actions). For now, we'd turn it off again to avoid further (maybe significant) modifications to our infra.

CGO was enabled to resolve a drop privileges problem (see [here](https://github.com/PelicanPlatform/pelican/issues/2485#issuecomment-3070860613) for details). Now, because CGO is disabled, `pelican-server` binary will not support drop privileges, while `pelican` binary still supports. This means, at least for now, drop privileges cannot work together with lotman.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -61,9 +61,6 @@ jobs:
         key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
         restore-keys: |
           ${{ runner.os }}-go-
-    - name: Install build tools
-      run: |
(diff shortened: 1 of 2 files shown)


patrickbrophy (commented):
I checked out to this branch and was able to successfully build for all platforms via goreleaser. I'll wait till Justin gives the OK on this.

jhiemstrawisc (approved):
What a fun adventure we've all been on together with CGO. Let's turn it off and hopefully never look back 😆

PR #2513, opened 2025-07-21
Improve deprecated parameter handling in `pelican config get`

h2zh:
When user tries to get a deprecated config via `pelican config get`, this PR:
1. Show deprecation messages for deprecated parameters instead of silently skipping them
2. Provide clear instructions on how to view deprecated parameters using `--include-deprecated` flag

It also gives users the flexibility to choose between broad discovery (substring) and precise targeting (exact match) when retrieving configuration parameters.

Expected behavior:
```
# pelican config get issuerkey
issuerkey: This parameter is DEPRECATED. If you still need to view it, run: pelican config get issuerkey --include-deprecated
issuerkeysdirectory: "issuer-keys"
# pelican config get issuerkey --exact-match
issuerkey: This parameter is DEPRECATED. If you still need to view it, run: pelican config get issuerkey --include-deprecated
No matching configuration parameters found.
# pelican config get issuerkey --exact-match --include-deprecated
issuerkey: "issuer.jwk"
```

### DIFF
--- a/cmd/config_printer/config_get.go
+++ b/cmd/config_printer/config_get.go
@@ -67,10 +67,6 @@ func configGet(cmd *cobra.Command, args []string) {
 				continue
 			}
 
-			if docParam.Deprecated && !includeDeprecated {
-				continue
-			}
-
 			if len(components) > 0 {
 				componentsCheckFailed := true
 				for _, c := range components {
@@ -105,6 +101,12 @@ func configGet(cmd *cobra.Command, args []string) {
 		}
 
 		if matchesFound {
+			// Check for deprecated parameter only when it matches the search criteria
+			if exists && docParam.Deprecated && !includeDeprecated {
+				fmt.Printf("%s: This parameter is DEPRECATED. If you still need to view it, run: pelican config get %s --include-deprecated\n", key, key)
+				continue
+			}
+
 			matches = append(matches, Match{
 				OriginalKey:      key,
 				HighlightedKey:   highlightedKey,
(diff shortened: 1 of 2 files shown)


cmd/config_printer/config_get.go
```diff
@@ -92,19 +88,39 @@ func configGet(cmd *cobra.Command, args []string) {
...
+						matchesFound = true
+					}
 				}
 			}
 		}
 
 		if matchesFound {
+			// Check for deprecated parameter only when it matches the search criteria
+			if exists && docParam.Deprecated && !includeDeprecated {
+				fmt.Printf("%s: This parameter is DEPRECATED. If you still need to view it, run: pelican config get %s --include-deprecated\n", key, key)
```
jhiemstrawisc, line 120: Three notes here:
- Since you're already doing this bit of special handling for deprecated configs, why not also do it for hidden components?
- Can you think of a way to replace the hard-coded `--include-deprecated` hint with something derived from the actual flag? That way, if we change the flag name in the future it'll update here as well. (if it's too hard to do this, it isn't a huge deal)
- Can you find a way to highlight the key's matching substring the way other keys are highlighted:
Screenshot 2025-07-22 at 14 59 08

h2zh: 1. Done!
2. I looked for solutions but didn't find a good one. I can change it to `cmd.Flags().Lookup("include-deprecated")` but the flag name ("include-deprecated") is still hard-coded.
3. There's no highlights on the first line because it is a prompt instead of part of the matching result. If there are multiple keys in the result matching the input keyword, all of them will be highlighted as below.
image

PR #2520, opened 2025-07-23
Fix non-existent `authfile-cache-generated` file errors during Pelican initial startup

h2zh:
- create a first-run mode for authfile emission (EmitAuthfile) during startup
- If the drop privileges feature is enabled, the first run should write the authfile as root user, and the subsequent runs will write the authfile as the unprivileged pelican user via the xrdhttp-pelican plugin.


h2zh:
Seems like these failed tests are universal in the current open PRs (see #2525).

PR #2526, opened 2025-07-24
Use the correct `ver` claim in scitoken verification

h2zh:
Based on the scitoken official [docs](https://scitokens.org/technical_docs/Verification) (scroll to the bottom), the correct behavior is for the `ver` claim is `scitoken:2.0` (singular). However, to maintain backward compatibility and typo tolerance, Pelican should keep accepting `scitokens:2.0` too.

The implementation is inspired by the `scitokens-cpp` [scitoken check](https://github.com/scitokens/scitokens-cpp/blob/709393585ae238397bdabe1e941541cc01044478/src/scitokens_internal.h#L553-L554) mentioned by @brianaydemir

### DIFF
--- a/client/acquire_token.go
+++ b/client/acquire_token.go
@@ -412,7 +412,11 @@ func tokenIsAcceptable(jwtSerialized string, objectName string, dirResp server_s
 		isWLCG = true
 	}
 	if ver, ok := tok.Get("ver"); ok {
-		if s, ok2 := ver.(string); ok2 && strings.HasPrefix(s, "scitokens:") {
+		// Based on the scitoken official docs(https://scitokens.org/technical_docs/Verification),
+		// the correct behavior for the `ver` claim is `scitoken:2.0` (singular). However, to maintain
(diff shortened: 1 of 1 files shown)


client/acquire_token.go
```diff
@@ -412,7 +412,7 @@ func tokenIsAcceptable(jwtSerialized string, objectName string, dirResp server_s
 		isWLCG = true
 	}
 	if ver, ok := tok.Get("ver"); ok {
-		if s, ok2 := ver.(string); ok2 && strings.HasPrefix(s, "scitokens:") {
+		if s, ok2 := ver.(string); ok2 && (strings.HasPrefix(s, "scitoken:2.0") || strings.HasPrefix(s, "scitokens:2.0")) {
```
patrickbrophy, line 415: Can you just add a comment linking to the check that Brian A. referenced? It would be nice to a have a reference explaining why we do this check.

patrickbrophy (changes requested):

PR #2529, opened 2025-07-25
Avoid ASCII color codes printed to non-terminals

bbockelm:
Additionally, removes a `println` statement that triggers during unit tests.


patrickbrophy (approved):
Tested on my machine and works as expected.

Issue #2475, opened 2025-07-08
"failed to verify size of downloaded file on disk" error for second cache pull


williamnswanson:
**Describe the bug**
_A clear and concise description of what the bug is._
On 7/7/2025, we noticed a large number of transfers fail due to a file being around 16kb (or 32kb, or 82 kb, but mostly ~16kb) larger on disk than Pelican expected it to be. The transfers in question had all previously attempted to pull the file from anther cache and failed due to "transfer to slow" errors. These transfers all tried at the same initial cache and again at the same secondary cache.

https://kibana.osg.chtc.io/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2025-07-06T05:00:00.000Z',to:'2025-07-08T05:00:00.000Z'))&_a=(columns:!(AttemptError,TransferUrl,TransferError),filters:!(),index:'9118b790-2822-11ef-a65d-4b3b1ae1a240',interval:auto,query:(language:kuery,query:'Endpoint:*osdf1.chic.nrp.internet2.edu*%20and%20not%20TransferSuccess:%20true%20and%20AttemptError:%20%22failed%20to%20verify%20size%22%20and%20TransferUrl.indexed:%20%22osdf:%2F%2F%2Fospool%2Fap21%2Fdata%2Fhroch%2Fsingularity_repos%2Fipglasmaframework_main_06_27_2025.sif%22%20and%20TransferError:%20osdfcache01.crc.nd.edu'),sort:!(!(RecordTime,desc)))

It seems that there were other similar errors

In the past 90 days, no transfer failed with "failed to verify size" on a third attempt, mostly on the second or a few on the first.

More investigation to come, but it seems strange enough to warrant a bug report.

PR #2533, opened 2025-07-25
Fix "failed to verify size" bug

patrickbrophy:
# `failed to verify size` bug investigation and solution
Described below is the investigation and solution of the bug described in #2475 
## Investigation
There were two critical bugs in the `downloadHTTP` function, manifesting in different ways:

1.  **Data Corruption:** A race condition could occur when a download was interrupted by a timeout (e.g., `SlowTransferError` or `StoppedTransferError`). The main function would return, but the background goroutine performing the `io.Copy` from the response body could remain active for a short period. This "zombie" goroutine could write one final buffer of data to the destination file *after* the function had exited. When the client logic initiated a subsequent transfer attempt (often from a different cache), it would resume from an incorrect offset, leading to duplicated data and a file that was larger than expected on disk. This explains the observed file size discrepancies.

2.  **Goroutine Leak:** The channel (`done`) used to signal the completion of the `io.Copy` goroutine was unbuffered. If the main function exited before the background goroutine finished, there would be no receiver on the channel.

## Solution

1.  **Guaranteed Execution:** A `defer` function is added at the top of `downloadHTTP`. This function is guaranteed to execute just before `downloadHTTP` returns, regardless of which code path is taken (success, timeout, or any other error).

2.  **Orderly Shutdown Sequence:** The deferred function enforces a strict and safe shutdown sequence for every exit:
    *   It first calls `cancel()` on the context, which signals the `io.Copy` goroutine to abort its operation immediately.
    *   It then waits (`<-done`) for the `io.Copy` goroutine to fully terminate and signal its completion. This is the crucial step that prevents the race condition, as it ensures no more bytes can be written to the file.
    *   Once the goroutine is confirmed to be stopped, the deferred function reads the final, accurate number of bytes transferred from the progress writer.

3.  **Accurate Final State:** The final byte count is used to update the named return variable `downloaded`. Furthermore, if the function is returning an error object that contains a byte count (like `SlowTransferError`), that object is also updated. This ensures that all callers and logs receive the true final state of the transfer, preventing incorrect resumption offsets.

This deferred cleanup model holistically solves both the data corruption race condition and the goroutine leak, and it makes the function more resilient to future modifications.

## Testing
In the TestSlowTransfers test, we now use the `goleak` package to test for the presence of a goroutine leak. Reproducing the data corruption was difficult but

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -2773,18 +2773,46 @@ func downloadHTTP(ctx context.Context, te *TransferEngine, callback TransferCall
 	// Run the GET request, reading from the response body, in a separate goroutine.
 	// The main go routine will wait for the download to finish and send out progress
 	// updates.
+
+	// This channel is used to signal the main goroutine that the download is done
+	// In a deferred function, we wait on the done channel to get the signal that the download is done
 	done := make(chan error, 1)
 	pw := &progressWriter{
 		writer: writer,
 	}
-	// IDEA: we can use the same context with the cancel to listen to the for done channel
+
 	go func() {
-		defer resp.Body.Close()
+		// Copy the response body to the progress writer
+		// When we are done, we will send an error (could be nil) to the done channel
+		// and then close the done channel
 		_, err := io.Copy(pw, resp.Body)
-		done <- err
-		close(done)
+		done <- err // send the error
+		close(done) // signal that the download is done
 	}()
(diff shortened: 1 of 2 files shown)


bbockelm (approved):
LGTM.  Did a quick rebase to fix a merge conflict.

PR #2536, opened 2025-07-25
Fix deadlock where `results` channel is not read after cancellation

bbockelm:
The results channel is not read after the TransferEngine's context has been cancelled.  Any time where we write to the channel, we must have a switch statement and bail if the context has been cancelled.

Fixes a deadlock/hang observed in the wild.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -1876,22 +1876,31 @@ func runTransferWorker(ctx context.Context, workChan <-chan *clientTransferFile,
 				}
 			}
 			if file.file.ctx.Err() == context.Canceled {
-				results <- &clientTransferResults{
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case results <- &clientTransferResults{
 					id: file.uuid,
 					results: TransferResults{
 						JobId: file.jobId,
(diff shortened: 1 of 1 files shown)


brianaydemir:
@bbockelm I'm still seeing a deadlock, but it's not as consistent (using the original reproducer, anyway).

- [stash_plugin.txt](https://github.com/user-attachments/files/21470535/stash_plugin.txt)
- [delve.txt](https://github.com/user-attachments/files/21470536/delve.txt)

Goroutine 189 is stuck on the following channel send:

https://github.com/bbockelm/pelican/blob/a62a20b05e19c1bd4b303389c852533ac5f571dc/client/handle_http.go#L1833-L1835

bbockelm:
@brianaydemir - the place where you note above is a second goroutine that runs in the client.  Previously, you hit issues in the routines that execute the transfers; this time, it was the one that converted transfer requests to individual objects (e.g., if you request a recursive download, a separate goroutine does nothing but walks directories and feeds the queues of stuff to do).

Indeed, found two more spots in that goroutine that didn't check for cancelation (most others were covered) and pushed a fix in the latest commit.  Please try again.

PR #2547, opened 2025-08-05
Don't change the perms and ownership of TLS key file

h2zh:
This PR replaces another PR to fix the linked issue (see the [reason](https://github.com/PelicanPlatform/pelican/pull/2540#issuecomment-3156084793))

With this PR, Pelican won't programmatically chmod/chown the `tls.key`, `tls.crt`, so the program won't hit the "read-only external file system" error. Admin needs to change the perms and ownership of them beforehand to the following values, to allow the TLS certs auto update in the drop-privs mode.

owned by: [user: pelican, group: root]. file perms: 0640


patrickbrophy (approved):
LGTM

Issue #2528, opened 2025-07-24
Enable Fireflies in XRootD to allow for network flow marking


patrickbrophy:
After a discussion with Shawn McKee, we talked about what was necessary to enable Fireflies in Pelican. It requires updating the XRootD config to use the `pmark` [configuration](https://xrootd.web.cern.ch/doc/dev57/xrd_config.htm#_Toc171719982) to enable fireflies.

PR #2555, opened 2025-08-07
Add fireflies config to cache and origin configs

patrickbrophy:
This PR addresses issue #2528. This PR enables network flow marking in both the Origin and Cache. This PR updates the XRootD configuration for both the Origin and the Cache.


bbockelm:
@patrickbrophy -

I know vaguely what the fireflies are meant to do but relatively little about what the configuration changes do.
- What sort of data does this generate?  Is it a bunch of unlabeled garbage or does it produce value?
- Are there any operational concerns?  Privacy issues?

Can we put this around a hidden config knob, defaulted to off, and then with the scitags team, selectively turn it on for a cache or two?

I'm a bit reluctant to go full-bore ahead until we understand that we're producing value for them!

Issue #2548, opened 2025-08-05
Can't recursively copy to /dev/null


matyasselmeci:
Pelican Version:
7.18.0

Pelican Service:
- [x] Client

Attempting to download recursively to `/dev/null` results in errors because `/dev/null` is not a directory:

```
$ pelican object get -c dtn-pas.denv.nrp.internet2.edu:8443 -r osdf:///pelicanplatform/test/ /dev/null

ERROR[2025-08-05T14:49:03-05:00] Error when attempting to transfer object /pelicanplatform/test//testfile-500M.md5 for client 01987bc7-b108-7954-965a-806f4e7d693f:
stat /dev/null/testfile-500M.md5: not a directory
```
It would be useful to be able to do this for testing.

PR #2560, opened 2025-08-08
Allow for Client to recursively download to `/dev/null`

patrickbrophy:
This PR addresses issue #2548. This PR allows the client to recursively download to `/dev/null`.


turetske (changes requested):
Can you write a test that creates a job with targetPath of os.DevNull and ensures that it actually works properly?

Issue #2534, opened 2025-07-25
`object get` with `?pack=auto` seg faults when the destination does not exist


brianaydemir:
### Pelican Version

```
[local:tmp]$ pelican --version
Version: 7.17.2
Build Date: 2025-07-02T16:50:11Z
Build Commit: 2fb9fbcfa85fa8fa4d8229c73c458a8661eed7ae
Built By: goreleaser
```

### Describe the bug

Exactly what the title says.

### To Reproduce

```
[local:tmp]$ ls -a
./	../
[local:tmp]$ pelican object get 'osdf://pelicanplatform?pack=auto' ./asdf
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x46e8ca0]

goroutine 178 [running]:
github.com/pelicanplatform/pelican/client.downloadObject(0xc000001140)
	/home/runner/work/pelican/pelican/client/handle_http.go:2050 +0xa40
github.com/pelicanplatform/pelican/client.runTransferWorker({0x5ce9ca8, 0xc000d8a320}, 0xc000dc4af0, 0xc000dac2a0)
	/home/runner/work/pelican/pelican/client/handle_http.go:1867 +0x207
github.com/pelicanplatform/pelican/client.NewTransferEngine.func1()
	/home/runner/work/pelican/pelican/client/handle_http.go:798 +0x2d
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.12.0/errgroup/errgroup.go:78 +0x50
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.12.0/errgroup/errgroup.go:75 +0x96
```

### Other Remarks

The problems are:

1. If/when we create the directory specified by `localPath`, we never update `info`.

2. We do not check whether `info` is `nil` at line 2050.

https://github.com/PelicanPlatform/pelican/blob/2fb9fbcfa85fa8fa4d8229c73c458a8661eed7ae/client/handle_http.go#L2031-L2050

PR #2561, opened 2025-08-08
Fix segfault in `object get` with `pack=auto` when destination directory doesn't exist

patrickbrophy:
This PR resolves issue #2534 by fixing a segmentation fault that occurred in the client when attempting to download an object with pack=auto into a non-existent directory.


turetske (changes requested):
Write a unit test that would have caused a segfault or the error before the fix and ensure the fix allows it to pass.

Issue #2472, opened 2025-07-07
Create "Commands Reference" page for the docs website


aowen-uwmad:
Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [x] Other: Docs

**Is your feature request related to a problem? Please describe.**
There is no standalone reference describing the available commands, options for the `pelican` command.

For example, if I don't remember if the flag is `--token` or `--cred` for the client, I have to dig through the manual until I find the [Additional Flags](https://docs.pelicanplatform.org/getting-data-with-pelican/client#additional-flags) section of the [Command Line Client](https://docs.pelicanplatform.org/getting-data-with-pelican/client) subpage of the [Getting Data with Pelican](https://docs.pelicanplatform.org/getting-data-with-pelican) page of the docs website.

**Describe the solution you'd like**
Top level page linking to a commands reference, analogous to the [HTCondor manual](https://htcondor.readthedocs.io/en/latest/man-pages/index.html). 

**Additional context**
Presume that said page could be autogenerated from the help text(s) code of the commands.

jhiemstrawisc:
For posterity, we're looking at exploring Cobra's built-in docs generator for Markdown:
https://github.com/spf13/cobra/blob/main/site/content/docgen/md.md

PR #2562, opened 2025-08-08
Add CLI documentation generator

patrickbrophy:
This PR addresses issue #2472. This PR allows for documentation to be automatically generated using the `go generate` command. A minor implementation note, I had to put the generation code in the `cmd` package because of how Go handles accessing exported variable inside the `main` package.

### DIFF
--- a/docs/app/commands-reference/server/_meta.js
+++ b/docs/app/commands-reference/server/_meta.js
@@ -1,3 +1,4 @@
 export default {
+    "database": "pelican server database",
     "set-logging-level": "pelican server set-logging-level",
 }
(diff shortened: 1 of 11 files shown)


h2zh:
The approach to auto-generating the CLI docs is really clever. Having said that, I think what's implemented in this PR doesn't totally reflect the requirements in the ticket created by @aowen-uwmad. My understanding was that the goal is to have the docs live on the documentation website directly. The current implementation seems to require users to run a manual command to generate and view them.

Btw, sorry for tagging you on a Friday before your vacation! Please correct me if I get it wrong.

patrickbrophy:
Yes the generator takes that into account. A command is "hidden" when the [`Command.Hidden`](https://pkg.go.dev/github.com/spf13/cobra#Command) field is set to true. Below is a screenshot showing that the `object delete` command is hidden:
image

patrickbrophy:
I have updated the layout to be less nested and a lowercase like how our commands are written in the terminal.

image

Issue #2565, opened 2025-08-11
Local Cache should ignore SIGHUP


patrickbrophy:
@timtheisen reported that when HTCondor is re-configuring, it sends a SIGHUP to all daemons, then pelican local cache dies and is restarted by HTCondor 10 seconds later. After discussion at the condor dev meeting, the consensus landed on that the Pelican local cache should gracefully handle/ignore a SIGHUP signal.

timtheisen:
Brian Bockelman said the pelican should restart (exec itself) when a SIGHUP is received.

PR #2567, opened 2025-08-12
Restart Pelican when SIGHUP is received

patrickbrophy:
This PR addresses issue #2565. This PR makes it so, when Pelican receives a SIGHUP, the process will restart.


patrickbrophy:
I've thought about this more, I think the best approach is to have the timeout. The issue behind this PR comes from the local cache and I believe it would be worth while to follow the same approach of draining the client requests.

h2zh (approved):

Issue #2483, opened 2025-07-10
Registry Naming Overhaul


h2zh:
Pelican Service:
- [X] Client
- [ ] Plugin
- [X] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

**Is your feature request related to a problem? Please describe.**
Currently, Pelican lacks an useful stable identifier for each server, and uses inconsistent naming. This creates problems for the upcoming broker system, which requires unique id for routing requests and token-based authentication, and many other scenarios. 

**Describe the solution you'd have**
Migrate server namespace to new table schemas as the diagram illustrates:
<img width="770" height="614" alt="Image" src="https://github.com/user-attachments/assets/fc824baa-de26-48ce-a938-754136d60036" />

PR #2589, opened 2025-08-25
Fix duplicate name handling in the registration process

h2zh:
This PR fixes the problems found by William in Pelican 7.19.0-rc.0 testing and other minor issues.

After the Registry ID Overhaul (#2483), a server could have more than one service (Origin/Cache). There are two ways to register a service on a server in the Registry: CLI auto-registration adn Registry WebUI.

When a request wants to register a new service (Origin/Cache) on an existing server, we need to verify they are allowed to do so by checking if they have the same public keys.

- For CLI auto-registration, we add an extra step: check if the signature they present can be verified by the existing server's keyset


turetske (approved):

turetske (changes requested):
One small change to a comment requested.

turetske (approved):

PR #2593, opened 2025-08-27
Token fetch command

turetske:
A hidden token fetch command.

This is basically a small command that patches together a pieces of token code that already exist to return an existing token from the client for a given namespace. Will call out to a remote issuer if a token doesn't exist where the client expects it and a remote issuer does exist.


jhiemstrawisc (changes requested):

cmd/token.go
```diff
@@ -57,10 +57,24 @@ var (
...
+		Hidden: true,
+	}
 )
 
 func init() {
 	tokenCmd.AddCommand(tokenCreateCmd)
+	tokenCmd.AddCommand(tokenFetchCmd)
+
+	// Token fetch capabilities
+	tokenFetchCmd.Flags().BoolP("write", "w", false, "Fetch a token with the ability to write to or modify the specified resource. ")
```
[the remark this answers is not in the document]
turetske: I'm basing this off of the token generation code, where if all there for capabilities is "read" or "write", so there isn't a point to a "modify" flag. It would just. become a "write" in the TokenGenerator.

cmd/token.go
```diff
@@ -370,3 +385,53 @@ func createToken(cmd *cobra.Command, args []string) error {
 	fmt.Println(tok)
 	return nil
 }
+
+// Fetch a token for a given Pelican URL using the TokenGenerator
```
jhiemstrawisc, line 389: Can you add more description here, e.g.
```
// Fetch a token for a given Pelican URL using the TokenGenerator.
// Tokens are first fetched from the client's on-disk token cache,
// and if none are available an OAuth2 flow is initiated against the
// discovered issuer
```

cmd/token.go
```diff
@@ -370,3 +385,53 @@ func createToken(cmd *cobra.Command, args []string) error {
...
+	method := http.MethodGet
+
+	if write {
+		oper = config.TokenWrite
+		method = http.MethodPut
+	}
+	if modify {
+		oper = config.TokenDelete
+		method = http.MethodPut
+	}
```
jhiemstrawisc, line 419: This assumes token operations are mutually exclusive and that you'll never want a token for more than one operation. If that's the case, the behavior should be documented in in the CLI and the flags should be marked as mutually exclusive.

cmd/token.go
```diff
@@ -50,17 +50,32 @@ var (
...
+		Hidden: true,
+	}
 )
 
 func init() {
 	tokenCmd.AddCommand(tokenCreateCmd)
+	tokenCmd.AddCommand(tokenFetchCmd)
+
+	// Token fetch capabilities
+	tokenFetchCmd.Flags().BoolP("write", "w", false, "Fetch a token with the ability to write to or modify the specified resource. ")
```
jhiemstrawisc, line 77: Let's also have a `--read` flag so that token scopes are explicitly requested (the user has to know what they're asking for).

In fact, if these are the same exact flags used with the token create command, create a set of flags and apply them to both commands so we don't have two sets to manage.

cmd/token.go
```diff
@@ -50,17 +50,32 @@ var (
 	}
 
 	tokenCreateCmd = &cobra.Command{
-		Use:   "create",
+		Use:   "create --read|-r | --write|-w | --modify|-m | --stage|-s <pelican-url>",
```
jhiemstrawisc, line 53: Can you remove the flags enumeration? Invoking the CLI with `-h` already indicates the flags. It's only the required pelican URL that wasn't stated.

cmd/token.go
```diff
@@ -50,17 +50,32 @@ var (
...
 		Short: "Create a token",
 		RunE:  createToken,
 		Args:  cobra.ExactArgs(1),
 		Example: "To create a read/write token for /some/namespace/path in OSDF: " +
 			"pelican token create --read --write pelican://osg-htc.org/some/namespace/path",
 	}
+
+	tokenFetchCmd = &cobra.Command{
+		Use:   "fetch --write|-w | --modify|-m <pelican-url>",
```
jhiemstrawisc, line 62: Same comment about flags here.

jhiemstrawisc (approved):

Issue #2596, opened 2025-08-29
Integrate XrdCl-Pelican metrics


patrickbrophy:
Now in the in xrdcl-pelican plugin, we get monitoring data produced at the location specified via the `XRD_CURLSTATISTICSLOCATION` environment variable. In the future this data will be consumable over the g-stream, but for now only file based access is supported. Below is an example of the data:
```json
{
  "event": "xrdclcurl",
  "start": 1754850407.891171,
  "now": 1754850498.00161,
  "file": {
    "prefetch": {
      "count": 5,
      "expired": 5,
      "failed": 0,
      "reads_hit": 1792,
      "reads_miss": 0,
      "bytes_used": 58720256
    }
  },
  "workers": {
    "oldest_op": 1754850496.995429,
    "oldest_cycle": 1754850496.995425,
    "http_HEAD_307_duration": 0.00053,
    "http_HEAD_307_finished": 5,
    "http_HEAD_405_duration": 0.003222,
    "http_HEAD_405_finished": 12,
    "http_HEAD_preheader_duration": 2.024531,
    "http_HEAD_started": 17,
    "http_GET_206_duration": 334.258429,
    "http_GET_206_pause_duration": 301.398048,
    "http_GET_206_bytes": 58720256,
    "http_GET_206_client_timeout": 5,
    "http_GET_preheader_duration": 0.591125,
    "http_GET_started": 5,
    "http_OPTIONS_200_duration": 0.000472,
    "http_OPTIONS_200_finished": 5,
    "http_OPTIONS_405_duration": 0.000621,
    "http_OPTIONS_405_finished": 5,
    "http_OPTIONS_preheader_duration": 0.741525,
    "http_OPTIONS_started": 10,
    "http_PROPFIND_207_duration": 0.001585,
    "http_PROPFIND_207_bytes": 11152,
    "http_PROPFIND_207_finished": 17,
    "http_PROPFIND_preheader_duration": 1.046161,
    "http_PROPFIND_started": 17,
    "conncall_error": 0,
    "conncall_started": 0,
    "conncall_success": 0,
    "conncall_timeout": 0
  },
  "queues": {
    "produced": 1821,
    "consumed": 1821,
    "pending": 0,
    "rejected": 0
  }
}```


We should consume this data and then integrate it into Prometheus metrics.

PR #2597, opened 2025-08-29
xrdcl-pelican metrics

patrickbrophy:
This PR addresses issue #2596. This PR consumes metrics from the xrdcl-pelican. It creates a new routine to periodically gather the new statistics and then funnels them along into Prometheus metrics.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -1718,7 +1718,7 @@ description: |+
 
   ```yaml
   Cache:
-    ClientStatisticsLocation: "$Cache.RunLocation/xrootd.stats"
+    ClientStatisticsLocation: "${Cache.RunLocation}/xrootd.stats"
   ```
 
   The XRootD process will periodically write JSON-formatted statistics to the specified file, which can be read by external monitoring (e.g., `jq`).
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -1718,7 +1718,7 @@ description: |+
 
   ```yaml
   Cache:
-    ClientStatisticsLocation: "/tmp/xrootd.stats"
+    ClientStatisticsLocation: "$Cache.RunLocation/xrootd.stats"
   ```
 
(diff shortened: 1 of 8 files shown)


metrics/xrdcl_curl_stats.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/pelicanplatform/pelican/param"
+)
+
+// This function should be used up until XRootD v6 is released
+// When XRootD v6 is released, these stats will be available over the g-stream
+// Until then the stats are consumed from Cache.ClientStatisticsLocation
+// When XRootD v6 is released, this function should be removed along with Cache.ClientStatisticsLocation
+func LaunchXrdCurlStatsMonitoring(ctx context.Context, egrp *errgroup.Group) {
+	egrp.Go(func() error {
+		ticker := time.NewTicker(time.Second * 30)
```
bbockelm, line 20: The stats are updated every 5 seconds.  Let's use that interval.

bbockelm (changes requested):

bbockelm:
Really minor changes requested - please use a directory controlled by Pelican.  Once that's done, let's merge.

docs/parameters.yaml
```diff
@@ -1718,7 +1718,7 @@ description: |+
 
   ```yaml
   Cache:
-    ClientStatisticsLocation: "/tmp/xrootd.stats"
+    ClientStatisticsLocation: "$Cache.RunLocation/xrootd.stats"
```
jhiemstrawisc, line 1721: We're not 100% consistent in our parameters yaml, but I think our convention is to use the `${}` bash-ism for describing multi-level configs elsewhere in our docs. Can you switch this to `${Cache.RunLocation}`?

jhiemstrawisc (approved):
Pre-approving because my change request is small.

jhiemstrawisc:
It looks to me like Patrick addressed BrianB's review comments, so I'll bypass protections to merge this instead of waiting for a re-review.

PR #2598, opened 2025-08-29
Use namespace prefix token issuer as X-Pelican-Token-Generation issuer

jhiemstrawisc:
Looking through various comments and past bugfixes, I'm really unsure what the guiding philosophy surrounding this header is supposed to be, and how that philosophy differs from the X-Pelican-Issuer header.

At a high level, there are issuers for a namespace and issuers for the Origin server itself, issuers for monitoring services and issuers for federation-related operations _at_ the Origin, but we've never formally explained how these interact, let alone the configuration mechanisms needed to cleanly split these as concepts. In any case, it seems like the token generation header should be using an issuer URL tied to the namespace itself.

This commit uses the first specified namespace issuer as the default issuer for clients to bootstrap tokens.


jhiemstrawisc:
Pinging @turetske to put this on the radar for a 7.19 release patch.

jhiemstrawisc:
@turetske, can you let me know when you patch this into 7.19?

Issue #2580, opened 2025-08-20
Update client to (almost) always hint at upload size


bbockelm:
In https://github.com/PelicanPlatform/pelican/pull/208, we stopped sending the uploaded object's content-length as part of the PUT request because we were seeing the golang HTTP client buffer the entire object in memory when we did this.

That doesn't make any sense to me: why would the golang client do such a thing?  It would surely be a strange behavior if we're the only folks on the internet that upload large files using Go!

We're finally at the point where we need to revisit & solve the mystery.  The S3 backend will soon need to know the upload size ahead of time because of other issues with multi-chunk uploads.

So,

1.  Let's re-enable `Content-Length` and see if we can figure out why the client was buffering the entire file and how to make it stop.  Hunt through golang documentation, test out the client for large uploads to reproduce, and see if there are workarounds (setting the content-length header instead of using the member variable?).
2.  If we can't re-enable `Content-Length`, XRootD can also take the size from the `oss.asize` query parameter (as in `/foo/bar.txt?oss.asize=1024` to declare an upload of 1024 bytes) and we can do that.

Given we're uploading a file on disk 99% of the time, we can always provide this as a hint.  The exception is when we're uploading a directory as a single compressed tarfile (since we don't know the resulting object size).  In that case, we can't provide a hint and we'll just have to fail for S3 backends.

PR #2601, opened 2025-08-29
Hint upload size

patrickbrophy:
This PR addresses issue #2580. After an extensive investigation, I had found that the client buffers the entire object into memory, when setting the `Request.ContentLength` field. I had also looked into setting the `Content-Length` header directly, and Go will ignore this. This PR adds the `oss.asize` query parameter to the upload to provide the hint. It only applies this query parameter if the object is greater than zero bytes and is not compressed.


bbockelm (approved):
Bummer that we had to do XRootD-specific query parameters -- but sounds like there aren't many other options available.

PR #2611, opened 2025-09-03
Bump xrdcl-pelican to v1.5.3

jhiemstrawisc:
This was on my mind after a ping from JIRA saying the rpms had beem promoted, so I figured I'd take care of it now. @turetske this should be patched into 7.19.

### DIFF
--- a/github_scripts/osx_install.sh
+++ b/github_scripts/osx_install.sh
@@ -54,7 +54,7 @@ ninja
 ninja install
 popd
 
-git clone --branch v1.5.1 https://github.com/PelicanPlatform/xrdcl-pelican.git
+git clone --branch v1.5.3 https://github.com/PelicanPlatform/xrdcl-pelican.git
 pushd xrdcl-pelican
 mkdir build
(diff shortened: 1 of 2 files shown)


patrickbrophy:
Thanks for taking care of this. The v7.18.2-rc.2 has the dependency updated in the Dockfile but not the osx installer. In any case this should align everything together once merged.

patrickbrophy:
Once the tests/builds pass I'll approve this.

jhiemstrawisc:
@patrickbrophy I think these failing MacOS tests are probably noise. I'm a bit concerned about them, but I double checked the diff between xrdcl-pelican  1.5.1 and 1.5.3 and didn't see anything obvious that would have the side effect of longer XRootD startup times only on Mac.

In the meantime, all the tests running against main are failing because xrdcl-pelican 1.5.1 is no longer in the repos. I think we should merge and watch future tests to make sure this isn't a real bug. We might just need to bump the default `WaitUntilWorking` timeout.

PR #2612, opened 2025-09-03
Comment out the Federation.DiscoveryUrl from `/etc/pelican/config.d/10-federation.yaml` from the `pelican-server` RPM

matyasselmeci:
We don't want it to override a valid Federation.DiscoveryUrl that was specified earlier in the config sequence, such as in `/usr/share/pelican/config.d`.

### DIFF
--- a/systemd/examples/10-federation.yaml
+++ b/systemd/examples/10-federation.yaml
@@ -5,9 +5,10 @@
 # Required configuration to integrate with your federation
 #####################################################
 
-Federation:
-  ## The main URL for your federation. You must specify this.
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Image builds are currently blocked on #2611 because the osg-testing repos no longer contain xrdcl-pelican 1.5.1 (they contain 1.5.3 instead); I'll talk to the other OSG folks about also including old versions of RPMs in osg-testing so we can avoid the situation later.

jhiemstrawisc (approved):
Short and sweet. Thanks for the fix!

Issue #2606, opened 2025-09-02
`osdf-*` RPM installs unset the OSDF Discovery URL configuration


jhiemstrawisc:
Pieced this together with @matyasselmeci's based on some config dumps provided by a user.

When installing v7.18 `osdf-cache`, the root of the config tree lives at `/etc/pelican/osdf-cache.yaml` and it defines:
```
ConfigLocations:
  - "/usr/share/pelican/config.d"
  - "/etc/pelican/config.d"
```

Pelican config files are parsed in the order they're defined, and if a config key is defined twice, the last value wins. There's a file at `/usr/share/pelican/config.d/10-osdf-defaults.yaml` that defines a `Federation.DiscoveryUrl`, but this value is unset by file parsed later on, `/etc/pelican/config.d/10-federation.yaml`, which defines:
```
Federation:
  ## The main URL for your federation. You must specify this.
  DiscoveryUrl:
```

It seems like the recent switch from `pelican` to `pelican-server` in more of our RPM installs is what introduced this second config file that unsets the Discovery URL. I haven't investigated which other RPMs this affects.

jhiemstrawisc:
Should this be patched into 7.19 since we haven't yet released it?

matyasselmeci:
Yeah I think this is bad enough that patching it into 7.19 would be a good idea

matyasselmeci:
@turetske can you backport it to 7.19? #2613 is the PR for it.

PR #2613, opened 2025-09-04
Comment out the Federation.DiscoveryUrl from `/etc/pelican/config.d/10-federation.yaml` from the `pelican-server` RPM

matyasselmeci:
We don't want it to override a valid Federation.DiscoveryUrl that was specified earlier in the config sequence, such as in `/usr/share/pelican/config.d`.

Backported from main (7.20), see #2606


turetske:
Closing this as I cherry-picked the actual original commit it (prefer that because it keeps the SHA the same)

Issue #2617, opened 2025-09-05
Add support to issuer to issue collections tokens and integrate group base acess


patrickbrophy:
We need to be able to issue collections tokens to able to access collections. We also need to be able to source groups in those tokens as well.

PR #2618, opened 2025-09-05
Groups-Collections Authorization Integration

patrickbrophy:
This PR addresses issue #2617. This PR allows for an authorization template to use a database defined group as a requirement for access. This PR also allows the Origin's built-in issuer to issue tokens with the collections scopes.

### DIFF
--- a/web_ui/oauth2_client.go
+++ b/web_ui/oauth2_client.go
@@ -168,19 +168,8 @@ func handleOAuthLogin(ctx *gin.Context) {
 // Given a user name, return the list of groups they belong to
 func generateGroupInfo(user string) (groups []string, err error) {
 	// Currently, only file-based and internal lookup is supported
-	if param.Issuer_GroupSource.GetString() != "file" || param.Issuer_GroupFile.GetString() != "internal" {
+	if param.Issuer_GroupSource.GetString() != "file" && param.Issuer_GroupFile.GetString() != "internal" {
 		return
-	}
-	if param.Issuer_GroupFile.GetString() == "internal" {
-		groupList, err := database.GetMemberGroups(database.ServerDatabase, user)
-		if err != nil {
-			return nil, err
-		}
-		groups = make([]string, 0, len(groupList))
-		for _, group := range groupList {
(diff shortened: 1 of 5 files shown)


web_ui/oauth2_client.go
```diff
@@ -166,8 +167,8 @@ func handleOAuthLogin(ctx *gin.Context) {
 
 // Given a user name, return the list of groups they belong to
 func generateGroupInfo(user string) (groups []string, err error) {
-	// Currently, only file-based lookup is supported
-	if param.Issuer_GroupSource.GetString() != "file" {
+	// Currently, only file-based and internal lookup is supported
+	if param.Issuer_GroupSource.GetString() != "file" && param.Issuer_GroupFile.GetString() != "internal" {
```
h2zh, line 171: It would be nice to convert both of them to lower cases before comparison. Also consider convert "file" and "internal" to const variables.

web_ui/oauth2_client.go
```diff
@@ -246,10 +247,25 @@ func generateUserGroupInfo(userInfo map[string]interface{}, idToken map[string]i
 				}
 			}
 		}
-	} else {
+	} else if param.Issuer_GroupSource.GetString() == "file" {
```
h2zh, line 250: same as above

web_ui/oauth2_client.go
```diff
@@ -246,10 +247,25 @@ func generateUserGroupInfo(userInfo map[string]interface{}, idToken map[string]i
...
 		groups, err = generateGroupInfo(user)
+		if err != nil {
+			return "", nil, err
+		}
+		return user, groups, nil
+	} else if param.Issuer_GroupSource.GetString() == "internal" {
```
h2zh, line 256: same as above

web_ui/oauth2_client.go
patrickbrophy: This actually looks like a logical error on my end, the second check in this statement _should_ be checking if the GroupSource is internal not if the GroupFile is internal. But also this check seem redundant to the caller making this check beforehand.

PR #2623, opened 2025-09-08
Autopopulate Federation Information with discoveryUrl Contents

### LINKED ISSUE #1476: Web UI with configured DiscoveryURL doesn't populate registry and director panes
When I run an origin setting only
```
Federation:
  DiscoveryUrl: foo
```
The web UI shows only the Discovery URL pane (lower right corner of UI). The UI for the origin and other services should use the discovered metadata to populate the other Director/Registry boxes.

CannonLock:
Updates the getFederationUrl to pull information from the discoveryUrl and then prefer displaying that information to those values configured directly.

Additionally no longer shows this value if in the federation and drops the topology url as well.

Closes #1476

### DIFF
--- a/web_ui/frontend/helpers/get/getFederationUrls.ts
+++ b/web_ui/frontend/helpers/get/getFederationUrls.ts
@@ -24,8 +24,6 @@ export const getFederationUrls = async () => {
       ? await discoverConfiguration(discoveryUrl)
       : {};
 
-    console.log(discoveredUrls);
-
     // Merge the two sets of URLs, with discovered URLs taking precedence
     return {
       ...federationUrls,
@@ -37,6 +35,10 @@ export const getFederationUrls = async () => {
   }
 };
 
+/**
+ * Pull federation URLs from the configuration and map to friendly names for display
+ * @param config
+ */
 const configurationToFederationUrls = (config: Config) => {
   return UrlData.reduce(
     (acc, { key, text }) => {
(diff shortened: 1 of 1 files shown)


web_ui/frontend/helpers/get/getFederationUrls.ts
```diff
@@ -0,0 +1,101 @@
...
+        acc[text] = url;
+      }
+      return acc;
+    },
+    {} as { [key: string]: string }
+  );
+};
+
+const UrlData = [
+  { key: ['Federation', 'NamespaceUrl'], text: 'Namespace Registry' },
```
h2zh, line 54: Where is `NamespaceUrl`? I checked `federationDiscoveryHandler` and didn't find this field.

web_ui/frontend/helpers/get/getFederationUrls.ts
```diff
@@ -0,0 +1,101 @@
...
+    // If there is a discovery URL, attempt to get the well-known configuration
+    const discoveryUrl = getObjectValue<string>(config, [
+      'Federation',
+      'DiscoveryUrl',
+    ]);
+    const discoveredUrls = discoveryUrl
+      ? await discoverConfiguration(discoveryUrl)
+      : {};
+
+    console.log(discoveredUrls);
```
h2zh, line 27: Debug logging left in

h2zh (approved):

PR #2624, opened 2025-09-08
Run condor_reconfig when the plugin is updated

### LINKED ISSUE #2110: Reconfig HTCondor when installing pelican-osdf-compat
Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [X] Other (please give the detail)

If pelican-osdf-compat is installed, HTCondor may not know that that a new version of Pelican is available.

I think that a post-install script in pelican-osdf-compat that would check if condor is running and if so run a condor_reconfig.

**Additional context**
_Add any other context or screenshots about the feature request here._

matyasselmeci:
Fixes #2110.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -307,6 +307,8 @@ nfpms:
     release: 1
     section: default
     priority: extra
+    scripts:
+      postinstall: "scripts/postinstall-plugin.sh"
     dependencies:
       - pelican
     provides:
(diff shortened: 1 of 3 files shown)


jhiemstrawisc:
> The GitHub interface for picking the linked issue did not let me pick 2110 no matter how I tried to enter it, so I added a comment in the PR description instead. This makes the PR validation fail; not sure what to do about that.

I've run into this before, and my hack is to link the two from the issue (which usually lets you find newer PRs). In any case, I think it's safe to ignore that failing action whenever we can see the issue is linked in another way.

Issue #2201, opened 2025-04-11
Error message handler UI doesn't wrap long text


h2zh:
Pelican Version:
7.15

Pelican Service:
- [ ] Client
- [ ] Plugin
- [x] Registry
- [x] Director
- [x] Origin
- [x] Cache
- [ ] Other (please give the detail)

**Describe the bug**
Error message surpasses the boundary of the error message handler UI.
![Image](https://github.com/user-attachments/assets/f2d2e35c-6077-417b-9883-35cd2fb61201)

**To Reproduce**
Trigger any error with a long error message.

**Expected behavior**
![Image](https://github.com/user-attachments/assets/91647877-2067-429d-9836-3231a4ad5cbe)

**Screenshots**
_If applicable, add screenshots to help explain your problem._

**Desktop (please complete the following information):**
 - OS: [e.g. Linux, MacOS]
 - Browser: [e.g. chrome, safari, if the bug relates to any web user interface]

**Additional context**
_Add any other context about the problem here._

PR #2626, opened 2025-09-08
Constraint width on alert so it doesn't go off screen

CannonLock:
100vw on phone and 70 on larger devices so that it doesn't go off the left of the screen.

Closes https://github.com/PelicanPlatform/pelican/issues/2201


h2zh:
Similar request to the other PR -  could you look at the linter issues and add labels?

PR #2630, opened 2025-09-09
Autofill Security Contact If From Origin

### LINKED ISSUE #2629: Autofill Registry only if coming from Origin
Autofill registry with the users cilogon only if they are coming from a origin registration.

CannonLock:
Only autofill the security contact if the user is coming from the Origin registration button.

This is a solid indicator that the person that is doing the registration is the one that is responsible for the service and therefore a good security contact.

Closes https://github.com/PelicanPlatform/pelican/issues/2629

### DIFF
--- a/web_ui/frontend/app/registry/components/Form.tsx
+++ b/web_ui/frontend/app/registry/components/Form.tsx
@@ -1,21 +1,10 @@
 import { Alert, Box, Button } from '@mui/material';
-import React, {
-  Dispatch,
-  SetStateAction,
-  useContext,
-  useState,
-  useEffect,
-} from 'react';
+import React, { Dispatch, SetStateAction, useContext, useEffect, useState, } from 'react';
 import useSWR from 'swr';
 
 import { RegistryNamespace } from '@/index';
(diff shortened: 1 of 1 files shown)


web_ui/frontend/app/registry/components/Form.tsx
```diff
@@ -71,10 +71,15 @@ const Form = ({ namespace, onSubmit }: FormProps) => {
...
-  // Auto-fill in the security contact if no security contact
+  // Auto-fill in the security contact if no security contact and request came from Origin
   const { data: user } = useSWR('getUser', getUser);
   useEffect(() => {
+    // If there is a fromUrl param then it came from the Origin
+    // We can assume this user is likely to be the security contact
+    const fromUrl = new URL(window.location.href).searchParams.get('fromUrl');
+
     if (
+      fromUrl &&
```
h2zh, line 82: I take a look at `getUser`. Maybe we should add two more conditions: `user.authenticated && user.user ` to double check?

h2zh (approved):

Issue #2633, opened 2025-09-09
Make director tests try multiple time with different issued tokens


patrickbrophy:
There was a backwards compatibility breakage in the multi director bug fixes. This break causes the old origins to reject tokens with the federation issuer. A quick fix would be to make multiple attempts, one with the federation issuer, and one with the director's external web url.

The fix should be patched into the 7.19.x series.

PR #2634, opened 2025-09-09
Fix director self-tests by making multiple attempts with different token issuers.

patrickbrophy:
This PR address issue #2633. This PR adds a second shot to the director tests with a different issued token. This should be patched into the 7.19.x series.

### DIFF
--- a/server_utils/test_file_transfer.go
+++ b/server_utils/test_file_transfer.go
@@ -71,10 +71,19 @@ func (t TestType) String() string {
 	return string(t)
 }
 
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (changes requested):
Can you lowercase the error strings?

jhiemstrawisc (approved):

h2zh (approved):
LGTM! For testing purpose, I tweaked the code locally to force it create token with`ExternalWebUrl` as `issuerUrl`. The Director test succeeds without 403 error.

Issue #2637, opened 2025-09-09
Panic in xrootd monitoring


patrickbrophy:
@brianaydemir observed the following panic:
```
DEBUG[2025-09-09T20:18:55Z] Received a summary statistics packet         
panic: counter cannot decrease in value

goroutine 474 [running]:
github.com/prometheus/client_golang/prometheus.(*counter).Add(0x1b24980?, 0xc0029c6840?)
	/root/go/pkg/mod/github.com/prometheus/client_golang@v1.17.0/prometheus/counter.go:128 +0xe9
github.com/pelicanplatform/pelican/metrics.HandleSummaryPacket({0xc001420000, 0x7ff, 0x10000})
	/app/ba-pelican/metrics/xrootd_metrics.go:1891 +0xcc8
github.com/pelicanplatform/pelican/metrics.handlePacket({0xc001420000?, 0x0?, 0x2cf6f80?})
	/app/ba-pelican/metrics/xrootd_metrics.go:1223 +0x126d
github.com/pelicanplatform/pelican/metrics.ConfigureMonitoring.func2()
	/app/ba-pelican/metrics/xrootd_metrics.go:1044 +0x1bf
created by github.com/pelicanplatform/pelican/metrics.ConfigureMonitoring in goroutine 1
	/app/ba-pelican/metrics/xrootd_metrics.go:1023 +0x5c5
```

This traces to this section of code https://github.com/PelicanPlatform/pelican/blob/277a1814908afc5c2ec776ef98badb30597413c7/metrics/xrootd_metrics.go#L1891

PR #2639, opened 2025-09-10
Fix panic in thread creations metric

patrickbrophy:
This PR addresses issue #2637. This PR changes the ThreadCreations metric from a counter to a gauge. This is because the underlying data from [XrootD is not monotonic](https://github.com/xrootd/xrootd/blob/f1de2038e0c5c10990769f5b7ff82200cc8c3d56/src/Xrd/XrdScheduler.cc#L700).

Note for reviewer:
Not sure if we should backport/patch this into 7.19, to include it into 7.20, or to wait and make it a 7.21 bug fix.

### DIFF
--- a/metrics/xrootd_metrics.go
+++ b/metrics/xrootd_metrics.go
@@ -473,9 +473,7 @@ var (
 		Help: "Number of scheduler threads",
 	}, []string{"state"})
 
-	// Unlink thread destructions this metric is not monotonic so we need to use a gauge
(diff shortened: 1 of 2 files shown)


patrickbrophy:
@bbockelm [this is what I was looking at](https://github.com/xrootd/xrootd/blob/f1de2038e0c5c10990769f5b7ff82200cc8c3d56/src/Xrd/XrdScheduler.cc#L730) when I made the switch to a gauge. 

Here you can see the counter be decremented if `retc` is non-zero.

If you really want this to be a counter then I can just add a simple check to only update when the change is positive.

PR #2650, opened 2025-09-15
Take out the script that auto formats webUI PR from GitHub Actions

CannonLock:
Currently the formatter runs on PR's that it cannot edit which is resulting in a error. This update checks that the PR is coming from a branch on this repository and not a fork before it tries to automatically fix things.

Only run the formatter if that action can write to the repository in question. 

If it comes form a fork then don't try to auto format.

Here is a example of the action running on a internal branch: https://github.com/PelicanPlatform/pelican/pull/2649


CannonLock:
@h2zh Yeah, it will barely ever run. I can also just take it out.

h2zh (approved):
LGTM. I rename this PR to reflect its up-to-date intent.

PR #2667, opened 2025-09-17
Update gorelease config to retain symbols in release builds

jthiltges:
Having symbols would be handy for debugging issues that appear in production deployments.

A build of the main branch with debug symbols gives a pelican-server binary of 76MB, vs 57MB in the v7.19.2 image. Note that this should leave optimization enabled.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -47,8 +47,6 @@ builds:
     ldflags:
       - &ldflags-metadata
         -X github.com/pelicanplatform/pelican/version.commit={{.Commit}} -X github.com/pelicanplatform/pelican/version.date={{.Date}} -X github.com/pelicanplatform/pelican/version.builtBy=goreleaser -X github.com/pelicanplatform/pelican/version.version={{.Version}}
-      - &ldflags-strip-symbols
-        -s -w
     ignore:
       - goos: windows
         goarch: arm64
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Let's keep `pelican-debug` around because we use that for the client (whereas John's change is for `pelican-server`).

.goreleaser.yml
```diff
@@ -47,8 +47,6 @@ builds:
     ldflags:
       - &ldflags-metadata
         -X github.com/pelicanplatform/pelican/version.commit={{.Commit}} -X github.com/pelicanplatform/pelican/version.date={{.Date}} -X github.com/pelicanplatform/pelican/version.builtBy=goreleaser -X github.com/pelicanplatform/pelican/version.version={{.Version}}
-      - &ldflags-strip-symbols
```
brianaydemir, line 50: The HTCondor dev team has concerns about this binary's size, so in the short term, we should continue to strip symbols from this binary.

brianaydemir (changes requested):
The primary goal of this PR is to retain symbols in the _server_ binary.

Due to concerns with the size of the client binary, I'd prefer that we continue stripping symbols from it, and instead ensure that we are actually installing `pelican-server` instead of `pelican`, where needed – for example, #2724. (We need to do that anyway in anticipation of those two binaries actually containing different functionality.)

.goreleaser.yml
jthiltges: I updated the PR to affect the symbols on the pelican-server build only. I'd really like to get a pelican-debuginfo at some point, but having server symbols is helpful progress!

Issue #2638, opened 2025-09-10
Improve 403 error classification and messages


patrickbrophy:
Current issue:
- 403 errors are not currently classified, making it difficult to query or filter them.
- The Pelican client already computes whether the token is expected to be authorized before use. The error message should include the result of that computation.

Expected improvement:
Error messages should distinguish between:
- Permission denied due to an expected condition (e.g., token expired).
- Permission denied when the token appeared valid but was rejected.

Issue #1895, opened 2025-01-14
Plain English messages for 403, 404 errors.


aowen-uwmad:
Pelican Service:
- [x] Client
- [x] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

**Is your feature request related to a problem? Please describe.**
Error messages for:

a) nonexistent namespace
b) existing namespace but no access permission
c) existing namespace with nonexistent object

are generally unhelpful for understanding and troubleshooting the error.

**Describe the solution you'd like**
Plain English explanation of the likely cause.

**Additional context**
When attempting to get/put data in cases (a) and (b), Pelican (usually) returns a series of HTTP 403 errors.
Most users are not aware that this means "permission denied" in http speak.
More importantly, because of (a), the problem may not be one of permissions but a typo in the namespace address.
Suggest adding something along the lines of "You either do not have permission to interact with this namespace or the namespace does not exist."

When attempting to get/put data in case (c), Pelican (usually) returns a series of HTTP 404 errors.
Most users are not aware that this means "not found" in http speak.
If occurring on GET, suggest adding something like "Unable to find the requested object with the specified namespace.".
(I assume the only PUT scenario this could occur in would actually return a 403 error: unable to write object to namespace, in which case the above message would be used.)

aowen-uwmad:
I'd like this to be addressed. I'm open to closing it as not planned, so long as it is explicit (and not because the bot did it).

PR #2669, opened 2025-09-18
Add Permission Denied error to Client/Plugin

patrickbrophy:
This PR addresses issue #2638. This PR adds a new permission denied error to the client. This data is consumed into the transfer ads produced by the plugin.


jhiemstrawisc:
@patrickbrophy this looks like a partial solution to #1895 as well. How much extra work would it be to add a classification for 404s to knock out a long-standing `Facilitation` issue?

Issue #2492, opened 2025-07-14
Ask consent before deleting namespace registrations


CannonLock:
On the registry before the api is called to delete a registration, ask for consent so that you don't misclick and delete something important.

PR #2670, opened 2025-09-18
Confirm close button

CannonLock:
Adds ConfirmClose Button to the UI. 

This is a button that needs two clicks for confirmation with the second being in a different place to prevent a quick double click from deleting two items.

https://github.com/user-attachments/assets/c1299ebf-8d39-4b17-a2e8-8fa4f504c663

Closes https://github.com/PelicanPlatform/pelican/issues/2492

Uses: https://path-cc.io/web-components/?path=/docs/components-confirmbutton--docs


h2zh:
Could you rebase this PR to take out the changes from the npm upgrade PR?

web_ui/frontend/components/Namespace/Card.tsx
```diff
@@ -143,25 +144,26 @@ export const Card = ({ namespace, authenticated, onUpdate }: CardProps) => {
                     </Tooltip>
                   )}
                   <Tooltip title={'Delete Registration'}>
-                    <IconButton
+                    <ConfirmButton
```
h2zh, line 147: This button looks smooth. Though I wonder is it possible to customize the text of this button? "Confirm" seems a little bit ambiguous here. An extreme scenario is that the user accidently clicks the delete icon (without realizing they do that), and the "Confirm" button happens to overlap other icons. User might think the "Confirm" button means approving this namespace. Maybe "Delete" or "Confirm Deletion" would be better?

web_ui/frontend/components/Namespace/DeniedCard.tsx
```diff
@@ -79,21 +80,22 @@ export const DeniedCard = ({ namespace, authenticated }: DeniedCardProps) => {
...
+											onClick={(e) => e.stopPropagation()}
+											onConfirm={async (e) => {
+												e.stopPropagation()
+												await alertOnError(
+													async () => await deleteNamespace(namespace.id),
+													'Could Not Delete Registration',
+													dispatch
+												);
+												await mutate(NAMESPACE_KEY);
+											}}
```
h2zh: This may be out of the scope of this PR. I notice a "size" variable in `Card.tsx` but not in this `DeniedCard.tsx`. I guess we should use it for all card components (card, deniedcard, pendingcard) to make the UI responsive
```
 const size = useMediaQuery((theme: Theme) => theme.breakpoints.down('md'))
    ? 'small'
    : 'medium';
```

h2zh (changes requested):

web_ui/frontend/components/Namespace/Card.tsx
CannonLock: Yeah that is a good idea. 

I also thought a containing background around the button and the Confirm would help clear up the association so I have modified the button to do this.

https://path-cc.io/web-components/?path=/docs/components-confirmbutton--docs

PR #2674, opened 2025-09-19
Unify labels on Geo Ip Error Reporting

CannonLock:


CannonLock:
From @jhiemstrawisc 

I think changing server_name --> name is okay, but we should keep project for the client metric -- this comes from the project set in HTCondor transfer ads

Issue #2648, opened 2025-09-15
Make local filesystem checks before attempting requests to the federation


williamnswanson:
Pelican Service:
- [X] Client
- [X] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

The client / plugin currently seems to attempt to contact the director about transfer outputs before checking if the file it's attempting to transfer exists and can be read.

Could the client / plugin check everything it can for the transfer before contacting the director / federation? (And make transfer Ads differentiable between failures with accessing output files and actual failures to contact the director)

PR #2677, opened 2025-09-19
Check local filesystem before contacting director on upload

patrickbrophy:
This PR addresses issue #2648. This PR makes the client check the local filesystem for the file to upload before contacting the director.

### DIFF
--- a/cmd/plugin_test.go
+++ b/cmd/plugin_test.go
@@ -903,10 +903,9 @@ func TestFailTransfer(t *testing.T) {
 		require.True(t, ok)
 		assert.NotEmpty(t, version)
 
-		// Check Attempts is 0 (no transfer attempts made)
 		attempts, ok := devDataMap["Attempts"]
 		require.True(t, ok)
-		assert.Equal(t, 0, attempts)
+		assert.Equal(t, 1, attempts)
 
 		// Check that TransferErrorData exists and has expected fields
 		transferErrorData, err := result.Get("TransferErrorData")


client/handle_http.go
```diff
@@ -3235,12 +3235,13 @@ func uploadObject(transfer *transferFile) (transferResult TransferResults, err e
 	if err != nil {
 		log.Errorln("Error checking local file ", transfer.localPath, ":", err)
 		if os.IsNotExist(err) {
-			err = error_codes.NewParameter_FileNotFoundError(err)
+			transferResult.Error = error_codes.NewParameter_FileNotFoundError(errors.Wrapf(err, "local file %q does not exist", transfer.localPath))
+		} else if os.IsPermission(err) {
+			transferResult.Error = error_codes.NewAuthorizationError(errors.Wrapf(err, "permission denied accessing local file %q", transfer.localPath))
 		} else {
-			err = error_codes.NewParameterError(err)
+			transferResult.Error = error_codes.NewSpecificationError(errors.Wrapf(err, "failed to stat local file %q", transfer.localPath))
```
turetske: NewParameterError

turetske (changes requested):
I have a few requests for changes, most focusing on the difference between a Parameter error vs a Specification error as well as less fragile way to do the tests.

client/director.go
```diff
@@ -267,11 +274,23 @@ func GetDirectorInfoForPath(ctx context.Context, pUrl *pelican_url.PelicanURL, h
...
-			err = errors.Wrapf(err, "error while querying the director at %s", pUrl.FedInfo.DirectorEndpoint)
-			// Wrap timeout errors specifically; leave other errors unwrapped until we know how to categorize them
-			var netErr net.Error
-			if errors.As(err, &netErr) && netErr.Timeout() {
-				err = error_codes.NewTransfer_DirectorTimeoutError(err)
+			// If not already a PelicanError, wrap it appropriately
+			var pe *error_codes.PelicanError
+			if !errors.As(err, &pe) {
+				// Check if this is a timeout error and use the appropriate retryable error type
+				if strings.Contains(err.Error(), "timeout") {
```
turetske: Why is this no longer a netErr.Timeout() check?

turetske (approved):
LGTM!

client/director.go
```diff
@@ -267,11 +275,19 @@ func GetDirectorInfoForPath(ctx context.Context, pUrl *pelican_url.PelicanURL, h
...
 			return
 		} else {
-			err = errors.Wrapf(err, "error while querying the director at %s", pUrl.FedInfo.DirectorEndpoint)
-			// Wrap timeout errors specifically; leave other errors unwrapped until we know how to categorize them
-			var netErr net.Error
-			if errors.As(err, &netErr) && netErr.Timeout() {
-				err = error_codes.NewTransfer_DirectorTimeoutError(err)
+			// If not already a PelicanError, wrap it appropriately
+			var pe *error_codes.PelicanError
+			if !errors.As(err, &pe) {
```
bbockelm: For future reference, I prefer the following approach

```
if foo {
   doSomething()
} else {
   doOther()
}
```

instead of

```
if !foo {
   doOther()
} else {
   doSomething()
}
```

That reduces the mental load of reading through the code.

PR #2679, opened 2025-09-22
Hardcode @chtc/webcomponents pre directory move

### LINKED ISSUE #2680: @chtc/web-components are not hardcoded
@chtc/web-components are '*' which has resulted in issue as it gets updated and Pelican relies on previous assumptions. Hardcode the version.

CannonLock:


h2zh (approved):
Discussed this offline after standup. LGTM once it gets an issue.

jhiemstrawisc:
I'm fine moving forward without an issue for this one -- it's an entirely internal change that fixes something that hasn't been released yet. If it doesn't end up in a metric, it should be okay.

PR #2686, opened 2025-09-22
Allow for unset issuer group source

patrickbrophy:
This PR fixes an issue observed by @williamnswanson. This bug was introduced by the new groups improvements merged in earlier in to the 7.20 release. 

@jhiemstrawisc We will need this in 7.20

### DIFF
--- a/oa4mp/serve.go
+++ b/oa4mp/serve.go
@@ -233,7 +233,7 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 
 	oidcAuthnUserClaim := param.Issuer_OIDCAuthenticationUserClaim.GetString()
 	groupSource := param.Issuer_GroupSource.GetString()
-	if groupSource != "" && groupSource != web_ui.GroupSourceTypeOIDC && groupSource != web_ui.GroupSourceTypeFile && groupSource != web_ui.GroupSourceTypeInternal {
+	if groupSource != "" && groupSource != "none" && groupSource != web_ui.GroupSourceTypeOIDC && groupSource != web_ui.GroupSourceTypeFile && groupSource != web_ui.GroupSourceTypeInternal {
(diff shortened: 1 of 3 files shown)


web_ui/oauth2_client.go
```diff
@@ -269,8 +269,8 @@ func generateUserGroupInfo(userInfo map[string]interface{}, idToken map[string]i
 		}
 
 	default:
-		err = errors.New("invalid group source: " + groupSource)
-		return "", nil, err
+		log.Debugf("No group source set, using none")
```
h2zh, line 272: Maybe a more descriptive comment would be better for debugging. For example, 
`err = errors.Errorf("No group source set, using %q by default", groupSource)`.

Actually, based on William's observation, we didn't set `groupSource` to `none`, but an empty string.

By the way, we should add "internal" to `Issuer.GroupSource` in `parameters.yaml`

oa4mp/serve.go
```diff
@@ -233,7 +233,7 @@ func ConfigureOA4MP() (launcher daemon.Launcher, err error) {
 
 	oidcAuthnUserClaim := param.Issuer_OIDCAuthenticationUserClaim.GetString()
 	groupSource := param.Issuer_GroupSource.GetString()
-	if groupSource != web_ui.GroupSourceTypeOIDC && groupSource != web_ui.GroupSourceTypeFile && groupSource != web_ui.GroupSourceTypeInternal {
+	if groupSource != "" && groupSource != web_ui.GroupSourceTypeOIDC && groupSource != web_ui.GroupSourceTypeFile && groupSource != web_ui.GroupSourceTypeInternal {
```
h2zh, line 236: According to `Issuer.GroupSource` in `parameters.yaml`, we should also rule out "none" in this check

h2zh (changes requested):

h2zh (approved):

Issue #2690, opened 2025-09-23
Users Table


CannonLock:
We have the concept of 'Users' of pelican but they are always stored as strings. 

I propose we create a `User` table to store these identities for the following reasons:

- Adds ability to attach more metadata from OIDC provider ( Like name ). So instead of just having a subiss, we have a name which we can more easily read.
- Makes the database tables easier to read. If I want to see where we are using the user I can grep for foreign keys.
- Enhances UI's with the ability to choose from a list of users when creating new entities.

The nice part is that this can be done completely under the table. Someone logs in with OIDC and we automatically create a user for them.

```
CREATE TABLE IF NOT EXISTS downtimes (
    ...
    created_by TEXT NOT NULL,
    updated_by TEXT NOT NULL,
    ...
);

CREATE TABLE group_members (
    ...
    member TEXT NOT NULL,
    added_by TEXT NOT NULL
    ...
);

CREATE TABLE collections (
    ...
    owner TEXT NOT NULL,
    ...
);

CREATE TABLE collection_members (
    ...
    added_by TEXT NOT NULL,
    ...
);

CREATE TABLE groups (
    ...
    created_by TEXT NOT NULL,
    ...
);
```

PR #2698, opened 2025-09-24
Split out group membership into new table

patrickbrophy:
This PR addresses issue #2690. This PR splits out group membership into a users table and adds QOL improvements to the group management APIs. This PR makes the `sub` and `issuer` the primary identifiers of a user and the `username` is a more user friendly identifier.

### DIFF
--- a/web_ui/oauth2_client.go
+++ b/web_ui/oauth2_client.go
@@ -228,6 +228,19 @@ func generateUserGroupInfo(userInfo map[string]interface{}, idToken map[string]i
 	}
 	user = userIdentifier
 
+	subIface, ok := claimsSource["sub"]
+	if !ok {
+		log.Errorln("User info endpoint did not return a value for the sub claim")
+		err = errors.New("identity provider did not return a subject for logged-in user")
+		return
+	}
+	sub, ok := subIface.(string)
+	if !ok {
+		log.Errorln("User info endpoint did not return a string for the sub claim")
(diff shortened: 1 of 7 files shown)


web_ui/oauth2_client.go
```diff
@@ -275,6 +307,13 @@ func generateUserGroupInfo(userInfo map[string]interface{}, idToken map[string]i
 		err = errors.Errorf("invalid group source: %s", groupSource)
 		return "", nil, err
 	}
+
+	// now that we have verified that the user belongs to a group we should create the user if it doesn't exist
+	_, err = database.GetOrCreateUser(database.ServerDatabase, user, sub, issuerClaimValue)
+	if err != nil {
+		return "", nil, err
+	}
+
```
CannonLock, line 316: We should create the user no matter what. That gives us more potential to have the user in our db when we go to add them to a group, making that action easier.

If they are already users then they should be enhanced to include the newly found sub + iss fields.

CannonLock (changes requested):
Looks good, two small tweaks.

h2zh (changes requested):
I like the way you create user from the login token. This is a good universal solution to persist user info in various services.

Throwing a bunch of comments here to make our future life easier.

Btw, it would nice to note it somewhere that there is another "group" concept in Pelican, representing the wlcg groups in the login cookie. See `GetUserGroups` func.

PR #2720, opened 2025-10-03
Added depth limited recursive listings to `pelican object ls`

patrickbrophy:
This PR adds the followings flags to the `pelican object ls` command:
- `--recursive`
- `--depth` 

These flags allow for recursive listings to be performed and to be depth limited.


patrickbrophy:
Unfortunately I don't think we can get paging on listings. This is due to the fact that we are directly querying XRootD using PROPFIND which afaik doesn't support pagination.

PR #2725, opened 2025-10-07
Switch to installing the `pelican-server` binary in all server images

brianaydemir:
Exactly what the PR's title says.

I tested this by building and running all eight server images, checking that they made it at least as far as executing the `pelican-server` binary.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -500,17 +500,17 @@ ENV JAVA_HOME="/usr/lib/jvm/jre" \
 #################################################################
 FROM pelican-software-base AS director
 ARG TARGETOS TARGETARCH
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
The changes overall look correct to me. My only question is why we're changing from `/usr/local/bin` to `/usr/local/sbin`? I don't know enough about linux packaging to know why we'd use one over the other.

brianaydemir:
@jhiemstrawisc `bin` vs. `sbin` is inherited from the fix for #1744 and the fact that that's where `entrypoint.sh` expects them to be.

`sbin` is the traditional directory for binaries used exclusively by the sys admin, if I recall correctly.

Issue #2708, opened 2025-09-30
Enable Persist on Successful Close (POSC) plugin


patrickbrophy:
The [POSC plugin](https://github.com/PelicanPlatform/xrootd-s3-http?tab=readme-ov-file#configure-the-posc-persist-on-successful-close-plugin) enables for objects only successfully closed to be visible under the namespace. This prevents incomplete uploads being visible under the namespace. The link above describes the behavior and necessary configuration to enable the plugin.

PR #2730, opened 2025-10-09
Enable Persist on Successful Close Plugin in Origin

patrickbrophy:
This PR addresses #2708. As discussed in the pelican dev meeting this prevents fully uploaded objects from appearing under the namespace.

### DIFF
--- a/server_utils/origin_posix.go
+++ b/server_utils/origin_posix.go
@@ -43,9 +43,31 @@ func (o *PosixOrigin) validateStoragePrefix(prefix string) error {
 	return validateFederationPrefix(prefix)
 }
 
-func (o *PosixOrigin) validateExtra(e *OriginExport, _ int) error {
-	// Check that UploadTempLocation is not under any StoragePrefix
-	// This prevents temporary upload files from being accessible via the federation namespace
+// validateStoragePrefixNotRoot rejects StoragePrefix values that resolve to "/".
+// Exporting the entire root filesystem is dangerous: anyone with write access could
+// overwrite security-policy files (e.g. auth/scitokens configuration), and every
+// absolute path on the system becomes reachable through the federation namespace.
+func (o *PosixOrigin) validateStoragePrefixNotRoot(e *OriginExport) error {
+	storageAbs, err := filepath.Abs(e.StoragePrefix)
+	if err != nil {
+		return errors.Wrapf(err, "unable to resolve absolute path for StoragePrefix '%s'", e.StoragePrefix)
+	}
+
+	if storageAbs == "/" {
+		return errors.Errorf("StoragePrefix '%s' resolves to the root filesystem '/' "+
+			"(export for federation prefix '%s'). Exporting '/' would make the entire filesystem accessible "+
+			"through the federation namespace, allowing overwrites of security-policy files and other sensitive data. "+
+			"Please use a more specific StoragePrefix.",
+			e.StoragePrefix, e.FederationPrefix)
+	}
+
+	return nil
+}
+
+// validateTempUploadLocation ensures that Origin.UploadTempLocation does not fall
+// under any export's StoragePrefix. If it did, in-progress upload files would be
+// reachable via the federation namespace, bypassing POSC's path-based isolation.
+func (o *PosixOrigin) validateTempUploadLocation(e *OriginExport) error {
 	uploadTempLocation := param.Origin_UploadTempLocation.GetString()
 	if uploadTempLocation == "" {
 		// If not set, it will use the default which is under RunLocation, so no conflict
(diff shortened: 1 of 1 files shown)


docs/parameters.yaml
```diff
@@ -913,6 +913,15 @@ root_default: /run/pelican/xrootd/origin
...
 components: ["origin"]
 ---
+name: Origin.InProgressLocation
+description: |+
+  A directory where objects being uploaded to the origin are temporarily stored until they are committed under the namespace.
+
+  This is only available for origins that have a StorageType of "posix".
+type: filename
+default: $Origin.RunLocation/in-progress
+components: ["origin"]
```
jhiemstrawisc: If you don't have an objection, can you mark this as hidden? Is there a concrete reason a generic Origin user would ever choose to use something other than the default, e.g. "this should be on a separate mount than the rest of the Origin's storage"?

docs/parameters.yaml
```diff
@@ -913,6 +913,15 @@ root_default: /run/pelican/xrootd/origin
 default: $XDG_RUNTIME_DIR/pelican/origin
 components: ["origin"]
 ---
+name: Origin.InProgressLocation
```
jhiemstrawisc: I'm not in love with this name because I don't think it hints strongly enough what process is "in progress". The problem is I don't know that I have a phenomenally better name... Maybe something like `Origin.POSCTempLocation`? `Origin.AtomicWritesTempLocation`? Other ideas?

e2e_fed_tests/posc_origin_test.go
```diff
@@ -0,0 +1,209 @@
...
+	// For unauthenticated uploads, files go under the "anonymous" subdirectory.
+	// Get the actual in-progress location from the running configuration.
+	inProgressLocation := param.Origin_InProgressLocation.GetString()
+	require.NotEmpty(t, inProgressLocation, "Origin.InProgressLocation should be set")
+	inProgressDir := filepath.Join(inProgressLocation, "anonymous")
+
+	t.Logf("In-progress directory: %s", inProgressDir)
+	t.Logf("Storage prefix: %s", storagePrefix)
+
+	// Create a large file with random content (1GB to ensure upload takes long enough to observe)
```
jhiemstrawisc: Test comment suggest `500MB` but this is `1GB`. I also wonder whether you can stream this onto disk as you generate the data instead of gobbling up 1GB of memory before writing it out.

xrootd/resources/xrootd-origin.cfg
```diff
@@ -59,6 +59,16 @@ all.adminpath {{.Origin.RunLocation}}
 all.pidpath {{.Origin.RunLocation}}
 {{if eq .Origin.StorageType "posix"}}
 oss.localroot {{.Xrootd.Mount}}
+
```
jhiemstrawisc: Can you leave a note in here about the explicit need to order these plugins the way you did?

xrootd/resources/xrootd-origin.cfg
```diff
@@ -59,6 +59,16 @@ all.adminpath {{.Origin.RunLocation}}
...
+{{- if .Origin.Multiuser}}
+ofs.osslib ++ libXrdMultiuser.so
+ofs.ckslib * libXrdMultiuser.so
+multiuser.checksumonwrite on
+{{- end}}
+
+ofs.osslib ++ libXrdOssPosc.so
+posc.trace {{.Logging.OriginOss}}
+posc.prefix /in-progress
```
jhiemstrawisc: Can you comment here on the relationship between `posc.prefix` and `oss.localroot`, as well as why you aren't grabbing the in progress location from the param?

xrootd/xrootd_config.go
```diff
@@ -265,6 +266,68 @@ func CheckOriginXrootdEnv(exportPath string, server server_structs.XRootDServer,
...
+
+		// Create the user specified in-progress directory with appropriate permissions
+		inProgressDir := param.Origin_InProgressLocation.GetString()
+
+		// Check if the directory exists
+		if stat, err := os.Stat(inProgressDir); os.IsNotExist(err) {
+			// Directory doesn't exist - create it with mode 1777 (world-writable + sticky bit)
+			log.Infof("Creating in-progress directory %v with mode 1777 (world-writable + sticky bit)", inProgressDir)
+			if err = config.MkdirAll(inProgressDir, 0777|os.ModeSticky, uid, gid); err != nil {
+				return errors.Wrapf(err, "Unable to create in-progress directory %v", inProgressDir)
```
jhiemstrawisc: Can you lowercase the error strings here and below?

xrootd/resources/xrootd-origin.cfg
```diff
@@ -59,6 +59,16 @@ all.adminpath {{.Origin.RunLocation}}
...
 {{if eq .Origin.StorageType "posix"}}
 oss.localroot {{.Xrootd.Mount}}
+
+{{- if .Origin.Multiuser}}
+ofs.osslib ++ libXrdMultiuser.so
+ofs.ckslib * libXrdMultiuser.so
+multiuser.checksumonwrite on
+{{- end}}
+
+ofs.osslib ++ libXrdOssPosc.so
```
jhiemstrawisc: We talked about making this optional with a config knob. I'll let you decide whether that's the route we want to go, since ultimately we really _do_ want this to be turned on.

e2e_fed_tests/posc_origin_test.go
```diff
@@ -0,0 +1,209 @@
...
+	_, err = os.Stat(finalFilePath)
+	assert.True(t, os.IsNotExist(err), "Final file should not exist after cancelled upload")
+
+	// Assert that a pelican object stat for the attempted uploaded file fails with a 404 error
+	_, err = client.DoStat(ft.Ctx, uploadUrl, client.WithTokenLocation(""))
+	require.Error(t, err, "Stat for cancelled upload file should return an error")
+	require.ErrorAs(t, err, &pe)
+	require.Equal(t, "Specification.FileNotFound", pe.ErrorType())
+
+	t.Log("POSC origin test passed: temporary files are properly managed during upload cancellation")
```
jhiemstrawisc: Since these log lines are usually only visible in our CI on failure, I don't know that this adds much.

server_utils/origin_posix.go
```diff
@@ -36,3 +42,42 @@ func (o *PosixOrigin) validateStoragePrefix(prefix string) error {
...
+	uploadTempLocation := param.Origin_UploadTempLocation.GetString()
+	if uploadTempLocation == "" {
+		// If not set, it will use the default which is under RunLocation, so no conflict
+		return nil
+	}
+
+	// Normalize paths for comparison
+	uploadTempAbs, err := filepath.Abs(uploadTempLocation)
+	if err != nil {
+		return errors.Wrapf(err, "unable to resolve absolute path for Origin.UploadTempLocation '%s'", uploadTempLocation)
```
jhiemstrawisc: Can you use the param name getter here?

server_utils/origin_posix.go
```diff
@@ -36,3 +42,42 @@ func (o *PosixOrigin) validateStoragePrefix(prefix string) error {
...
+	rel, err := filepath.Rel(storageAbs, uploadTempAbs)
+	if err != nil {
+		// This shouldn't happen if both paths are absolute, but handle it gracefully
+		return errors.Wrapf(err, "unable to determine relationship between StoragePrefix '%s' and UploadTempLocation '%s'", e.StoragePrefix, uploadTempLocation)
+	}
+
+	// If rel doesn't start with "..", then uploadTempLocation is equal to or under storagePrefix
+	// rel == "." means they're equal, and rel without ".." prefix means uploadTempLocation is under storagePrefix
+	if !strings.HasPrefix(rel, "..") {
+		return errors.Errorf("Origin.UploadTempLocation '%s' cannot be equal to or under StoragePrefix '%s' (export for federation prefix '%s'). "+
```
jhiemstrawisc: And here

PR #2739, opened 2025-10-14
Registry API updates to avoid stale server/service entry

h2zh:
This PR:

1. Rule out the stale entries in the Server API results
2. Provide a Swagger Docs for WebUI development
3. Check against the approved server name in auto-registration
4. See the commit messages for other changes

### Relevant PRs
3 PRs will work together to fix this the root cause behind the `/ndp/public` [problem](https://opensciencegrid.atlassian.net/browse/OPS-438) (will get into Pelican v7.21). The other two are: 
#2745
#2748

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -373,20 +373,7 @@ func keySignChallengeCommit(ctx *gin.Context, data *registrationData) (bool, map
 			}
 			// Record not found, continue with registration
 		} else if existingServer != nil && existingServer.ID != "" {
-			// Server exists, check approval status
-			// If the server has multiple service registrations, check if any of them is approved. If not, return an error.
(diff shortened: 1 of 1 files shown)


registry/registry_db.go
```diff
@@ -768,10 +808,48 @@ func setRegistrationPubKey(prefix string, pubkeyDbString string) error {
...
+	// Default: do not delete server
+	deleteServer := false
+	serverID := ""
+	if serverReg != nil && serverReg.ID != "" {
+		serverID = serverReg.ID
+		// If the server has exactly one service registration, then deleting it should also delete the server entry.
+		if len(serverReg.Registration) == 1 {
+			deleteServer = true
+		}
+	}
```
CannonLock: This is made redundant by the code inside the transaction.

h2zh: Good catch!

CannonLock (approved):
LGTM

PR #2745, opened 2025-10-15
Auto cleanup stale entries in `servers` and `services` tables

h2zh:
This PR:

1. Remove database mover code
  Because Registry DB migration (to pelican.sqlite) was completed in v7.19
2.  Automatically cleanup the stale entries in the `servers` and `services` tables

- They are caused by the damaged foreign key constraint in `services` table and we didn't update webUI to use server deletion API in time.
- See https://opensciencegrid.atlassian.net/browse/OPS-438

### Relevant PRs
3 PRs will work together to fix this the root cause behind the `/ndp/public` [problem](https://opensciencegrid.atlassian.net/browse/OPS-438) (all under review, will get into Pelican v7.21): 
#2745 
#2739
#2748

### DIFF
--- a/database/server.go
+++ b/database/server.go
@@ -107,9 +107,9 @@ func cleanupStaleServerEntries() error {
 		var staleServerIDs []string
 		rawQuery := `
 			SELECT DISTINCT servers.id
-			FROM services
+			FROM servers
+			LEFT JOIN services ON servers.id = services.server_id
 			LEFT JOIN registrations ON services.registration_id = registrations.id
-			LEFT JOIN servers ON servers.id = services.server_id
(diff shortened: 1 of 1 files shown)


h2zh:
@brianaydemir Right, that deletion might deviate from users' expectations for our backward compatibility support policies, and cause trouble if they bump the Registry from <=7.18 to 7.21. The migration code was introduced in 7.19 to consolidate the `registry.sqlite` into `pelican.sqlite`. I wanted to delete it because I thought we're the only entity that runs a Registry server. I reverted the deletion to maintain the backward compatibility support for a longer time.

PR #2747, opened 2025-10-15
Use federation's discovery URL as token issuer for Director/Federation tests

jhiemstrawisc:
This changes the issuer URL in Caches/Origins that's used to accept tokens needed for director/federation tests.

The core of the change was a simple change, but I had to do some extra work to handle the way I personally run test federations, where I don't set a federation root, but do hardcode the Director/Registry. Now, all services should discover a federation discovery URL regardless of how they're initially configured because they can fallback and discover things from the Director itself.

### DIFF
--- a/web_ui/authentication_test.go
+++ b/web_ui/authentication_test.go
@@ -50,6 +50,23 @@ import (
 	"github.com/pelicanplatform/pelican/token_scopes"
 )
 
+func migrateTestDB(t *testing.T) {
+	err := database.ServerDatabase.AutoMigrate(&database.Collection{})
+	require.NoError(t, err, "Failed to migrate DB for collections table")
+	err = database.ServerDatabase.AutoMigrate(&database.CollectionMember{})
+	require.NoError(t, err, "Failed to migrate DB for collection members table")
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -355,12 +355,43 @@ func GetAllPrefixes() []ConfigPrefix {
 
 // We can't parse a schemeless hostname when there's a port, so check for a scheme and add one if none exists.
 func wrapWithHttpsIfNeeded(urlStr string) string {
+	// The fact that we don't overwrite http:// with https:// appears to be an artifact from testing.
```
patrickbrophy: Good call on adding a warning when URLs are missing a scheme. That said, it might be worth considering making this a Debugf or Infof instead of Warningf, since test setups or local dev environments might intentionally omit schemes.

jhiemstrawisc:
For posterity, I'm commenting that Patrick was absolutely correct to request a unit test. I'd assumed I was minimally changing a function that already had some testing because of how important it is, but after Patrick's request I realized that `discoverFederationImpl` previously had _no_ tests. Writing one caught a small logic bug in my change. Unfortunately, that small logic bug, once fixed, required changing a lot of other tests that previously did not have a means to handle the new discovery requirement.

PR #2748, opened 2025-10-15
Enforce Foreign Key constraint on every database connection

h2zh:
This PR appends a foreign-keys parameter to the DSN at initialization to ensure the Foreign Key constraint is enforced in all database connections.

Analysis: SQLite PRAGMAs are per-connection. A one-time PRAGMA foreign_keys=ON in `InitServerDatabase` only affects the connection that executed it. Under k8s, more pooled connections are opened and may never run that PRAGMA, so cascade deletes don’t fire. DSN parameters ensure every connection has FKs on by default.

Test: This PR needs to be tested in a Registry component running in the k8s pod to ensure the Foreign Key constraint is enforced there.

Other PRs for this `/ndp/public` origin [problem](https://opensciencegrid.atlassian.net/browse/OPS-438):

- #2745 (done)
- #2739 (in progress, as of Oct 15 1PM)

### DIFF
--- a/database/server_test.go
+++ b/database/server_test.go
@@ -1,6 +1,9 @@
 package database
 
 import (
+	"database/sql"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


patrickbrophy (approved):
Thanks for this! LGTM

Issue #2350, opened 2025-05-23
Set server logging level default to "info"


brianhlin:
Pelican Service:
- [ ] Client
- [ ] Plugin
- [X] Registry
- [X] Director
- [X] Origin
- [X] Cache
- [ ] Other (please give the detail)

**Is your feature request related to a problem? Please describe.**
Pelican server logs are eerily quiet by default, to the point where administrators wonder if the service is actually *doing* anything.

**Describe the solution you'd like**
Let's set the default debug level to `info` for servers across all types of installations (container, RPMs, etc).

Issue #2336, opened 2025-05-21
Deprecate `Debug` configuration variable


brianhlin:
Pelican Service:
- [X] Client
- [X] Plugin
- [X] Registry
- [X] Director
- [X] Origin
- [X] Cache
- [ ] Other (please give the detail)

**Is your feature request related to a problem? Please describe.**
There are too many ways to control log levels and they interact in surprising ways: `Debug` overrides subsystem specific log levels, which override `Logging.Level`. The latter two make sense, the behavior in `Debug` is odd.

For example, set `Debug: true` and `Logging.Level: trace` and run `pelican config get logging` and you'll see that all the subsystem log levels are set to `debug`

**Describe the solution you'd like**
Announce deprecation of `Debug: true` and then retire it.

PR #2750, opened 2025-10-15
Deprecate `Debug` param and modify default client/server log levels

jhiemstrawisc:
As the title says, this attempts two things:
- The `Debug` config param is deprecated per @brianhlin's request in #2336
- Also per Brian's request in #2350 , the default log level for servers is cranked from `error` to `info` (and the default for clients is set to `warn`, primarily so that it still outputs deprecation warnings without getting too chatty).

I recommend reading commit messages for the larger commits because understanding some of my choices requires knowledge of how we handle logging internally. In my messages, I tried to explain which internal limitations I was working around.

Since the effects of config overlays (CLI flags vs yaml-provided config vs env vars) can be hard to probe in our testing framework, here are some of the tests I ran by hand to convince myself I achieved the desired behavior:

Servers:
- Run server with no explicit log level set (no `--debug`, no logging.level, etc) --> should resolve to info level
- Run server with `--debug` --> should resolve to debug level, but does not print warning about deprecated `Debug` param
- Run server with Debug in config --> should resolve to debug level and print deprecation warning 
- Run server with log level configured in pelican.yaml --> should get that config, no deprecation warning
- Run server with Debug and Logging.Level error --> should resolve to debug from override and print deprecation warning
- Run server with `--debug` and Logging.Level: error --> should resolve to debug and print no deprecation warning

Client:
- Run client with deprecated config key (e.g. DisableHttpProxy: true) -- you should get the warnings because internal log level is warning (you won't see explicitly in a config dump or in log messages unless you use a deprecated param because clients don't print the config unless you're at debug or higher)
- Run client with `--debug` --> should get debug output but no deprecation warning
- Run client with Debug in config --> should resolve to debug level and print deprecation warning
- Run client with log level configured in pelican.yaml  --> should get that config
- Run client with Debug and Logging.Level: error --> should resolve to debug from override and print warning
- Run client with `--debug` and Logging.Level: error --> should resolve to debug and print no deprecation warning

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -196,6 +196,7 @@ description: |+
 
   The default logging level for clients is "warn", whereas the default for servers is "info".
 type: string
+default: none
 client_default: warn
 server_default: info
 components: ["*"]


brianhlin:
> (and the default for clients is set to warn, primarily so that it still outputs deprecation warnings without getting too chatty).

It's not 100% clear to me if this is an increase in the default debug level. If so, is there a way to get a list of all the `warn` level messages to see if they're appropriate?

brianhlin:
Outside of the messages with "Error" in them, I think those all look like reasonable warnings

config/config.go
```diff
@@ -1814,6 +1843,11 @@ func ResetClientInitialized() {
 func SetClientDefaults(v *viper.Viper) error {
 	configDir := v.GetString("ConfigDir")
 
+	// Duplicate setting a default logging level so that this function picks picks up
+	// the one case where we need to set client/server defaults differently directly in
+	// InitConfigInternal. We need to do that because internal logging levels are set by
+	// InitServer/InitClient before we call SetClientDefaults/SetServerDefaults.
+	v.SetDefault(param.Logging_Level.GetName(), "warning")
```
h2zh: Though we already normalize “warn” vs “warning” in the code, we use “warn” in `Logging_Level`'s docs so I think we should stick to it.

config/config.go
```diff
@@ -764,17 +769,22 @@ func SetBaseDefaultsInConfig(v *viper.Viper) {
...
-		log.Warnf("Debug is set as a flag or in config, this will override anything set for '%s' within your configuration", param.Logging_Level.GetName())
-	} else {
-		logLevel := param.Logging_Level.GetString()
-		level, err := log.ParseLevel(logLevel)
-		if err != nil {
-			return errors.Wrapf(err, "failed to parse value of config param %s", param.Logging_Level.GetString())
-		}
-		SetLogging(level)
+		warnDebugOnce.Do(func() {
+			log.Warnf("The config param '%s' is set in your configuration, which will override any values set for '%s' ", param.Debug.GetName(), param.Logging_Level.GetName())
```
h2zh: There is a more elegant way for putting a string inside the quotes `'%s'` - check out `%q`

config/config.go
```diff
@@ -615,6 +616,10 @@ func handleDeprecatedConfig() {
 			if viper.IsSet(deprecated) {
 				if replacement[0] == "none" {
 					log.Warningf("Deprecated configuration key %s is set. This is being removed in future release", deprecated)
+				} else if deprecated == param.Debug.GetName() {
+					// Special case for the Debug key; we handle mapping it to Logging.Level: debug in
+					// `setLoggingInternal()` because that has already been executed by the time we get here.
+					log.Warningf("The configuration key '%s' is being deprecated. While debug logging has been enabled, you should set '%s' to 'debug' to achieve this behavior in the future.", deprecated, param.Logging_Level.GetName())
```
h2zh: This does not necessarily mean debug logging has been enabled - when user sets "Debug: false", this warning message will still be triggered.

config/config.go
```diff
@@ -1066,6 +1080,11 @@ func SetServerDefaults(v *viper.Viper) error {
 	configDir := v.GetString("ConfigDir")
 	v.SetConfigType("yaml")
 
+	// Duplicate setting a default logging level so that this function picks picks up
+	// the one case where we need to set client/server defaults differently directly in
+	// InitConfigInternal. We need to do that because internal logging levels are set by
+	// InitServer/InitClient before we call SetClientDefaults/SetServerDefaults.
+	v.SetDefault(param.Logging_Level.GetName(), "info")
```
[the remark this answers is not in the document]
jhiemstrawisc: Great catch. Since the `default` tag in our parameters yaml doesn't let me easily describe a client/server split, I'll use `info` for the `default` tag but describe the separate values in the text.

docs/parameters.yaml
```diff
@@ -193,8 +193,10 @@ description: |+
   A string defining the log level of the client. Options include (going from most info to least): Trace, Debug, Info, Warn, Error, Fatal, Panic.
 
   Log levels are inherited by all components unless explicitly overridden. Levels are case-insensitive.
+
+  The default logging level for clients is "warn", whereas the default for servers is "info".
 type: string
-default: error
+default: info
```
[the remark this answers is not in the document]
jhiemstrawisc: I pinged Cannon and he says the new `{client,server}_default` tags need to be added manually.

jhiemstrawisc:
The remaining test failures change on every GHA run and seem unrelated to the actual PR. I'm going to merge because I think we're chasing flaky tests here. Happy to own the consequences if that turns out not to be the case!

PR #2762, opened 2025-10-21
Allow passing of GORELEASER_CURRENT_TAG env down through containers

jhiemstrawisc:
This lets us use `--build-arg` to pass an argument to Docker for injecting a custom version into the Pelican binary when building a new image without the need to create tags.

For example, one could do something like this from the repo root: docker build -t jhiemstra/test-pelican:v7.21.3-debug.ABC1 --build-arg \
    GORELEASER_CURRENT_TAG=v7.21.3-debug.ABCD1 --target origin -f images/Dockerfile .

This is in order to help developers provide version info when they need to build/deploy custom versions.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -36,6 +36,7 @@ ARG OSG_REPO=release
 
 # For compiling Pelican.
 ARG GO_VER=1.24.7
+ARG GORELEASER_CURRENT_TAG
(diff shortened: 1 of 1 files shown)


brianaydemir (changes requested):
The design looks good to me. Just a couple of nitpicks.

PR #2773, opened 2025-10-28
Token and Authorization docs

jhiemstrawisc:
A few notes about this:
1. This section of docs is currently incomplete -- I'm checkpointing what I have until I can finish the second "Advanced" section of token docs that this page already links to.
2. This section is dependent on some new formatting I worked on for the upstream docs repo. That formatting lives at [this PR](https://github.com/PelicanPlatform/docs/pull/41), and this shouldn't be merged until that's upstreamed.

### DIFF
--- a/docs/app/advanced-concepts/auth/page.mdx
+++ b/docs/app/advanced-concepts/auth/page.mdx
@@ -196,7 +196,7 @@ Specifications that describe the extra contents of a JWT are called *profiles*, 
 The two profiles used in the Pelican ecosystem are the **WLCG Profile** and the **SciTokens** profile.
 
 #### **WLCG Profile**
-The Worldwide Large Hadron Collider (WLCG) auth group maintains the WLCG token profile.
+The [Worldwide LHC Computing Grid](https://wlcg.web.cern.ch/) (WLCG) auth group maintains the WLCG token profile.
 Some of the WLCG scopes/capabilities Pelican uses include:
 	- `storage.read:/path/to/resource`: grants the ability to *read* the specified resource
 	- `storage.create:/path/to/resource`: grants the ability to *create* but **not** modify the specified resource


jhiemstrawisc:
For the reviewers, please think carefully about the content of what I'm saying, especially in the "advanced" section -- while I used some AI to help with formatting and clarity, 99% of the content first traveled through my fingers. I likely made mistakes! In particular, it'd be useful for others to think about the descriptions/diagrams in the section about establishing trust, which turned out to be pretty tricky and I'm not totally sure I got it right.

I also highly recommend you actually build these docs to see formatting, double check my links, etc. I tried to double check everything before requesting a review, but there's a lot of content and I'm becoming blind to it because of how long I've been staring at it...

jhiemstrawisc:
I should also note that these docs still aren't a complete "everything auth in Pelican" -- I'm punting on things like "how to set up the built-in issuer". I think some of those topics can/should be added in a followup PR.

docs/app/advanced-usage/auth/page.mdx
```diff
@@ -0,0 +1,409 @@
...
+	*   `exp`   (Expiration): When does this token expire?
+	*   `iat`   (Issued At): When was this token created?
+	*   `nbf`   (Not Before): When does the lifetime of this token start?
+    *   `scope` (Scope): Which actions can be performed on which resources, i.e. what is this token allowed to do?
+	*   `jti`   (JWT Identifier): A unique identifier that is specific to this token.
+	*   Other fields can be defined according to specific token *profiles* (see ["Token Profiles"](#token-profiles) for more information)
+3.  **Signature:** The cryptographic proof that ensures the token was created by the right issuer and hasn't been altered. Signatures do not decode to JSON
+
+<Callout type="example">
+Below are JSON objects that represent an example JWT header and payload:
```
aowen-uwmad:
```suggestion
Below are JSON objects that represent an example JWT **header** and **payload**:
```

docs/app/advanced-usage/auth/page.mdx
```diff
@@ -0,0 +1,409 @@
...
+
+<Callout type="warning">
+Be careful using tokens that let you modify object contents after the object has been written -- Pelican objects should be treated as immutable, so these tokens can get you in trouble!
+</Callout>
+
+## Issuers
+
+While discussing what tokens are and how they work, it's been unavoidable to briefly discuss the concept of *issuers*, but up until now we haven't rigorously explained what issuers are and why it's difficult to precisely describe them.
+
+At its simplest, an **Issuer** is the authority that creates/mints, signs, and provides means to validate tokens.
```
aowen-uwmad:
```suggestion
At its simplest, an **Issuer** is the authority that creates (or "mints"), signs, and provides means to validate tokens.
```

aowen-uwmad (commented):
Some writing feedback. Did not investigate the technical details; someone else should confirm those before approving.

brianaydemir:
This probably deserves a conversation in the proverbial hallway (rather than me trying to type an essay into GitHub), but the reason our story for "issuers" is a mess is because it invents terminology where it doesn't need to.

> *   `iss`   (Issuer): Who created this token?  
> …  
> While discussing what tokens are and how they work, it's been unavoidable to briefly discuss the concept of *issuers*, but up until now we haven't rigorously explained what issuers are and why it's difficult to precisely describe them.  
> …  
> At its simplest, an **Issuer** is the authority that creates/mints, signs, and provides means to validate tokens.

After looking at some RFCs, in the context of Pelican, I think this is what I'd start with: [An "issuer" is an **identifier** for an authorization server](https://datatracker.ietf.org/doc/html/rfc8414#section-2), in the sense of OAuth2, that supports [metadata discovery](https://datatracker.ietf.org/doc/html/rfc8414#section-3). Pelican expects this metadata to include a URL to the signing key(s) for validating tokens issued by the authorization server. Depending on the authorization server's configuration, as conveyed in its metadata, Pelican may be able to use standard OAuth2 flows to obtain access tokens from the authorization server.

An "issuer" doesn't _do_ anything, because identifiers don't _do_ anything. They just are.

The active thing is an "authorization server." It is the thing that creates and signs tokens. To do this, sometimes, it hands a key over to a third party it trusts and has the third party create JWTs in its name. Sometimes, it publishes a collection of endpoints for an OAuth2 flow.

jhiemstrawisc:
Pinging @brianaydemir -- can you take a peek at my latest commit? I thought your comment about authorization servers was insightful and I tried rewriting some of the docs to reflect it.

Issue #2764, opened 2025-10-22
Add size based retention policy to Prometheus


patrickbrophy:
The operations team would like to configure Prometheus' data retention policy beyond just retention time. The axis Prometheus offers is the size of the blocks stored on disk. This is configurable through the `storage.tsdb.retention.size` flag. 

In Prometheus internals it requires setting `cfg.tsdb.MaxBytes`.

PR #2779, opened 2025-10-29
Make Prometheus data retention policy configurable

patrickbrophy:
This PR addresses issue #2764. This PR adds the `Monitoring.DataRetentionSize` configuration parameter. This will allow operators to adjust the data retention policy in the same ways you could with a traditional Prometheus server (see [docs](https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects)).

### DIFF
--- a/web_ui/prometheus.go
+++ b/web_ui/prometheus.go
@@ -471,8 +471,10 @@ func ConfigureEmbeddedPrometheus(ctx context.Context, engine *gin.Engine) error 
 	cfg.tsdb.RetentionDuration = retention
 
 	maxBytes, err := units.ParseBase2Bytes(param.Monitoring_DataRetentionSize.GetString())
-	if err != nil {
-		return fmt.Errorf("failed to parse Monitoring.DataRetentionSize as a byte value: %v", err)
+	if err != nil || maxBytes < 0 {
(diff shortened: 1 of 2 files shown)


web_ui/prometheus.go
```diff
@@ -470,6 +470,12 @@ func ConfigureEmbeddedPrometheus(ctx context.Context, engine *gin.Engine) error
 	}
 	cfg.tsdb.RetentionDuration = retention
 
+	maxBytes, err := units.ParseBase2Bytes(param.Monitoring_DataRetentionSize.GetString())
+	if err != nil {
+		return fmt.Errorf("failed to parse max bytes for retention size: %v", err)
```
jhiemstrawisc: I think this should explicitly point at the config knob so the user knows what they need to change.

web_ui/prometheus.go
```diff
@@ -470,6 +470,12 @@ func ConfigureEmbeddedPrometheus(ctx context.Context, engine *gin.Engine) error
...
+	}
+	cfg.tsdb.MaxBytes = maxBytes
```
jhiemstrawisc: I tried configuring a value of `-10B` and the server started happily with no warnings or errors. Do you have any ideas what Prometheus does behind the scenes? Should handling of this number be moved into `config/config.go` so we can validate the input more rigorously?

PR #2780, opened 2025-10-29
Design and implement a local client API

bbockelm:
At some level, this is also an experiment in providing significant standalone functionality through the use of agentic AI ... I was driving the code and overseeing the implementation but didn't write any of the generated code.

This PR provides a "local client server"; listening on a domain socket, the client-api server listens for incoming requests and performs them.  Multiple clients can connect to the server and benefit from it executing work in the background.

### DIFF
--- a/cmd/plugin_htcondor_test.go
+++ b/cmd/plugin_htcondor_test.go
@@ -160,9 +160,15 @@ Origin:
 	// Get the pelican binary (built once via sync.Once)
 	pelicanBinary := getPelicanBinary(t)
 
+	// Create temporary plugin directory in /tmp
+	pluginDir, err := os.MkdirTemp("/tmp", "pelican-libexec-*")
+	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -106,7 +106,7 @@ func Execute() error {
 	ctx := context.WithValue(egrpCtx, config.EgrpKey, egrp)
 	exeErr := rootCmd.ExecuteContext(ctx)
 	if exeErr != nil {
-		log.Errorln("Fatal error occurred at the start of the program. Cleanup started:", exeErr)
```
patrickbrophy: Why change the log level?

bbockelm:
Considering the original source of a lot of this work, I thought it appropriate to ask Gemini for relevant quips upon merge.

So, @patrickbrophy, @jhiemstrawisc --

- "Il dado è tratto"
- "La fortuna aiuta gli audaci"
- "Per aspera ad astra"
- Поехали!
- Кто не рискует, тот не пьёт шампанского

🚀

Issue #2302, opened 2025-05-08
Add Options for setting the UID/GIDs for the Pelican and XRootD users in containers


williamnswanson:
Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [X] Origin
- [X] Cache
- [X] Other (Container Images)

**Is your feature request related to a problem? Please describe.**
_A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]_

At the moment the UID and GID of the `pelican` and `xrootd` users in container images are hardcoded (to 10941:10941 and 10940:10940 respectively), and while the upcoming DropPrivs feature seems like it will allow for setting a user to drop to, it doesn't seem to allow for setting UIDs/GIDs or for changing the `xrootd` user's ids independently (or in a non-DropPrivs container).

Sometimes we'll need to make sure any files that the Pelican/Xrootd processes create will be written by a user with specific UIDs/GIDs, and while it's possible to new mount /etc/passwd and /etc/group files with the changed ids for `pelican` and `xrootd` (and set a user to run as in the `docker run` command), it would be nice if root-ly pelican could do this itself.

**Describe the solution you'd like**
_A clear and concise description of what you want to happen._
Add configuration options to specify the UIDs and GIDs for the `pelican` and `xrootd` users, and an option to have root-ly Pelican write it's own (non-secure?) dirs/files as the `pelican` user (and not attempt to chown things to 0:0 if it's been told to write as `pelican`)

jhiemstrawisc:
@williamnswanson, we just merged some code that lets you specify a UID:GID for the `pelican` and `xrootd` users in our containers _at build time_. With the change, you can produce an image that uses an arbitrary UID:GID for those users, but it requires building your own container (as opposed to grabbing a container from Harbor that comes out of the Pelican build process).

I think what you're asking is slightly different than what I implemented - that we add configuration to `pelican.yaml` that lets an admin like yourself change the UID:GID for these users after the container was built with the default 10940/10941. Is that correct?

I view my solution as "fix it at build time" while I think you're asking for "fix it at runtime"

[...]

PR #2784, opened 2025-10-30
Make xrootd/pelican user UID:GID in images configurable

jhiemstrawisc:
As requested by William -- this provides a Docker build arg for setting the UID/GIDs of the pelican and xrootd users in our images.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -39,6 +39,12 @@ ARG GO_VER=1.24.7
 ARG GORELEASER_CURRENT_TAG
 ARG NODEJS_VER=20
 
+# For controlling the UID/GID of users in the container images.
+ARG PELICAN_UID=10941
+ARG PELICAN_GID=10941
+ARG XROOTD_UID=10940
+ARG XROOTD_GID=10940
+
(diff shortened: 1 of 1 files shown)


brianaydemir (approved):
For the record, what @williamnswanson wanted was configuration knobs in the proverbial `pelican.yaml` that's used to otherwise configure an origin or cache.

For an RPM-based install of Pelican, you can create manually create the `xrootd` and `pelican` users prior to installing any Pelican or XRootD packages.

For our container images, that is in fact exactly what we do — because the XRootD packages create various files that end up being owned by `xrootd`. Changing `xrootd`'s  UID and GID after the fact is certainly possible, but then you're looking at crawling the entire filesystem to `chown` files while not massacring volume mounts. If someone really wants to do that, they can take the container images we publish, and change the entrypoint to a script that does all that work before invoking our own entrypoint.

The alternative is what this PR provides — the ability to set the UID and GID when building the image in the first place.

jhiemstrawisc:
Since I'm unlinking the issue to avoid automatic closure, I'll re-link them with this comment: #2302

PR #2785, opened 2025-10-31
Persist downtime set at Origin/Cache in Registry

h2zh:
See the commits.
persist-server-set-dt drawio (1)

### DIFF
--- a/director/cache_ads.go
+++ b/director/cache_ads.go
@@ -663,16 +663,16 @@ func getRegistryDowntimes(ctx context.Context) ([]server_structs.Downtime, error
 	return downtimes, nil
 }
 
-// categorizeServerDowntimes categorizes downtimes based on whether they're for a running or offline server.
-// The Director still cares about offline servers because it needs to display their downtime info too.
+// formatServerDowntimes processes registry downtimes to align server names with the Director's in-memory state.
+// For running servers, it converts ServerID to the actual server name from serverAds.
+// For offline servers, it keeps the original downtime entry for display purposes.
 //
-// It returns two slices:
-//  1. runningServerDowntimes: downtimes for currently running servers (need to apply filtering)
-//  2. allDowntimes: all downtimes including non-running servers (for display in Director's downtime page)
-func categorizeServerDowntimes(registryDowntimes []server_structs.Downtime, ads map[string]*ttlcache.Item[string, *server_structs.Advertisement]) (runningServerDowntimes, allDowntimes []server_structs.Downtime, err error) {
+// All downtimes are returned and used for both display and filtering. Filtering offline servers
+// is harmless - the filter entries sit unused until the server comes online, at which point
+// they ensure the server is immediately blocked without waiting for the next registry poll.
+func formatServerDowntimes(registryDowntimes []server_structs.Downtime, ads map[string]*ttlcache.Item[string, *server_structs.Advertisement]) ([]server_structs.Downtime, error) {
 	// Build a prefix→name map to convert legacy downtime entries' server name (prefix)
-	// to actual server name (meanwhile checking if the server is currently running).
-	// Similarly, build a serverID→name map to convert serverID to server name.
+	// to actual server name. Similarly, build a serverID→name map.
 	prefixToName := make(map[string]string, len(ads))
 	serverIDToName := make(map[string]string, len(ads))
 	for _, ad := range ads {
@@ -682,48 +682,36 @@ func categorizeServerDowntimes(registryDowntimes []server_structs.Downtime, ads 
 		}
 	}
 
-	runningServerDowntimes = make([]server_structs.Downtime, 0)
-	allDowntimes = make([]server_structs.Downtime, 0)
+	allDowntimes := make([]server_structs.Downtime, 0, len(registryDowntimes))
 
 	for i := 0; i < len(registryDowntimes); i++ {
 		downtime := registryDowntimes[i]
 		// If this entry doesn't have a server ID, it means it's an old downtime entry (recorded before v7.22),
 		// where the server name is the prefix. This block is for backward compatibility.
-		// In the Registry, downtime.serverName is prefix, not server name (because Registry doesn't know the server name)
-		// So we need to find its corresponding server name in the serverAds and use it to overwrite downtime.serverName
 		if downtime.ServerID == "" {
 			name, found := prefixToName[downtime.ServerName]
-			if !found {
-				// Server is not currently running, but we still want to display the downtime
-				log.Tracef("Server with prefix %s is not currently running in the Director, but keeping downtime for display.", downtime.ServerName)
-				allDowntimes = append(allDowntimes, downtime)
-				continue
+			if found {
+				// Convert the server name from prefix to actual server name
+				downtime.ServerName = name
 			}
-			// Convert the server name from prefix to actual server name
-			downtime.ServerName = name
-			// End of backward compatibility block
+			// If not found, keep original ServerName (prefix) for display
 		} else {
 			name, found := serverIDToName[downtime.ServerID]
-			if !found {
-				// Server is not currently running, but we still want to display the downtime
-				log.Tracef("Server with ID %s is not currently running in the Director, but keeping downtime for display.", downtime.ServerID)
-				allDowntimes = append(allDowntimes, downtime)
-				continue
+			if found {
+				downtime.ServerName = name
 			}
-			downtime.ServerName = name
+			// If not found, keep original ServerName for display
 		}
-		// Server is currently running
-		runningServerDowntimes = append(runningServerDowntimes, downtime)
 		allDowntimes = append(allDowntimes, downtime)
 	}
 
-	return runningServerDowntimes, allDowntimes, nil
+	return allDowntimes, nil
 }
 
 // applyDowntimeFilters builds new filter and federation downtime maps based on the current registry data.
 // Callers remain responsible for handling synchronization (lock) and updating the shared state.
 func applyDowntimeFilters(
-	runningServerDowntimes, allDowntimes []server_structs.Downtime,
+	allDowntimes []server_structs.Downtime,
 	currentFilters map[string]filterType,
 	currentFederationDowntimes map[string][]server_structs.Downtime,
 ) (map[string]filterType, map[string][]server_structs.Downtime) {
@@ -739,17 +727,16 @@ func applyDowntimeFilters(
 	newFederationDowntimes := make(map[string][]server_structs.Downtime, len(currentFederationDowntimes))
 	currentTime := time.Now().UTC().UnixMilli()
 
-	// First, save ALL downtimes (including non-running servers) for display in Director's downtime page
+	// Save all downtimes for display and apply filtering
+	// Filtering offline servers is harmless - entries sit unused until the server comes online,
+	// at which point applyActiveDowntimeFilter uses them to immediately block the server
 	for _, downtime := range allDowntimes {
 		newFederationDowntimes[downtime.ServerName] = append(newFederationDowntimes[downtime.ServerName], downtime)
-	}
 
-	// Then, apply filtering only to currently running servers
-	for _, downtime := range runningServerDowntimes {
 		// Check existing downtime filter
 		originalFilterType, hasOriginalFilter := newFilters[downtime.ServerName]
 
-		// If this server is already put in downtime (and not tempAllowed), we don't need to do anything
+		// If this server is already put in downtime (and not tempAllowed), skip
 		if hasOriginalFilter && originalFilterType != tempAllowed {
 			continue
 		}
@@ -782,17 +769,14 @@ func updateDowntimeFromRegistry(ctx context.Context) error {
 
 	ads := serverAds.Items() // pull the cached server ads slice once
 
-	// Categorize server downtimes based on whether they're for a running server or not
-	// (not in the cached server ads = not running). The Director still cares about servers
-	// that aren't running because it needs to display downtime info for those servers.
-	runningServerDowntimes, allDowntimes, err := categorizeServerDowntimes(registryDowntimes, ads)
+	// Format downtimes to align server names with Director's in-memory state
+	allDowntimes, err := formatServerDowntimes(registryDowntimes, ads)
 	if err != nil {
 		return err
 	}
 
 	// Create new filters as needed and apply them
 	newFilters, newFederation := applyDowntimeFilters(
-		runningServerDowntimes,
 		allDowntimes,
 		filteredServers,
 		federationDowntimes,
--- a/director/cache_ads_test.go
+++ b/director/cache_ads_test.go
@@ -533,21 +533,20 @@ func TestApplyDowntimeFilters(t *testing.T) {
 		EndTime:    now + 1_000,
 	}
 
-	running := []server_structs.Downtime{
-		activeDowntime,
-		tempAllowedDowntime,
-		futureDowntime,
-		indefDowntime,
-		permFilteredDowntime,
-	}
-
 	offlineOnly := server_structs.Downtime{
 		ServerName: "offline-only",
 		StartTime:  now - 10_000,
 		EndTime:    now - 5_000,
 	}
-	allDowntimes := append([]server_structs.Downtime{}, running...)
-	allDowntimes = append(allDowntimes, offlineOnly)
+
+	allDowntimes := []server_structs.Downtime{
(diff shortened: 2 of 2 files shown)


database/server.go
```diff
@@ -504,8 +504,8 @@ func GetServiceName() (string, error) {
 }
 
 // Retrieve all service names from most recent to oldest
-func GetServiceNameHistory() ([]server_structs.ServiceName, error) {
-	var entries []server_structs.ServiceName
+func GetServiceNameHistory() ([]server_structs.ServerName, error) {
```
jhiemstrawisc: Should the name of this function and its docstring change to match the fact that it takes as input a server name, not a service name?

registry/registry.go
```diff
@@ -1270,6 +1276,47 @@ func getAllowedPrefixesForCachesHandler(ctx *gin.Context) {
...
+			Msg:    "Invalid prefix. Prefix cannot be empty",
+		})
+		return
+	}
+
+	server, err := getServerByPrefix(prefix)
+	if err != nil {
+		ctx.JSON(http.StatusInternalServerError, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    "Failed to get server by prefix",
```
jhiemstrawisc: Do you think it's worth passing the error message back to the request initiator? I worry that as written, when this error condition is hit we neither log it in the Registry nor tell the user about what went wrong. How will we debug it?

cache/advertise.go
```diff
@@ -51,7 +51,7 @@ type (
 // Can use this mechanism to override the minimum for the sake of tests
 var MinFedTokenTickerRate = 1 * time.Minute
 
-func (server *CacheServer) CreateAdvertisement(name, originUrl, originWebUrl string) (*server_structs.OriginAdvertiseV2, error) {
+func (server *CacheServer) CreateAdvertisement(name, id, originUrl, originWebUrl string) (*server_structs.OriginAdvertiseV2, error) {
```
jhiemstrawisc: This question also applies to the Origin interface's implementation of this method -- why are some of the values needed to create the advertisement passed as input (name, ID, etc) while other needed values are fetched internally by the function?

For example, the `OriginAdvertiseV2` struct that gets set up in this sets the downtimes, but those aren't passed as input, they're fetched internally? I don't think it's a huge deal to use one approach or the other, but it's not clear to me why we need a mix of the two. I get nervous whenever we need to change the signatures for exported functions like this, especially because that means you have to touch many lines of code to make a simple change.

director/cache_ads.go
```diff
@@ -573,24 +573,57 @@ func updateDowntimeFromRegistry(ctx context.Context) error {
...
+			// End of backward compatibility block
+		} else {
+			name, found := serverIDToName[downtime.ServerID]
+			if !found {
+				// Server is not currently running, but we still want to display the downtime
+				log.Tracef("Server with ID %s is not currently running in the Director, but keeping downtime for display.", downtime.ServerID)
+				allDowntimes = append(allDowntimes, downtime)
+				continue
+			}
+			downtime.ServerName = name
```
jhiemstrawisc: This comment is for the overall `updateDowntimeFromRegistry` function -- right now the function is about 130 lines of very dense, complicated logic. Can you consider breaking it up into smaller functions whose names explain each step of the logic? I'm thinking something like:
```
func updateDowntimeFromRegistry(ctx context.Context) error {
    // Get all downtime information from the registry
    registryDowntimes, err := getRegistryDowntimes(ctx)
    if err != nil {
        // return
    }

    // Categorize server downtimes based on whether they're for a running server or a server the Director doesn't currently know about. The Director still cares about servers that aren't running because it needs to display downtime info for those servers.
    runningServerDowntimes, allDowntimes, err := categorizeServerDowntimes(...)
    if err != nil {
        // do something
    }
    
    // Create new filters as needed and apply them
    if err := applyDowntimeFilters(...); err != nil {
        // do something
    }
    return nil
}
```

This would be helpful because you could write more targeted unit tests for each of these individual functions as well.

token/token_verify.go
```diff
@@ -187,6 +189,52 @@ func (a AuthCheckImpl) checkLocalIssuer(c *gin.Context, strToken string, expecte
...
+	// Look up the server's public key and metadata based on the subject
+	jwks, err := GetJWKSFromServerID(ctx, subject)
+	if err != nil {
+		return errors.Wrap(err, "Failed to lookup registered server")
+	}
+
+	// Now verify the token with the server's public key
+	parsed, err := jwt.Parse([]byte(strToken), jwt.WithKeySet(jwks))
+	if err != nil {
+		return errors.Wrap(err, "Failed to verify JWT with server's registered public key")
```
jhiemstrawisc: I'd use the phrase "key set" instead of "public key" -- we should always be assuming there are multiple keys whenever we work with a JWKS.

web_ui/server_api.go
```diff
@@ -197,10 +240,20 @@ func HandleCreateDowntime(ctx *gin.Context) {
...
 	}
+
+	if err := database.CreateDowntime(&downtime); err != nil {
+		if errors.Is(err, gorm.ErrDuplicatedKey) {
+			log.Debugf("Downtime already exists: %v; skipping creation", downtime)
+		} else {
+			ctx.JSON(http.StatusInternalServerError, server_structs.SimpleApiResp{Status: server_structs.RespFailed, Msg: "Failed to create downtime: " + err.Error()})
+			return
+		}
+	}
```
jhiemstrawisc: I think this is the section we talked about changing in a follow-up PR -- the Origin/Cache should not rely on a response from the Registry to populate its own downtime table.

In the meantime, in case we don't get to that PR for the upcoming release, can you add a comment before this section that explicitly states why this code waits for a registry response before recording the downtime? That way if this doesn't change right away, we still have a comment explaining why you chose this design.

web_ui/server_api.go
```diff
@@ -266,39 +330,123 @@ func HandleUpdateDowntime(ctx *gin.Context) {
...
+	ctx.JSON(http.StatusOK, updatedDowntime)
 }
 
 func HandleDeleteDowntime(ctx *gin.Context) {
 	uuid := ctx.Param("uuid")
+	existingDowntime, err := database.GetDowntimeByUUID(uuid)
+	if err != nil {
+		ctx.JSON(http.StatusNotFound, server_structs.SimpleApiResp{Status: server_structs.RespFailed, Msg: "Downtime record not found"})
+		return
+	}
```
jhiemstrawisc: How do you handle the different cases of "there is no downtime entry for this UUID" and "querying the downtime table for this UUID produced an error" -- right now you treat the error condition as a "not found", but I wonder whether an error could happen because of a database issue.

web_ui/server_api.go
```diff
@@ -266,39 +330,123 @@ func HandleUpdateDowntime(ctx *gin.Context) {
...
+			return nil
+		}
+	}
+
+	regURL.Path = path.Join(regURL.Path, "api", "v1.0", "downtime", id)
+
+	tokCfg := token.NewWLCGToken()
+	tokCfg.Lifetime = 2 * time.Minute
+	tokCfg.Subject = dt.ServerID
+	tokCfg.AddAudienceAny()
```
jhiemstrawisc: In this case, we explicitly know who is going to use this token -- the Registry. Can you set the audience to the registry's discovered hostname:port?

director/cache_ads.go
```diff
@@ -442,15 +442,25 @@ func updateLatLong(ad *server_structs.ServerAd) error {
...
+	// If the server is >= v7.22, it sets the serverID in the downtime entry.
+	// Its downtimes are already handled by the Registry polling (updateDowntimeFromRegistry),
+	// so we don't need to do anything here
+	for _, dt := range downtimes {
+		if dt.ServerID != "" {
+			log.Debugf("The downtimes set by the admin of Server %q have been propagated to the Director through polling the Registry; skipping processing the downtimes in the server ads", serverName)
+			return
+		}
+	}
+
```
jhiemstrawisc: This comment isn't for this exact snippet of code, but one of the things we talked about in our team discussion of downtime management was having Origins/Caches always report downtimes directly to the Director as well as through the Registry. We brought up the "CalDAV" protocol and mentioned that part of what the Director should be doing is figuring out how to merge two potentially-conflicting downtimes when one comes from the Origin/Cache and the other comes from the Registry.

Do you think we should take a stab at that in this PR so we can retain the functionality where downtimes are sent by Origins/Caches to the Director, or do you think that should be a followup to be tackled in the future?

token/token_verify.go
```diff
@@ -405,6 +491,68 @@ func GetJWKSFromIssUrl(issuer string) (*jwk.Set, error) {
...
+	if registryEndpoint == "" {
+		return nil, errors.New("registry endpoint unknown; set Federation.RegistryEndpoint or Server.ExternalWebUrl")
+	}
+
+	registryURL, err := url.Parse(registryEndpoint)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse registry endpoint")
+	}
+
+	registeredServerPubKeyURL, err := url.JoinPath(registryURL.String(), "api", "v1.0", "registry_ui", "servers", serverID, "pubkey")
```
jhiemstrawisc: I had a quick chat with BrianB about this, and I think there needs to be a change of approach here -- whenever we verify a token that's associated with an "issuer", we need to make sure we follow good "issuer" and "openid" protocols.

In this case that would mean using the Server ID to construct a proper Issuer URL like `https://registry.com/api/v1.0/registry_ui/<the ID>` or `https://registry.com/api/v1.0/registry/<the ID>` (depending on where it needs to live). From the issuer URL we visit the `/.well-known/openid-configuration` path, i.e. `https://registry.com/api/v1.0/registry/<the ID>/.well-known/openid-configuration` to download the OIDC config JSON. That will have a `jwks_uri` key that points at the location of the key.

In other words, you can still use this as the URL that supplies the key, but you should discover this URL by following the OIDC discovery process. Part of what's needed for this should already be close to functional. For example, the Registry Issuer URL for a namespace is `https://registry.com/api/v1.0/registry/<namespace>` -- this already supports the `.well-known` lookup stuff so you might consider extending this behavior to work with Server IDs as well, although I think we're moving toward registering the Server ID as the namespace (like /caches/<Server ID> instead of /caches/<Sitename>).

server_utils/sitename.go
```diff
@@ -106,15 +106,21 @@ func GetServiceName(ctx context.Context, server server_structs.ServerType) (name
...
-			log.Errorf("Failed to get sitename from the registry for the origin. Will fallback to using %s: %v", param.Xrootd_Sitename.GetName(), err)
+			log.Errorf("Failed to get metadata from the registry for the origin. Will fallback to using %s: %v", param.Xrootd_Sitename.GetName(), err)
+		} else {
+			nameFromReg = serverReg.Name
+			idFromReg = serverReg.ID
 		}
 	} else if server.IsEnabled(server_structs.CacheType) {
 		cachePrefix := server_structs.GetCacheNs(param.Xrootd_Sitename.GetString())
-		nameFromReg, err = getSitenameFromReg(ctx, cachePrefix)
+		serverReg, err = getServerMetadataFromReg(ctx, cachePrefix)
```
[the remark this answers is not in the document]
h2zh: The type-specific part is already handled by `GetServerMetadata`, I think generics would be over-engineering.

database/server.go
```diff
@@ -458,19 +457,19 @@ func ShutdownDB() error {
 	return err
 }
 
-// Create or update a record for the given serverName
+// Create or update a record for the given server id
```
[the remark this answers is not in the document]
h2zh: Added the definitions for server id and name in this docstring.

cache/advertise.go
h2zh: I think the reason why {id, name, url, webUrl} are passed in `CreateAdvertisement` as input params is that they share the same getter functions between origin and cache. 

Then the code evolved organically. The fields with same getter logics should've passed in as input params, but didn't (for example, `downtime`). For `downtimes` in particular, I address it in this PR.

director/cache_ads.go
```diff
@@ -573,24 +573,57 @@ func updateDowntimeFromRegistry(ctx context.Context) error {
...
 	defer filteredServersMutex.Unlock()
 
-	// In the Registry, downtime.serverName is prefix, not server name (because Registry doesn't know the server name)
-	// So we need to find its corresponding server name in the serverAds and use it to overwrite downtime.serverName
-
 	ads := serverAds.Items() // pull the ads slice once
-	// Build a prefix→name map
+	// Build a prefix→name map to convert legacy downtime entries' server name (prefix)
+	// to actual server name (meanwhile checking if the server is currently running);
+	// and a set of running server names to check if a server is currently running
```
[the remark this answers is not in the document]
h2zh: I simplify this verbose comment

registry/registry_ui.go
```diff
@@ -911,6 +924,60 @@ func getServerHandler(ctx *gin.Context) {
...
+		ctx.JSON(http.StatusNotFound, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    "Server not found"})
+		return
+	}
+
+	if len(server.Registration) == 0 {
+		ctx.JSON(http.StatusNotFound, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    "Server registration not found"})
```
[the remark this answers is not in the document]
h2zh: Server and Registration are two tables in Registry. They are associated by a bridge table Service. 
Each server may has multiple services (Origin, Cache); each service has an one-to-one relationship with a registration.

This error represents a scenario that in theory would never happen because of the DB rules we set - a server doesn't has its corresponding registration.

web_ui/server_api.go
h2zh: Good catch - I separate the "not found" error from other errors.

database/server.go
```diff
@@ -489,8 +492,8 @@ func UpsertServiceName(serverName string, typ server_structs.ServerType) error {
 }
 
 // Retrieve the server name in use - lookup the entry whose UpdatedAt is the most recent
-func GetServiceName() (string, error) {
-	var entry server_structs.ServiceName
+func GetServerName() (string, error) {
```
CannonLock: What mechanism makes the most recently updated serverName the one for this server? If this function is used on the registry would it give the intended results?

Noting this isn't used anywhere might delete it to save time.

web_ui/ui.go
```diff
@@ -597,11 +597,12 @@ func configureCommonEndpoints(engine *gin.Engine) error {
 
 	downtimeAPI := engine.Group("/api/v1.0/downtime")
 	{
-		downtimeAPI.POST("", AuthHandler, AdminAuthHandler, HandleCreateDowntime)
+		downtimeAPI.POST("", DowntimeAuthHandler, HandleCreateDowntime)
+		downtimeAPI.POST("/:uuid", DowntimeAuthHandler, HandleCreateDowntime)
```
CannonLock: Why this endpoint? How would I know the id of the downtime before it is created?

director/cache_ads.go
```diff
@@ -442,15 +442,25 @@ func updateLatLong(ad *server_structs.ServerAd) error {
 	return nil
 }
 
-// Apply downtimes provided by Origin/Cache servers to the director's in-memory state for the given server.
+// Apply downtimes provided by Origin/Cache servers (<= v7.21) to the director's in-memory state for the given server.
 // This function treats a nil or empty slice as "no downtimes", clearing any stale
 // serverFiltered entry and removing cached server downtimes.
 func applyServerDowntimes(serverName string, downtimes []server_structs.Downtime) {
+	// If the server is >= v7.22, it sets the serverID in the downtime entry.
```
CannonLock: Can you add a loud TODO to indicate this can get removed when all listeners are above this version?

director/cache_ads.go
```diff
@@ -530,70 +540,107 @@ func getCachedDowntimes(serverName string) ([]server_structs.Downtime, error) {
...
-		if !found {
-			log.Infof("Unable to find server name for prefix %s in the Director. The server with the given prefix is not running now.", latestFedDowntimes[i].ServerName)
-			continue
+
+	runningServerDowntimes = make([]server_structs.Downtime, 0)
+	allDowntimes = make([]server_structs.Downtime, 0)
+
+	for i := 0; i < len(registryDowntimes); i++ {
+		downtime := registryDowntimes[i]
+		// If this entry doesn't have a server ID, it means it's an old downtime entry (recorded before v7.22),
```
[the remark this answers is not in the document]
h2zh: same as above.

PR #2786, opened 2025-11-03
Enable web UI adjustment

turetske:
Adjusted the code so that Server.EnableUI is now ONLY a check for whether to start up the webpage.

Any other code that has backend implications/use has been moved outside of a check involving the Server.EnableUI parameter.

For now, we will not be implementing separate checks for enabling/disabling these code blocks.

### DIFF
--- a/launchers/registry_serve.go
+++ b/launchers/registry_serve.go
@@ -29,7 +29,6 @@ import (
 	"github.com/pelicanplatform/pelican/config"
 	"github.com/pelicanplatform/pelican/database"
 	"github.com/pelicanplatform/pelican/metrics"
-	"github.com/pelicanplatform/pelican/oauth2"
 	"github.com/pelicanplatform/pelican/param"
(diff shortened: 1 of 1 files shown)


docs/parameters.yaml
```diff
@@ -939,13 +939,6 @@ type: duration
 default: 15s
 components: ["origin"]
 ---
-name: Origin.EnableUI
-description: |+
-  Indicate whether the origin should enable its web UI.
```
h2zh: User and admin may get confused about the scope of **Server**.EnableUI, so it would be nice to also update the description, for example:
Indicate whether a server should enable its web UI. This only controls the serving of web UI resources and pages. Backend functionality such as OIDC authentication, OAuth endpoints, and API routes will remain enabled regardless of this setting.

h2zh (approved):
This PR is clear and well-structured. I got a few nice-to-have comments so I just pre-approve it.

h2zh (approved):

PR #2791, opened 2025-11-10
Collections CLI and Authz

patrickbrophy:
This PR introduces the collections CLI and fixes the authorization around collections.

### DIFF
--- a/origin/collections.go
+++ b/origin/collections.go
@@ -296,6 +296,7 @@ func handleListCollections(ctx *gin.Context) {
 			Status: server_structs.RespFailed,
 			Msg:    fmt.Sprintf("Failed to get user from context: %v", err),
 		})
+		return
 	}
 	if user == "" {
 		ctx.JSON(http.StatusInternalServerError, server_structs.SimpleApiResp{


h2zh (changes requested):
It's a well-structured OAuth2 device flow implementation. Though I got the following error when I try to run some basic collection commands (after I approve it in the webUI). Could you take a look? FYI, I use the oa4mp origin config yaml you sent me before

```
pelican origin collections list/create

ERROR[2025-11-24T18:52:07Z] Fatal error occurred at the start of the program. Cleanup started: API error (status 403): cannot verify token with server issuer: Failed to verify the scope of the token. Require [web_ui.access collection.read (.create when I run the collection create command)]: no scope is present; required for authorization
cannot verify token with API token issuer: Invalid API token format 
```

Also other comments as below.

origin/collections.go
```diff
@@ -82,12 +280,8 @@ type GetCollectionRes struct {
...
 
 func handleListCollections(ctx *gin.Context) {
-	authOption := token.AuthOption{
-		Sources: []token.TokenSource{token.Cookie, token.Header},
-		Issuers: []token.TokenIssuer{token.LocalIssuer, token.APITokenIssuer},
-		Scopes:  []token_scopes.TokenScope{token_scopes.WebUi_Access, token_scopes.Collection_Read},
-	}
-	status, ok, err := token.Verify(ctx, authOption)
+	// List collections doesn't target a specific collection, so pass empty string
+	status, ok, err := verifyTokenWithCollectionScope(ctx, token_scopes.Collection_Read, "")
```
h2zh: Could you add a `return` on line 299

PR #2792, opened 2025-11-10
Add BrokerURL to the cache ad when broker is enabled 

h2zh:
This PR consolidates the BrokerURL getter logics for Origin and Cache in their advertising process, and adds a new fed test for Broker-enabled Cache.

See the commit messages for details.

Not sure if this PR conflicts with the current Broker work you're doing @patrickbrophy

### DIFF
--- a/cache/broker_test.go
+++ b/cache/broker_test.go
@@ -22,11 +22,13 @@ package cache_test
 
 import (
 	_ "embed"
+	"net/url"
 	"testing"
+	"time"
 
-	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/require"
 
+	"github.com/pelicanplatform/pelican/director"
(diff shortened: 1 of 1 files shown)


cache/broker_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/pelicanplatform/pelican/server_utils"
+)
+
+var (
+	//go:embed resources/broker-config.yaml
+	brokerConfig string
+)
+
+// Spin up a federation and verifies that the cache broker infrastructure is properly configured.
+func TestBrokerApi(t *testing.T) {
```
patrickbrophy: This seems like a strange test. It appears that we are just manually incrementing the counter ourselves and then checking that it has increased? This doesn't really seem to test that the connection was actually made.

patrickbrophy (changes requested):

PR #2798, opened 2025-11-13
Add a "site-local" mode for caches

jhiemstrawisc:
The goal of this new mode is to let groups start Pelican caches that are intended to serve _only_ their local clients. This means the cache does not register, does not advertise, and does not otherwise engage in federation-type activities. It's a barebones XRootD server with a Pelican WebUI sitting on top of it.

Initially I was unsure whether these caches -- which federation operators will likely remain unaware of -- should implement all of the federation auth policies, in part because this breaks the federation operators' ability to convey the urgency of security fixes to cache administrators (you can't talk to someone if you don't know they exist). Since Caches are charged with implementing the authZ policies of all namespaces/Origins in a federation, it makes me a bit nervous that we'll have no clue what's going on with these services, yet we'll happily let them hold onto any data that passes through them. After talking to Brian, I agree it's still better to let the site-local caches implement the authz policies to the best of their ability, but to accept that it's beyond our control what happens after that. If a site-local cache admin breaks something or otherwise leaks all the data in the cache, that's no different than anyone else downloading the data and exposing it.

I'll note that I don't think my implementation here is particularly robust -- the hallmark of a "site-local" cache is that it does not partake in federation-type activities. However, the way this is implemented means that anyone implementing _new_ federation activities in the future will need to know to exclude them for the site-local mode because the triggers for turning on this feature are sprinkled all over the place, and experience shows this sets us up for a game of whack-a-mole. However, restructuring the codebase to formalize "joining a federation" to have a single entry point (rather than an ad hoc collection of actions that are sprinkled through cache startup) that can be wrapped with a single `if param.Cache_EnableSiteLocalMode.GetBool()` would be a very large overhaul.

### DIFF
--- a/github_scripts/site_local_cache_test.sh
+++ b/github_scripts/site_local_cache_test.sh
@@ -235,9 +235,6 @@ echo "Site-local cache is running!"
 echo "Test 1: Checking Registry to ensure site-local cache did not register..."
 REGISTRY_SERVERS_URL="https://$HOSTNAME:8444/api/v1.0/registry_ui/servers"
 
-# # Make a token for authentication
-# ./pelican token create pelican://$HOSTNAME:8444/test --read --write --audience "https://wlcg.cern.ch/jwt/v1/any" --issuer "https://`hostname`:8444" --subject "origin"  --profile "wlcg" --lifetime 600 > site_local_test/test-token.jwt
-
 # echo "Token created for testing"
 
 REGISTRY_RESPONSE=$(curl -k -s -X GET "$REGISTRY_SERVERS_URL")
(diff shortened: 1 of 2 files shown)


github_scripts/site_local_cache_test.sh
```diff
@@ -0,0 +1,377 @@
...
+
+echo "Site-local cache is running!"
+
+# Now run the tests
+
+# Test 1: Verify site-local cache did not register at the Registry
+echo "Test 1: Checking Registry to ensure site-local cache did not register..."
+REGISTRY_SERVERS_URL="https://$HOSTNAME:8444/api/v1.0/registry_ui/servers"
+
+# # Make a token for authentication
```
turetske: Is this commented block a copy/paste artifact that should be deleted?

turetske (approved):
Small cleanup request in the end-to-end test script.

PR #2801, opened 2025-11-14
Trigger degraded state with a configurable ratio of server concurrency

jhiemstrawisc:
Before this change, Caches/Origins entered the degraded state whenever their active IO exceeded their concurrency limit. However, this meant the Director didn't penalize servers until they were already likely to be causing errors.

With this commit, the Cache/Origin admin can configure the ratio of active IO to concurrency limit that triggers the degraded state. The hope is that this helps pump the brakes _before_ throttling kicks in.


config/config.go
patrickbrophy: So here we allow the range to be [0, 100] but in `server_structs/server_utils.go` we all for the range to be [1, 99]. This seems like a small oversight but one worth thinking about.

patrickbrophy (approved):
Thanks for the quick fix. LGTM

PR #2813, opened 2025-11-18
Make tls.AlertError retryable

turetske:
Ensure tls.AlertError is retryable


patrickbrophy (approved):
Simple enough, lets wait to merge till Harbor is back up.

Issue #2820, opened 2025-11-20
Weird behavior in director when updating keys in registry


patrickbrophy:
We have observed strange behavior where a Origin jwk was updated in both the registry and Origin but the director was rejecting the advertisements from the cache. Below is a high level view of what was happening:
```
T0: Initial state
- Director: key-v1 (expires at T15)
- Registry: key-v1
- Origin: key-v1

T1: Update keys in cache and registry
- Director: key-v1
- Registry: key-v2
- Origin: key-v2

T2: Origin advertises to Director and fails to verify because using the cached key-v1
... (Origin continues to fail to advertise to Director)
```

What was odd is that the Director's TTL cache never evicted the old key.

The solution to this problem should be to make sure the TTL cache doesn't refresh the expiration of key on touch.  This way the key will eventually be evicted from the cache and then when the service advertises again, the director will get the key stored in the registry and store it in the TTL cache.

PR #2821, opened 2025-11-20
Fix namespace keys TTL cache in director

patrickbrophy:
This PR addresses issue #2820. This PR does the following:
- Disable touch on hit in namespace keys TTL cache
- Added test to verify expected cache invalidation

### DIFF
--- a/director/origin_api_test.go
+++ b/director/origin_api_test.go
@@ -20,9 +20,6 @@ package director
 
 import (
 	"context"
-	"crypto/ecdsa"
-	"crypto/elliptic"
-	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -32,7 +29,6 @@ import (
 	"time"
 
 	"github.com/jellydator/ttlcache/v3"
-	"github.com/lestrrat-go/jwx/v2/jwa"
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


director/origin_api_test.go
```diff
@@ -188,3 +193,182 @@ func TestNamespaceKeysCacheEviction(t *testing.T) {
...
+// TestNamespaceKeysCacheTTLExpiration tests that the namespaceKeys cache
+// properly expires keys after the TTL period and that WithDisableTouchOnHit
+// prevents TTL refresh on cache access. This validates the fix for the bug
+// where stale keys would remain cached indefinitely if accessed frequently.
+func TestNamespaceKeysCacheTTLExpiration(t *testing.T) {
+	ctx, cancel, egrp := test_utils.TestContext(context.Background(), t)
+	defer func() { require.NoError(t, egrp.Wait()) }()
+	defer cancel()
+
+	server_utils.ResetTestState()
```
jhiemstrawisc: Can you either defer a test state cleanup or add one as a `t.Cleanup()`?

director/origin_api_test.go
```diff
@@ -188,3 +193,182 @@ func TestNamespaceKeysCacheEviction(t *testing.T) {
...
+
+	// Use a shorter TTL for testing (2 seconds instead of 15 minutes)
+	originalTTL := param.Director_AdvertisementTTL.GetDuration()
+	defer viper.Set(param.Director_AdvertisementTTL.GetName(), originalTTL)
```
jhiemstrawisc: You shouldn't need to defer re-setting viper variables -- as a matter of principle, all viper config should be cleaned up by the test itself.

director/origin_api_test.go
```diff
@@ -188,3 +193,182 @@ func TestNamespaceKeysCacheEviction(t *testing.T) {
...
+	require.NoError(t, err, "Failed to generate new private key")
+	newKey, err := jwk.FromRaw(newKeyRaw)
+	require.NoError(t, err, "Failed to convert new key to JWK")
+	err = jwk.AssignKeyID(newKey)
+	require.NoError(t, err)
+	err = newKey.Set(jwk.AlgorithmKey, jwa.ES256)
+	require.NoError(t, err)
+
+	newPublicKey, err := newKey.PublicKey()
+	require.NoError(t, err, "Failed to create new public key")
```
jhiemstrawisc: I think it'd be preferable to use an existing function to generate the keys -- see `config/init_server_creds.go` and see if you can use one of the key generation functions there. Similarly, `test_utils/utils.go` provides `GenerateJWK()` and `GenerateJWKS()` that might be useful.

The first sign here that a helper function probably exists is that creating keys is a fairly common thing for us to do, and it's too easy to forget the bit about assigning a key ID (nice job remembering that, btw!). Whenever there's something common that's also nuanced/tricky to get right, there's probably a helper function somewhere. If there isn't, it's one of those situations where I wouldn't hesitate to add one and tell the rest of the dev team about it.

director/origin_api_test.go
```diff
@@ -188,3 +193,182 @@ func TestNamespaceKeysCacheEviction(t *testing.T) {
...
+	require.NoError(t, err)
+	keyLoc, err := server_utils.GetJWKSURLFromIssuerURL(issuerUrl)
+	require.NoError(t, err)
+
+	// Create a token signed with the old key
+	advTokenCfg := token.NewWLCGToken()
+	advTokenCfg.Lifetime = time.Minute
+	advTokenCfg.Issuer = issuerUrl
+	advTokenCfg.Subject = "test-cache"
+	advTokenCfg.AddAudiences("https://director.example.com")
```
jhiemstrawisc: It's a mistake that we don't enforce the audience when the Director verifies the advertise token -- let's not fix that here, but we should open another issue for it. In the mean time, this test should use an audience that corresponds to the Director configured via the mock fed root because once we fix that, it'll cause this test to fail otherwise.

jhiemstrawisc (changes requested):

director/origin_api_test.go
```diff
@@ -188,3 +193,182 @@ func TestNamespaceKeysCacheEviction(t *testing.T) {
...
+	tok, err := advTokenCfg.CreateTokenWithKey(oldKey)
+	require.NoError(t, err, "Failed to create token with old key")
+
+	// First verification - should fetch keys from registry (jwksCallCount = 1)
+	ok, err := verifyAdvertiseToken(ctx, tok, "/test-namespace")
+	require.NoError(t, err)
+	assert.True(t, ok, "Token verification should succeed with old key")
+	assert.Equal(t, 1, jwksCallCount, "JWKS should be fetched once on first verification")
+
+	// Verify the key is cached
```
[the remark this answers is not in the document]
patrickbrophy: Yes, we cache the key if the ttl cache misses.

jhiemstrawisc:
@h2zh, can you add this one to your patch radar?

h2zh:
Sure!

PR #2825, opened 2025-11-21
Error wrap 7.22

turetske:
Update the error wrapping for release 7.22

The idea is to wrap more errors as their appropriate pelican error type. The commits are split by error type, generally, and tend to be 2-3 commits per errors (1-2 for the error wrapping and any new error generation, and 1 for new/altering tests).

### DIFF
--- a/error_codes/error_codes.go
+++ b/error_codes/error_codes.go
@@ -258,7 +258,7 @@ func NewTransfer_ChecksumMismatchError(err error) *PelicanError {
 		exitCode:    9,
 		code:        6006,
 		retryable:   true,
-		description: "The client successfully transferred the file but the checksum computed by the client did not match the checksum reported by the server. This indicates data corruption during transfer or a mismatch between client and server checksum calculations.",
+		description: "The client successfully transferred the file but the checksum computed by the client did not match the checksum reported by the server.",
 		err:         err,
 	}
 }


docs/error_codes.yaml
```diff
@@ -164,37 +171,46 @@ retryable: true
...
   This indicates the director did not respond before the timeout threshold.
 retryable: true
+---
+type: Transfer.ChecksumMismatch
+code: 6006
+clientExitCode: 9
+description: >-
+  The client successfully transferred the file but the checksum computed by the client
+  did not match the checksum reported by the server. This indicates data corruption
+  during transfer or a mismatch between client and server checksum calculations.
```
patrickbrophy: The "mismatch between client and server checksum calculations" is already implied by the error type.

client/handle_http.go
```diff
@@ -2762,6 +2502,12 @@ func downloadHTTP(ctx context.Context, te *TransferEngine, callback TransferCall
...
 			// ENOMEM is error from os for unable to allocate memory
 			err = &allocateMemoryError{Err: err}
+		} else if isTLSCertificateValidationError(err) {
+			// TLS certificate validation error - wrap as SpecificationError (configuration issue, not retryable)
+			err = error_codes.NewSpecificationError(err)
+		} else if isContextDeadlineError(err) || isDNSError(err) || (isTLSError(err) && !isTLSCertificateValidationError(err)) || isDialError(err) {
+			// Connection setup errors (timeout, DNS, TLS handshake, dial) - wrap as ConnectionSetupError (retryable)
+			err = &ConnectionSetupError{Err: err}
 		} else {
 			err = &ConnectionSetupError{Err: err}
```
patrickbrophy: Is it correct to always fallback to a connection error?

patrickbrophy (changes requested):
Just a minor change requested

client/handle_http.go
turetske: No, I think I was overzealous in trying to wrap everything. I'll remove it and hopefully most things will still be wrapped and this way in this next release I can see what isn't getting wrapped and address those separately.

Issue #2819, opened 2025-11-20
PELICAN_CLIENT_PREFERREDCACHES ignored


astroclark:
Pelican Version:
_Run `./pelican --version` to get version number._

```
$ pelican --version
Version: 7.20.2
Build Date: 2025-10-03T20:49:50Z
Build Commit: d83d6a89277765c4075f3be04e0180f43a6604da
Built By: goreleaser
```

Pelican Service:
- [x] Client
- [x] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache
- [ ] Other (please give the detail)

**Describe the bug**
_A clear and concise description of what the bug is._

The variable `PELICAN_CLIENT_PREFERREDCACHES` to specify a specific cache for e.g. testing as [documented](https://docs.pelicanplatform.org/advanced-usage/plugin#setting-environment-variables-for-the-plugin-as-part-of-htcondor-job) is ignored by both the pelican client and the HTCondor plugin.

The older `PELICAN_NEAREST_CACHE` is respected by the client and the plugin.

**To Reproduce**
_Steps to reproduce the behavior._

From the [docs]() use:
```
environment = "PELICAN_CLIENT_PREFERREDCACHES=https://address.to.desired.cache:port"
```
in an HTCondor job or just set it at execution time with e.g.:
```
$ PELICAN_PREFERRED_CACHES="https://address.to.desired.cache:port" pelican object get osdf:///some/path
```

**Expected behavior**
_A clear and concise description of what you expected to happen._

The client and plugin for this version (7.20.2) should use the cache specified in `PELICAN_PREFERRED_CACHES` (as they do for `PELICAN_NEAREST_CACHE`)

**Screenshots**
_If applicable, add screenshots to help explain your problem._

**Desktop (please complete the following information):**
 - OS: [e.g. Linux, MacOS]
 - Browser: [e.g. chrome, safari, if the bug relates to any web user interface]

OS is linux `Rocky Linux release 8.10` for CLI test and similarly rocky:8 containers for HTCondor tests.

**Additional context**
_Add any other context about the problem here._

Initially noted in https://opensciencegrid.atlassian.net/browse/HTCONDOR-3411

jhiemstrawisc:
[...]

PR #2828, opened 2025-12-01
Fix plugin environment config

patrickbrophy:
This PR addresses isssue #2819. This PR fixes a bug where `PELICAN_CLIENT_PREFERREDCACHES` environment variable was being ignored by the plugin, while the older and deprecated `PELICAN_NEAREST_CACHE` worked. The cause was that we were switching Viper's environment prefix from `pelican` to `osdf`.

### DIFF
--- a/config/env.go
+++ b/config/env.go
@@ -93,7 +93,7 @@ func bindLegacyClientEnv() {
 	// Handle TOPOLOGY_NAMESPACE_URL
 	for _, prefix := range prefixes {
 		if val, isSet := os.LookupEnv(prefix.String() + "_TOPOLOGY_NAMESPACE_URL"); isSet {
-			viper.Set("Federation.TopologyNamespaceURL", val)
+			viper.Set(param.Federation_TopologyNamespaceUrl.GetName(), val)
 			break
 		}
 	}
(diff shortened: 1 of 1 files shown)


config/config_test.go
```diff
@@ -930,3 +930,127 @@ func TestDiscoverFederationImpl(t *testing.T) {
 		})
 	}
 }
+
+// TestPreferredCachesEnvVar tests that PELICAN_CLIENT_PREFERREDCACHES environment variable
+// works correctly in both regular client mode (PELICAN prefix) and plugin mode (OSDF prefix).
+// This test addresses issue #2819 where PELICAN_CLIENT_PREFERREDCACHES was ignored in plugin mode.
+func TestPreferredCachesEnvVar(t *testing.T) {
```
jhiemstrawisc: A few generic notes about the tests:
1. I think the bug you should be solving isn't specific to the `PELICAN_CLIENT_PREFERREDCACHES` param, but is generic to _all_ params supplied as environment variables. That is, this should be testing generic env var mapping into the Pelican config structs. Some of these tests are for specific behavior related to this param, like the last test checking that we correctly parse spaces in the value of this param when it's supplied as an environment variable. I think that's actually a bit unrelated to the bug at hand (I'm pretty sure that parsing is actually handled by viper itself, so there's not much we can do about it).
2. When the fix for this bug is refactored to bring some of the initialization stuff under a common function instead of permitting skew between clients/servers, these tests should run through setting up env var configs for both binary modes.
3. These tests each share a common protocol, so they should be consolidated with a test struct and a test runner, e.g.:
```
type testCase struct {
    name string
    binaryPrefix ConfigPrefix
    envVarPrefix string // either PELICAN or OSDF, or an unknown value
    envVarSuffix string // e.g. CLIENT_PREFERREDCACHES
    configValue string // the value to set for the env var
    binaryMode // server or client
    configShouldApply bool // whether or not the config should be picked up by our param package
}

// to run each test
// 1. Set the binary prefix
// 2. Construct the env var and set the value
// 3. Init config for either server or client
// 4. Check that calling `param.<param name>.Get()` has the expected output
```
4. Instead of jumping through hoops to reset env vars, use `t.Setenv()` which uses a `Cleanup` to do that for you.
5. Using `ResetConfig()` should already reset the preferred prefix so you can remove that too.

jhiemstrawisc (changes requested):

jhiemstrawisc (changes requested):

PR #2834, opened 2025-12-03
Origin use registered prefix when polling broker

patrickbrophy:
When the Director ConnectToService it passes the registered prefix of the Origin to connect to it. When the Origin polls the broker it makes two polls, one for the web endpoints and the other for xrootd. The problem lies in that the xrootd poll was using its own xrootd url, which is unregistered, in the polling token. Instead, we should poll using the registered prefix. This way, both the director and origin (web and xrootd) can all talk on the same channel.

### DIFF
--- a/cache/broker_client.go
+++ b/cache/broker_client.go
@@ -39,6 +39,7 @@ import (
 
 	"github.com/pelicanplatform/pelican/broker"
 	"github.com/pelicanplatform/pelican/config"
+	"github.com/pelicanplatform/pelican/database"
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/server_structs"
 )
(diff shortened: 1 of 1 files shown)


origin/broker_client.go
```diff
@@ -141,13 +141,19 @@ func LaunchBrokerListener(ctx context.Context, egrp *errgroup.Group, engine *gin
 		return
 	}
 
+	// The origin is registered under its external web URL, so we use that
+	// as the prefix for token authentication regardless of which address
+	// we're polling for. This ensures the broker can verify our tokens
+	// against the registered namespace.
+	registeredPrefix := server_structs.GetOriginNs(externalWebUrl.Host)
```
h2zh: At first, I am worried that there might be some exceptions in the prod db. However, I soon realize that if the registered prefix is not the same as `GetOriginNs(externalWebUrl.Host)`, this prefix will fail to advertise - which means there's no exception for this pattern. What do you think?

h2zh (changes requested):
Nice PR that clears the blind spot! I left two comments/questions

h2zh (approved):
LGTM

PR #2836, opened 2025-12-04
Redact JWT signature blocks in logs

brianaydemir:
Pelican was already redacting JWT signature blocks in server logs. This PR adds the same functionality to the client.

In addition, the client no longer logs the contents of the refresh token it receives from the device code flow. (As far as I'm aware, the refresh token can be an arbitrary string, so we have to parse through the response from the server.)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2145,8 +2145,8 @@ func InitClient() error {
 
 	var printClientConfigErr error
 	printClientConfigOnce.Do(func() {
-		if log.GetLevel() == log.DebugLevel {
-			PrintPelicanVersion(os.Stdout)
+		if GetEffectiveLogLevel() >= log.DebugLevel {
+			LogPelicanVersion()
 			printClientConfigErr = PrintClientConfig()
 		}
 	})


brianaydemir:
> It seems like `initFilterLogging` really belongs somewhere central like `InitConfigInternal()`.

@jhiemstrawisc You might believe in this. It might be a good idea. Perhaps a great one, even. Heavens knows I gave it a try.

But I'm too naïve to make it work. Empirically, [FlushLogs](https://github.com/PelicanPlatform/pelican/blob/c8edb77b4da955a53743c453943479c78f6972d2/logging/logging.go#L113-L114) must be called before [initFilterLogging](https://github.com/PelicanPlatform/pelican/blob/c8edb77b4da955a53743c453943479c78f6972d2/config/logging.go#L126) (or cannot be called after it?) in order for the regex filter to actually be applied to the logs' output.

There must be something I'm not grokking that explains why the ordering of these two functions matters here.

jhiemstrawisc:
@brianaydemir I think I got to the bottom of at least one of these GHA failures...

The crux of the failing CI test in the logs is [here](https://github.com/PelicanPlatform/pelican/actions/runs/20608548831/job/59832274810?pr=2836#step:17:824). The lines following that, up until the `***`, hinted that something about this change started dumping Pelican version information into the token file, which means it cannot be passed directly to the client via `-t`.

Why, you might ask? The `initFilterLogging()` function cranks up the log level to Debug [here](https://github.com/PelicanPlatform/pelican/blob/cb166399a3431245b19362c6f55eaaa731d9c7ed/config/logging.go#L135) to ensure the filter hooks see all messages, but I think this has the side effect of causing the version info to be printed to stdout during client operations like `token create` (see [here](https://github.com/PelicanPlatform/pelican/blob/cb166399a3431245b19362c6f55eaaa731d9c7ed/config/config.go#L2144-L2149)). Perhaps the solution is to use `config/logging.go::GetEffectiveLogLevel()` instead of `log.GetLevel()`?

Thoughts?

PR #2843, opened 2025-12-05
Remove 'Content-Disposition: attachment' header from Origins

jhiemstrawisc:
This header is no longer needed by the web client, and it prevents people from browsing a directory tree at their Origins (erm... a... collections tree? Whatever, I'm talking about a POSIX Origin where it really _is_ a directory). By removing this, the directory tree is rendered as a web page instead of downloaded as HTML.

Video proof:

https://github.com/user-attachments/assets/68b91797-bfb0-4e5d-b321-e87910746fba


brianaydemir (approved):
LGTM. I'll trust in your testing.

PR #2846, opened 2025-12-08
Fix token discovery logic

### LINKED ISSUE #2812: Strange behavior from token discovery code
I'm working with our friends at JLab, and we're seeing some strange behavior related to the `v7.21` client's token validation logic (on ap23 if that matters).

Consider this debug output from a _successful_ transfer that was invoked with `pelican object get osdf://jlab-osdf/gluex/<rest of the path> /dev/null`:
```
time="2025-11-18T12:49:56-06:00" level=debug msg="Token is not acceptable; clearing cache"
time="2025-11-18T12:49:56-06:00" level=debug msg="Using token from XDG_RUNTIME_DIR"
time="2025-11-18T12:49:56-06:00" level=debug msg="Opening token file: <file redacted>"
time="2025-11-18T12:49:56-06:00" level=debug msg="Out of token locations to search"
time="2025-11-18T12:49:56-06:00" level=warning msg="Using provided token  even though it does not appear to be acceptable to perform transfer"
```

Two things to note:
1. There's an extra space in the last message at "token  even". The line of code producing this error message lives [here](https://github.com/PelicanPlatform/pelican/blob/7a08b44e9c308ce7d2c1148a978c56ec688ee092/client/acquire_token.go#L367) and it _should_ be providing a token location. The first bug in this line is that the token location is empty, the second bug is that we should always be wrapping inputs to `%s`-formatted strings with quotes so it's obvious when the values are empty (better yet, use the `%q` formatter that @h2zh told me about recently, which does this automatically).
2. I was able to eyeball the token stored in the token file, and it appears completely valid (note that timestamps have been converted to human-readable form):
```
{
  "aud": "ANY",
  "sub": <redacted>,
  "ver": "scitoken:2.0",
  "nbf": "Tue Nov 18 12:12:21 PM CST 2025",
  "scope": "read:/gluex/ write:/gluex/",
  "iss": "https://cilogon.org/jlab",
  "exp": "Tue Nov 18 03:12:26 PM CST 2025",
  "iat": "Tue Nov 18 12:12:26 PM CST 2025",
  "jti": <redacted,
  "group": <redacted>
}
```
I double checked this against the information I can get via the Director and it all checks out. Furthermore, the token _did_ work, so the bug must live in our token detection code.

patrickbrophy:
Fixes #2812 - Token discovery was incorrectly rejecting valid SciTokens and logging empty token location. The cause of this was that token validation only checked for WLCG tokens, causing valid SciTokens to be incorrectly flagged as unacceptable.

### DIFF
--- a/client/main_test.go
+++ b/client/main_test.go
@@ -20,9 +20,6 @@ package client
 
 import (
 	"context"
-	"crypto/ecdsa"
-	"crypto/elliptic"
-	"crypto/rand"
 	"fmt"
 	"net"
 	"net/url"
@@ -31,10 +28,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
-	"time"
 
-	"github.com/lestrrat-go/jwx/v2/jwa"
-	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -46,8 +40,6 @@ import (
 	"github.com/pelicanplatform/pelican/server_structs"
 	"github.com/pelicanplatform/pelican/server_utils"
 	"github.com/pelicanplatform/pelican/test_utils"
-	"github.com/pelicanplatform/pelican/token"
-	"github.com/pelicanplatform/pelican/token_scopes"
 )
 
 // TestGetIps calls main.get_ips with a hostname, checking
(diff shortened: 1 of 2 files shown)


client/acquire_token.go
```diff
@@ -496,46 +514,106 @@ func tokenIsAcceptable(jwtSerialized string, objectName string, dirResp server_s
...
+		if matchesResource(targetResource, resource, opts.Operation) {
 			return true
 		}
 	}
 	return false
 }
 
+// matchesResource checks if the target resource matches the scope resource.
+// For shared URLs, exact matches are preferred, but prefix matching is also acceptable.
+func matchesResource(targetResource, scopeResource string, operation config.TokenOperation) bool {
```
turetske: It would be nice to have unit tests for these helper functions. Helps identify more specifically when/if we introduce bugs that cause token verification issues.

client/main_test.go
turetske: Looking at this because you altered the test. This really feels like it should be in acquire_token_test.go file rather than main.

PR #2863, opened 2025-12-11
Add support for admin groups

### LINKED ISSUE #2838: Allow to specify web admin group
Currently you can specify a list of users that should be treated as admins. We should support providing a group(s) and any user who is a member of that group is an admin.

patrickbrophy:
Resolves #2838. This PR adds a new configuration parameter, `Server.UIAdminGroups`. With this new parameter it is possible to specify admin groups. This PR also adds tests to ensure that the behavior is correct.


h2zh (changes requested):

h2zh (approved):
LGTM!

h2zh:
This PR may need a rebase before merging @patrickbrophy

PR #2881, opened 2025-12-16
Validate filesystem perms against namespace capabilities for posix origins

### LINKED ISSUE #2200: Try to validate fs permissions against ns capabilities for posix origins
Pelican happily lets an origin start with a capability like "PublicReads" even when the `xrootd` user doesn't have the filesystem permissions to read anything under the storage prefix. This results in hard-to-diagnose runtime errors that require log diving and prior experience to solve.

Instead, we could try to turn this into a startup failure for the origin by checking the storage prefix and its permissions against the capabilities assigned for that storage prefix. If you say Pelican/XRootD should be able to read `/foo` but the ownership/permissions on `/foo` aren't correct for that to happen, throw and error.

h2zh:
Check if the xrootd user has the required filesystem permissions for the configured capabilities

- Read+execute permissions for Reads, PublicReads, and Listings capabilities
- Write+execute permissions for Writes capability
- Also require storage prefix exists and is a directory

Also update tests to replace made-up storage prefixes with temp directories that have proper permissions, because:

- Tests run as root, but the permission check simulates what the xrootd daemon user would have. Since xrootd is neither the owner nor in root's group, it falls into the "Others" category. Setting 0777 ensures the "others" bits grant full access (rwx).

Here are some hand test ideas:
You may need to tweak the following two configs to observe the behaviors of an Origin in different scenarios.
```
  Exports:
    - StoragePrefix: /example
      Capabilities: [...] 
```

Scenarios:
-Storage prefix does not exist
-Storage prefix exists but xrootd user has no permissions
-Storage prefix directory uses 0755 (others have r-x but not w). Set `Capabilities: ["PublicReads", "Listings"]`. The origin is expected to run sucessfully because `r-x` is all that Reads and Listings require. Then add Writes back and re-run, where it should fail
-Storage prefix exists with correct permissions (happy path)


turetske (changes requested):

jhiemstrawisc (changes requested):
This may be mostly fleshed out via merge conflicts (which is why I'm not commenting on individual lines), but after a rebase, please double check this for any new instances of the old-style param setter `param.Set(ParamName string, value string)`

These should be updated to `param.ParamName.Set(value)`

turetske (approved):

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
LGTM. Thanks for the cleanup!

h2zh:
I talked to @williamnswanson today, this validation feature will not work with multi-user setup so should be disable when `Origin.Multiuser` is true. I'll create a patch for that in the 7.25 release cycle.

PR #2887, opened 2025-12-17
Add param that disables Director's hosting of federation metadata

jhiemstrawisc:
There is an entire class of bugs that revolves around having the
Director host a copy of federation metadata when it isn't also the
root discovery URL for the federation. We keep finding these bugs
in a trickle, but we haven't been able to squash them all because
of how entangled these concepts are.

Adding this knob doesn't _solve_ those bugs, but it should help us
detect them more easily by allowing us to turn of Director metadata
when we don't want people using the Director for that purpose.

### DIFF
--- a/e2e_fed_tests/director_test.go
+++ b/e2e_fed_tests/director_test.go
@@ -258,12 +258,12 @@ func TestDirectorFedTokenCacheAPI(t *testing.T) {
 
 			tok, err := jwt.ParseInsecure([]byte(tokStr))
 			require.NoError(t, err, "Failed to parse token")
-			// In this case, the "fed issuer" is the director because we're running as fed-in-a-box.
-			// However, that need not be true in general wherever the Director has a configured Federation.DiscoveryUrl.
+			// The fed-test utility uses a separate HTTP server for hosting federation metadata,
+			// and sets it as the Discovery endpoint -- tokens need to be issued by that endpoint
 			fedInfo, err := config.GetFederation(ctx)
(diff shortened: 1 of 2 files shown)


jhiemstrawisc:
@williamnswanson This new param also feels relevant to you, since we'll probably want to come up with a plan for turning it on in the OSDF (hopefully without exposing too many bugs at once...).

e2e_fed_tests/director_test.go
```diff
@@ -273,3 +275,151 @@ func TestDirectorFedTokenCacheAPI(t *testing.T) {
...
+			viper.Set(param.Director_EnableFederationMetadataHosting.GetName(), tt.enableHosting)
+			_ = fed_test_utils.NewFedTest(t, bothPubNamespaces)
+
+			ctx := context.Background()
+			ctx, _, _ = test_utils.TestContext(ctx, t)
+
+			fedInfo, err := config.GetFederation(ctx)
+			require.NoError(t, err)
+
+			fmt.Printf("\n\n\nFED INFO: %+v\n\n\n", fedInfo)
```
turetske: Is this an old debug statement?

turetske (changes requested):
One minor issue regarding a println that was left in.

jhiemstrawisc:
Added @patrickbrophy as a second reviewer because of the new Prometheus metric this adds.

patrickbrophy (commented):
The metric looks good 👍

PR #2929, opened 2025-12-28
Enable containers to add CAs to the trust store on start up

brianaydemir:
This PR makes two improvements to our container images.

The primary improvement is to `entrypoint.sh`, which now scans `/certs` for certificates (`*.crt` files) and adds them to the container's trust store. During development and testing, this should help minimize, if not eliminate, the need to resort to `TLSSkipVerify: true` or similar, especially for those programs that might not be aware of Pelican's configuration (e.g., `curl`).

> [!NOTE]
> There is hypothetically a backwards compatibility concern given that I chose to scan `/certs` based on where We™ currently seem to bind mount certificates into containers. That said, I would find it passing strange for a container to mount a certificate ***it*** doesn't trust but presumably expects the rest of the world to trust.

The secondary improvements are to the container image build process (the Dockerfile), which are intended to facilitate repeated image builds by improving layer caching and not building unnecessary binaries.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -726,6 +726,7 @@ RUN --mount=type=cache,id=dnf-${TARGETPLATFORM},target=/var/cache/dnf,sharing=lo
         libcap-devel \
         libuuid-devel \
         openssl-devel \
+        python3-devel \
         sqlite-devel \
         tinyxml2-devel \
         zlib-devel \


brianaydemir:
@jhiemstrawisc 

> I did a quick poll and we're all using the VSCode devcontainers extension, which doesn't go through the entrypoints defined here. At least in my case, I also rely on Pelican to generate a new CA on its first run whenever I need to restart my container, at which point it's too late to do something clever with entrypoints.

To quote Luke Skywalker, ["Every word of what you just said was wrong."](https://www.youtube.com/watch?v=2sRS1dwCotw)

Obviously, not literally so, but… I am aware of no production service that is expected to create its own certificate or CA on startup (or restart), and Pelican's released container images use these entrypoints, so the notion that there's a developer out there that doesn't go through these entrypoints is… What are they actually testing?

Issue #2928, opened 2025-12-28
Director's stat feature leaks goroutines


brianaydemir:
## Summary

The director maintains a cache of results from attempting to stat objects at origins and caches. Over time, the director can build up an increasing number of goroutines that are waiting on the following channel send:

https://github.com/PelicanPlatform/pelican/blob/074a9766f8c06e281c4fbfeb98a205f021288b56/director/stat.go#L490-L494

## Data

I've put together a Docker Compose-framework for reproducing and investigating this issue:

- [2025-12-28-test-director-stat.tar.gz](https://github.com/user-attachments/files/24361299/2025-12-28-test-director-stat.tar.gz)

Out of the box, the framework sets up 8 origins, all serving the same public namespace, and all serving the same set of objects.

Following the instructions in the included README, I observe the following:

<img width="1536" height="960" alt="Image" src="https://github.com/user-attachments/assets/8bedb5a3-55d7-4c3b-a2f0-0db03f03f1dd" />

The count of goroutines keeps climbing, as does memory usage. Examining the stacks of the goroutines

- [goroutine-1.txt](https://github.com/user-attachments/files/24361306/goroutine-1.txt)
- [goroutine-2.txt](https://github.com/user-attachments/files/24361308/goroutine-2.txt)

we can see that a distressing number of them are waiting on the channel send noted above. This suggests that hits to the stat cache are the immediate cause.

If I reconfigure the director to minimize hits to the stat cache

    CachePresenceCapacity: 1

and then re-run the test, I observe the following:

<img width="1536" height="960" alt="Image" src="https://github.com/user-attachments/assets/b3d0954a-702a-4757-92c1-3032c325fe1c" />

The count of goroutines stabilizes, as does memory usage.

If I disable the stat functionality entirely, I observe the following:

<img width="1536" height="960" alt="Image" src="https://github.com/user-attachments/assets/ac17a14c-cd4e-464b-8421-5377e33589b6" />

Again, the count of goroutines stabilizes, as does memory usage.

## Commentary

We have a strong indication that hits to the stat cache are problematic. In a production environment, it would not be surprising if the ever increasing pile of goroutines leads to a memory explosion in the director.

[...]

PR #2935, opened 2026-01-01
Fix director's stat feature to no longer "leak" goroutines

brianaydemir:
## Overview

Previously, when the director's "stat" feature encountered a hit to its cache, it would attempt to communicate the result via a channel. Unfortunately, the goroutine sending on the channel was also the same goroutine responsible for receiving on the channel, so the goroutine could become blocked on itself if the channel's buffer was at capacity. Goroutines might be inexpensive, but pile up enough of them, and memory usage will become noticeable.

Now, we'll ensure that channels' buffers have sufficient capacity.

> [!NOTE]
> An alternative solution here is to communicate the result of a stat cache hit in its own goroutine. It's not immediately obvious to me what incurs a greater performance penalty: allocating larger channel buffers, or spawning more goroutines. That said, these hypothetical stat cache hit goroutines would do so little work that it seems silly to make them synchronize on a channel send, which means we would want the channels to be buffered, anyway. And that means we'd need to decide on how large to make the buffers…

I'm also taking this opportunity to make log messages more consistent between the various outcomes, moving said log messages to "trace", and fixing what appears to be an issue with double counting in Prometheus metrics.

## My Testing

Behold, my new favorite toy, Docker Compose and a bunch of scripts and config files for starting up a data federation (this isn't substantively different from what I posted on #2928):

- [2026-01-02-pelican-test-framework.tar.gz](https://github.com/user-attachments/files/24409586/2026-01-02-pelican-test-framework.tar.gz)

Out of the box, this is what I see in Grafana:

Screenshot 2026-01-02 at 09 57 02

We can see that goroutines are no longer exploding in number.

## Testing Advice

You can tweak the above framework to force an unusual number of hits to the stat cache, few or no hits to the stat cache, etc. etc.

By editing `environment.cfg`, you can can quickly toggle between an official release of your choice, and containers built from this PR — useful for validating that you have a scenario that triggers the bug (or not), and that the PR made a difference (if necessary).

It is very much worth pulling up

* https://localhost:9000/api/v1.0/debug/pprof/goroutine?debug=1, and
* https://localhost:9000/api/v1.0/debug/pprof/goroutine?debug=2,

and searching the output for `stat.go`. The former link groups and counts goroutines by their stacks; the latter lists each one separately, along with how long it's been waiting on something. In both cases, you're keeping an eye out for goroutines that might be piling up in `stat.go` but at a rate that's too low to be seen clearly in Grafana (I encountered this scenario while developing this PR…).

### DIFF
--- a/director/stat.go
+++ b/director/stat.go
@@ -350,11 +350,12 @@ func (stat *ObjectStat) queryServersForObject(ctx context.Context, objectName st
 		return
 	}
 	timeout := param.Director_StatTimeout.GetDuration()
-	// Note there is a small buffer in each channel; in the case of a cache hit, we write
-	// to the channel from within this goroutine.
-	positiveReqChan := make(chan *objectMetadata, 5)
-	negativeReqChan := make(chan error, 5)
-	deniedReqChan := make(chan *headReqForbiddenErr, 5) // Requests with 403 response
+	// We both send to *and* receive from these channels within this
+	// goroutine, so ensure that they have sufficiently large buffers
+	// to prevent the goroutine from blocking on itself.
+	positiveReqChan := make(chan *objectMetadata, len(ads))
+	negativeReqChan := make(chan error, len(ads))
+	deniedReqChan := make(chan *headReqForbiddenErr, len(ads)) // Requests with 403 response
 	// Cancel the rest of the requests when requests received >= max required
 	maxCancelCtx, maxCancel := context.WithCancel(ctx)
 	numTotalReq := 0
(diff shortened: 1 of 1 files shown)


brianaydemir:
I revised my framework to make caches usable, which means you can test with more "realistic" configurations such as 2 origins and 6 caches:

- [2026-01-04-pelican-test-framework.tar.gz](https://github.com/user-attachments/files/24424158/2026-01-04-pelican-test-framework.tar.gz)

It turns out I was generating certificates badly, and that I wasn't giving the caches enough time to actually pull the correct auth info for the namespaces.

I also added in creating a token, for testing with transfers that require authorization.

PR #2936, opened 2026-01-02
Stubs for two prometheus endpoints

h2zh:
The quick fix of the linked issue is removing Grafana's access to Prometheus' `/rules` and `/alerts` endpoints, which is already done. This PR is a permanent fix:

Return empty results instead of nil to prevent nil pointer panics when external tools (like Grafana) query Prometheus' `/rules` or `/alerts` endpoints


patrickbrophy (changes requested):
@h2zh Could you write a small test to ensure that hitting these endpoints doesn't cause any undesirable behavior? Otherwise the stub implementation looks good.

PR #2945, opened 2026-01-05
Prevent test suite from killing Docker init process

### LINKED ISSUE #2944: Fix test suite to prevent docker container exiting while running tests
When running `go test ./...` in a dev container environment, the test suite will run for a while and then the dev container will exit with code 0.

patrickbrophy:
Fixes #2944. This PR prevents the XRootD restart feature from killing PID 1 and updates the test to use dummy PIDs.


jhiemstrawisc:
This fix works as advertised, and with it my dev container doesn't die anymore. @bbockelm, is there a reason this test should be sending a `SIGKILL` to PID 1 or is this change fine? Pointing `StoreRestartInfo` at PID 1 came from one of the few commits in the original PR that was authored by you and not Copilot, so I want to make sure changing this isn't breaking something else.

jhiemstrawisc (approved):

PR #2958, opened 2026-01-07
Implement API token generation command

### LINKED ISSUE #2956: API key generation command
It would be a nice feature to be able to create an API key from the command line. This could be used in instances where a tool like Puppet is setting up an origin. 

The command should be something like `pelican api-key generate --scopes monitoring.query,monitoring.scrape`. The implementation should be similar to the downtime commands.

patrickbrophy:
Fixes #2956. This PR implements a new command `pelican api-key generate`. This command generates an API token for a specified server. You must specify the scopes of the token and other fields have defaults if they are not specified.

### DIFF
--- a/cmd/api_key_test.go
+++ b/cmd/api_key_test.go
@@ -1,0 +1,442 @@
+/***************************************************************
+ *
+ * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License.  You may
+ * obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


cmd/api_key.go
```diff
@@ -0,0 +1,43 @@
+/***************************************************************
+ *
+ * Copyright (C) 2025, Pelican Project, Morgridge Institute for Research
```
h2zh:
```suggestion
 * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
```

cmd/api_key_generate.go
```diff
@@ -0,0 +1,192 @@
...
+	// Generate default name if not provided
+	name := apiKeyName
+	if name == "" {
+		name = fmt.Sprintf("cli-generated-%d", time.Now().Unix())
+	}
+
+	// Validate expiration format if provided
+	expiration := apiKeyExpiration
+	if expiration == "" {
+		expiration = "never"
```
h2zh: Same as above comment

h2zh (changes requested):
Code quality is good. Some unit tests would be appreciated!

PR #2969, opened 2026-01-09
Add new fields to AuthorizationTemplateForm.tsx

CannonLock:
Closes # 2 9 6 8


h2zh (approved):
LGTM!

Btw, I noticed a potential UX problem on the Config setting webUI: after user modifies any config's value and clicks "save", the config value doesn't change immediately. Instead, it takes approximately 1 min (and a manual refresh) to reflect the new value on webUI. This might need to communicate with facilitator and/or user before the fix.

PR #2972, opened 2026-01-10
Add GitHub action to check for large objects in PRs

bbockelm:
After a near-miss where a large file was committed (and I had to do some git surgery), this adds a PR validation to avoid large objects.


bbockelm:
Ok, self-merging.  7.23 just forked so there's minimal negative impact from that.

PR #2983, opened 2026-01-12
Pelican sync fixes

turetske:
This PR addresses multiple issues, all related to sync functionality.

The first, an issue where Pelican was creating directories with the incorrect permissions. This is now fixed with a masking issue.

The second, an issue where uploads on syncs where the file already existed were causing the whole sync to fail rather than skipping the uploads.

Another to have sync create temporary files on downloads until done.

### DIFF
--- a/client/util_test.go
+++ b/client/util_test.go
@@ -154,7 +154,9 @@ func TestGenerateTempPathFormat(t *testing.T) {
 	// Use filepath to construct platform-specific pattern
 	expectedDir := filepath.FromSlash("/data")
 	expectedBase := `.myfile\.dat\.[a-zA-Z0-9]{6}`
-	pattern := regexp.MustCompile(fmt.Sprintf(`^%s%c%s$`, regexp.QuoteMeta(expectedDir), filepath.Separator, expectedBase))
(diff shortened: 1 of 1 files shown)


patrickbrophy (approved):
Not sure why the integration test failed, but the code looks good and well tested. LGTM

bbockelm:
@turetske / @patrickbrophy - need a quick follow-up here for two things before this gets released:

1.  No arbitrary sleeps in unit tests.  That's how you grow a flakey unit test.  If you're waiting for something to occur (like a randomly-named file popping up in a directory that's a temp file), use `require.Eventually` which handles all the polling / timeout logic for you.  Anything with this pattern is surefire way to a future headache:

   ```go
   	// Give download time to start
	time.Sleep(50 * time.Millisecond)
   ```

2. Never randomly generate a pathname when creating a temporary file (see https://github.com/PelicanPlatform/pelican/pull/2983/changes#diff-1d422c608d8259c7eb511edc34aff9eac59b6fea0837f6a322f81afa01669677R46).  This is always a race condition and sometimes a security vulnerability; it's an anti-pattern we should scrub out.  Instead, use `os.CreateTemp`.

PR #2990, opened 2026-01-14
Fall back to system config location for non-root users if user config is not provided

jhiemstrawisc:
I thought this was a feature previously, but it turns out it probably wasn't.

In any case, Pelican now falls back to its standard system config location '/etc/pelican' in the event that a non-root user doesn't supply their own config.

This behavior is needed for site-local caches because HTCondor jobs don't run as root, and system admins need a way to configure which Cache an EP will use without pre-generating config for each individual user of the EP.

### DIFF
--- a/docs/app/parameters/page.mdx
+++ b/docs/app/parameters/page.mdx
@@ -1,44 +1,13 @@
-import parameters from '/public/static/parameters.json'; import Parameters from "/components/Parameters"; import { Callout } from 'nextra/components'
+import parameters from '/public/static/parameters.json';
+import Parameters from "/components/Parameters";
+import { Callout } from 'nextra/components'
 
 # Configuration
 
 ## Yaml Configuration
-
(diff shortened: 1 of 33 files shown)


docs/app/parameters/page.mdx
```diff
@@ -5,8 +5,38 @@ import { Callout } from 'nextra/components'
...
+### Configuration File Location
+
+Pelican automatically searches for its configuration file in different locations depending on the user running the process:
+
+- **Root User**:
+  - The default location is `/etc/pelican/pelican.yaml` (or `C:\ProgramData\pelican\pelican.yaml` on Windows).
+
+- **Non-Root User**:
+  - Pelican first checks for a user-specific configuration at `$HOME/.config/pelican/pelican.yaml`.
+  - If that file does not exist, it falls back to checking `/etc/pelican/pelican.yaml`.
```
turetske: Do you want to specify the windows fallback location here as well?

turetske (approved):
LGTM!

PR #2992, opened 2026-01-14
Use a stock AlmaLinux image as the base for container images

brianaydemir:
## Summary

Pelican's container images are now based on a stock AlmaLinux 9 image and not OSG's `software-base` image.

Users should see no difference in behavior. Admins who regularly `exec` into running containers might find differences in the software packages that are installed.

## Details

The problem in the linked issue seems to occur only when using QEMU emulation to build the ARM64 images. I did not find a way to determine exactly what problem `tar` was running into; I only determined that a `dnf distro-sync` in OSG's `software-base` image seemed to be the cause.

While there are a variety of improvements that we could make to how we build Pelican's container images — notably, to use GitHub's ARM64 runners rather than emulation — in the interest of not delaying the 7.23 release, the solution here is to simply switch out the base image.

Along the way, I've taken a moment to combine some of the layers in earlier build stages. Apparently too many layers [can lead to performance problems](https://github.com/opensciencegrid/docker-software-base/issues/71), and arguably, the build process is a little clearer if we can consolidate the "scripts" for each stage. I've also made `set -o pipefail` a more consistent feature of these scripts; it's one of those things that's easily forgotten and will bite us when least expected.

I also took a moment to clean-up the `.dockerignore` file. One of my pet peeves is an "ignore" file that mysteriously lists a bunch of paths with no justification for why.

For testing, there was a moment in time when this PR hacked to the GitHub Actions to force ARM64 builds (we usually build only x86 for PRs). The runs of interest can be found here:

* https://github.com/PelicanPlatform/pelican/actions/runs/21008220110/job/60398686972?pr=2992
* https://github.com/PelicanPlatform/pelican/actions/runs/21008220110/job/60398687038?pr=2992

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,6 +1,6 @@
 # ***************************************************************
 #
-#  Copyright (C) 2025, Pelican Project, Morgridge Institute for Research
+#  Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
 #
 #  Licensed under the Apache License, Version 2.0 (the "License"); you
 #  may not use this file except in compliance with the License.  You may
@@ -30,6 +30,7 @@
 # ARG TARGETARCH
 
 # For selecting OSG-maintained packages.
+ARG ALMALINUX_TAG=9
 ARG BASE_OS=el9
 ARG OSG_SERIES=24
 ARG OSG_REPO=release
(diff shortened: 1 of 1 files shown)


brianaydemir:
@patrickbrophy I handed this one to you because you'll likely be getting the PR that upgrades and redoes how we install OA4MP, so I figured you might as well start taking a look at how the `Dockerfile` works.

brianaydemir:
For reference: [the Dockerfile for OSG's `software-base` image](https://github.com/opensciencegrid/docker-software-base/blob/master/Dockerfile).

PR #2995, opened 2026-01-15
Switch to v1.25

bbockelm:
For the POSIXv2 work, we need golang v1.25 to get `os.Root` functionality (to prevent symlinks from getting data outside the storage root).

However, this requires a two-step dance: one PR for the v1.25 upgrade to update the testing container then #2960 for the POSIXv2 functionality itself (whose GitHub Actions are executed inside the container).


brianaydemir (approved):
LGTM.

PR #2999, opened 2026-01-16
Fix http.Transport copy-by-value causing race conditions

turetske:
Replace Transport dereference pattern with Clone() to prevent concurrent map write panics. When http.Transport is copied by value (by dereferencing a pointer), it creates separate mutex copies that no longer protect the shared internal map, leading to race conditions.

Fixed instances:
- oa4mp/proxy.go: Use Clone() instead of dereferencing *GetTransport() when customizing DialContext for Unix socket communication
- web_ui/engine_test.go: Use Clone() for test transport customization

Also added `go vet` to the .pre-commit checker

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,10 +57,3 @@ repos:
         args: [--wrap, "no"]
         files: \.md$
         exclude: \.mdx$
--   repo: local
-    hooks:
-    -   id: go-vet
-        name: go vet
-        entry: bash -c 'for dir in $(printf "%s\n" "$@" | xargs -n1 dirname | sort -u); do go vet "./$dir"; done' --
-        language: system
-        types: [go]


.pre-commit-config.yaml
```diff
@@ -57,3 +57,10 @@ repos:
         args: [--wrap, "no"]
         files: \.md$
         exclude: \.mdx$
+-   repo: local
+    hooks:
+    -   id: go-vet
+        name: go vet
+        entry: bash -c 'for dir in $(printf "%s\n" "$@" | xargs -n1 dirname | sort -u); do go vet "./$dir"; done' --
+        language: system
+        types: [go]
```
h2zh: Could you explain why we need to trigger `go-vet` here explicitly? I believe it is already incorporated in `golangci-lint` on line 34 https://golangci-lint.run/docs/linters/ 

Golangci-lint enables a default set of linters that includes `govet`. Since it's not disabled in this configuration `/.golangci.yaml`, it will run as part of the golangci-lint checks.

Issue #3003, opened 2026-01-19
Run all tests against POSIXv2 backend


bbockelm:
The POSIXv2 backend tests only cover a subset of functionality; the general test suite is more thorough.

Create a way to default tests to use POSIXv2 and fix any issues arising.

PR #3004, opened 2026-01-19
Fix minor POSIXv2 backend issues

bbockelm:
I added a small switch that allows you to move all federation tests from POSIX to POSIXv2.  This revealed some minor implementation errors / inconsistencies between POSIXv1 (XRootD) and POSIXv2.

With this PR, all tests pass with POSIXv2!

Fixes: #3003


bbockelm:
@turetske - these minor fixes are worth backporting into the next v7.23 RC.

PR #3008, opened 2026-01-20
Periodic metadata discrepancy check for the Director

h2zh:
This feature monitors configuration discrepancies between the Director's locally-served federation metadata and the metadata provided by the Discovery URL. It compares three critical components:

1. **Director endpoint**
2. **Registry endpoint**
3. **JWKS keys**

#### Director and Registry Endpoints
For the Director and Registry endpoints, a discrepancy is straightforward: **the URLs must match exactly**. Any mismatch indicates a configuration issue that should be flagged to the administrator.

#### JWKS Keys (Special Definition)
For JWKS keys, the definition of discrepancy is different due to operational flexibility requirements:

- **Discrepancy exists**: When there are **zero overlapping keys** between the Director's JWKS and the Discovery URL's JWKS
- **No discrepancy**: When there is **at least one overlapping key** between both key sets

**Rationale**: In production environments, administrators may configure different public keys at the Director and Discovery endpoints for various operational reasons (e.g., key rotation, multi-tenant scenarios). The federation can operate correctly as long as **at least one key is shared** between both endpoints, allowing for cross-validation and trust establishment.

### Warning Behavior

A warning banner is displayed when any of the following scenarios happens:
- Director endpoint URLs don't match exactly
- Registry endpoint URLs don't match exactly
- JWKS key sets have **no overlapping keys** (zero intersection)

### Note for reviewers

This is a full-stack PR so I kindly request two reviewers to screen it from backend and frontend respectively.

Screenshot 2026-01-17 at 11 45 25 AM

Here are the relevant configs:
```
  ###################
  # Loc Federation  #
  ###################
  RegistryUrl: https://<YOUR_LOCAL_Registry_HOSTNAME>:port
  DirectorUrl: https://<YOUR_LOCAL_Director_HOST>:port
  # You can switch between Director URL and Discovery URL to trigger the banner in this PR
  DiscoveryUrl: https://<YOUR_LOCAL_Director_HOST>:port OR https://osg-htc.org
  JwkUrl: https://<YOUR_LOCAL_Director_HOST>:port/.well-known/issuer.jwks OR "https://osdf-director.osg-htc.org/.well-known/issuer.jwks"
```

### DIFF
--- a/director/metadata_comparison_test.go
+++ b/director/metadata_comparison_test.go
@@ -20,17 +20,12 @@ package director
 
 import (
 	"context"
-	"crypto/ecdsa"
-	"crypto/elliptic"
-	"crypto/rand"
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 	"time"
 
-	"github.com/lestrrat-go/jwx/v2/jwa"
-	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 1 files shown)


CannonLock (approved):
LGTM

director/metadata_comparison_test.go
```diff
@@ -0,0 +1,532 @@
...
+			} else {
+				_, _ = w.Write([]byte(jwksJson2))
+			}
+			callCount++
+		}))
+		defer server.Close()
+
+		hasOverlap, err := compareJwksKeys(ctx, server.URL+"/jwks1", server.URL+"/jwks2")
+		require.NoError(t, err)
+		assert.False(t, hasOverlap, "Different keys should return no overlap")
```
turetske: I think the turn "overlap" is misleading. I would say use "discrepancy" and say "Different keys should return a discrepancy" - "overlap" can be a partial term, which isn't really what we are looking for here. Especially since the overall PR is about discrepancies.

PR #3009, opened 2026-01-20
Sync JWKS URI

h2zh:
This PR syncs the JWKS URI in federation discovery endpoint `/.well-known/pelican-configuration` with the JWKS URI in Pelican process' global federation metadata `pelican_url.FederationDiscovery`

- Before this PR, the JWKS URI in federation discovery endpoint is just appending `/.well-known/pelican-configuration` to Director URL, even though the Discovery URL is set. This means the JWKS URI provided by Discovery URL may contain a different set of keys.
~~But this may be intentional! Director should advertise its own key endpoint rather than potentially pointing to a different server's JWKS URI.~~

> A quick recap: In `discoverFederationImpl` func, `fedInfo.JwksUri` will first fill with the value of the `Federation.JwkUrl` config param. If it is not set, that function will probe Discovery Endpoint to get the `jwks_uri` there. At last, if `fedInfo.JwksUri` is still empty, it will fallbacks to Director's externalUrl + "/.well-known/issuer.jwks"

- This PR will work with #3008. [How does they work?](https://github.com/PelicanPlatform/pelican/pull/3009#issuecomment-3786059290)
- ~~This PR was intended to replace #3008. But I realize maybe the good practice for a server is using its own JWKS URI, instead of pointing to another's, even though there might be discrepancy for the keys between them! We should still use a dedicated goroutine to monitor the discrepancy, as in #3008. So I just close this PR.~~

### DIFF
--- a/director/discovery.go
+++ b/director/discovery.go
@@ -184,10 +184,10 @@ func federationDiscoveryHandler(ctx *gin.Context) {
 
 	jwksUri := fedInfo.JwksUri
 	if jwksUri == "" {
-		log.Error("Bad server configuration: fail to get JwksUri: ", err)
+		log.Error("Bad server configuration: Director is missing federation keys (jwks_uri is empty)")
 		ctx.JSON(http.StatusInternalServerError, server_structs.SimpleApiResp{
 			Status: server_structs.RespFailed,
-			Msg:    "Bad server configuration: unable to get JwksUri",
+			Msg:    "Bad server configuration: Director is missing federation keys (jwks_uri is empty)",
 		})
 		return
 	}


h2zh:
After one day of consideration, my plan is moving forward with both this PR and #3008.

This PR can ensure the same keys are provided via the `/.well-known/pelican-configuration` endpoints at Director and discovery URL (just like the 1. you mentioned). 

And #3008 can warn the Director admin for rare scenarios when they overwrite the config related to the metadata. Beside your example in 2., they can customize the `jwks_uri` with
```
Federation:
  JwkUrl: <>
```

director/discovery.go
```diff
@@ -182,12 +182,12 @@ func federationDiscoveryHandler(ctx *gin.Context) {
 
 	brokerUrl := fedInfo.BrokerEndpoint
 
-	jwksUri, err := url.JoinPath(directorUrl.String(), directorJWKSPath)
-	if err != nil {
-		log.Error("Bad server configuration: fail to generate JwksUri: ", err)
+	jwksUri := fedInfo.JwksUri
+	if jwksUri == "" {
+		log.Error("Bad server configuration: fail to get JwksUri: ", err)
```
jhiemstrawisc: Two comments here:
1. I don't think you want to provide this `err` here because it's coming from the previous `registryUrl, err := url.Parse(registryUrlStr)` and should thus always be nil
2. I think we should avoid using the internal variable name `JwksUri` here. Instead, can you say something like `Bad server configuration: Director is missing federation keys (\"jwks_uri\")`? I like this a little better because someone who looks at the metadata JSON will see `jwks_uri`, not `JwksUri`.

jhiemstrawisc (approved):

PR #3014, opened 2026-01-21
Add support for prestaging API in the cache

### LINKED ISSUE #3013: Use prestage API from xrdhttp-pelican
The xrdhttp-pelican plugin provides a prestaging API that causes the cache to bring in a particular object; this is an improvement over the `pelican object prestage` we currently have where prestage is interpreted as a download to `/dev/null`.

Detect when the plugin is present and have the transfer engine use it if it is.

Note: I just released xrdhttp-pelican v0.0.9 which fixes a permission bug that would prevent this from working.  We'll need to merge in the work on this issue _after_ the base container upgrades to at least that version.

This is a follow-up of #2066 which was closed due to inactivity (basically, that ticket requested two things - prestage and evict; this only targets prestage).

bbockelm:
We auto-detect whether the cache has the prestaging API loaded; if so, switch to that instead of downloading to /dev/null as a way to simulate prestaging.

Note that the tests need xrdhttp-pelican v0.0.9 or later installed into the development container to pass.

Fixes #3013

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -2214,6 +2214,7 @@ func downloadObject(transfer *transferFile) (transferResults TransferResults, er
 
 		// Traditional prestage: download to /dev/null
 		localPath = os.DevNull
+		writeDestination = os.DevNull
 		fileWriter = io.Discard
 	}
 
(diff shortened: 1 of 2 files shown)


bbockelm:
Ok, I think this is ready to roll!

PR #3016, opened 2026-01-21
Suppress record not found logs

h2zh:
- GORM would generate a log for each DB query, no matter success  (level: debug) or failure (level: error)
- However, when the user starts an Origin/Cache for the first time, the DB is empty so the DB query will definitely get a "record not found" error
- This error will be logged by GORM by default as an error log. But this may be misleading for the user because this is an expected behavior. The "record not found" is already handled by the subsequent [code](https://github.com/PelicanPlatform/pelican/blob/513a13a4f891b4503b453750d732539e856a4058/registry/registry_db.go#L318).
- So this PR aims to suppress query logs on these two DB operations to avoid confusions.

### DIFF
--- a/database/server.go
+++ b/database/server.go
@@ -11,6 +11,7 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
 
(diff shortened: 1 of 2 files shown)


patrickbrophy (approved):
LGTM

Issue #2997, opened 2026-01-15
Improve how GitHub Actions build multi-architecture container images


brianaydemir:
The current GitHub Actions for building Pelican's container images:

1. Take an inordinate amount of time for pushes to `main` and semver tags.
2. Do not attempt multi-architecture builds for pull requests, due to (1).
3. Depend on 3+ GiB of the GitHub Actions cache persisting for the duration of the run.

This all results in issues such as #2993 not being caught in a timely fashion, and mysterious failures with "more than a few" concurrent merges to `main`, pushes to PR branches, and the like.

Issue #2502, opened 2025-07-16
PRs that touch `Dockerfile` should run tests in a container image based on that `Dockerfile`


jhiemstrawisc:
One consequence of the way we choose an image to run our PR tests in was exposed in #2490 -- this PR updated the way several container dependencies were installed, but in doing so broke the `libLotMan` installation.

However, because the LotMan tests were run in an image generated off the tip of main, the tests passed when they should have failed. We only caught this after the PR was merged and tests in other PRs started failing because of the new image.

I think we've discussed this in the past, but whenever a PR touches `images/Dockerfile`, that PRs tests should really be run in the new container. The downside is these tests will take much longer because the action needs to re-build the image instead of pulling.

Are there other files that, when touched, should trigger this behavior?

brianaydemir:
@jhiemstrawisc So, I'm blocked on the following:

In any action run triggered by a pull request, we do not have access to any repository secrets. This means that while we can build a new `pelican-test` image, we cannot push it anywhere. As a result, we cannot persist the image beyond the lifetime of the current job/runner instance.

This is a problem because we run the test suite in a _new_ job/runner instance. We'd have to re-engineer the suite so that it could be run via something like `docker run <image> <script>`.

brianaydemir:
I'm closing this issue / declaring bankruptcy. Short of comprehensively rewriting Pelican's testing workflow, I'm not seeing how we can make any forward progress on this issue.

PR #3022, opened 2026-01-22
Enable the build and test workflow to push images to non-Pelican repos

brianaydemir:
## Summary

The main innovation of this PR is introducing a scheme by which a developer can run Pelican's build-and-test workflow on a branch in their fork of the main PelicanPlatform/pelican repo. This includes running the workflow in a version of the testing container that has been updated to take into account changes the branch has made to Pelican's dependencies and the workflows themselves.

As a demonstration and test of this scheme, this PR includes an update to the definition of the testing container so that the build-and-test workflow actually runs the `TestHTCondorPlugin` unit test, and also fixes the test so that it runs as intended.

## The Scheme

Requirements:

* The developer has a project on hub.opensciencegrid.org that they can:

    1. Pull/push container images to
    2. Create a "robot" account with similar privileges

The basic idea:

0. The developer updates `build-and-test.yml` with their GitHub fork and corresponding project on hub.opensciencegrid.org. (This needs to be done only once, as the change can be merged into the PelicanPlatform/pelican repo.)

1. The developer enables the running of GitHub Actions on their fork. (They are disabled by default.)

2. The developer adds `PELICAN_HARBOR_ROBOT_USER` and `PELICAN_HARBOR_ROBOT_PASSWORD` repository secrets to their fork, with contents befitting of their names.

3. As needed, the developer creates a new (pre-)release on their fork. The tag should be a semvar tag following Pelican's release tagging scheme. The branch should be the one they will be using to create pull requests against PelicanPlatform/pelican.

> [!NOTE]
> To be more precise, the developer will still need to make two PRs against PelicanPlatform/pelican: One to update dependencies or similar, so that a new version of the testing container is built. And a second with all of the changes that need to be tested in that updated container. The idea and point here is to provide a way that the developer can test the combined effects of both PRs _before_ submitting them.

While this isn't quite what #2502 asked for, I feel like it's close enough.

## Other Details

The entirety of the build-and-test workflows could use a bit of love, which I intend to take care of when I tackle #2997.

In the meantime, the number of files that #2995 had to touch is disappointing, so I've gone through and updated everything else to use `go.mod` and `package.json` as the authoritative version specifiers for Go and Node.

### DIFF
--- a/.github/workflows/check-go-generate.yml
+++ b/.github/workflows/check-go-generate.yml
@@ -21,6 +21,7 @@ jobs:
         uses: actions/setup-go@v6
         with:
           go-version-file: "go.mod"
+          check-latest: true
           cache: false
 
       - name: Run Go Generate
--- a/.github/workflows/pre-commit-linter.yml
+++ b/.github/workflows/pre-commit-linter.yml
@@ -21,6 +21,7 @@ jobs:
         uses: actions/setup-go@v6
         with:
           go-version-file: "go.mod"
+          check-latest: true
           cache: false
 
       - name: Set up Python
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,6 +36,7 @@ jobs:
         uses: actions/setup-go@v6
         with:
           go-version-file: "go.mod"
+          check-latest: true
 
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v6
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -52,6 +52,7 @@ jobs:
       uses: actions/setup-go@v6
       with:
         go-version-file: "go.mod"
+        check-latest: true
     - name: Install gotestsum
       run: |
         go install gotest.tools/gotestsum@latest
--- a/.github/workflows/test-macos-windows.yml
+++ b/.github/workflows/test-macos-windows.yml
@@ -51,6 +51,7 @@ jobs:
       uses: actions/setup-go@v6
       with:
         go-version-file: "go.mod"
+        check-latest: true
     - name: Create Go temporary directories (Windows workaround)
       if: runner.os == 'Windows'
       shell: pwsh
@@ -65,6 +66,7 @@ jobs:
       uses: actions/setup-go@v6
       with:
         go-version-file: "go.mod"
+        check-latest: true
     - name: Install macOS Dependencies
       run: ./github_scripts/osx_install.sh
       if: runner.os == 'macOS'


go.mod
```diff
@@ -5,6 +5,8 @@ retract [v1.0.0, v1.0.5]
 
 go 1.25.0
 
+toolchain go1.25.5
```
jhiemstrawisc: I'm going through commit-by-commit like you suggested, so maybe I'll see the answer to this later, but why is pinning the toolchain to this specific version needed?

I ask because it looks like BrianB recently made the explicit choice to _remove_ the toolchain definition. See:
https://github.com/PelicanPlatform/pelican/commit/2f798fab1813ec9eba70f32b4614447c14473959#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R6

.github/scripts/validate-defaults/main.py
```diff
@@ -1,43 +1,44 @@
...
+    errors = []
+    for key, default in defaults.items():
+        if key not in parameters:
+            errors.append(f"Key for default value is not in the parameters file: {key}")
+            continue
+
+        # If the defaults file has a value, it must match the parameters
+        # file. The reverse need not be true (the default value may be set
+        # via other means, i.e., code).
+        if default != parameters[key]["default"]:
```
jhiemstrawisc: This script still feels inadequate to me since the bulk of our defaults aren't actually defined in `config/resources/defaults.yaml`.

What if instead the solution did something like:
1. Build Pelican.
2. In a clean environment with no existing Pelican configuration, run `pelican config dump` or `pelican config dump --json` (whichever format is easier to work with).
3. Check the output of this against `docs/parameters.yaml`.

Not only might this work with params that aren't defined via `config/resources/defaults.yaml`, but because some of our defaults use bashisms that reference other parameters:
```
name: Origin.Url
description: |+
  The origin's configured URL, as reported to XRootD. This is the file transfer endpoint for the origin.
type: url
default: https://${Server.Hostname}:${Origin.Port}
components: ["origin"]
```
we could actually evaluate correctness in these more complex cases.

Feel free to say "out of scope," but I thought I'd bring it up since you took the time to clean things up here.

.github/scripts/validate-parameters/main.py
```diff
@@ -1,63 +1,80 @@
...
-                errors.append(f"Parameter component is not a list: {parameter['name']}")
-                continue
+        if parameter["type"] == "object" and parameter["name"] not in VERIFIED_OBJECT_STRUCTURES:
+            errors.append(f"Parameter's object structure not verified for configuration via the web UI: {parameter['name']}")
 
-            if parameter['type'] not in ENUMERATIONS['type']:
-                errors.append(f"Parameter type is not a recorded type. Please add web dev as a reviewer: {parameter['name']}")
-                continue
+    if errors:
+        raise RuntimeError("\n" + "\n".join(errors))
```
jhiemstrawisc: Another place where I think a quick reference to the filepath `docs/parameters.yaml` could be useful.

images/build-stages.svg
jhiemstrawisc: Wizardry!

images/Dockerfile
```diff
@@ -640,54 +642,44 @@ CMD ["serve"]
...
-
-# Create alice user for multi-user testing
-RUN <<ENDRUN
-  set -eux
-  set -o pipefail
-
-  useradd -m -s /bin/bash alice
+  # Add some normal user accounts for testing.
+  # The inspiration for their names: https://en.wikipedia.org/wiki/Alice_and_Bob#Cast_of_characters.
+  for name in alice bob carol dave; do
```
jhiemstrawisc: Apparently you don't care about security because you didn't let Eve join in the fun 😛

.github/workflows/build-and-test.yml
```diff
@@ -33,31 +33,67 @@ on:
...
+              REGISTRY_REPO="pelican_platform"
+              IS_PRIMARY_GITHUB_REPO=true ;;
+
+            # Some testing scenarios require the ability to push container
+            # images to a registry. Here, we list developers who have forked
+            # the main PelicanPlatform/pelican repo and are able to push
+            # images somewhere.
+
+            brianaydemir/pelicanplatform-pelican)
+              REGISTRY_REPO="brian.aydemir.r1" ;;
```
jhiemstrawisc: Little hard to piece this together -- is `REGISTRY_REPO` always going to be tied `hub.opensciencegrid.org` later on?

brianaydemir: As a practical matter, I would bet yes.

As a theoretical matter, there's no need to assume it. In fact, the way this workflow constructs image names in general is in need of cleaning up, which I'm doing in #3051. Look forward to the "fix" there.

go.mod
brianaydemir: So, the goal here is to have _one_ location that defines the Go version to use, and `go.mod` seems as good a place as any because it already expresses an opinion on that.

I think the actual problem is that unless Go and/or GoReleaser automatically update to the latest version of the toolchain, even when the currently installed version already satisfies any stated requirements (e.g., the `go` line in `go.mod`), we've effectively been pinning the toolchain version in `images/Dockerfile` since at least #1877.

That said, there's an API that can be used to look up what the latest version is.

.github/scripts/validate-defaults/main.py
```diff
@@ -1,43 +1,44 @@
...
+                result[f"{key}.{key2}"] = value2
         else:
-            out[key] = value
-    return out
+            result[key] = value
+    return result
+
+
+def main():
+    with open(DEFAULTS_FILE, mode="r", encoding="utf-8") as fp:
```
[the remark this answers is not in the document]
brianaydemir: On the one hand, I don't think anything other `default` shows up at https://docs.pelicanplatform.org/parameters.

On the other hand, what's the worst that could happen if I add the check:

```
./config/resources/osdf.yaml: Xrootd.ManagerHost: Expected 'redirector.osgstorage.org', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Xrootd.SummaryMonitoringHost: Expected 'xrd-report.osgstorage.org', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Xrootd.DetailedMonitoringHost: Expected 'xrd-mon.osgstorage.org', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Federation.DiscoveryUrl: Expected 'osg-htc.org', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Federation.TopologyNamespaceUrl: Expected 'https://topology.opensciencegrid.org/osdf/namespaces?production=1', but found 'https://topology.opensciencegrid.org/osdf/namespaces' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Topology.DisableDowntime: Expected 'False', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Topology.DisableCacheX509: Expected 'False', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Topology.DisableOriginX509: Expected 'False', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Topology.DisableCaches: Expected 'False', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
./config/resources/osdf.yaml: Topology.DisableOrigins: Expected 'False', but found 'None' for 'osdf_default' in './docs/parameters.yaml'
```

As for the rest, see [my responses to other comments](https://github.com/PelicanPlatform/pelican/pull/3022#discussion_r2751316918).

images/Dockerfile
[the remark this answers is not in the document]
brianaydemir: I'll try to clarify the comment.

Basically, I'm too lazy to open up a web browser, so it's sometimes nice to have a "quick reference" or "cheatsheet" of sorts in the file itself.

And if you're not me, I imagine that it might be nice to have all the possible build args listed in a single location, at the top of the file, even if you don't need to ever set them. (The only thing worse than needing to update a magic string is… updating a magic string buried in the middle of a file you otherwise don't care about.)

images/Dockerfile
brianaydemir: ~6~4~0k~ users should be enough for anyone.

go.mod
bbockelm: Honestly, not 100% sure what Go's behavior is here.  One of the nifty (scary?) things introduced last year is the ability to have the runtime and compiler be independent -- go version N compiler can compile code for N+2, for example.

So, despite the CI/CD using 1.25.5 at https://github.com/PelicanPlatform/pelican/blob/main/images/Dockerfile#L39, it's a touch unclear what the behavior is.

If we do want to start freezing the toolchain (which I'm still not 100% convinced on), it seems like we need another metafile so we can set it in precisely one place (contrast with the various XRootD plugins...)

PR #3034, opened 2026-01-27
Add single-object groups/users APIs for groups and users and fix groups members field

### LINKED ISSUE #2741: Group + User Additional Endpoints
Couple items are needed to enhance the groups UI.

I want to be able to update the existing groups and users.

PATCH /groups/<groupId>
PATCH /users/<userId>

I want to be able to get a single group or user based on its ID. 

GET /groups/<groupId>
GET /user/<userId>

Additionally currently the members attribute in the groups json response is always null.

patrickbrophy:
Closes #2741.

This PR does the following:
- Add `GET/PATCH` endpoints for groups and users
- Extend DB helpers and preload Members.User so Group.members is always an array and not null
- Wire new routes in ui.go and add tests to cover the new endpoints and updated group behavior

### DIFF
--- a/web_ui/groups.go
+++ b/web_ui/groups.go
@@ -200,7 +200,12 @@ func handleUpdateGroup(ctx *gin.Context) {
 		})
 		return
 	}
-	isAdmin, _ := CheckAdmin(user, groups)
+	isAdmin, _ := CheckAdmin(UserIdentity{
+		Username: user,
+		ID:       userId,
+		Groups:   groups,
+		Sub:      ctx.GetString("OIDCSub"),
+	})
 
 	if err := database.UpdateGroup(database.ServerDatabase, id, req.Name, req.Description, userId, isAdmin); err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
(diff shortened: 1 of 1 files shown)


swagger/pelican-swagger.yaml
```diff
@@ -3968,6 +3968,88 @@ paths:
...
+          schema:
+            $ref: "#/definitions/ErrorModelV2"
+        "404":
+          description: Group not found
+          schema:
+            $ref: "#/definitions/ErrorModelV2"
+        "500":
+          description: Internal server error
+          schema:
+            $ref: "#/definitions/ErrorModelV2"
```
h2zh: I didn't see the delete endpoints for both user and group. Is it an intentional redaction?

database/collection.go
```diff
@@ -685,12 +715,39 @@ func GetGroupWithMembers(db *gorm.DB, groupId string) (*Group, error) {
...
 func ListGroups(db *gorm.DB) ([]Group, error) {
 	groups := []Group{}
-	if err := db.Find(&groups).Error; err != nil {
+	if err := db.Preload("Members.User").Find(&groups).Error; err != nil {
 		return nil, err
 	}
 	return groups, nil
 }
 
+func UpdateGroup(db *gorm.DB, id string, name, description *string) error {
```
h2zh: Since `DeleteGroup` allows admin to delete group, I think we should allow admin to edit group too. What do you think?

database/collection.go
```diff
@@ -770,3 +827,67 @@ func GetAllCollections(db *gorm.DB) ([]Collection, error) {
...
+
+		return nil
+	})
+}
+
+// DeleteUser deletes a user and cleans up any collection ACL entries that reference the
+// user's implicit personal group name ("user-"+username).
+func DeleteUser(db *gorm.DB, userID string) error {
+	var user User
+	if err := db.First(&user, "id = ?", userID).Error; err != nil {
```
h2zh: Same as above

h2zh (changes requested):
A few comments. They should be straightforward to fix.

PR #3038, opened 2026-01-27
Bump version of xrootd-s3-http to v0.6.3

patrickbrophy:
Bump xrootd-s3-http to 0.6.3 and fix OSX installer


h2zh:
It would be nice to have a accompanying issue for this PR, since now it is more than just a simple version bump @patrickbrophy

PR #3051, opened 2026-01-29
Use architecture-specific runners for multi-architecture builds

brianaydemir:
## Summary

This PR aims to address point (1) of #2997 by using architecture-specific GitHub Actions runners for building Pelican's container images.

If that turns out to work well in practice, then point (2) is a straightforward change to `.github/workflows/build-and-test.yml`.

Point (3) is still a bit of a mystery to me, wherein I'll need to observe the behaviors of our revised GitHub Actions workflows before deciding upon further improvements. That said, I'm hoping the various changes made in this PR and #3022 will result in fewer "useless" caches.

## Details

As with #3022, @jhiemstrawisc I would suggest reviewing this PR commit-by-commit, because the commit messages contain more information about what I had in mind.

I'm also more than happy to trigger, on demand (ask), a particular scenario in [my own fork](https://github.com/brianaydemir/pelicanplatform-pelican/).

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -62,6 +62,9 @@ jobs:
         env:
           JUNIT_FILE: junit-${{ matrix.binary_name }}.xml
         run: |
+          # Building the web UI on Windows merely creates an empty file at
+          # web_ui/frontend/out/index.html. Nevertheless, we'll go through
+          # the same steps as for other OSes, in case that changes.
           echo "::group::Building web UI"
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
Overall this looks really clean, thank you for the specific commits with detailed messages. After poring over everything, I had one request for an extra comment, one question about workflow triggers and one request for a follow up after this gets merged.

Since these are all minor, I'll pre-approve and let you address them as you see fit.

PR #3054, opened 2026-01-30
Add rate limiter & metrics to the POSIXv2 implementation

bbockelm:


metrics/http_metrics.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// Server type constants for labeling
+const (
+	ServerTypeOrigin = "origin"
+	ServerTypeCache  = "cache"
+)
+
+// HTTP method tracking
+const (
```
patrickbrophy: Some of these methods are defined already in [`net/http`]( https://pkg.go.dev/net/http#pkg-constants)

metrics/http_metrics.go
```diff
@@ -0,0 +1,98 @@
...
+		Name: "pelican_http_large_transfer_bytes_total",
+		Help: "Total bytes in large HTTP transfers (>100MB)",
+	}, []string{"server_type", "direction", "method"})
+)
+
+// Server type constants for labeling
+const (
+	ServerTypeOrigin = "origin"
+	ServerTypeCache  = "cache"
+)
```
patrickbrophy: Could you use `server_structs.OriginType.String()` and `server_structs.OriginType.String()` instead of redefining these strings?

PR #3067, opened 2026-02-03
Don't reference nonexistent CLI flag in help text

jhiemstrawisc:
The `--scope` flag existed with the previous CLI tool of `pelican origin token create`, but it does not exist for `pelican token create`, at least not currently.

Remove its reference in the help text.


jhiemstrawisc:
This change only affects help text in the CLI, so I'm not worried about the currently-failing test (which is known to be flaky).

Issue #3050, opened 2026-01-28
The web UI should never use a user's username as an identifier


brianaydemir:
As of v7.23.0-rc.2…

## Summary

Suppose that we've added the following user to an origin's internal user database:

<img width="583" height="219" alt="Image" src="https://github.com/user-attachments/assets/242356dd-a414-4ad4-8c0b-0aae271676bd" />

Suppose also that we've configured this origin with `Issuer.OIDCAuthenticationUserClaim: sub`, which is the default value for this parameter. Accordingly, anywhere we've had occasion to refer to this user — `Issuer.AuthorizationTemplates`, `Server.UIAdminUsers`, etc. — we've used their `sub`, i.e., the string `http://cilogon.org/serverA/users/279521`.

The origin proceeds to treat this user in a most unkind way: When they log into the origin — which they might do if they're an admin or attempting to use the built-in issuer to obtain an access token for a namespace — the `login` cookie that gets set in their browser records the following claims:

```
  "sub": "Brian Aydemir",
  "oidc_iss": "https://cilogon.org",
  "oidc_sub": "http://cilogon.org/serverA/users/279521",
```

Unfortunately for this poor user, it's `"sub": "Brian Aydemir"` that rules the day, so they might as well not exist in any place that referenced `http://cilogon.org/serverA/users/279521`.

Furthermore, we can now get the origin to save a baffling configuration: The origin is kind enough to let the admin refer to human-readable usernames when configuring authorization templates…

<img width="355" height="353" alt="Image" src="https://github.com/user-attachments/assets/eb215bec-ed79-4a19-8b09-dec7249db3a0" />

…but the actual YAML-based configuration that gets saved to disk uses that human-readable username instead of the user's `sub`, which is what the admin would expect when that YAML-based configuration specifies `Issuer.OIDCAuthenticationUserClaim: sub`:

```
[root@origin-0 pelican]# cat /etc/pelican/web-config.yaml 
issuer:
    authorizationtemplates:
        - actions:
            - read
            - create
            - modify
          prefix: /data
          users:
            - Brian Aydemir
```

## Details

Baring a tragic misunderstanding of the model for Pelican's internal user database (#3043, #3044), a user's (user)name should never appear anywhere we need a stable identifier.

Furthermore, so long as a user's `sub` functions as their unique identifier (assuming that Pelican can only ever be configured with one OIDC issuer, as is currently the case)…

```
sqlite> .schema users
CREATE TABLE users (
[...]

Issue #3049, opened 2026-01-28
In the web configuration UI, only a group's creator can edit the group


brianaydemir:
As of v7.23.0-rc.2…

## Summary

Exactly what the title says. To reproduce:

1. As admin Alice, log into an origin's web UI, and use it to create a new group, with users and all.
2. Log out.
3. As admin Bob, log into the same origin's web UI, and attempt to make any edit to the group Alice created.
4. Observe that no edits succeed.

## Other remarks

A member of `Server.UIAdminUsers` or the built-in `admin` user should be all powerful, whether or not we inhabit a world with fine-grained permissions (#1694).

PR #3076, opened 2026-02-04
Groups fixes

### LINKED ISSUE #3044: Users that are automatically added to the origin's internal user database are given non-sensical (user)names
## Summary

As of Pelican v7.23.0-rc.2, when a user logs into an origin and a record is automatically created in the origin's internal user database, their (user)name is simply a copy of their OIDC `sub` — which is generally not readable or meaningful to a human.

Image

## Desired behavior

While there's no guarantee that the OIDC provider will provide a claim that could be used as a readable name for the user, there are a [slew of standard claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) that the origin could check for and try to use before falling back on `sub`.

Some of those claims could also be used to automatically populate additional fields (#3043).

patrickbrophy:
Closes #3044, #3049, #3050. This PR does the following:

1. Refactor Admin authz:
- Introduce UserIdentity struct to encapsulate user idenitifiers
- Updated CheckAdmin to verify privileges against all stable
  idenitifies. This ensures admins configured via OIDC subject retain
  access even if their display name changes

2. Fix Group Management Permissions:
- Updated AddGroupMember and RemoveGroupMember in the database layer to
  accept an isAdmin flag.
- Used isAdmin flag as an override

3. Improve OIDC username generation
- When generating a username, look up human-readable OIDC claims (name,
  email, preferred_username). We fall back to sub if necessary. This
  prevents users from being created with opaque, unfriendly identifiers
  by default

### DIFF
--- a/web_ui/authentication_test.go
+++ b/web_ui/authentication_test.go
@@ -272,6 +272,9 @@ func TestPasswordResetAPI(t *testing.T) {
 		loginCookieTokenCfg.AddAudiences(param.Server_ExternalWebUrl.GetString())
 		loginCookieTokenCfg.Subject = "user" // general user
 		loginCookieTokenCfg.AddScopes(token_scopes.WebUi_Access, token_scopes.Monitoring_Query, token_scopes.Monitoring_Scrape)
+		loginCookieTokenCfg.Claims = map[string]string{
+			"user_id": "user",
+		}
 
 		// CreateToken also handles validation for us
 		tok, err := loginCookieTokenCfg.CreateToken()
--- a/web_ui/ui_test.go
+++ b/web_ui/ui_test.go
@@ -428,6 +428,9 @@ func TestServerHostRestart(t *testing.T) {
 		tk.Lifetime = 5 * time.Minute
 		tk.AddAudiences("https://example.com")
 		tk.AddScopes(token_scopes.WebUi_Access)
+		tk.Claims = map[string]string{
+			"user_id": "regular-user",
+		}
 		tok, err := tk.CreateToken()
 		require.NoError(t, err)
 


web_ui/groups.go
```diff
@@ -184,15 +184,24 @@ func handleAddGroupMember(ctx *gin.Context) {
 		return
 	}
 
-	_, userId, _, err := GetUserGroups(ctx)
-	if err != nil || userId == "" {
+	user, userId, groups, err := GetUserGroups(ctx)
+	if err != nil {
```
h2zh: I think the `userId == ""` guard should be retained so the subsequent DB operations could have this "addedBy" attribute. Maybe if you have other reason to remove it?

h2zh (changes requested):
Thanks for bringing these enhancements in, though bundling everything in one commit makes the review really time-consuming. Is it possible to break it into 3-4 commits (one for each ticket)?

I'm sure there's a better way to solve #3050. The issue says "a user's name should never appear anywhere we need a stable identifier." This PR adds stable identifiers (internal ID, OIDC sub) as supplementary fields that travel alongside the username, and uses them in a few specific places (admin checks, authz template matching). But the username remains the primary identity that flows through GetUserGroups -> ctx.Set("User") -> every handler -> every database operation. 

I don't have a perfect plan either - I'm thinking of changing the JWT sub to be the internal ID or OIDC sub, and treating username as purely display. But it would require tons of work. Maybe it would be worth to take some time to ponder and talk to Brian_s_ to hear they say.

web_ui/oauth2_client.go
```diff
@@ -250,30 +250,51 @@ func generateUserGroupInfo(userInfo map[string]interface{}, idToken map[string]i
...
-	if !ok {
-		log.Errorln("User info endpoint did not return a value for the user claim", userClaim)
-		err = errors.New("identity provider did not return an identity for logged-in user")
-		return
+
+	var displayName string
+	// If the configured claim is "sub" (default), try to find a more human-readable username from standard claims
+	// This addresses Issue #3044 where users get non-sensical usernames like "http://cilogon.org/..."
+	if userClaim == "sub" {
+		usernameCandidates := []string{"preferred_username", "name", "nickname", "email"}
```
jhiemstrawisc: Would be useful to link to https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims as a way to provide breadcrumbs for future code readers as to where these values were picked from.

jhiemstrawisc (changes requested):
I'm giving this a very light review, mostly for style, because I'm not at all familiar with this section of code and Howard already gave a precursory review. As much as I'd love to dive into it, the `v7.24` crunch is preventing me from getting as in-depth as I'd like.

jhiemstrawisc (approved):
My minor change request was taken care of, so I'm okay with this getting merged once everyone else is ready.

PR #3077, opened 2026-02-05
Addition of a new SSH-based backend for the origin

bbockelm:
Permits the origin to launch a helper over SSH which connects back and allows the origin to serve out the helper's filesystem.

### DIFF
--- a/server_utils/afero_webdav.go
+++ b/server_utils/afero_webdav.go
@@ -1,6 +1,6 @@
 /***************************************************************
  *
- * Copyright (C) 2025, Pelican Project, Morgridge Institute for Research
+ * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License.  You may
--- a/server_utils/osrootfs.go
+++ b/server_utils/osrootfs.go
@@ -1,6 +1,6 @@
 /***************************************************************
  *
- * Copyright (C) 2025, Pelican Project, Morgridge Institute for Research
+ * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License.  You may
--- a/ssh_posixv2/auth_test.go
+++ b/ssh_posixv2/auth_test.go
@@ -2,7 +2,7 @@
 
 /***************************************************************
  *
- * Copyright (C) 2025, Pelican Project, Morgridge Institute for Research
+ * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License.  You may


origin/advertise.go
brianaydemir: This logic is inconsistent with `launchers/origin_serve.go`.

Here, we always advertise the data URL as `/api/v1.0/origin/data`. In `launchers`, we register handlers for that route only if the director is also enabled for the server.

If the federation's director and the origin are running as separate services, while a naïve `curl` to the origin works as expected

```
[root@dev app]# curl https://origin-0:8444/public/data/0.0
0.0.28177
```

the director sends the client to a non-working endpoint:

```
[root@dev app]# ./pelican object get --debug --direct pelican://director:8444/public/data/0.0 asdf
...
DEBUG[2026-02-16T21:13:53Z] Trying the object servers: [https://origin-0:8444/api/v1.0/origin/data/public/data/0.0] 
...
DEBUG[2026-02-16T21:13:53Z] Failed to download from https://origin-0:8444/api/v1.0/origin/data/public/data/0.0 : request failed (HTTP status 404): 404 page not found: Specification.FileNotFound Error: Error code 5011: server returned 404 Not Found  job=019c684d-782f-764b-a861-9c2bee9ec718 url="https://origin-0:8444/api/v1.0/origin/data/public/data/0.0"
```

launchers/origin_serve.go
brianaydemir: While we configure OA4MP irrespective of the storage backend, we only actually launch OA4MP if the storage backend makes use of XRootD.

This feems unnecessarily restrictive in the presence of the `posixv2` and `ssh` backends.

cmd/origin_ssh_auth.go
```diff
@@ -0,0 +1,178 @@
...
+	// Status command uses same origin flag
+	sshAuthStatusCmd.Flags().StringVar(&sshAuthOrigin, "origin", "", "Origin URL to check (auto-detected if not specified)")
+
+	// Add sub-commands
+	sshAuthCmd.AddCommand(sshAuthLoginCmd)
+	sshAuthCmd.AddCommand(sshAuthStatusCmd)
+}
+
+// getOriginURL returns the origin URL from the flag, address file, or config
+func getOriginURL() (string, error) {
```
brianaydemir: Both `runSSHAuthLogin` and `runSSHAuthStatus` suffer from a problem where they rely on this function to return the origin's URL, the problem being that there's no guarantee that Viper has been configured, which these functions rely on (indirectly).

Or in the words of Copilot:

```
The origin ssh-auth login command is calling config.ReadAddressFile() which uses getServerRuntimeDir().
  This function reads from viper.GetString(param.RuntimeDir.GetName()), but viper is not initialized when running the CLI
  command.
  The problem is:
   1. The ssh-auth login command runs as a standalone CLI command
   2. It calls config.ReadAddressFile() at line 126 of /Users/baydemir/Ivalice/GitHub/pelican/cmd/origin_ssh_auth.go
   3. ReadAddressFile() uses getServerRuntimeDir() which relies on viper configuration
   4. But the CLI command hasn't initialized the configuration, so RuntimeDir is empty
   5. This causes ReadAddressFile() to fail with "runtime directory is not configured"
  The fix: The CLI command needs to initialize the configuration before trying to read the address file. You need to call 
  config.InitClient() or similar configuration initialization in the runSSHAuthLogin and runSSHAuthStatus functions before
  calling getOriginURL().
```

Emperically, Copilot is not entirely wrong. Where I disagree with it: I think `InitServer` is more appropriate.

ssh_posixv2/helper.go
```diff
@@ -0,0 +1,620 @@
...
+					time.Since(lastPong), timeout)
+				return errors.New("helper keepalive timeout")
+			}
+		}
+	}
+}
+
+// StopHelper stops the remote helper process.
+// It first tries a clean shutdown via stdin message, then falls back to signals.
+func (c *SSHConnection) StopHelper(ctx context.Context) error {
```
patrickbrophy: The bug described in `ssh_posixv2/backend.go` triggers a race condition in `StopHelper`.
1.  When `runConnection` passes `ctx` (expired session contex) to `StopHelper`, so the `cleanShutdownCtx` derived from it is immediately expired. The 3-second grace period for clean shutdown never actually happens.
2. After the SIGKILL path, StopHelper sets `c.helperIO = nil` without waiting for the errgroup goroutines to finish. The `readHelperStdout` goroutine, still running between its ctx.Done() check and the `c.helperIO` dereference, hits a nil pointer.

patrickbrophy (changes requested):
While running the docker compose setup, I noticed after a little while that the Origin panicked with the following:
```
origin-ssh-1  | panic: runtime error: invalid memory address or nil pointer dereference
origin-ssh-1  | [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x123926c]
origin-ssh-1  |
origin-ssh-1  | goroutine 2975 [running]:
origin-ssh-1  | github.com/pelicanplatform/pelican/ssh_posixv2.(*SSHConnection).readHelperStdout(0x4000fecf70, {0x2996660, 0x4001332d70})
origin-ssh-1  |         /pelican-build/ssh_posixv2/helper.go:231 +0xcc
origin-ssh-1  | github.com/pelicanplatform/pelican/ssh_posixv2.(*SSHConnection).StartHelper.func2()
origin-ssh-1  |         /pelican-build/ssh_posixv2/helper.go:185 +0x24
origin-ssh-1  | golang.org/x/sync/errgroup.(*Group).Go.func1()
origin-ssh-1  |         /root/go/pkg/mod/golang.org/x/sync@v0.18.0/errgroup/errgroup.go:93 +0x4c
origin-ssh-1  | created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 369
origin-ssh-1  |         /root/go/pkg/mod/golang.org/x/sync@v0.18.0/errgroup/errgroup.go:78 +0x90
```

The linked comments describe how this likely happened.

PR #3085, opened 2026-02-07
Add support for setup of rclone

bbockelm:
This commit adds new sub-commands for installing and configuring rclone to work with a Pelican data federation.

Particularly, rclone has the ability to invoke a command to renew the token *if* a 401 is sent; this is done and the director now returns a 401 on expiration.

### DIFF
--- a/docs/app/commands-reference/_meta.js
+++ b/docs/app/commands-reference/_meta.js
@@ -13,6 +13,7 @@ export default {
     "object": "pelican object",
     "origin": "pelican origin",
     "plugin": "pelican plugin",
+    "rclone": "pelican rclone",
     "registry": "pelican registry",
     "server": "pelican server",
     "token": "pelican token",
--- a/docs/app/commands-reference/page.mdx
+++ b/docs/app/commands-reference/page.mdx
@@ -40,6 +40,7 @@ across multiple dataset providers.
 * [pelican object](/commands-reference/object/)	 - Interact with objects in the federation
 * [pelican origin](/commands-reference/origin/)	 - Operate a Pelican origin service
 * [pelican plugin](/commands-reference/plugin/)	 - Plugin management for HTCSS
+* [pelican rclone](/commands-reference/rclone/)	 - Commands for integrating Pelican with rclone
 * [pelican registry](/commands-reference/registry/)	 - Interact with a Pelican registry service
 * [pelican server](/commands-reference/server/)	 - Manage server operations
 * [pelican token](/commands-reference/token/)	 - Interact with tokens used to interact with objects in Pelican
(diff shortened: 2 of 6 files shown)


CannonLock:
@bbockelm Sounds good, I will tack on a PR tomorrow to sort the config piece.

bbockelm:
One thing to note - this only affects a "hidden" knob currently.  I don't have a feel for whether that impacts the web UI code.

cmd/rclone_install.go
```diff
@@ -0,0 +1,280 @@
...
+	"runtime"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"golang.org/x/term"
+)
+
+var (
+	rcloneInstallCmd = &cobra.Command{
```
turetske: Do we really want to add in a download and install `rclone` command? Shouldn't attempting to rclone without the installation cause and error and the user should install it on its own? This seems like it's outside the bounds of what pelican should do/be responsible for.

cmd/rclone_setup.go
```diff
@@ -0,0 +1,468 @@
+/***************************************************************
```
turetske: Mark this as don't build on windows

director/director.go
```diff
@@ -719,6 +720,64 @@ func generateRedirectResponse(ctx *gin.Context, chosenAds []server_structs.Serve
...
+// validateClientToken inspects the client's bearer token, if one is present, and
+// returns 401 Unauthorized if the token is expired. This enables clients like rclone
+// to re-run their bearer_token_command to obtain a fresh token.
+//
+// If no token is present, the request is allowed through — the Director does not
+// require tokens. The origin or cache will enforce authorization as needed.
+//
+// Returns (http.StatusOK, nil) if no token is present or the token is not expired.
+// Returns (http.StatusUnauthorized, err) if a token is present but expired.
+func validateClientToken(ctx *gin.Context, nsAd server_structs.NamespaceAdV2, requestId uuid.UUID) (int, error) {
```
turetske: Why is this including nsAd? It seems unused.

director/director.go
```diff
@@ -719,6 +720,64 @@ func generateRedirectResponse(ctx *gin.Context, chosenAds []server_structs.Serve
...
+	// period because the token could expire between the redirect and the
+	// client reaching the origin/cache. Use 10 seconds, or half the token
+	// lifetime if the lifetime is shorter than 10 seconds, to also account
+	// for clock skew.
+	exp := parsed.Expiration()
+	if !exp.IsZero() {
+		iat := parsed.IssuedAt()
+		grace := 10 * time.Second
+		if !iat.IsZero() {
+			lifetime := exp.Sub(iat)
```
turetske: Is it ever possible for exp < iat? I don't think so, but if it could happen then lifetime would be < 0, and grace would then be -, which would incorrectly cause the token to pass.

PR #3092, opened 2026-02-09
Origin and Caches now check for a minimum xrootd version before starting

### LINKED ISSUE #2991: Require minimum XRootD version in the binary
Pelican Service:
- [ ] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [x] Origin
- [x] Cache

The cache and origin require a minimum version of XRootD to work properly (or even start up for that matter). Currently, this is 5.8.2 (primarily for the purge plugin and critical bug fixes). People that do not use the packages do not know about the version requirement, leading to time spent debugging crashes.

Have the XRootD-requiring components of Pelican check the version of XRootD -- this can be done by running `xrootd -v`, which will print `v5.9.1` or something following that pattern _to stderr_. If the xrootd version is too old, exit with a clear error message. This requires keeping `.goreleaser.yml` and the version check in sync, so each spot should have a comment cross-referencing the other.

turetske:
Adds a "CheckXrootdVerion()" function which runs `xrootd -v` to get the version and then parses and normalizes the version string.

Compares against a MinXrootdVersion constant.

Returns an error if the version is insufficient or the binary isn't found.

Cases tested outside the unit tests:
xrootd not found (stops and outputs an error saying it can't find an xrootd version)
xrootd below minimum version (stops and outputs an errors comparing the versions and informing the user to update)
xrootd exactly the minimum version (works properly)
xrootd above the minimum version (works properly)

This will close #2991

### DIFF
--- a/xrootd/version.go
+++ b/xrootd/version.go
@@ -86,7 +86,7 @@ func CheckXrootdVersion() error {
 	if xrootdVersion.LessThan(requiredVersion) {
 		return errors.Errorf("XRootD version %s is insufficient (minimum required: %s). "+
 			"Please upgrade XRootD to version %s or later. "+
-			"This requirement is necessary for proper operation of Pelican cache and origin servers.",
+			"This requirement is necessary for proper operation of Pelican Cache and Origin servers.",
 			xrootdVersion.String(), MinXrootdVersion, MinXrootdVersion)
 	}
 


jhiemstrawisc (approved):
Simple enough, and cleanly implemented. One small nit, and disabling the new tests on Windows. I'll approve ahead of time so you can merge once they're done without re-requesting a review.

PR #3103, opened 2026-02-11
Put POSC behind a feature flag

patrickbrophy:
This PR puts POSC behind a feature flag `Origin.EnableAtomicUploads`. This is needed due to users that may not have `xrootd-s3-http` `v0.6.4` on their systems.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -1015,8 +1015,23 @@ components: ["origin"]
 ---
 name: Origin.EnableAtomicUploads
 description: |+
-  A boolean indicating whether the origin should enable atomic uploads. When true, the origin will enable atomic uploads.
+  A boolean that, when true, enables atomic uploads for the Origin. When true, the origin will enable atomic uploads.
   When false, the origin will not enable atomic uploads. Atomic uploads are only available for origins that have a StorageType of "posix".
+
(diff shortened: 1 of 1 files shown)


docs/parameters.yaml
```diff
@@ -1013,8 +1013,19 @@ root_default: /run/pelican/xrootd/origin
 default: $XDG_RUNTIME_DIR/pelican/origin
 components: ["origin"]
 ---
+name: Origin.EnableAtomicUploads
+description: |+
+  A boolean indicating whether the origin should enable atomic uploads. When true, the origin will enable atomic uploads.
```
jhiemstrawisc: Two notes here:
- Maybe say "A boolean that, when true, enables atomic uploads for the Origin"
- Since you're now making the POSC integration a toggle, it feels more appropriate to explain what atomic writes are and what problems they solve at this toggle than in the temp directory knob. Can you move the description here?

docs/parameters.yaml
```diff
@@ -1013,8 +1013,19 @@ root_default: /run/pelican/xrootd/origin
...
+  When false, the origin will not enable atomic uploads. Atomic uploads are only available for origins that have a StorageType of "posix".
```
jhiemstrawisc: You should also explain the conditions that allow admins to flip this switch, i.e. "enabling this requires xrootd-s3-http version ...."

PR #3111, opened 2026-02-11
Integrate N2N plugin to map FederationPrefix to StoragePrefix for XRoot storage backends

### LINKED ISSUE #3110: Allow for prefix name translation on Origin's with storage backed by XrootD
Currently when using Origin with an XRootD storage backend, the `StoragePrefix` must match the `FederationPrefix`. This is due to how we proxy the request from Pelican to the remote XRootD storage. However with the introduction of the  [N2N Prefix plugin](https://github.com/PelicanPlatform/xrootd-s3-http?tab=readme-ov-file#configure-the-n2n-prefix-plugin) we can map one prefix to another. 

So to make this all work, we can set up a rule that maps the `FederationPrefix` to an arbitrary `StoragePrefix` and remove the previously mentioned constraint. In practice this would mean that an Origin admin could have a namespace `/foo` but an XRootD storage backend that exports `/bar`.

patrickbrophy:
Closes #3110

With this PR it is now possible to map a FederationPrefix to an arbitrary StoragePrefix for XRootD storage backends. This enables an origin to export `/foo` and map it to `/bar` on the remote XRootD storage.

### DIFF
--- a/server_utils/origin_xroot.go
+++ b/server_utils/origin_xroot.go
@@ -42,15 +42,12 @@ func (o *XRootOrigin) validateStoragePrefix(prefix string) error {
 }
 
 func (o *XRootOrigin) validateExtra(e *OriginExport, numExports int) (err error) {
-	// FederationPrefix is the path clients use; StoragePrefix is the path on the backend XRootD.
-	// When they differ, the xrootd-origin config uses the Name2Name (N2N) plugin to map
-	// federation path -> backend path (prefixn2n.rule). Only xroot storage type supports this.
 	xRootServiceUrl := param.Origin_XRootServiceUrl.GetString()
 	if xRootServiceUrl == "" {
-		return errors.New("Origin.XRootServiceUrl is required for XRoot origins")
+		return errors.Errorf("%s is required for XRoot origins", param.Origin_XRootServiceUrl.GetName())
 	}
 	if _, err = url.Parse(xRootServiceUrl); err != nil {
-		return errors.Wrapf(err, "unable to parse Origin.XRootServiceUrl '%s'", xRootServiceUrl)
+		return errors.Wrapf(err, "unable to parse %s value '%s'", param.Origin_XRootServiceUrl.GetName(), xRootServiceUrl)
 	}
 	return
 }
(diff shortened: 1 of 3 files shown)


server_utils/origin_xroot.go
```diff
@@ -42,10 +42,9 @@ func (o *XRootOrigin) validateStoragePrefix(prefix string) error {
 }
 
 func (o *XRootOrigin) validateExtra(e *OriginExport, numExports int) (err error) {
-	if e.FederationPrefix != e.StoragePrefix {
-		return errors.Errorf("FederationPrefix and StoragePrefix must be the same for XRoot origins, but you configured '%s' and '%s'", e.FederationPrefix, e.StoragePrefix)
-	}
-
+	// FederationPrefix is the path clients use; StoragePrefix is the path on the backend XRootD.
+	// When they differ, the xrootd-origin config uses the Name2Name (N2N) plugin to map
```
jhiemstrawisc: Small technical nit on this comment -- the code uses the N2N plugin whether or not the two paths are different. The comment makes it seem like we only invoke the plugin when the two paths differ.

On a bigger note, I think this comment belongs near the XRootD config you changed.

jhiemstrawisc (changes requested):
Small nit, otherwise I tested by spinning up a local Origin with this config:
```
Origin:
  StorageType: "xroot"
  XRootServiceUrl: "eospublic.cern.ch:1094"
  Exports:
    - FederationPrefix: "/my-super-cool-cern-data"
      StoragePrefix: "/eos/opendata/cms"
      Capabilities: ["PublicReads", "Writes", "Listings"]
```

I was able to perform a successful object listing against it.

Can you also add one last bit of documentation that's missing? In `docs/parameters.yaml` under the `Origin.Exports` description, we list out a few extra knobs relevant to the `s3` and `globus` backends, but we don't list `Origin.XrootServiceUrl` for the `xroot` backend (if you look at the params table, I think you'll see what I'm after). This is probably because previously we didn't anticipate this storage backend getting first class support. It seems like that ship has sailed so we should give a bit more information about how to set it up.

Issue #3109, opened 2026-02-11
Federation config knobs that aren't gettable/settable via `param` should still be named in param


jhiemstrawisc:
Even if we don't want people getting/setting these values through the param package, we should still provide the param names.

[Here's one good reason](https://github.com/PelicanPlatform/pelican/blob/bbc3d38447d0bc4c683435a4e3838013f942ca33/config/config.go#L1588-L1591) wherein a typo of `Federation_DirectorUrl` instead of `Federation.DirectorUrl` makes me question whether a) this line does what it's supposed to and b) if it doesn't, why this doesn't cause noticeable issues.

This is related to my code exploration for #3108.

Issue #3108, opened 2026-02-11
`discoverFederationImpl` in dire need of rework


jhiemstrawisc:
Our ops team was recently able to start and run the ITB `v7.22` Director in a way where the Director, acting as the root of federation metadata discovery, didn't know its own discovery URL:
```
$ pelican-server config dump
<abridged>
Federation:
    BrokerUrl: https://osdf-itb.osg-htc.org/
    DirectorUrl: ""
    DiscoveryUrl: ""
    JwkUrl: ""
    RegistryUrl: https://osdf-itb-registry.osg-htc.org/
    TopologyDowntimeUrl: ""
    TopologyNamespaceUrl: https://topology.opensciencegrid.org/osdf/namespaces?include_downed=1&include_inactive=1
    TopologyReloadInterval: 0s
    TopologyUrl: ""
<abridged>

ERROR[2026-02-11T16:14:18Z] Error getting federation info: could not discover federation services from '': Discovery URL must have a host, but you provided ''
```
That this Director started at all is a bug.

This surprised me because my train of work in #2747 implied Pelican should almost _always_ be performing discovery. Apparently I didn't go far enough, though. The requirement was further eroded by #2897, which skips discovery if no discovery URL is resolved.

Overall, the logic in `discoverFederationImpl()` function has gotten really dense and hard to follow, especially with respect to Directors acting as fallback Discovery endpoints in select cases and the deferred normalization we do. Given its importance, this function should be reworked to be easier to follow, and the function's requirements should be both documented and tested thoroughly.

Here's my first pass at what I think it needs to accomplish in the rework, and I may further flesh this out after more thought:
1. A service must never start without resolving its federation metadata. The bare minimum for all services includes the Discovery, Director and Registry endpoints. Failure to deduce any of these should be a startup error. Guaranteeing that these values are always accessible and normalized in the global fed info object also renders unnecessary the safety checks for empty values and valid URLs, as is done [here](https://github.com/PelicanPlatform/pelican/blob/bbc3d38447d0bc4c683435a4e3838013f942ca33/xrootd/xrootd_config.go#L484-L486) and in a million other places.
2. All values set in the global fed info object should be normalized to include a scheme and to strip certain ports (e.g. 443).
[...]

Issue #3107, opened 2026-02-11
Much needed cleanup of test file transfer utilities


jhiemstrawisc:
While working on #3094, I'm finding it necessary to modify some of [this code](https://github.com/PelicanPlatform/pelican/blob/a890acea7ee9169ffdfa7de5c9532ec5a8d40ddd/server_utils/test_file_transfer.go#L77) to generate tokens for Director/Origin/Cache self tests.

However, I've been staring at some of this for awhile and I noticed this is connected in principle to [this interface](https://github.com/PelicanPlatform/pelican/blob/a890acea7ee9169ffdfa7de5c9532ec5a8d40ddd/server_utils/test_file_transfer.go#L42)... which is defined but used nowhere in the project.

To confuse code readers even more, these are completely isolated from [this _separate_ function](https://github.com/PelicanPlatform/pelican/blob/a890acea7ee9169ffdfa7de5c9532ec5a8d40ddd/xrootd/self_monitor.go#L283) with an identical name but different signature.

While my fix for #3094 will be minimal, the file transfer utilities _really_ need to be cleaned up. I'm not convinced any of this really needs to be an interface, especially at the expense of code readability/maintainability.

jhiemstrawisc:
FYI, I'm grabbing this as an outside contributor soon :)

PR #3112, opened 2026-02-11
Prefer Normalized Discovery URL via `config.GetFederation()` over unnormalized version via `param`

### LINKED ISSUE #3094: Setting schemeless `Federation.DiscoveryUrl` causes token verification issues
Reported by a user and reproduced by me. The following config with schemeless Discovery URL:
```
Federation:
  DiscoveryUrl: hostname.com:4443
```

renders as having a scheme with `pelican config dump`:
```
Federation:
  DiscoveryUrl: https://hostname.com:4443
```

but causes token validation for services started with it, e.g. from an Origin:
```
DEBUG[2026-02-10T14:23:34Z] Cannot verify token: cannot verify token with federation issuer: Token issuer https://hostname:4443 does not match the issuer from the federation. Expecting the issuer to be hostname.com:4443 
```

I was wondering whether it's acceptable to strip the protocol when checking JWTs, since it's ultimately the hostname that's protected by a certificate. Here's what the [JWT RFC](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.1) says of the `iss` claim:
```
   The "iss" (issuer) claim identifies the principal that issued the
   JWT.  The processing of this claim is generally application specific.
   The "iss" value is a case-sensitive string containing a StringOrURI
   value.  Use of this claim is OPTIONAL.
```

Since it makes the point that `iss` is case sensitive and can be either a string or a URI, that suggests to me it's not kosher to strip schemes. The fix for this should live in normalizing Discovery/Director URLs on startup such that functions retrieving these values are given the full URI.

jhiemstrawisc:
The core of the bug I was working on here came from the fact that a token
producer and a token consumer might both be configured with the same
Discovery host, but might be constructing an issuer URL from that host
in different ways when one gets a normalized version (containing an 'https'
scheme) and the other gets only the hostname. This was affecting Director
tests for a few OSDF services and is probably and old bug.

This fixes the acute bug and also looks through the rest of the repo
to find other places where this was happening.

In particular, this is only a bandaid, and these issues describe
longer-term cleanup related to the bug:
- https://github.com/PelicanPlatform/pelican/issues/3107
- https://github.com/PelicanPlatform/pelican/issues/3108
- https://github.com/PelicanPlatform/pelican/issues/3109

Because this is a bandaid, I restricted my cleanup to actual source
code files and did not chase anything into test files unless I had to 
in order to fix a test.

I also cleaned up a few other small bugs that caught my eye while
working on the main fix.

Closes #3094

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -283,7 +283,7 @@ func TestInitConfig(t *testing.T) {
 	// Check if server address is correct by defaults.yaml
 	assert.Equal(t, "0.0.0.0", param.Server_WebHost.GetString())
 	// Check that Federation Discovery url is correct by osdf.yaml
-	assert.Equal(t, "osg-htc.org", param.Federation_DiscoveryUrl.GetString())
+	assert.Equal(t, "https://osg-htc.org", param.Federation_DiscoveryUrl.GetString())
 
 	require.NoError(t, param.Set("Server.WebHost", "1.1.1.1")) // should write to temp config file
 	if err := viper.WriteConfigAs(tempCfgFile.Name()); err != nil {
--- a/origin/disk_usage_test.go
+++ b/origin/disk_usage_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/server_structs"
 	"github.com/pelicanplatform/pelican/server_utils"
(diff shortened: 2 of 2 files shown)


jhiemstrawisc:
Also pinging @turetske, since we'll want to patch this into `v7.23` once it's approved/merged.

PR #3122, opened 2026-02-14
Suppress CodeQL false positives for filepath fields

brianaydemir:
Add a custom replacement for the built-in go/clear-text-logging query that allowlists S3SecretKeyfile, UIPasswordFile, and PasswordLocation. These struct fields store file paths, not secrets, so logging them is safe.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,14 @@
+name: "Pelican CodeQL Configuration"
+
+# Run the default queries plus our custom replacements.
+queries:
+  - uses: ./.github/codeql/go
+
+query-filters:
+  # Exclude the built-in go/clear-text-logging query. Our replacement in
+  # .github/codeql/go/CleartextLogging.ql keeps the same @id and logic but
+  # adds an allowlist for struct fields that hold file *paths* rather than
+  # secrets (S3SecretKeyfile, UIPasswordFile, PasswordLocation).
+  - exclude:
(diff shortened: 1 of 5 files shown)


brianaydemir:
According to Copilot:

```
● Here's a summary of the changes:
  Files created:
   1. .github/codeql/codeql-config.yml — CodeQL configuration that loads our custom Go queries and excludes the built-in 
  go/clear-text-logging query (matched by query path: /^Security\//) so only our custom version runs.
   2. .github/codeql/go/qlpack.yml — CodeQL query pack definition declaring a dependency on codeql/go-all.
   3. .github/codeql/go/CleartextLogging.ql — Custom replacement for the built-in go/clear-text-logging query. It adds an 
  isSafeFilepathField predicate that allowlists S3SecretKeyfile, UIPasswordFile, and PasswordLocation — struct fields that 
  store file paths rather than secrets. The query otherwise preserves all the original logic.
  Files modified:
   1. .github/workflows/codeql-analysis.yml — Added config-file: ./.github/codeql/codeql-config.yml to the CodeQL init step.
  This eliminates the ~25+ false-positive go/clear-text-logging alerts (like #333) while keeping the query active for any
  genuinely sensitive data.
```

brianaydemir:
For posterity, my "conversation" with Copilot: [GitHub Copilot CodeQL fixes.txt](https://github.com/user-attachments/files/25318336/GitHub.Copilot.CodeQL.fixes.txt).

PR #3144, opened 2026-02-18
Require Server.EnableUI = true for Globus backend

### LINKED ISSUE #3139: Globus Origins _require_ functioning Web UI but don't check for it
While reading through #3125, it occurred to me that Globus Origins truly require that the web UI is enabled, but I couldn't find anywhere in the startup code where we check for [`Server.EnableUI`](https://docs.pelicanplatform.org/parameters#Server-EnableUI) for this backend.

This should be added as a validation step. At first I thought it should go somewhere in https://github.com/PelicanPlatform/pelican/blob/main/server_utils/origin_globus.go, but this validation isn't really for any export itself. Rather, I'm inclined to think it should happen in the config initialization in `config/config.go`, but I'm open to being told otherwise.

whwjiang:
A Globus origin requires a functioning web UI to perform OAuth flow. However, this wasn't checked previously. This PR adds a validation check. This fixes #3139.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1746,6 +1746,9 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
 				return errors.Wrap(err, "unable to parse Origin.HTTPServiceUrl as a URL")
 			}
 		case "globus":
+			if !param.Server_EnableUI.GetBool() {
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (changes requested):
Can you drop the commits that pertain to documentation? I'd like to merge the config validation fix and the docs independently.

Issue #3157, opened 2026-02-25
`pelican server set-logging-level` messages don't warn when state won't change or what bumped levels will revert to


jhiemstrawisc:
This was reported by @williamnswanson, who ran `pelican server set-logging-level debug 10m --param Logging.Cache.PssSetOpt` in a `v7.23.1-rc.0` Cache container.

The CLI reported:
```
Log level for Logging.Cache.PssSetOpt successfully changed to 'debug' for 600 seconds
Change ID: 2f67688a-e10f-4d28-9b0e-a3920bd5fc58
Will revert at: 2026-02-25T16:23:55Z
```
but William did not observe XRootD rebooting.

It turns out the issue was that `Logging.Cache.PssSetOpt` was already at `debug`.

Two things I dislike about this:
1. The CLI should warn when you send a command that has no effect -- it should have told William that this log level was already set to debug
2. The CLI should tell you what level will be used once the reversion takes place

PR #3160, opened 2026-02-25
xrootd: fix new daemon being terminated immediately after restart on log-level change

### LINKED ISSUE #3156: `pelican server set-logging-level` sometimes fails to restart xrootd
While debugging an issue with OSDF caches, I tried out the new hotness that is `pelican server set-logging-level` to bump a few XRootD logging knobs.

From a Cache running `v7.23.0`, I ran `pelican-server server set-logging-level debug 60m --param Logging.Cache.PssSetOpt`. Because the param toggles log levels in XRootD, the Cache must restart XRootD. When I ran the command, I got the expected output from the CLI indicating it had succeeded in changing the param value, and that the value would revert in the future.

However, between the OSDF Director and the Cache's logs, I could see something wasn't right; director tests were failing, and the logs started filling up with self test errors saying "connection refused". @brianaydemir found the culprit, which was that Pelican attempted to restart XRootD but encountered an error:
```
time="2026-02-25T14:32:51Z" level=info msg="Temporary log level change requested" change_id=ef787c2d-c428-4385-9400-aa60889d15d9 duration=1h0m0s fields.level=trace user=admin
time="2026-02-25T14:32:51Z" level=info msg="Beginning XRootD restart sequence"
time="2026-02-25T14:32:51Z" level=debug msg="Sending SIGTERM to existing XRootD processes"
time="2026-02-25T14:32:51Z" level=info msg="Served Request" client=128.105.82.98 daemon=gin fields.time=7.24492ms method=POST resource=/api/v1.0/logging/level status=200
time="2026-02-25T14:32:51Z" level=info msg="Daemon \"xrootd.cache\" exited during expected restart: signal: terminated"
time="2026-02-25T14:32:51Z" level=debug msg="Reconfiguring XRootD runtime directory"
...
time="2026-02-25T14:32:52Z" level=info msg="------ xrootd cache@uw-osdf-pelican-cache-544c7c7f98-sh9gr:8443 initialization completed." daemon=xrootd.cache
time="2026-02-25T14:32:52Z" level=info msg="Cache startup complete on port 8443"
time="2026-02-25T14:32:52Z" level=info msg="XRootD restart complete with new PIDs: [58172]"
time="2026-02-25T14:32:52Z" level=error msg="Failed to restart XRootD after logging configuration change" error="XRootD restart already in progress"
time="2026-02-25T14:32:52Z" level=info msg="Daemon \"xrootd.cache\" with pid 58172 was killed"
```
The fact that the server didn't shout when it failed to xrootd surprises me -- I'd have expected something around [this section of code](https://github.com/PelicanPlatform/pelican/blob/365f04393a5b0153f397b587bf08948a26e1a29f/daemon/launch_unix.go#L285-L287) would have caught the issue, as it's responsible for crashing Pelican if XRootD crashes/fails to start. It's probably a good idea to integrate these bits of code; for xrootd-based Caches/Origins, failure to start xrootd is a critical bug that should probably result in server shutdown.

### LINKED ISSUE #3158: `pelican server set-logging-level` doesn't hook into existing "restart" mechanics that cause graceful server drain
While I was happy to see that `pelican server set-logging-level` does stick the server's health status in the `StatusShuttingDown` state, it'd be nice if the command used some of the same code paths hit by an admin using the `Restart` button in the Web UI (i.e. [`handleGracefulShutdown`](https://github.com/PelicanPlatform/pelican/blob/365f04393a5b0153f397b587bf08948a26e1a29f/launchers/launcher.go#L486).

In particular, the Web UI's `Restart` button two nice things:
- it causes an immediate advertisement to the Director to tell the federation the service is about to disappear, preventing additional redirects
- it waits 1 minute before restarting the components to give ongoing client transfers a chance to complete

As is, a command that needs to reboot XRootD like `pelican server set-logging-level 10m --param Logging.Cache.PssSetOpt` will kill ongoing client requests without any grace two times -- once on the initial restart and once on the revert.

brianaydemir:
Closes PelicanPlatform/pelican#3156.
Closes PelicanPlatform/pelican#3158.

**Not** considered (read as: intentionally ignored): PelicanPlatform/pelican#3157

## Fix for #3156 — new daemon terminated immediately after restart

Running `pelican server set-logging-level` with an XRootD-specific parameter triggers a full XRootD restart. The restart context was created inside `handleXrootdLoggingChange` with a short timeout, so when the function returned and `defer cancel()` fired, Go sent SIGTERM to the newly-started daemon.

**Fix**: the `ctx`, `egrp`, and `callback` fields in `restartInfo` are replaced with a single `launch` closure that captures the server-lifetime context at registration time. `RestartXrootd` calls `info.launch(newLaunchers)` so the new daemon's lifetime is bound to the server context, not a short-lived restart context.

## Fix for #3158 — no graceful drain before restart

Previously `RestartXrootd` sent SIGTERM to old processes immediately, cutting off in-flight client transfers and without advertising the impending downtime to the Director.

**Fix**: `RestartXrootd` now runs a global pre-restart phase once per restart operation. Before sending any signals, it advertises `StatusShuttingDown` and waits for the configured drain period, so restart-triggered logging changes use the same graceful drain behavior as the normal server restart flow.

## Fix for post-restart 404 — server not re-advertised to Director after restart

The `preRestartHook` removes the server from the Director's routing table. Without a corresponding post-restart advertisement the server would be invisible to clients until the next periodic advertisement cycle.

**Fix**: `RestartXrootd` now runs a global post-restart phase after all replacement daemons are running. Launcher code injects the advertisement behavior, and the restart flow re-advertises restarted servers before returning so they become available to the Director immediately.

## Minimum duration enforcement for XRootD logging param changes

With a 60 s drain period, `set-logging-level 30s --param Logging.Origin.Cms` would never take effect: the `LogLevelManager` timer fires at T=30 s and reverts the param while the drain is still running, so XRootD restarts with the old config.

**Fix**: `HandleSetLogLevel` returns HTTP 400 when the requested duration is ≤ `Xrootd_ShutdownTimeout` for any param that requires an XRootD restart. The response includes `requiresRestart` and `effectiveAt` so the caller knows when the change will take effect and for how long; the CLI prints this information when `requiresRestart` is true.


fed_test_utils/fed.go
```diff
@@ -175,6 +175,9 @@ func NewFedTest(t *testing.T, originConfig string) (ft *FedTest) {
 	require.NoError(t, param.Set(param.Cache_EnableEvictionMonitoring.GetName(), false))
 	require.NoError(t, param.Set(param.Cache_StorageLocation.GetName(), filepath.Join(tmpPath, "xcache-data")))
 	require.NoError(t, param.Set(param.Cache_DbLocation.GetName(), filepath.Join(t.TempDir(), "cache.sqlite")))
+	// In tests, skip the drain-wait period before XRootD restarts so tests
+	// don't time out waiting for PIDs to change.
+	require.NoError(t, param.Set(param.Xrootd_ShutdownTimeout.GetName(), 0))
```
brianaydemir: This sounds plausible to me, but will this interfere with tests outside of `launchers` and `xrootd`?

jhiemstrawisc (changes requested):
This diff exposes some unintuitive and maybe incorrect behavior for changes to log levels that require XRootD restarts.

Consider this command/response:
```
$ pelican server set-logging-level info 30s --param Logging.Origin.Cms --server https://`hostname`:8444

Log level for Logging.Origin.Cms successfully changed to 'info' for 30 seconds
Change ID: 70f2125a-ba8b-4b9c-b614-30afb388d287
Will revert at: 2026-03-05T19:14:16Z
```

Because the param being changed requires an XRootD restart, I expected the message received by the client to be incorrect for two reasons:
1. The logging levels for XRootD's CMS component haven't yet been changed because XRootD is still draining for another minute.
2. By the time XRootD _does_ restart, the in-memory param will have reverted so the intended change is never picked up.

I say this is what I expected, but the reality is another subtle bug was exposed -- because the param revert tries to restart XRootD 30s into the drain, an error is thrown about the second restart running into the first restart that's still draining:
```
ERROR[2026-03-05T19:44:16Z] Failed to restart XRootD after logging configuration change  error="XRootD restart already in progress"
```

When XRootD logging params are modified, the message sent back to the client should include information about when the change will take effect, and we should think about what it means to provide time durations shorter than the XRootD drain period.

Finally, we should also think about how we can pass the restart context around without embedding it in a struct.

brianaydemir:
*Comment by GitHub Copilot:*

Addressing @jhiemstrawisc's [review](https://github.com/PelicanPlatform/pelican/pull/3160#pullrequestreview-3897596728):

1. **Timing info**: The response now includes `requiresRestart`, `effectiveAt`, and `effectiveDurationSeconds`. The CLI prints a note about when the change will take effect and for how long.
2. **Short durations**: `HandleSetLogLevel` now returns HTTP 400 when `duration ≤ Xrootd_ShutdownTimeout` for XRootD logging params, with a message explaining the minimum allowed duration.
3. **Context in struct**: `restartInfo.ctx` was removed. A `launch func(launchers []daemon.Launcher) ([]int, error)` closure captures the server-lifetime context at registration time, so `RestartXrootd` never stores a bare `context.Context`.

jhiemstrawisc (changes requested):

fed_test_utils/fed.go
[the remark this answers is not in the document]
brianaydemir: I addressed with this via a rebase on `main`.

brianaydemir:
@jhiemstrawisc This is back in your court.

As my first foray into getting Copilot to edit a real code base through multiple iterations of potential fixes… this PR is definitely not perfect. That said, I think the overall change to code looks okay(-ish).

web_ui/logging_api.go
```diff
@@ -114,6 +117,33 @@ func HandleSetLogLevel(ctx *gin.Context) {
...
+				Msg: fmt.Sprintf(
+					"XRootD logging parameters require a restart with a %s drain period. "+
+						"The requested duration (%s) must be longer than %s.",
+					drainPeriod, duration, drainPeriod),
+			})
+			return
+		}
+	}
+
 	if err := manager.AddChange(changeID, parameterName, level, duration); err != nil {
```
jhiemstrawisc, line 147: I think this duration should be the requested duration in addition to the drain time. For example, with a configured `Xrootd.ShutdownTimeout` of `10s`, I tested by running:
```
$ date && echo "" && pelican server set-logging-level trace 12s --param Logging.Origin.Scitokens --server https://`hostname`:8444
Mon Apr 27 21:25:06 UTC 2026

Change ID: eef89ed1-1b46-48c6-8076-343894193d9a
Note: A restart of XRootD is required.
Change will take effect at: 2026-04-27T21:25:16Z
Change will revert at: 2026-04-27T21:25:28Z
Log level for Logging.Origin.Scitokens successfully changed to 'trace' for 12 seconds (pending restart of XRootD)
```

Watching the Origin logs shows that we only get 2 seconds of the new config before XRootD reboots a second time:
```
INFO[2026-04-27T21:25:06Z] Applied log level change                      new_level=trace parameter=Logging.Origin.Scitokens previous_level=fatal
INFO[2026-04-27T21:25:06Z] Temporary log level change requested          change_id=eef89ed1-1b46-48c6-8076-343894193d9a duration=12s fields.level=trace user=admin

INFO[2026-04-27T21:25:16Z] Shutdown grace period elapsed; proceeding with shutdown and discarding incomplete transfers 
INFO[2026-04-27T21:25:16Z] Daemon "xrootd.origin" exited during expected restart: signal: terminated 

INFO[2026-04-27T21:25:18Z] Applied log level change                      new_level=fatal parameter=Logging.Origin.Scitokens previous_level=trace
INFO[2026-04-27T21:25:18Z] Beginning XRootD restart sequence            
INFO[2026-04-27T21:25:18Z] Waiting 10s for in-flight transfers before shutting down 
```

This also makes me wonder -- for the params that need an xrootd reboot, should we start that second drain period one `Xrootd.ShutdownTimeout` before the param is scheduled to be reverted? If we don't, the param technically doesn't truly revert until one drain period after what's reported back to the client. But maybe it's fair to say the "process" of reverting has started because in theory the server told the Director not to send it new clients.

web_ui/logging_api.go
```diff
@@ -131,6 +161,14 @@ func HandleSetLogLevel(ctx *gin.Context) {
 		Remaining:     req.Duration,
 	}
 
+	if requiresRestart {
```
jhiemstrawisc, line 164: This block can be bypassed using an edge case I concocted where I run the same command twice but with different durations:
```
$ date && echo "" && pelican server set-logging-level trace 30s --param Logging.Origin.Scitokens --server https://`hostname`:8444
Mon Apr 27 21:40:54 UTC 2026

Change ID: 418b0af4-db13-4a2b-8d8e-b49bcc3fbb27
Note: A restart of XRootD is required.
Change will take effect at: 2026-04-27T21:41:04Z
Change will revert at: 2026-04-27T21:41:34Z
Log level for Logging.Origin.Scitokens successfully changed to 'trace' for 30 seconds (pending restart of XRootD)

$ date && echo "" && pelican server set-logging-level trace 12s --param Logging.Origin.Scitokens --server https://`hostname`:8444
Mon Apr 27 21:40:55 UTC 2026

Change ID: 45dd00f9-f3e2-45b2-adbe-02e8411e8e63
Change will revert at: 2026-04-27T21:41:07Z
Log level for Logging.Origin.Scitokens successfully changed to 'trace' for 12 seconds
```

It seems like the first command wins and the second command doesn't do anything other than give a false report to the client. I guess I'd expect the second command to override the first command, or if nothing else, report that a change is already in progress and cannot be updated.

This block is bypassed because `requiresRestart` is determined by comparing the proposed value against the current in-memory param value. In this case, both are the same, so `DetectXrootdLoggingChange` returns false.

xrootd/restart.go
jhiemstrawisc, line 150: Not your code, but this was flagged by Claude -- should this have a read lock?

xrootd/restart.go
```diff
@@ -56,23 +57,50 @@ var (
...
+//
+// This keeps xrootd restart logic decoupled from launcher_utils (which owns the
+// concrete Director advertisement behavior) and provides a test seam for stubbing.
+func SetRestartAdvertiseFn(fn func(context.Context, []server_structs.XRootDServer) error) {
+	if fn == nil {
+		advertiseServersFn = func(_ context.Context, _ []server_structs.XRootDServer) error { return nil }
+		return
+	}
+	advertiseServersFn = fn
+}
```
jhiemstrawisc, line 94: I'm a bit worried about this setter function because of `pelican serve` mode where both the Cache and the Origin end up calling it via `cache_serve.go` and `origin_serve.go`. Conveniently, they set the same advertise fn, so the goroutine race is benign. However, this strikes me as the kind of thing that could break in the future in a hard-to-diagnose way.

xrootd/restart.go
```diff
@@ -231,19 +263,50 @@ func RestartXrootd(ctx context.Context, oldPids []int) (newPids []int, err error
...
+	if len(servers) == 0 {
+		return nil
+	}
+	// This advertisement assumes RestartXrootd has already set component health
+	// to "shutting down", so Director sees restart-in-progress state immediately.
+	log.Infof("Waiting %s for in-flight transfers before shutting down", param.Xrootd_ShutdownTimeout.GetDuration().String())
+	if advErr := advertiseServersFn(serverCtx, servers); advErr != nil {
+		log.Errorf("Failed to advertise before shutdown: %v", advErr)
+	}
+	time.Sleep(param.Xrootd_ShutdownTimeout.GetDuration())
```
jhiemstrawisc: One thing that's recently occurred to me is that we now technically have mutable config. I'm not sure how much we need to guard against multi-read race conditions, but it might be worth sampling the value of `param.Xrootd_ShutdownTimeout` only once in this section to guarantee the sleep value and the logged value can't drift.

cmd/logging_set_level.go
```diff
@@ -157,21 +157,33 @@ func setLogLevel(cmd *cobra.Command, args []string) error {
...
+	if response.RequiresRestart {
+		if response.EffectiveAt != nil {
+			fmt.Printf("Note: A restart of XRootD is required.\n")
+			fmt.Printf("Change will take effect at: %s\n", response.EffectiveAt.Local().Format(time.RFC3339))
+		}
+	}
+	fmt.Printf("Change will revert at: %s\n", response.EndTime.Local().Format(time.RFC3339))
+	fmt.Printf("Log level for %s successfully changed to '%s' for %d seconds", response.ParameterName, response.Level, response.Remaining)
+	if response.RequiresRestart {
+		fmt.Printf(" (pending restart of XRootD)")
```
jhiemstrawisc: It feels strange to me to say the level was "successfully changed" followed by "pending restart" -- the first indicates something has already happened, while the second indicates it is scheduled for the future. What do you think about something that is more explicit about what's going on under the hood, e.g. "Log level change ... was successfully registered" followed by "value should take affect after restart of XRootD"?

launchers/origin_serve.go
```diff
@@ -191,11 +191,15 @@ func OriginServe(ctx context.Context, engine *gin.Engine, egrp *errgroup.Group,
 		originServer.SetPids(pids)
 
 		// Store restart information after PIDs are known
-		xrootd.StoreRestartInfo(launchers, pids, egrp, portStartCallback, false, useCMSD, privileged)
+		launch := func(ls []daemon.Launcher) ([]int, error) {
+			return xrootd.LaunchDaemons(ctx, ls, egrp, portStartCallback)
+		}
```
jhiemstrawisc, line 206: This concept of closure was new to me -- a neat trick to have in the toolbox for solving these types of problems!

jhiemstrawisc (changes requested):

web_ui/logging_api.go
brianaydemir: Oddly enough, I actually had this conversation with Copilot during development:

```
  The timeline for a request with duration=D, drainPeriod=P:

  ┌───────┬────────────────────────────────────────────────────────────┐
  │ Time  │ Event                                                      │
  ├───────┼────────────────────────────────────────────────────────────┤
  │ T=0   │ AddChange called; timer set to fire at T=D; param.Set(new) │
  │       │ fires callback; RestartXrootd begins; preRestartHook       │
  │       │ drains for P seconds                                       │
  ├───────┼────────────────────────────────────────────────────────────┤
  │ T=P   │ Initial restart complete — new level now active in XRootD  │
  ├───────┼────────────────────────────────────────────────────────────┤
  │ T=D   │ Revert timer fires; param.Set(old) fires callback; second  │
  │       │ RestartXrootd begins; preRestartHook drains again for P    │
  │       │ seconds                                                    │
  ├───────┼────────────────────────────────────────────────────────────┤
  │ T=D+P │ Revert restart complete — old level restored               │
  └───────┴────────────────────────────────────────────────────────────┘

  The new level is active in XRootD from T=P to T=D+P, which is exactly
  D seconds — the full requested duration. The drain periods are
  symmetric overhead on both ends; they don't reduce the effective
  window.
```

This analysis ignores the unknowable jitter from the amount of time it takes XRootD to actually restart, but I think that's okay.

And I think it's reasonable to include the drain period as part of the change's duration because the purpose of the drain period is to let existing clients complete their transfers — during which time XRootD might have things to log.

xrootd/restart.go
brianaydemir: I don't think it can hurt, and it's easy enough to include this in the read lock we're already taking a few lines down.

xrootd/restart.go
brianaydemir: Ayup. The calls can be moved into `LaunchModules`, where we're already deciding what one-time things need doing based on which modules are being started.

xrootd/restart.go
brianaydemir: Agreed.

web_ui/logging_api.go
brianaydemir: For this PR, I'd like to limit the scope to, "Make a single log-level change, with nothing else going on, work reasonably."

There is code that intends to ensure that multiple, overlapping changes do something reasonable (e.g., at any given moment, the most verbose log level wins), but as you've pointed out, there are gaps in when it is reasonable for overlapping changes to be registered in the first place.

Then there are potential bugs around what happens when config, is say, reloaded.

I'll leave it up to you whether to write this up as an separate issue.

brianaydemir:
@jhiemstrawisc Back in your court. In addition to responding various comments, I went ahead and fixed the bug where the log level never actually appeared to revert.

jhiemstrawisc (approved):
I gave this a quick re-review and nothing new jumps out to me. Thanks for the work on this!

PR #3175, opened 2026-02-28
Client modifications needed for the local cache

bbockelm:
The refactored local cache work uses the client for all its "heavy lifting"; the goal is to keep all download, upload, and stat work centralized into a single location so any bugs (or improvements) only need to be written once.

Accordingly, we really had to "bulk up" some of the functionality of the client to make it work better as a library and not just a download/upload tool.

The two potentially-breaking changes in this work are:

- Director responses are cached by the transfer engine in a TTLCache. Previously, each separate item transferred incurred the cost of a director lookup.
- The internal "stat" function prefers to "stat" against a cache in cases where we will be downloading from the cache.  If the object turns out to be a collection, it'll redo the stat with the collections URL.

Other important changes include:

- Makes the `ETag` and other headers evailable to the client even before the data transfer starts.  This will be used by the cache to determine where / how to store the object.
- Some `ETag` awareness in the download paths: if different caches provide the same object with different `ETag`s, abort the transfer.
- Downloads can be done for a given byte-range instead of a whole object.
- If the PUT response includes an ETag, then this is included in the transfer results returned to the client.
- Uploads and downloads can be done against a Reader/Writer instead of a file on the filesystem.
- The client can be put in "cache mode", contacting origins directly even for origins that refuse direct reads from end-users.  For obvious reasons, this is not exposed to the CLI.
- The client can provide a second, "federation token", that the cache uses to provide it's not an end-user to origins that reject.
- Token generation is now an interface, `TokenProvider`, in case the caller has a separate implementation of the logic.

Further, this dulls the edge on - but doesn't fix - a medium-sized design flaw in the client: both the transfer client and transfer engine try to close the results channel as if they own it.  So, if you are careless about the ordering of shutdown, you can trigger a panic on close.  This fixes the panic but not the underlying "shared ownership" problem.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -1327,9 +1327,10 @@ func (tc *TransferClient) NewTransferJob(ctx context.Context, remoteUrl *url.URL
 
 	tj.directorUrl = copyUrl.FedInfo.DirectorEndpoint
 
-	// Use the director response cache with singleflight coalescing.
-	// Concurrent jobs for paths under the same namespace will share
-	// a single director query.
+	// Use the director response cache with singleflight coalescing for
+	// concurrent jobs requesting the same object.  Note we cannot coalesce
+	// at the namespace level because we don't know the namespace an object
+	// is in a-priori.
 	var dirResp server_structs.DirectorResponse
 	directorFailed := false
 	if tc.engine != nil && tc.engine.dirRespCache != nil {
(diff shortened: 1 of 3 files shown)


bbockelm:
In the name of shrinking the local cache rewrite, I'm going ahead with this one...

PR #3189, opened 2026-03-05
HA Director Forwarding Fix for 3+ Directors 

### LINKED ISSUE #3185: HA Directors infinite forwarding loop with 3+ directors and response body leak
Two correctness issues in the high-availability director implementation:
1. Infinite forwarding loop when there are 3 or more directors, service ads (origin/cache registrations) can be forwarded indefinitely between directors instead of converging.

2. HTTP response body leak in director-to-director advertisement, when a peer director returns a non-200 response, the response body is never closed, leaking connections over time.

## 1. Forwarding Loop:
Current behavior:
forwardServiceAd only skips the immediate sender (originatorName). With three directors A, B, C:
- Origin registers with A → A forwards to B and C.
- B receives from A, forwards to C (A excluded).
- C receives from A, forwards to B (A excluded).
- C receives from B, forwards to A (B excluded).
- B receives from C, forwards to A (C excluded).
- The same ad keeps circulating; the process never terminates.

The coalescing buffer reduces duplicate work but does not stop the loop; once the buffer drains, the next copy is accepted and forwarded again. In practice this is rate-limited by HTTP latency but remains unbounded and wastes CPU and network.

### Expected behavior:
- Each director should forward an ad at most once (after adding itself to a “seen” set). Forwarding should stop once every director has seen the ad.

### Scope:
For a 2 director setup, where the current check is sufficient. The bug only appears with 3+ directors.

### Current behavior:
On non-200 response:
```go
if resp.StatusCode != http.StatusOK {
    log.Errorln(...)
    // read body for logging
    return  // resp.Body is never closed
}
defer resp.Body.Close()  // only reached on 200
```
The body (and underlying connection) is never closed, so the connection is not returned to the pool. Under repeated failures (e.g. unhealthy or misconfigured peer), this leaks connections.

### Expected behavior:
resp.Body must always be closed (e.g. via defer resp.Body.Close() immediately after a successful client.Do()), regardless of status code.

patrickbrophy:
Closes #3185 

This PR fixes the infinite forwarding bug for a HA director setup with 3+ directors. The fix is implemented by using a `seenBy` set which tracks directors that have seen the advertisement. The `seenBy` set is forwarded along with the rest of the advertisement so that when a director receives the ad it can add itself to the list and forward the advertisement to peers that have yet to see the advertisement. 

In addition to the fix, this PR adds a suite of tests to ensure that forwarding logic and other properties of the HA directors is correct. In `director/forward_service_test.go`, I added a series of tests that will ensure that the forwarding logic works end to end across director networks of varying sizes. This PR also adds a new set of property based tests in `director/ha_property_test.go`. These tests cover properties of forwarding, advertisement ordering, advertisement staleness, and time skew correction. 

I should note that I used AI to help write these tests and I am feeling confident on the efficacy of them.

### DIFF
--- a/director/ha_property_test.go
+++ b/director/ha_property_test.go
@@ -291,8 +291,9 @@ func makeTestServiceAd() *server_structs.OriginAdvertiseV2 {
 // The test exercises 100 random configurations: varying numbers of directors
 // (2–8) and varying sizes of the initial seenBy set.
 func TestSeenByExclusionIsExact(t *testing.T) {
+	config.ResetConfig()
+	t.Cleanup(config.ResetConfig)
 	require.NoError(t, param.Set(param.Server_ExternalWebUrl.GetName(), "http://test-self.example.com"))
-	t.Cleanup(func() { _ = param.Set(param.Server_ExternalWebUrl.GetName(), "") })
 
 	setupForwardingState(t)
 
@@ -363,8 +364,9 @@ func TestSeenByExclusionIsExact(t *testing.T) {
 // If this invariant breaks—e.g., D forgets to add itself, or adds duplicates—
 // the exclusion check in a downstream director becomes incorrect.
 func TestSeenByGrowsByExactlyOnePerHop(t *testing.T) {
+	config.ResetConfig()
+	t.Cleanup(config.ResetConfig)
 	require.NoError(t, param.Set(param.Server_ExternalWebUrl.GetName(), "http://test-self.example.com"))
-	t.Cleanup(func() { _ = param.Set(param.Server_ExternalWebUrl.GetName(), "") })
 
 	setupForwardingState(t)
 
@@ -412,8 +414,9 @@ func TestSeenByGrowsByExactlyOnePerHop(t *testing.T) {
 // entries after repeated forwarding. Duplicates would bloat the payload and
 // could cause bugs in downstream directors that use seenBy for set membership.
 func TestSeenByNoDuplicates(t *testing.T) {
+	config.ResetConfig()
+	t.Cleanup(config.ResetConfig)
 	require.NoError(t, param.Set(param.Server_ExternalWebUrl.GetName(), "http://test-self.example.com"))
-	t.Cleanup(func() { _ = param.Set(param.Server_ExternalWebUrl.GetName(), "") })
 
 	setupForwardingState(t)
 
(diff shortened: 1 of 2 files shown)


director/ha_property_test.go
```diff
@@ -0,0 +1,647 @@
...
+		}
+		assert.LessOrEqual(t, diff, tolerance,
+			"trial %d: skew correction failed to preserve lifetime (lifetime=%v skew=%v)", trial, lifetime, skew)
+	}
+}
+
+// TestTimeSkewCorrectionBelowThresholdIsNoop verifies that small clock
+// differences (≤100ms) do not trigger any correction. The ad's expiry should
+// remain exactly as the sender set it.
+func TestTimeSkewCorrectionBelowThresholdIsNoop(t *testing.T) {
```
jhiemstrawisc: I'm struggling to figure out what this test is actually testing. It looks like it's testing its own logic. Should it be calling some function?

director/ha_property_test.go
```diff
@@ -0,0 +1,647 @@
...
+		}
+	}
+}
+
+// TestSeenByNoDuplicates verifies that the seenBy list never contains duplicate
+// entries after repeated forwarding. Duplicates would bloat the payload and
+// could cause bugs in downstream directors that use seenBy for set membership.
+func TestSeenByNoDuplicates(t *testing.T) {
+	require.NoError(t, param.Set(param.Server_ExternalWebUrl.GetName(), "http://test-self.example.com"))
+	t.Cleanup(func() { _ = param.Set(param.Server_ExternalWebUrl.GetName(), "") })
```
jhiemstrawisc: And here.

director/ha_property_test.go
```diff
@@ -0,0 +1,647 @@
...
+				for _, entry := range fwd.SeenBy {
+					seen[entry]++
+				}
+				for entry, count := range seen {
+					assert.Equal(t, 1, count,
+						"trial %d: seenBy entry %q appears %d times in payload for %s",
+						trial, entry, count, name)
+				}
+			default:
+				// director might have been skipped due to seenBy; that's fine
```
jhiemstrawisc: I'm not convinced I agree with this comment. For `names[1:]`, presumably these should _never_ be skipped because `forwardServiceAd` should generate a `seenBy` slice containing only `names[0]`, no?

jhiemstrawisc (approved):
LGTM. Assuming tests pass, let's merge it!

PR #3198, opened 2026-03-06
First pass at site-local cache docs

jhiemstrawisc:


docs/app/operating-a-federation/cache/site-local-cache/page.mdx
```diff
@@ -0,0 +1,167 @@
...
+Then launch the Cache as usual:
+
+```bash
+pelican cache serve --config /path/to/pelican.yaml
+```
+
+<Callout>
+  Even though the site-local Cache does not join the federation, it still needs to know the federation discovery endpoint so it can discover federation-managed Caches to pull objects from on a Cache miss.
+  Note that site-local Cache attempt to fetch missing objects from federation Caches, not directly from Origins.
+</Callout>
```
williamnswanson: This seems like a repeat of the end of the `How It Works` section? If this is intended to be a "say it three times so the reader definitely gets the point" that might be fine.

docs/app/operating-a-federation/cache/site-local-cache/page.mdx
```diff
@@ -0,0 +1,167 @@
...
+Client:
+  PreferredCaches:
+    - https://my-site-local-cache.example.com:8442
+```
+
+Or via the environment variable (space-separated):
+
+```bash
+export PELICAN_CLIENT_PREFERREDCACHES="https://my-site-local-cache.example.com:8442"
+```
```
williamnswanson:
````suggestion
Or via the environment variable (space-separated):

```bash
export PELICAN_CLIENT_PREFERREDCACHES="https://my-site-local-cache.example.com:8442"
```

For persistent configuration, set [`Client.PreferredCaches`](/parameters#Client-PreferredCaches) in your Pelican configuration file or environment variable:

```yaml
Client:
  PreferredCaches:
    - https://my-site-local-cache.example.com:8442
```
````

This groups the "temp / one-off" configurations together and have the one persistent config by itself.

PR #3202, opened 2026-03-09
Add knob to disable client X509 in Caches, error if enabled with bad cert

jhiemstrawisc:
This addresses the problem that most CAs will no longer issue certificates with both client/server auth in the EKU. With this diff, we disable client x509 by default.

If the user wishes to turn client x509 back on, they may do so using the `Cache.DisableClientX509` knob -- but if they do, their certificate MUST contain the appropriate clientAuth attr (otherwise Pelican will refuse to start).


h2zh (changes requested):
It's a big win for finding this bug! Some tiny comments for your fix as below : )

utils/ca_utils.go
```diff
@@ -238,3 +239,37 @@ func isSameDirSymlink(f fs.DirEntry, dir string) bool {
...
+		block, pemBytes = pem.Decode(pemBytes)
+		if block == nil {
+			break
+		}
+		if block.Type != "CERTIFICATE" {
+			continue
+		}
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			break
```
h2zh: Why this is `break` instead of returning an error?

xrootd/launch.go
```diff
@@ -174,6 +175,29 @@ func makeUnprivilegedXrootdLauncher(daemonName string, xrootdRun string, configP
...
+		// required because modern CAs (e.g. Let's Encrypt) no longer issue certs
+		// with the clientAuth Extended Key Usage, causing SSL errors when the Cache
+		// acts as a client against an Origin on Cache misses.
+		if param.Cache_DisableClientX509.GetBool() {
+			result.ExtraEnv = append(result.ExtraEnv, "XRD_CURLDISABLEX509=1")
+		} else {
+			// The operator has explicitly opted into sending the host cert. Validate
+			// that the configured TLS certificate actually includes clientAuth EKU;
+			// if it doesn't, the Cache will likely fail to communicate with Origins.
+			if certFile := param.Server_TLSCertificateChain.GetString(); certFile != "" {
```
h2zh: We can remove this check to flatten the if-else block, because `Server_TLSCertificateChain` cannot be an empty string here, thanks to the checks in `ValidateHostCertificateHostname` func. Or, we can keep it here as a defensive check.

PR #3205, opened 2026-03-10
Add Read Only Middleware

CannonLock:
Configured via Server.ReadOnly this boolean will indicate whether unsafe or modifying HTTP requests will be allowed.

Requests that do not fit that will be rejected with a clear error message.


CannonLock:
@h2zh I can't attach the relevant issue because it is so old it won't come up in the drop down.

Additionally I am going to hit the UI parts of this in another PR. It is going to be a massive amount of work that I doubt gets in this release. As things are the UI displays everything just fine and will just error out when people try to change things.

PR #3219, opened 2026-03-12
Force Login on Oauth Flow

### LINKED ISSUE #2766: There is no good way for a non-admin user to log themselves out of an origin
### The Problem

Imagine that you're a user who wants to access a protected namespace, and that the origin for the namespace has the built-in issuer enabled for granting tokens that authorize such access. You run `pelican object {get,put} ...` and get told to visit https://origin.example.com/api/v1.0/issuer/device?user_code=123_456_789, which you dutifully do.

Now, suppose that the authentication provider is misconfigured, so it sends an incorrect list of group memberships to the origin. At this point, the origin sets a cookie in your web browser that indicates you are "logged in." This cookie also contains that incorrect list of group memberships.

Unfortunately, it doesn't matter if someone fixes the configuration flaw with the authentication provider and asks you to try `pelican object {get,put} ...` again. When you visit https://origin.example.com/api/v1.0/issuer/device?user_code=ABC_DEF_123, lo and behold, the origin thinks that you're already logged in, does _not_ send you to the authentication provider, and continues to use the incorrect list of group memberships from the cookie.

Maybe we can be clever and log ourselves out by visiting https://origin.example.com/:

---
Image

---

While that page technically gets the job done, that's a pretty bad user experience.

### What I'd expect

I can imagine a couple of options:

1. Make the origin's web UI usable for non-admin users.

2. Always force the user to re-authenticate when they visit `/api/v1.0/issuer/device`.

The first option is something that needs to happen regardless. The second option I think is what a standalone OA4MP instance behaves like; we should double check and mimic it, if so.

CannonLock:
Users are reused on issuer endpoints which means you can't try and relogin with new groups.

Make the endpoint smarter so that if you are not coming in from a recent login it forces a new login so that your data is as fresh as possible.

### DIFF
--- a/oa4mp/proxy.go
+++ b/oa4mp/proxy.go
@@ -29,6 +29,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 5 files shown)


brianaydemir (changes requested):

h2zh:
Doing some housekeeping for 7.27... Could you follow up this ticket when you're back? @CannonLock

CannonLock:
@brianaydemir Where are you at with this issue? Shall we just close this and the attached issue since we are dropping OA4MP in v26?

PR #3224, opened 2026-03-12
Fix isProxyEnabled()

### LINKED ISSUE #2371: disableHTTPProxy isn't really being considered

turetske:
-- Checks Client_DisableHttpProxy first
-- Also now checks all 4 standar env vars
-- Creates tests to ensure proper handling

Closes #2371

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -696,7 +696,7 @@ func isProxyEnabled() bool {
 		return false
 	}
 	for _, envVar := range []string{"http_proxy", "HTTP_PROXY", "https_proxy", "HTTPS_PROXY"} {
-		if _, isSet := os.LookupEnv(envVar); isSet {
+		if val, isSet := os.LookupEnv(envVar); isSet && val != "" {
 			return true
 		}
 	}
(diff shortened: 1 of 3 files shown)


client/handle_http.go
```diff
@@ -588,13 +588,15 @@ func mergeCancel(ctx1, ctx2 context.Context) (context.Context, context.CancelFun
...
 func isProxyEnabled() bool {
-	if _, isSet := os.LookupEnv("http_proxy"); !isSet {
-		return false
-	}
 	if param.Client_DisableHttpProxy.GetBool() {
 		return false
 	}
-	return true
+	for _, envVar := range []string{"http_proxy", "HTTP_PROXY", "https_proxy", "HTTPS_PROXY"} {
+		if _, isSet := os.LookupEnv(envVar); isSet {
```
h2zh: Do we need to consider these variable is set to an empty string?

PR #3234, opened 2026-03-13
Add other backends (HTTP/WebDAV, Globus, S3/blob) to internal Pelican

### LINKED ISSUE #3245: Convert other backends to be embedded inside the Pelican process
Let's go ahead and take a swing at the other backends (HTTP/WebDAV, Globus, S3) and add support embedding them inside Pelican.

bbockelm:
This is an initial draft of moving other backends to be embedded inside the Pelican process.  Quite a bit left to go for code review and testing - but I thought it'd be good to open for other eyeballs ASAP.

@whwjiang - thought you'd have fun playing with this one.


origin_serve/backend_blob.go
```diff
@@ -0,0 +1,694 @@
...
+		blobURL, err = buildS3BlobURL(opts)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// If per-export S3 credentials were provided, set them in the environment
+	// so the gocloud AWS credential chain picks them up.
+	//
+	// FIXME(deploy): this mutates global process environment, which has two
```
turetske: Is there an existing issue for this FIXME? Otherwise please create one as I worry this will just get lost.

origin/globus.go
```diff
@@ -134,8 +134,9 @@ func InitGlobusBackend(exps []server_utils.OriginExport) error {
 		return errors.Wrap(err, "failed to initialize Globus backend: failed to get xrootd gid")
 	}
 
-	if server_structs.OriginStorageType(param.Origin_StorageType.GetString()) != server_structs.OriginStorageGlobus {
-		return errors.Errorf("failed to initialize Globus backend: Origin.StorageType is not Globus: %s",
+	ost := server_structs.OriginStorageType(param.Origin_StorageType.GetString())
```
turetske, line 131: Small nitpick, but xrootdGid doesn't apply if using OriginStorageGlobusv2 and should be behind a conditional, I think.

origin_serve/backend_blob.go
```diff
@@ -0,0 +1,694 @@
...
+	_ = fs.bucket.Delete(ctx, key+"/")
+	return firstErr
+}
+
+// Rename implements webdav.FileSystem.
+func (fs *blobFileSystem) Rename(ctx context.Context, oldName, newName string) error {
+	oldKey := blobKey(oldName)
+	newKey := blobKey(newName)
+
+	if err := fs.bucket.Copy(ctx, newKey, oldKey, nil); err != nil {
```
turetske, line 461: This isn't a recursive copy. I believe this will rename oldKey to newKey, but any child objects in the collection will be under oldKey. Which is obviously bad. If this will ever be used on a collection rather than an individual "leaf" object, it needs to iterate recursively through and do a copy/delete.

launchers/origin_serve.go
```diff
@@ -220,7 +231,11 @@ func OriginServeFinish(ctx context.Context, egrp *errgroup.Group, engine *gin.En
 
 	// Handle POSIXv2 and SSH-specific initialization now that the web server is running
 	storageType := param.Origin_StorageType.GetString()
-	useXRootD := storageType != string(server_structs.OriginStoragePosixv2) && storageType != string(server_structs.OriginStorageSSH)
+	useXRootD := storageType != string(server_structs.OriginStoragePosixv2) &&
```
turetske: This is an identical repeated code segment to lines 59-63. Should useXRootD become part of a helper function to avoid repeated code?

origin/globus_client.go
```diff
@@ -61,13 +61,36 @@ var (
...
-	globusTransferServer          = "transfer.api.globus.org"  // The resource name for the Globus transfer API server
-	globusTransferEndpointBaseUrl = "https://transfer.api.globusonline.org/v0.10/"
-	globusTransferBaseScope       = "urn:globus:auth:scope:transfer.api.globus.org:all"
+	globusTransferServer    = "transfer.api.globus.org" // The resource name for the Globus transfer API server
+	globusTransferBaseScope = "urn:globus:auth:scope:transfer.api.globus.org:all"
 )
 
+// globusIssuerURL returns the configurable Globus OIDC issuer URL.
+func globusIssuerURL() string {
+	if v := param.Origin_GlobusIssuerURL.GetString(); v != "" {
```
turetske, line 80: This is set in defaults.yaml, is it possible for it to be an empty string? Unless someone sets it explicitly, which seems odd.

origin/globus_client.go
```diff
@@ -61,13 +61,36 @@ var (
...
+		return v
+	}
+	return "https://auth.globus.org/"
+}
+
+// globusTransferAPIBaseURL returns the configurable Globus Transfer API base URL.
+func globusTransferAPIBaseURL() string {
+	if v := param.Origin_GlobusTransferAPIBaseUrl.GetString(); v != "" {
```
turetske, line 88: Same comment as above.

go.mod
```diff
@@ -53,6 +53,7 @@ require (
 	github.com/youmark/pkcs8 v0.0.0-20201027041543-1326539a0a0a
 	github.com/zsais/go-gin-prometheus v0.1.0
 	go.uber.org/atomic v1.11.0
+	gocloud.dev v0.45.0
```
turetske, line 63: So, I think this is pretty large. Since we are splitting the client/server binaries (and that's merged into main already, but may not be part of this PR) and this should only be in the server binaries, make sure this dependency is only pulled in there as the whole point of the split is to reduce the client binary size.

origin_serve/backend_blob.go
```diff
@@ -0,0 +1,694 @@
...
+	// Check if this is a "directory" by peeking at a prefix listing.
+	dirPrefix := key
+	if dirPrefix != "" && !strings.HasSuffix(dirPrefix, "/") {
+		dirPrefix += "/"
+	}
+	peekIter := fs.bucket.List(&blob.ListOptions{Prefix: dirPrefix, Delimiter: "/"})
+	if _, peekErr := peekIter.Next(ctx); peekErr == nil {
+		// It is a directory — return a lazy dir handle (a fresh iterator
+		// will be created when Readdir is called).
+		return &blobDirFile{name: name, bucket: fs.bucket, prefix: dirPrefix}, nil
```
turetske: Should this store the context for the later Readdir call?

origin_serve/backend_blob.go
```diff
@@ -0,0 +1,694 @@
...
+}
+
+// BlobFileSysInfo is returned by blobFileInfo.Sys() when metadata is available.
+type BlobFileSysInfo struct {
+	ETag string
+}
+
+func (fi *blobFileInfo) Name() string      { return fi.name }
+func (fi *blobFileInfo) Size() int64       { return fi.size }
+func (fi *blobFileInfo) Mode() os.FileMode { return 0444 }
```
turetske: Are we always read only? If so, this is fine, but I want to check.

turetske (changes requested):
Please deal with the merge conflicts as well.

origin_serve/backend_blob.go
bbockelm: Not necessarily - just tweaked it to 0644 to be a bit more clear.

(It's all faked regardless with the backend object store...)

go.mod
bbockelm: Split is indeed done.

launchers/origin_serve.go
bbockelm: Extracted an `originUsesXRootD()` helper and used it in both `OriginServe` and `OriginServeFinish`.

origin_serve/backend_blob.go
bbockelm: lingering FIXME got addressed; no longer uses env vars to inject creds.

origin/globus.go
bbockelm: Done. It's behind a conditional now: tokenGid defaults to the pelican user's own group, and only the XRootD-backed OriginStorageGlobus path overrides it with the daemon gid

origin_serve/backend_blob.go
bbockelm: Fixed — Rename now moves the whole subtree (paginated, best-effort).  Definitely not going to have atomic semantics like a FS -- but that should be OK in an object store?

origin/globus_client.go
bbockelm: Guarded against it: globusIssuerURL() returns the configured value only when it's non-empty and otherwise falls back to the canonical https://auth.globus.org/.

(and the handling of defaults.yaml changed dramatically to boot so even this should be relatively unreachable)

origin/globus_client.go
bbockelm: Yup, same fix!  Good catch.

origin_serve/backend_blob.go
bbockelm: Agreed.

blobDirFile now carries the request ctx and uses it for Readdir, so the lazy directory read does not use a default `context.Background()`.

bbockelm:
Ok - beyond a current Mac OS X GHA outage, I think this is ready to go... can pick things up later if there are real OS X failures.

PR #3258, opened 2026-03-20
Add missing generated code

jhiemstrawisc:
This adds some generated code that was never committed. To commit the generated code, I also had to fix a missing eof newline.


CannonLock:
@jhiemstrawisc Looking at what I have for the parameters I think my mistake here was the omission of `web_ui/frontend/public/data/error_codes.json` in the `.gitignore`

PR #3266, opened 2026-03-23
Run nightly tests with race detector and add nightly test badge

### LINKED ISSUE #3263: Incorporate Go's builtin race detector into nightly test runs, add repo banner
PR #3119 introduced a nightly run of tests to aggregate potentially-flaky tests. One thing I'd like to add to this is go's builtin `-race` detector flag, which could help surface various race conditions throughout the code base. In fact, it helped me surface the issue I solved in #3259. In particular, I don't want to run tests with race detection on PR tests because it incurs a time overhead penalty of 2-3x.

I'll pull at the thread just a bit more -- in addition to running the tests nightly with race detection, let's add a banner to the repo to surface the test results more easily (I wasn't even aware the nightly tests had been merged because you have to hunt for the results).

jhiemstrawisc:
Since the race detector is good at... finding race bugs, I decided it was worth running nightly. I don't want to run as part of every PR's suite of actions because of the extra time overhead (2-3x from what I can tell).

In addition, I'm adding a repo badge that will turn red when nightly tests fail. I more or less expect this to be red for the foreseeable future, and it'll be a badge of shame until we finally clean up all of our lousy tests.

Closes #3263

### DIFF
--- a/.github/workflows/test-linux-scheduled.yml
+++ b/.github/workflows/test-linux-scheduled.yml
@@ -16,6 +16,7 @@ jobs:
     uses: ./.github/workflows/test-linux.yml
     with:
       image: hub.opensciencegrid.org/pelican_platform/pelican-test:latest-itb
+      race_detection: true
 
   analyze-runs:
     name: Analyze Runs
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -16,6 +16,10 @@ on:
       image:
         required: true
         type: string
+      race_detection:
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   run-tests:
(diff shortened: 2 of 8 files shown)


brianaydemir (changes requested):
The nightly Linux tests don't work:

```
=== Errors
go: -race requires cgo; enable cgo by setting CGO_ENABLED=1

DONE 0 tests, 1 error in 0.003s
```

[The race detector requires a C compiler](https://go.dev/doc/articles/race_detector#Requirements), which we don't install in the  Linux test container.

brianaydemir (approved):
LGTM.

Being honest: I've forgotten what I did originally to determine the issue with a missing C compiler. The testing process that uncovered that is something that should be documented, but that's not on this PR.

PR #3277, opened 2026-03-25
Generate .goreleaser.yml from template

### LINKED ISSUE #2479: Generate .goreleaser.yaml from template
There's some duplication in `.goreleaser.yml` and steps that need to be done to update it in the release process (such as updating the version number in RPM requirements, see #2438 ).

1. Remove `.goreleaser.yml` from the repository; in its place, create a template and a script that makes `.goreleaser.yml` from that template
2. Modify the GitHub Action to add a step that runs the script
3. Modify the Makefile to run the script
4. Modify the Dockerfile to run the script

matyasselmeci:
This pull request adds a new script and Makefile target to generate a `.goreleaser.generated.yml` file from a template. This is used to allow us to substitute the version into places where GoReleaser's own template substitution doesn't take effect (e.g., in RPM package dependencies). The Dockerfile and GitHub Actions are updated to use the new generation script and the resulting config. The original `.goreleaser.yml` is deleted to make sure it doesn't get used by accident.

The placeholders `%VERSION%` and `%RPMVERSION%` in the template file are substituted with the raw version and RPM-safe version (wherein `-` characters are replaced with `~`); the version should be derived using the same logic GoReleaser would use (use `GORELEASER_CURRENT_TAG` if it's set (such as in the Dockerfile), or use the current tag, or use the nearest tag).

Note: This includes the change from #3236


jhiemstrawisc (approved):
Looks okay to me. I tested creating `goreleaser.generated.yml` locally and was able to build from it (after modifying my build scripting to use it).

I also tested building a Docker image with a custom `v123.456` version and verified that Pelican reported the correct version on the other end.

PR #3295, opened 2026-03-31
Auto fill in GitHub-OAuth-specific config param values

### LINKED ISSUE #3296: Simplify the config param setup for origin with Github OAuth as IDP
Too many config parameters have to be set in this case, which is inconvenient and error-prone.

h2zh:
- When `Issuer.GroupSource: github` is set, GitHub OAuth endpoints, scopes, and claim names should be set automatically
- So admin only needs to supply the OAuth App credentials and site-specific settings
- Values that the user has explicitly set (i.e. differ from the CILogon defaults that defaults.yaml seeds) are never overwritten

> Note: if a param is set in defaults.yaml, `IsSet()` returns true. I introduce a comparison here to tell what OIDC params are actually set by the user

- Also update the docs
- I span up a Github OAuth origin in the local federation to make sure it functions the same, and webUI, `pelican config get` can see the correct value.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1781,13 +1781,13 @@ func ApplyLogLevelInheritance(v *viper.Viper) {
 // values show up as IsSet()=true even when the user never wrote them. We use these
 // constants to distinguish "came from defaults.yaml" from "user explicitly set".
 var cilogonOIDCDefaults = struct {
-	issuer              string
-	authorizationEndpt  string
-	tokenEndpt          string
-	userInfoEndpt       string
-	deviceAuthEndpt     string
-	scopes              []string
-	authUserClaim       string
+	issuer             string
+	authorizationEndpt string
+	tokenEndpt         string
+	userInfoEndpt      string
+	deviceAuthEndpt    string
+	scopes             []string
+	authUserClaim      string
 }{
 	issuer:             "https://cilogon.org",
 	authorizationEndpt: "https://cilogon.org/authorize",
--- a/config/config.go
+++ b/config/config.go
@@ -1243,6 +1243,11 @@ func initConfigInternalImpl(logLevel log.Level) {
 		cobra.CheckErr(err)
 	}
 
+	// Auto-fill GitHub OAuth2 endpoints and claim names when GroupSource is "github".
+	// This runs after all config sources are merged so GroupSource is readable,
+	// and early enough that pelican config get also reflects the filled values.
+	cobra.CheckErr(applyGitHubOAuthDefaults())
+
 	// Use configuration to set the logging level, which must be fed to
 	// the logging library and isn't accessed directly through viper
 	err = setLoggingInternal()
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -1620,6 +1625,94 @@ func SetServerDefaults(v *viper.Viper) error {
...
+	return p.IsSet() && p.GetString() != defaultVal
+}
+
+// applyGitHubOAuthDefaults sets GitHub-specific OAuth2 endpoint and claim defaults
+// when Issuer.GroupSource is set to "github". This spares operators from having to
+// manually specify every fixed GitHub endpoint and claim name.
+// Values that the user has explicitly set (i.e. differ from the CILogon defaults
+// that defaults.yaml seeds) are never overwritten.
+func applyGitHubOAuthDefaults() error {
+	if strings.ToLower(param.Issuer_GroupSource.GetString()) != "github" {
```
jhiemstrawisc: I think the "class" of problems you're getting at here is the idea of there being "flavors" of auth. We now have a GitHub flavor and a CILogon flavor. What I'd like to see is a function that reads the type and dispatches to set all the proper defaults based on that type. Something like (feel free to change what you see fit):

```
type OAuthFlavor string

const (
	OAuthFlavorGitHub  OAuthFlavor = "github"
	OAuthFlavorCILogon OAuthFlavor = "cilogon"
)

func ParseOAuthFlavor(s string) (OAuthFlavor, error) {
	switch OAuthFlavor(strings.ToLower(s)) {
	case OAuthFlavorGitHub:
		return OAuthFlavorGitHub, nil
	case OAuthFlavorCILogon:
		return OAuthFlavorCILogon, nil
	default:
		return "", errors.Errorf("unknown OAuth flavor: %q", s)
	}
}

func (f OAuthFlavor) String() string {
	return string(f)
}

// setOAuthFlavorDefaults applies viper.SetDefault for the OIDC params
// appropriate to the configured OAuth flavor. Call from SetServerDefaults().
func setOAuthFlavorDefaults(v *viper.Viper) error {
	groupSource := param.Issuer_GroupSource.GetString()
	if groupSource == "" {
		return nil
	}

	flavor, err := ParseOAuthFlavor(groupSource)
	if err != nil {
		return err
	}

	switch flavor {
	case OAuthFlavorCILogon:
		v.SetDefault(param.OIDC_Issuer.GetString(), "https://cilogon.org")
		v.SetDefault(param.OIDC_AuthorizationEndpoint.GetString(), "https://cilogon.org/authorize")
		v.SetDefault(param.OIDC_TokenEndpoint.GetString(), "https://cilogon.org/oauth2/token")
		v.SetDefault(param.OIDC_UserInfoEndpoint.GetString(), "https://cilogon.org/oauth2/userinfo")
		v.SetDefault(param.OIDC_DeviceAuthEndpoint.GetString(), "https://cilogon.org/oauth2/device_authorization")
		v.SetDefault(param.OIDC_Scopes.GetString(), []string{"openid", "email", "profile"})
		v.SetDefault(param.Issuer_OIDCAuthenticationUserClaim.GetString(), "sub")
	case OAuthFlavorGitHub:
		v.SetDefault(param.OIDC_Issuer.GetString(), "https://github.com")
		v.SetDefault(param.OIDC_AuthorizationEndpoint.GetString(), "https://github.com/login/oauth/authorize")
		v.SetDefault(param.OIDC_TokenEndpoint.GetString(), "https://github.com/login/oauth/access_token")
		v.SetDefault(param.OIDC_UserInfoEndpoint.GetString(), "https://api.github.com/user")
		v.SetDefault(param.OIDC_DeviceAuthEndpoint.GetString(), "https://github.com/login/device/code")
		v.SetDefault(param.OIDC_Scopes.GetString(), []string{"user", "read:org"})
		v.SetDefault(param.Issuer_OIDCAuthenticationUserClaim.GetString(), "login")
		v.SetDefault(param.Issuer_OIDCSubjectClaim.GetString(), "id")
	}
	return nil
}
```

Assuming you move the defaults out of `config/resources/defaults.yaml` (owing to the fact that they're now dynamic) I like an approach like this for a few reasons:
- `OAuthFlavor` is a typed string with well-known constants.
- `ParseOAuthFlavor` gives a single validation point for incoming strings, with a clear error for unrecognized values.
- It can be called directly in `config/config.go::SetServerDefaults()`, so things like the `pelican config` tool can pick up runtime-derived configurations.
- It greatly simplifies the need to do all the user override checks this function currently does because that can be entirely handled with `v.SetDefault()`

jhiemstrawisc (changes requested):

h2zh:
To whoever review this PR: You should just look at commit [41914e2](https://github.com/PelicanPlatform/pelican/pull/3295/commits/41914e23f108d45313815be39c0f4da15e44355e), or the total diff. I think this "OAuth flavor dispatcher" mechanism is the most elegant way to handle this situation. Most importantly, it allows `pelican config` commands get the correct values.
I ran local manual tests for origins with CILogon, Github oauth respectively. Ans also tested the `pelican config` commands.

h2zh:
@melody0123 This PR will get you familiar with high-level auth in pelican, how to set up an origin with built-in issuer, how to use GitHub as the auth provider for origin.

config/config.go
```diff
@@ -1771,6 +1775,69 @@ func ApplyLogLevelInheritance(v *viper.Viper) {
...
+		param.OIDC_Issuer.GetName():                        "https://github.com",
+		param.OIDC_AuthorizationEndpoint.GetName():         "https://github.com/login/oauth/authorize",
+		param.OIDC_TokenEndpoint.GetName():                 "https://github.com/login/oauth/access_token",
+		param.OIDC_UserInfoEndpoint.GetName():              "https://api.github.com/user",
+		param.OIDC_DeviceAuthEndpoint.GetName():            "https://github.com/login/device/code",
+		param.OIDC_Scopes.GetName():                        []string{"user", "read:org"},
+		param.Issuer_OIDCAuthenticationUserClaim.GetName(): "login",
+		param.Issuer_OIDCSubjectClaim.GetName():            "id",
+	},
+}
```
melody0123, line 1823: I really love the concept of flavor, which is elegant and scalable. However, hard-encoded values into binary looks fishy to me. Currently, we have these default values stores in 3 places: here, parameters.yaml, and page.mdx. However, as @jhiemstrawisc mentioned earlier, there is no mechanism to enforce synchronization among all 3 files. My suggestion is to keep the default values in parameters.yaml, let compiler generate them into a go source file / generate them into another yaml file and read them in at runtime.

Another idea just now came into my mind: Now we have two flavors and each flavor speaks its own set of parameters. Do we have any mechanism to prevent different set of parameters from being mixed up? (e.g., When flavor is GitHub, `Issuer.OIDCSubjectClaim` is `sub`)

melody0123 (changes requested):
The PR is in good shape, but it reveals several deeper-level questions of the project, which may not be solved by this PR alone.
1. (Action needed) Hard-encoded (conditional) default values in the source file (config.go): the (conditional) default values (e.g., GitHub and OIDC endpoint URLs) are spread across multiple files: .go source files, page.mdx and parameters.yaml. I think the best way to synchronize them is to just keep them in the parameters.yaml and generate .go from it during compile time. However, this is an architecture-level change that may not be feasible in this PR.
2. (No action needed, for info only) Admins are free to edit config parameters in WebUI and create inconsistent state (e.g. change GroupSource to github but leave endpoint URLs as CILogon). Just point out the problem here. It should be solved in a separate PR.

melody0123:
> (No action needed, for info only) Admins are free to edit config parameters in WebUI and create inconsistent state (e.g. change GroupSource to github but leave endpoint URLs as CILogon). Just point out the problem here. It should be solved in a separate PR.

This issue is already brought up here: https://github.com/PelicanPlatform/pelican/issues/3647
Let's just focus on the hard-encoded default values then

melody0123:
Yet another rough idea:
The default config parameters are conditional because they depend on the Issuer.GroupSource value. The conditional branch is a hierarchical structure, and YAML file is also hierarchical. I was wondering if we can take advantage of the isomorphism between the two concepts to centralize the ground truth to the parameter.yaml

config/config.go
h2zh: I don't think creating another config param value yaml file is a good idea, as we recently intentionally removed `config/resources/defaults.yaml`. I remembered file-seeded defaults caused some problems in Pelican before.

Besides, as long as the system admin follows the guide here and the instruction of our staff, the mix up won't happen. It only happens when system admin explicitly change the value, where they should hold the responsibility for whatever happens.

h2zh:
Agreed on the goal: a single source of truth for these values. Two clarifications on where we actually stand:

**The compile-time generation you're describing already exists.** `docs/parameters.yaml` is compiled into Go by `go generate` (`config/parameter_defaults.go`), and the `check-go-generate` CI job fails any PR where they drift. 

**What this PR does to keep the drift surface minimal so no changes needed:**
- The affected params are `default: none` in parameters.yaml, so the generator emits no competing default: the flavor map in `config.go` is the *only* place these values exist in code, not a second copy.
- The per-flavor values are documented in each param's description in parameters.yaml, so the docs site and `pelican config describe` show them; page.mdx now describes behavior rather than restating URLs.
- This is the pattern the repo already sanctions — see the "GUIDANCE FOR DEVELOPERS" comment above `SetServerDefaults`, category "conditional/runtime-computed defaults".

config/config.go
melody0123: > I don't think creating another config param value yaml file is a good idea, as we recently intentionally removed config/resources/defaults.yaml. I remembered file-seeded defaults caused some problems in Pelican before.

I agree this part. If we previously had issues with extra yaml config files, let's avoid them.

> Besides, as long as the system admin follows the guide here and the instruction of our staff, the mix up won't happen. It only happens when system admin explicitly change the value, where they should hold the responsibility for whatever happens.

I agree that system admins do hold responsibility (to some degree). However, I think as developers, we should implement mechanisms/policies to minimize the possibility to make mistakes. Both sysadmins and we are human, and human make mistakes. If we can rely on machine-enforced checks, we should do that.

One of the simplest fix is to: 1) make config parameters in WebUI read-only; 2) Implement config parameter sanity check on server startup. I feel that both of them go beyond the scope of this PR and are worth a new ticket.

melody0123 (approved):
> The affected params are default: none in parameters.yaml, so the generator emits no competing default: the flavor map in config.go is the only place these values exist in code, not a second copy.

I agree that there is only one place (config.go) for the values to exist in the executable binary. However, my point is that these values are also written down as documentation (non-executable) under `description` keys in parameters.yaml. So here comes the problem: How to make sure the values under executable part keep aligned with non-executable part? I feel that this question is too big to address in this PR, because it boils down to a difficult question: How to make machine to enforce the alignment between code and documentation? This question may be worth a standalone toolchain project to address.

> The per-flavor values are documented in each param's description in parameters.yaml, so the docs site and pelican config describe show them; page.mdx now describes behavior rather than restating URLs.

> This is the pattern the repo already sanctions — see the "GUIDANCE FOR DEVELOPERS" comment above SetServerDefaults, category "conditional/runtime-computed defaults".

Checked and agreed.

Only thing to mention for future: avoid using force push to commit requested changes for a review.

PR #3312, opened 2026-04-03
Install sssd-client in all origin containers

### LINKED ISSUE #3310: `origin` containers dropped `sssd-client`, needed by UWDF multiuser Origins
Some of the refactoring in [this commit](https://github.com/PelicanPlatform/pelican/commit/1858ea80902a37f70cd4cd4486c9f73303d6eb4a) removed `sssd-client` from the Origin software base because it looked like the package was specific to OSDF. However, it turns out this is also needed by UWDF multiuser Origins, so the package should be restored in the software base.

### LINKED ISSUE #3440: Patch #3311 into 7.25 to restore Origin `sssd`
See:
https://github.com/PelicanPlatform/pelican/issues/3311

brianaydemir:
It would also seem that there's no actual need for an `sssd` user account.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -573,6 +573,9 @@ RUN --mount=type=bind,source=oa4mp/resources,target=/context \
   set -eux
   set -o pipefail
 
+  # Some XRootD multi-user setups require sssd-client.
+  dnf install -y sssd-client
+
   # OA4MP and Apache Tomcat both require a Java runtime.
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
👍

jhiemstrawisc:
Pinging @h2zh -- I don't see an issue for tracking 7.25 patches, but this should be patched in.

PR #3313, opened 2026-04-05
Address items found during AI code review for security

### LINKED ISSUE #3314: Security review for embedded OAuth2 issuer
@brianaydemir kindly did a security review (well, kindly forced various AI agents to do a security review...) of the upcoming OAuth2 embedded issuer code.  They found a list of things that reasonably should be fixed prior to release.

bbockelm:
Note - these are security-relevant items for *unreleased*, off-by-default code, hence the public branch.  @h2zh - though the impact here is modest, I'd like these merged prior to v7.25.0.  In terms of reviewing, about 70% is in regression tests.

Particularly:
- Ensure the different issuers have better separation (client of issuer A has no power in issuer B).
- Admin auth middleware for the admin client API.
- Tweak the DCRP rate limiter to be server-wide, not per-issuer.


oauth2/issuer/storage.go
```diff
@@ -133,15 +133,18 @@ func (s *OIDCStorage) CreateClient(ctx context.Context, client *fosite.DefaultCl
...
 		Scopes:        string(scopes),
 		Public:        client.Public,
-		Namespace:     s.Namespace,
 	}
 
+	// With the composite primary key (ID, Namespace), ON CONFLICT only
+	// triggers for the same (client_id, namespace) pair, preventing
+	// cross-namespace overwrites.
 	return s.db.WithContext(ctx).Clauses(clause.OnConflict{
 		UpdateAll: true,
```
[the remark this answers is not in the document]
bbockelm: Actually, this code overwrites existing clients, which is going to be potentially confusing if done in error.  It'd be better to force administrators to explicitly delete and recreate the client if that's what they want!

Let's take a different tack: let's fail on conflict if we can.

oauth2/issuer/handlers.go
```diff
@@ -95,21 +95,15 @@ func RegisterRoutesWithMiddleware(engine *gin.Engine, registry *ProviderRegistry
...
-	_ = registry
-	_ = middleware
-	// Admin endpoints are served by the same dispatch handler. The caller
-	// must register admin-specific auth middleware at a higher level or use
-	// a separate approach. For now, admin routes are sub-paths of the
-	// namespace dispatch — see handleDispatch which checks for /admin/ prefix.
+// RegisterAdminRoutes registers admin middleware for admin client-management endpoints.
+// The middleware is stored on each provider and enforced in the dispatch handler
+// whenever an admin/* action is requested.
+func RegisterAdminRoutes(registry *ProviderRegistry, middleware ...gin.HandlerFunc) {
```
h2zh: I'm not familiar with customizing Gin with `runAdminMiddleware` so I spent some time to read up. It turns out it might be a risky move. I found out a way to leverage Gin's native group feature to avoid this customization for your consideration.

### Why it is likely a problem

`runAdminMiddleware` called each middleware (`gin.HandlerFunc`) directly via `mw(ctx)`. When a middleware calls `ctx.Next()` internally, Gin advances `ctx.index` against its own handler chain, instead of the temporary list in `runAdminMiddleware`. This design has at least the following side effects:

- With the invocation of `ctx.Next()` in AuthHandler, AdminAuthHandler (outside of the handler chain), Gin's `ctx.index` is out of sync with the handler it points to. Middlewares would silently misbehave. 
- The length of the handler chain `len(ctx.handlers)` cannot reflect the correct number of handlers, because AuthHandler, AdminAuthHandler are not in `ctx.handlers`. 
- Customized middleware handling logics have no concurrency protection
- Expected middleware order: [userIdentityMw, NamespaceMiddleware, AuthHandler, AdminAuthHandler, handleDispatch]
- Actual middleware order: [userIdentityMw, NamespaceMiddleware, handleDispatch,AuthHandler, AdminAuthHandler]

### The Intention
I guess `runAdminMiddleware` is introduced here because the /*namespace wildcard route makes it impossible to attach per-path middleware via Gin groups on the same URL. So I'm proposing a fix avoids the problem entirely by using a separate URL prefix for admin routes. Since the built-in issuer hasn't been used in production, I think it's ok to change the routing architecture.

### Proposed fix
It would be better to register middlewares through Gin's native route groups, so that `c.Next()` advances through the correct chain in the correct order.

First, move `/admin/` prefix before the `namespace` segment, which lets Gin distinguish admin from non-admin routes at the routing-tree level. 

Before: /api/v1.0/issuer/ns/{namespace}/admin/clients[/{id}]
After:  /api/v1.0/issuer/admin/ns/{namespace}/clients[/{id}]

Then use a Gin routing group to handle all admin client-management endpoints.

```
func RegisterAdminRoutes(engine *gin.Engine, registry *ProviderRegistry, middleware ...gin.HandlerFunc) {
	allMiddleware := append([]gin.HandlerFunc{NamespaceMiddleware(registry)}, middleware...)
	adminGroup := engine.Group("/api/v1.0/issuer/admin/ns", allMiddleware...)
	adminGroup.GET("/*namespace", handleAdminDispatch)
	adminGroup.POST("/*namespace", handleAdminDispatch)
	adminGroup.PUT("/*namespace", handleAdminDispatch)
	adminGroup.DELETE("/*namespace", handleAdminDispatch)
}

func handleAdminDispatch(ctx *gin.Context) {
...
// Move all admin client-management actions here
	action := strings.TrimPrefix(ActionSuffix(ctx), "/")
...

Delete runAdminMiddleware 
```

oauth2/issuer/storage.go
```diff
@@ -592,20 +596,20 @@ func (s *OIDCStorage) DeleteOpenIDConnectSession(ctx context.Context, signature
 
 func (s *OIDCStorage) RevokeRefreshToken(ctx context.Context, requestID string) error {
 	return s.db.WithContext(ctx).Model(&OIDCRefreshToken{}).
-		Where("request_id = ?", requestID).
+		Where("request_id = ? AND namespace = ?", requestID, s.Namespace).
```
h2zh, line 599: Do we need to add the namespace param into the query on line 561 to avoid signature collision between two namespaces?

database/origin_migrations/20260401120000_oidc_clients_composite_pk.sql
```diff
@@ -0,0 +1,80 @@
...
+
+DROP TABLE oidc_clients;
+ALTER TABLE oidc_clients_new RENAME TO oidc_clients;
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+
+-- Revert to single-column primary key on id.
```
h2zh: Maybe we can remove this downgrade block since this script is incorrect: if duplicates exist, the downgrade will just stuck. Because the new primary key is (id, namespace), while the old primary key is just id.

oauth2/issuer/admin_handlers.go
```diff
@@ -445,14 +445,17 @@ func handleTokenExchange(ctx *gin.Context, provider *OIDCProvider) {
...
 	// (a) what the subject token grants
 	// (b) what the client is allowed to request
+	// No scopes are auto-allowed — including standard ones like
+	// offline_access — so that operators who omit a scope from a client's
+	// configuration can enforce that restriction.
 	grantedScopes := make([]string, 0, len(requestedScopes))
 	for _, scope := range requestedScopes {
 		// Check the scope is within the subject token's grants.
-		if !scopeAllowed(scope, subjectGrantedScopes) && !isStandardScope(scope) {
+		if !scopeAllowed(scope, subjectGrantedScopes) {
```
h2zh, line 454: Do we also need to the other invocation of `isStandardScope` on line 721 in `/oauth2/issuer/handlers.go`? We need to keep this consistent in auth code (handleAuthorizeSubmit), device flow (handleDeviceVerifySubmit), and token exchange (handleTokenExchange).

h2zh (changes requested):

oauth2/issuer/security_regression_test.go
```diff
@@ -0,0 +1,976 @@
...
+		"alpha's client must not be overwritten by beta's registration")
+
+	// Beta must NOT find alpha's client.
+	betaFound, err := betaStorage.GetClient(context.Background(), clientID)
+	require.NoError(t, err)
+	assert.Equal(t, []string{"https://beta/callback"}, betaFound.GetRedirectURIs(),
+		"beta must have its own client record")
+}
+
+// --- Finding #10: PKCE enforcement for public clients ---
```
h2zh, line 645: Missing Finding # 9

oauth2/issuer/security_regression_test.go
```diff
@@ -0,0 +1,976 @@
...
+	_, err = alphaStorage.GetRefreshTokenSession(context.Background(), "alpha-rt-sig", &WLCGSession{})
+	assert.NoError(t, err, "alpha's refresh token must not be revoked by beta's revocation call")
+}
+
+// ---- Helpers ----
+
+// secBcryptHash bcrypt-hashes a secret string.
+func secBcryptHash(secret string) ([]byte, error) {
+	return bcrypt.GenerateFromPassword([]byte(secret), bcrypt.DefaultCost)
+}
```
h2zh: Bumped into a [blog post](https://anti-pattern.com/tales-from-the-bcrypt): Unit tests related to password generation can be 50% faster, if you lower the cost of BCrypt hashing from DefaultCost(10) to `bcrypt.MinCost` (4)

oauth2/issuer/security_regression_test.go
```diff
@@ -0,0 +1,976 @@
...
+	require.NoError(t, dbutils.MigrateDB(sqlDB, database.EmbedUniversalMigrations, "universal_migrations"))
+	require.NoError(t, dbutils.MigrateServerSpecificDB(sqlDB, database.EmbedOriginMigrations, "origin_migrations", "origin"))
+
+	alphaStorage := NewOIDCStorage(db, "/alpha")
+	betaStorage := NewOIDCStorage(db, "/beta")
+
+	// Create clients in both namespaces.
+	for _, s := range []*OIDCStorage{alphaStorage, betaStorage} {
+		client := &fosite.DefaultClient{
+			ID: "rev-client", Secret: []byte("$2a$10$fakehashfakehashfakehashfakehashfakehashfakehashfake"),
```
h2zh: Maybe it's better to use `secBcryptHash` to set this secret?

If that's a better approach, replace other "$2a$10$fakehashfakehashfakehashfakehashfakehashfakehashfake" in this file

h2zh (changes requested):
All the things in this section are nice-to-have

oauth2/issuer/security_regression_test.go
bbockelm: Done in latest commit.

oauth2/issuer/security_regression_test.go
bbockelm: Done in the latest commit.

oauth2/issuer/security_regression_test.go
bbockelm: Added back in latest commit.

database/origin_migrations/20260401120000_oidc_clients_composite_pk.sql
bbockelm: Ok, deleted!

oauth2/issuer/handlers.go
bbockelm: Ok - I think I've got this right.  Please re-review.

oauth2/issuer/storage.go
```diff
@@ -540,8 +543,11 @@ func (s *OIDCStorage) DeleteRefreshTokenSession(ctx context.Context, signature s
 // than immediately revoking it, allowing brief reuse during network issues.
 func (s *OIDCStorage) RotateRefreshToken(ctx context.Context, requestID string, refreshTokenSignature string) error {
 	return s.db.WithContext(ctx).Model(&OIDCRefreshToken{}).
-		Where("request_id = ? AND active = ?", requestID, true).
-		Update("first_used_at", gorm.Expr("COALESCE(first_used_at, ?)", time.Now().UTC())).Error
+		Where("signature = ? AND namespace = ? AND active = ?", refreshTokenSignature, s.Namespace, true).
+		Updates(map[string]interface{}{
+			"first_used_at": gorm.Expr("COALESCE(first_used_at, ?)", time.Now().UTC()),
+			"active":        false,
+		}).Error
```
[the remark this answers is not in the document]
bbockelm: Addressed in latest commit.

oauth2/issuer/storage.go
bbockelm: Done!

oauth2/issuer/admin_handlers.go
bbockelm: Good catch, we should remove it there as well.

h2zh (approved):
LGTM! They’re well-considered. I’ve finished hand testing the new binary as well.

PR #3317, opened 2026-04-06
Cleanup downtime when a server is deleted

### LINKED ISSUE #3215: When a Service Registration is deleted from the Registry, also cleanup existing downtimes for them
A pending Registration for a server was deleted after it had set a downtime for itself in the Registry. This left behind and orphaned Downtime for "" (empty string) service.

Image

Image

The Registry should ensure that any database entries that reference a deleted Registration (or the registered server, services, ect.) also get cleaned up.

h2zh:
Besides downtime cleanup during server removal, this PR also fixes bugs in the SQL query for `DELETE /api/v1.0/registry_ui/servers/:id` (though this endpoint is never used in WebUI)

### DIFF
--- a/registry/registry_db.go
+++ b/registry/registry_db.go
@@ -828,6 +828,10 @@ func deleteRegistrationByID(id int) error {
 				if err := tx.Where("id = ? AND NOT EXISTS (?)", svc.ServerID, subq).Delete(&server_structs.Server{}).Error; err != nil {
 					return err
 				}
+				// Delete all downtimes associated with this server (no FK constraint, so must be done explicitly)
+				if err := tx.Where("server_id = ?", svc.ServerID).Delete(&server_structs.Downtime{}).Error; err != nil {
+					return errors.Wrapf(err, "failed to delete downtimes for server %s", svc.ServerID)
+				}
 			} else {
 				// Update server flags to reflect remaining service types
 				var hasOrigin, hasCache bool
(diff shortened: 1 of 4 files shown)


registry/registry_db.go
```diff
@@ -864,17 +868,21 @@ func deleteRegistrationByPrefix(prefix string) error {
...
 func deleteServerByID(id string) error {
+	// Fetch registrations before opening the transaction. getServerByID uses the
+	// global database.ServerDatabase rather than a tx-scoped connection; calling it
+	// inside the Transaction callback would require a second connection from the pool,
+	// which deadlocks on a single-connection SQLite pool (common in tests).
+	serverRegistration, err := getServerByID(id)
+	if err != nil {
+		return errors.Wrap(err, "failed to get server by ID")
+	}
+
```
CannonLock: It would be safer to not use this function and get the servers registrations another way inside the transaction block.

Doubt it happens but this is open to a registration being added after you get the existing ones, but before you delete the server.

registry/registry_db.go
```diff
@@ -828,6 +828,10 @@ func deleteRegistrationByID(id int) error {
 				if err := tx.Where("id = ? AND NOT EXISTS (?)", svc.ServerID, subq).Delete(&server_structs.Server{}).Error; err != nil {
 					return err
 				}
+				// Delete all downtimes associated with this server (no FK constraint, so must be done explicitly)
```
CannonLock, line 831: Might be worth it to create a single delete server function that can be called in a transaction to save duplicate code in the two places you delete a server.

CannonLock (changes requested):
Don't care about creating the deleteServer function, but we should get the registrations inside the transaction to prevent loose ends.

registry/registry_db.go
h2zh: Done. Elimate a potential race condition here. Thanks for the note!

registry/registry_db.go
h2zh: I considered this, but the two functions have fairly different control flow — deleteRegistrationByID conditionally deletes the server (or updates is_origin/is_cache if services remain), while deleteServerByID unconditionally removes the server and all its registrations. The only truly duplicated bit is the downtime deletion, which is ~3 lines.
Also worth noting: the endpoint that calls deleteServerByID is never actually invoked by webUI.

CannonLock (approved):
LGTM

PR #3327, opened 2026-04-06
Remove stale pending server registration requests automatically

### LINKED ISSUE #1428: Automatically reject or ignore registration attempts without valid DNS names
Pelican Service:
- [ ] Client
- [ ] Plugin
- [x] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

The OSDF registry is getting registration requests with random hostnames, e.g. `/origins/4a43a3f359c0:8444`.  These are probably automatic registrations from containers that don't have real hostnames.  They will likely never be useful so it would be nice to avoid clutter and have the registry ignore or at least automatically reject registrations with names that do not resolve in DNS.

h2zh:
This PR implements a timeout-based cleanup based on observed Origin/Cache server activity.

### How it works

A new goroutine (LaunchInactiveRegistrationCleanup) runs every `Registry.InactiveRegistrationCleanupInterval` (default 5m). It finds servers whose `last_seen` is older than `Registry.InactiveRegistrationTimeout` (default 20m), deletes their linked Pending server registration, then removes the server rows. 

`last_seen` gets updated in the `servers` table in Registry database every time when the Origin/Cache server polls metadata from Registry.

### Backward compatibility

Existing DB rows on upgrade: Migration backfills last_seen from updated_at/created_at; active servers refresh within minutes. NULL rows are excluded from WHERE last_seen < ? (SQL three-valued logic), so they are never spuriously deleted.
Old origin/cache + new registry: Poll metadata from Registry on a regular basis → last_seen refreshes normally. No client changes needed.
New origin + old registry: Heartbeat calls succeed harmlessly; no cleanup goroutine exists on old registry.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -3257,9 +3257,12 @@ components: ["registry"]
 ---
 name: Registry.InactiveRegistrationTimeout
 description: |+
-  The duration after which a server row is considered stale if its metadata has not been polled.
-  The registry updates servers.last_seen every time a server polls its metadata. The cleanup goroutine
-  deletes every registration linked to stale servers (including the server itself).
+  How long a pending server registration can linger in the registry if it stops running after
+  registering itself. While an Origin or Cache is awaiting admin approval, the registry
+  records the last time the server polled its own registration metadata. If no contact is
+  received within this timeout, the registration and all its associated namespace entries are
+  deleted. Only registrations still in the Pending state are affected; registrations that have
+  already been approved or denied are never removed by this cleanup.
 type: duration
 default: 20m
 components: ["registry"]
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -3259,7 +3259,7 @@ name: Registry.InactiveRegistrationTimeout
 description: |+
   The duration after which a server row is considered stale if its metadata has not been polled.
   The registry updates servers.last_seen every time a server polls its metadata. The cleanup goroutine
-  deletes every registration linked to stale servers (including the server itself). 
+  deletes every registration linked to stale servers (including the server itself).
 type: duration
 default: 20m
 components: ["registry"]
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -3255,6 +3255,22 @@ default: false
 osdf_default: true
 components: ["registry"]
 ---
+name: Registry.InactiveRegistrationTimeout
+description: |+
+  The duration after which a server row is considered stale if its metadata has not been polled.
+  The registry updates servers.last_seen every time a server polls its metadata. The cleanup goroutine
+  deletes every registration linked to stale servers (including the server itself). 
+type: duration
+default: 20m
+components: ["registry"]
+---
+name: Registry.InactiveRegistrationCleanupInterval
+description: |+
+  How often the registry runs the inactive server registration cleanup job.
+type: duration
+default: 5m
+components: ["registry"]
+---
 ############################
 #   Server-level configs   #
 ############################
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -1282,7 +1282,7 @@ func getServerByPrefixHandler(ctx *gin.Context) {
 
 	// Refresh servers.last_seen so the cleanup goroutine knows this server is still active.
 	if err := updateServerLastSeen(server.ID); err != nil {
-		log.Warningf("Failed to update last_seen for server %s (prefix %s): %v", server.ID, prefix, err)
+		log.Warningf("Failed to update last_seen for server %q (prefix %q): %v", server.ID, prefix, err)
 	}
 
 	ctx.JSON(http.StatusOK, server)
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -1280,6 +1280,11 @@ func getServerByPrefixHandler(ctx *gin.Context) {
 		return
 	}
 
+	// Refresh servers.last_seen so the cleanup goroutine knows this server is still active.
+	if err := updateServerLastSeen(server.ID); err != nil {
+		log.Warningf("Failed to update last_seen for server %s (prefix %s): %v", server.ID, prefix, err)
+	}
+
 	ctx.JSON(http.StatusOK, server)
 }
 
--- a/registry/registry_db.go
+++ b/registry/registry_db.go
@@ -674,7 +674,9 @@ func AddRegistration(ns *server_structs.Registration) error {
 				if isCache {
 					existingServer.IsCache = true
 				}
-				existingServer.UpdatedAt = time.Now()
+				now := time.Now().UTC()
+				existingServer.UpdatedAt = now
+				existingServer.LastSeen = now
 
 				if err := tx.Save(&existingServer).Error; err != nil {
 					return errors.Wrapf(err, "failed to update server service(s): %s", ns.AdminMetadata.SiteName)
@@ -692,6 +694,21 @@ func AddRegistration(ns *server_structs.Registration) error {
 
 		return nil
 	})
+}
+
+// updateServerLastSeen sets servers.last_seen = now (UTC). Called from getServerByPrefixHandler
+// when a server polls its metadata (e.g. each advertisement cycle).
+func updateServerLastSeen(serverID string) error {
+	if serverID == "" {
+		return errors.New("server ID is required to update last_seen")
+	}
+	result := database.ServerDatabase.Model(&server_structs.Server{}).
+		Where("id = ?", serverID).
+		Update("last_seen", time.Now().UTC())
+	if result.Error != nil {
+		return errors.Wrapf(result.Error, "failed to update last_seen for server %q", serverID)
+	}
+	return nil
 }
 
 func updateRegistration(ns *server_structs.Registration) error {
--- a/registry/registry_db.go
+++ b/registry/registry_db.go
@@ -239,6 +239,7 @@ func buildServerRegistration(server server_structs.Server, services []server_str
 		Note:      server.Note,
 		CreatedAt: server.CreatedAt,
 		UpdatedAt: server.UpdatedAt,
+		LastSeen:  server.LastSeen,
 	}
 
 	for _, service := range services {
@@ -303,6 +304,7 @@ func getServerByRegistrationID(registrationID int) (*server_structs.ServerRegist
 		Note:      service.Server.Note,
 		CreatedAt: service.Server.CreatedAt,
 		UpdatedAt: service.Server.UpdatedAt,
+		LastSeen:  service.Server.LastSeen,
 	}
 
 	for _, svc := range allServices {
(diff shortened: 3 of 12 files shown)


docs/parameters.yaml
```diff
@@ -2893,6 +2893,22 @@ default: false
 osdf_default: true
 components: ["registry"]
 ---
+name: Registry.InactiveRegistrationTimeout
+description: |+
+  The duration after which a server row is considered stale if its metadata has not been polled.
+  The registry updates servers.last_seen every time a server polls its metadata. The cleanup goroutine
```
jhiemstrawisc: I don't think the average admin reading this will know what `servers.last_seen` refers to from this description, and I doubt they care to know about goroutines, etc. Can you make the description less mountain dewy?

docs/parameters.yaml
```diff
@@ -2893,6 +2893,22 @@ default: false
...
+  deletes every registration linked to stale servers (including the server itself).
+type: duration
+default: 20m
```
jhiemstrawisc, line 3267: @williamnswanson, what do you think about this default? My gut says 20m is short, since some registrations might sit around on the order of days before Fabio finally asks for you to approve them. But this technically controls how long a registration can sit around without sending any pings, so maybe a short value is okay?

registry/registry.go
```diff
@@ -1314,6 +1314,11 @@ func getServerByPrefixHandler(ctx *gin.Context) {
 		return
 	}
 
+	// Refresh servers.last_seen so the cleanup goroutine knows this server is still active.
+	if err := updateServerLastSeen(server.ID); err != nil {
+		log.Warningf("Failed to update last_seen for server %s (prefix %s): %v", server.ID, prefix, err)
```
jhiemstrawisc: `updateServerLastSeen` has handling for an empty `server.ID`, so I assume it's somehow possible this is called with the empty value. Given that, can you either wrap the %s format specifiers in quotes or use %q?

registry/registry_db.go
```diff
@@ -692,6 +696,57 @@ func AddRegistration(ns *server_structs.Registration) error {
...
+}
+
+// deleteStaleServerRegistrations uses listServers to find servers whose last_seen
+// is older than cutoff and whose every registration is still Pending, then removes
+// each one via deleteServerByID. Servers with any non-Pending registration are
+// skipped; standalone namespace registrations (no servers row) are untouched.
+func deleteStaleServerRegistrations(cutoff time.Time) (regsDeleted, serversDeleted int64, err error) {
+	servers, err := listServers()
+	if err != nil {
+		return 0, 0, errors.Wrap(err, "failed to list servers for stale cleanup")
```
jhiemstrawisc: In the error handle, can you say something that mentions what's stale, e.g. "failed to list servers for stale pending registration cleanup"? This makes it explicit for the admin who might otherwise wonder what a stale cleanup is.

registry/registry_db.go
```diff
@@ -692,6 +696,57 @@ func AddRegistration(ns *server_structs.Registration) error {
...
+	}
+
+	for _, srv := range servers {
+		if srv.LastSeen.IsZero() || !srv.LastSeen.Before(cutoff) {
```
jhiemstrawisc, line 747: I stumbled over the `srv.LastSeen.IsZero()` at first -- can you add a comment explaining it?

registry/registry_db.go
```diff
@@ -692,6 +696,57 @@ func AddRegistration(ns *server_structs.Registration) error {
...
+				allPending = false
+				break
+			}
+		}
+		if !allPending {
+			continue
+		}
+
+		if err := deleteServerByID(srv.ID); err != nil {
+			log.Warningf("Failed to delete stale server %s (%s): %v", srv.ID, srv.Name, err)
```
jhiemstrawisc: I wouldn't call them stale servers, I'd say they're stale server registrations.

registry/registry_db_server_test.go
```diff
@@ -688,3 +691,526 @@ func TestServerWithMultipleServices(t *testing.T) {
...
+	err := database.ServerDatabase.Model(&server_structs.Server{}).
+		Where("id = ?", serverID).
+		Update("last_seen", ts).Error
+	require.NoError(t, err)
+}
+
+func TestUpdateServerLastSeen(t *testing.T) {
+	t.Cleanup(test_utils.SetupTestLogging(t))
+	setupMockRegistryDB(t)
+	defer teardownMockRegistryDB(t)
```
jhiemstrawisc: Should these test function wide cleanups be in a test cleanup function? It's not always clear to me when it's okay to use `defer` and when it's not.

registry/registry_db_server_test.go
```diff
@@ -688,3 +691,526 @@ func TestServerWithMultipleServices(t *testing.T) {
...
+	t.Run("ErrorOnEmptyID", func(t *testing.T) {
+		err := updateServerLastSeen("")
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "server ID is required")
+	})
+
+	t.Run("NoErrorOnNonExistentServerID", func(t *testing.T) {
+		// The function updates 0 rows but does not return an error (GORM doesn't
+		// treat zero-row updates as errors by default).
+		err := updateServerLastSeen("doesnotexist")
```
jhiemstrawisc, line 893: What does it imply if the actual source code calls `updateServerLastSeen` on a non-existent entry? How could we get to that point?

registry/registry_db.go
```diff
@@ -692,6 +696,57 @@ func AddRegistration(ns *server_structs.Registration) error {
...
+		return errors.Wrapf(result.Error, "failed to update last_seen for server %q", serverID)
+	}
+	return nil
+}
+
+// deleteStaleServerRegistrations uses listServers to find servers whose last_seen
+// is older than cutoff and whose every registration is still Pending, then removes
+// each one via deleteServerByID. Servers with any non-Pending registration are
+// skipped; standalone namespace registrations (no servers row) are untouched.
+func deleteStaleServerRegistrations(cutoff time.Time) (regsDeleted, serversDeleted int64, err error) {
```
jhiemstrawisc: Any reason why you think we need to use an int64 here? Do you expect that many pending servers, or are you trying to future proof something?

registry/registry_db.go
```diff
@@ -864,17 +919,21 @@ func deleteRegistrationByPrefix(prefix string) error {
 }
 
 func deleteServerByID(id string) error {
+	// Fetch registrations before opening the transaction. getServerByID uses the
```
jhiemstrawisc: Your tests for this function mention there's a cascade behavior associated with deleting a server. That kind of behavior (e.g. what other fields this removes) should live in a docstring explaining what it means to delete a "server" from the database.

registry/registry_db_server_test.go
```diff
@@ -688,3 +691,526 @@ func TestServerWithMultipleServices(t *testing.T) {
...
+		assert.NoError(t, err)
+	})
+
+	t.Run("UpdatesTimestampForExistingServer", func(t *testing.T) {
+		defer resetMockRegistryDB(t)
+
+		pubkey, err := test_utils.GenerateJWKS()
+		require.NoError(t, err)
```
jhiemstrawisc: I think you could cut down on some code duplication by rolling this into `createPendingServerReg`, since the key is always passed to the function. The only one I'm not sure about is where you create both an Origin/Cache with the same key to test deleting a multi-service server. I'm guessing the server resolution for that delete doesn't depend on the same key being used for both?

registry/registry_db.go
```diff
@@ -692,6 +696,57 @@ func AddRegistration(ns *server_structs.Registration) error {
...
+	result := database.ServerDatabase.Model(&server_structs.Server{}).
+		Where("id = ?", serverID).
+		Update("last_seen", time.Now().UTC())
+	if result.Error != nil {
+		return errors.Wrapf(result.Error, "failed to update last_seen for server %q", serverID)
+	}
+	return nil
+}
+
+// deleteStaleServerRegistrations uses listServers to find servers whose last_seen
```
jhiemstrawisc, line 722: Something to double check -- it's been awhile since I had a deep dive in the registry, but if I recall, setting `Registry.RequireCacheApproval` or `Registry.RequireOriginApproval` to false (the default) expose the services' keys, which is why the Director can verify the advertisement token and add the server ads to its ttlcache. However, I _think_ those servers stay in the pending state, unless those knobs also control auto-approving them. How does this new feature of removing pending approvals work with those existing knobs?

registry/registry_db_server_test.go
```diff
@@ -688,3 +691,526 @@ func TestServerWithMultipleServices(t *testing.T) {
...
+	defer teardownMockRegistryDB(t)
+
+	t.Run("EmptyDatabase", func(t *testing.T) {
+		regs, servers, err := deleteStaleServerRegistrations(time.Now().UTC())
+		require.NoError(t, err)
+		assert.Equal(t, int64(0), regs)
+		assert.Equal(t, int64(0), servers)
+	})
+
+	t.Run("DeletesPendingStaleServer", func(t *testing.T) {
```
jhiemstrawisc, line 1063: I left an earlier comment asking how this new feature works with `Registry.RequireCacheApproval` and `Registry.RequireOriginApproval` because I suspect that when these are false, the servers are left in the pending state and never auto approved. Regardless of whether that ends up being true, you should include a test that probes it.

registry/registry_db_server_test.go
```diff
@@ -688,3 +691,526 @@ func TestServerWithMultipleServices(t *testing.T) {
...
+
+		// last_seen is recent (5 minutes ago), cutoff is 1 hour ago.
+		setServerLastSeenDirect(t, srv.ID, time.Now().UTC().Add(-5*time.Minute))
+
+		_, serversDeleted, err := deleteStaleServerRegistrations(time.Now().UTC().Add(-1 * time.Hour))
+		require.NoError(t, err)
+		assert.Equal(t, int64(0), serversDeleted, "recently active pending server should not be deleted")
+	})
+
+	t.Run("SkipsServerWithZeroLastSeen", func(t *testing.T) {
```
jhiemstrawisc, line 1169: What could cause a last seen value of zero?

jhiemstrawisc (changes requested):

jhiemstrawisc:
One thing I wasn't sure of in the review is whether an Origin/Cache already has a routine or loop that will hit the right endpoint to trigger `updateServerLastSeen`. Do you know whether/how that happens? Maybe there could be an E2E test that starts a registry with short pending server expirations, starts an Origin/Cache to trigger a pending registration, makes sure the Origin/Cache ping at least a few times and the registry doesn't delete the registration, then finally kills the Origin/Cache to make sure everything is cleaned up?

docs/parameters.yaml
williamnswanson: So long as a server can remain in Pending ~forever while it's active and reachable, having a short time-since-last-seen purge is fine.

registry/registry_db.go
h2zh: Right, this is overengineering

registry/registry_db.go
h2zh: See commit ba2072a436d11db1f8d452449136fbfd256eef08
I didn't go deep to differenciate RequireCacheApproval or RequireOriginApproval because in practice, the Registry admin set them both to the same value. Even when they forget to set them both, here we check the "either" condition - either origin or cache doesn't need approval, the pending registration auto cleanup will not run.

registry/registry_db_server_test.go
h2zh: Same as above commit. I append a helper func `withApprovalRequired` for all relevant unit tests, and add a dedicated subtest (SkipsCleanupWhenApprovalNotRequired) for the scenario where approval is not required

h2zh:
`updateServerLastSeen` was hit by every server ad, which is a feature being well-tested. Besides, the current unit tests already provide enough coverage for this new feature.

registry/registry_db_server_test.go
h2zh: Good catch! I remember either you or BrianA once mentioned the nuance between defer and t.Cleanup. defer is not parallel-safe. So I changed all 11 occurrences of `defer teardownMockRegistryDB(t)` to `t.Cleanup(func() { teardownMockRegistryDB(t) })`
Having said that, these two `t.Cleanup` cannot be combined. `t.Cleanup(test_utils.SetupTestLogging(t))` works because the argument is evaluated immediately. It sets up the test logger right away and registers the returned restore function for later.

registry/registry_db_server_test.go
h2zh: Good idea! I refactor `createPendingServerReg` to receive an optional JWKS

registry/registry_db_server_test.go
h2zh: A pending server registration migrated from a DB version before the last_seen column was introduced will have a zero timestamp for last_seen. Such registrations should not be treated as stale.

registry/registry_db_server_test.go
h2zh: It means a TOCTOU race: the server row existed when getServerByPrefix read it, but was deleted before the update committed.

h2zh: LaunchInactiveRegistrationCleanup fires on a ticker and calls deleteStaleServerRegistrations → deleteServerByID. A stale server can be deleted at the same moment it polls. If the handler's read lands before the delete and the updateServerLastSeen write lands after it, you get the 0-row update.

jhiemstrawisc (approved):
I used Claude to re-review this and scanned some of the recent changes myself. Nothing jumped out to me as needing another round of revision. Really nice work with this one!

PR #3340, opened 2026-04-10
Cli docs

### LINKED ISSUE #1648: Add other command line information to the client documentation

turetske:
Updated the CLI docs with current 'pelican object' commands.

Close #1648

### DIFF
--- a/docs/app/about-pelican/core-concepts/page.mdx
+++ b/docs/app/about-pelican/core-concepts/page.mdx
@@ -17,6 +17,13 @@ or OSDF). This is because the underlying cache mechanism, powered by [XRootD](ht
 an object's contents change at the origin, the cache will remain unaware and continue to deliver the old object. In the worst case, when the cache only has a partial object,
 it may attempt to combine its stale version with whatever exists at the origin. Use object names wisely!
 
+### Collections
+A *collection* is a grouping of objects within a namespace, analogous to a directory on a filesystem.
+Collections may contain objects as well as other collections (called *subcollections*).
+Unlike directories, collections do not carry filesystem metadata such as owners or permissions.
+They exist purely as organizational structure within the federation's namespace.
+For example, in the object path `/pelicanplatform/test/hello-world.txt`, the paths `/pelicanplatform` and `/pelicanplatform/test` are both collections.
+
 ### Federations
 Objects in Pelican belong to *federations*, which are aggregations of data that are exposed to other individuals in the federation. Each Pelican
 federation constitutes its own global namespace of objects and each object within a federation has its own path, much like files on a computer.
--- a/docs/app/getting-data-with-pelican/client/page.mdx
+++ b/docs/app/getting-data-with-pelican/client/page.mdx
@@ -497,7 +497,6 @@ These flags are available on every `pelican` subcommand, including all `object` 
 - **`-f` or `--federation`:** Takes a URL that indicates to Pelican which federation the request should be made to.
 - **`-d` or `--debug`:** Takes no argument, but runs Pelican in debug mode, which provides verbose output for debugging purposes.
 - **`--config`:** Takes a filepath and indicates to Pelican the location of the Pelican configuration file.
-- **`--json`:** Takes no argument and outputs results in JSON format.
 - **`-l` or `--log`:** Takes a string that specifies a file location to output the pelican logs to rather than the stderr.
 - **`--version`:** Prints the version of Pelican and exits.
 
--- a/docs/app/getting-data-with-pelican/client/page.mdx
+++ b/docs/app/getting-data-with-pelican/client/page.mdx
@@ -367,9 +367,12 @@ This will upload objects from the local directory into the remote collection, cr
 A token is required because writing to a federation always requires authorization.
 
 <Callout type="warning">
-When uploading, if the source is a single file and the destination path ends with `/`, Pelican interprets the destination as a collection.
-If that collection does not already exist at the remote, the destination will be treated as an object name instead.
-To avoid confusion, omit the trailing `/` when the destination is intended to be an object, or ensure the collection already exists.
+When uploading a single file, be careful with trailing slashes on the destination path.
+A trailing `/` signals that the destination is a collection, and the file should be placed inside it.
+However, if that collection does not already exist at the remote, the trailing `/` is ignored and the entire path (including what looks like a collection name) becomes the object's name.
+
+For example, syncing `myfile.txt` to `pelican://osg-htc.org/my-namespace/new-dir/` will place `myfile.txt` inside `new-dir` if that collection exists, but if `new-dir` does not exist, it will create an object literally named `new-dir/` instead.
+To avoid this, either ensure the destination collection exists beforehand, or omit the trailing `/` so the destination is treated as an explicit object name.
 </Callout>
 
 ### Sync with a Specific Cache
@@ -405,6 +408,11 @@ This is useful for sharing access to objects without distributing long-lived tok
 Sharing is **prefix-based**, meaning all objects whose names match the given prefix will be accessible via the generated URL.
 For example, sharing `pelican://osg-htc.org/my-namespace/data/` grants access to all objects under that path.
 </Callout>
+
+The generated URL contains an embedded authorization token scoped to specific operations:
+
+- **Read-only (default):** The token grants the `storage.read` scope. Recipients can use `object get`, `object stat`, and `object ls` against matching object names.
+- **Read-write (`--write`):** The token grants the `storage.create` scope in addition to read. Recipients can also use `object put` and `object sync` to upload objects under the shared prefix.
 
 ### Generate a Read-Only Sharing URL
 
@@ -483,7 +491,9 @@ For a complete reference of all flags for every subcommand, see the [Commands Re
 
 ### Global Flags:
 
-- **`-h` or `--help`:** Takes no argument and can be used with any Pelican sub command for more information about the sub command and additional supported flags.
+These flags are available on every `pelican` subcommand, including all `object` subcommands:
+
+- **`-h` or `--help`:** Displays usage information for the current subcommand, including all supported flags. Available on every Pelican command and subcommand.
 - **`-f` or `--federation`:** Takes a URL that indicates to Pelican which federation the request should be made to.
 - **`-d` or `--debug`:** Takes no argument, but runs Pelican in debug mode, which provides verbose output for debugging purposes.
 - **`--config`:** Takes a filepath and indicates to Pelican the location of the Pelican configuration file.
--- a/docs/app/getting-data-with-pelican/client/page.mdx
+++ b/docs/app/getting-data-with-pelican/client/page.mdx
@@ -1,4 +1,5 @@
 import ExportedImage from "next-image-export-optimizer";
+import { Callout } from 'nextra/components'
 
 # Pelican's Command Line Client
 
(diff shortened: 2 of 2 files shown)


docs/app/about-pelican/core-concepts/page.mdx
```diff
@@ -13,10 +13,17 @@ Pelican should be thought of as a tool that works with federated *objects* as op
...
 an object's contents change at the origin, the cache will remain unaware and continue to deliver the old object. In the worst case, when the cache only has a partial object,
 it may attempt to combine its stale version with whatever exists at the origin. Use object names wisely!
 
+### Collections
+A *collection* is a grouping of objects within a namespace, analogous to a directory on a filesystem.
+Collections may contain objects as well as other collections (called *subcollections*).
+Unlike directories, collections do not carry filesystem metadata such as owners or permissions.
+They exist purely as organizational structure within the federation's namespace.
+For example, in the object path `/pelicanplatform/test/hello-world.txt`, the paths `/pelicanplatform` and `/pelicanplatform/test` are both collections.
+
```
jhiemstrawisc, line 26: I don't think this has been true for awhile -- Patrick's work over the last year introduced collections as a first-class resource management system with their own API, data model and access control mechanisms, so they do in fact have owners/ACLs and other metadata.

This gets messy because we now have two different definitions of "collections" in our codebase -- the heavy weight collections Patrick worked on, and the synonym for a namespace prefix that we reference in our various "ls" or "stat" commands. I don't think it's on you to clean this up, but we shouldn't keep digging.

Since I don't think your CLI docs really pertain specifically to collections, you could just get rid of this section. Otherwise, it should be updated to reflect the current state of the collections API. Then maybe in your actual docs, be vague about or avoid the term "collections" so you don't create linkages in the readers mind between the two concepts in our code that have drifted apart. Thoughts?

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
 pelican object copy <path/to/local/file> pelican://<federation-url></namespace-prefix></path/to/destination> -t </path/to/token/file>
 ```
 
+## Listing Objects in a Collection
```
jhiemstrawisc, line 168: If you choose to get rid of the "collections" terms here, you could swap it out for "namespace prefix", e.g. I think this line could be "Listing Objects Under a Namespace Prefix"

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+
+To list objects in a remote collection, use Pelican's `object ls` subcommand.
+This command queries a federation for all objects and subcollections at a given path.
+
+<Callout type="info">
+A **collection** is a grouping of objects in a namespace, analogous to a directory on a filesystem.
+Collections may contain other collections (subcollections) as well as objects.
+For a detailed definition, see [Core Concepts](../../about-pelican/core-concepts#collections).
+The `ls` command requires that the origin serving the namespace has the **Listing** capability enabled.
```
jhiemstrawisc, line 177: Can you capitalize the Pelican nouns like "Origin", "Cache", "Director", etc? This statement is also slightly off -- the namespace at the Origin needs to have listings enabled, not just the Origin itself.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+
+```bash copy
+pelican object ls pelican://osg-htc.org/pelicanplatform/test
+```
+
+This will display the objects and subcollections at the `/pelicanplatform/test` path, similar to running `ls` on a local directory.
+
+You can verify whether an origin supports listings before running `ls` by checking the Director.
+The Director's web UI shows each origin's capabilities as colored chips — look for a green **Listing** chip.
+Alternatively, query the Director's API:
```
jhiemstrawisc, line 193: I think this is off in a similar way -- you want to check the namespace/federation prefix capabilities, not necessarily the Origin's.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+
+```bash copy
+curl https://<director-url>/api/v1.0/director_ui/servers?server_type=origin
+```
+
+Look for `"Listing": true` in the `capabilities` field of the relevant origin.
```
jhiemstrawisc, line 199: This is really mountain dewy to include in our client documentation. I think it's plenty to tell the user to check the Director.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+
+```bash copy
+# List only collections (directories)
+pelican object ls -C pelican://osg-htc.org/pelicanplatform/test
+
+# List only objects (files)
+pelican object ls -O pelican://osg-htc.org/pelicanplatform/test
+```
+
+### Recursive Listing
```
jhiemstrawisc, line 229: Somewhere in here, it's worth mentioning the user should be cautious here -- requesting a listing over a very wide or very deep prefix might be costly for both the Origin and the Client.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+```console
+       Name: /pelicanplatform/test
+       Size: 0
+    ModTime: 2024-09-19 12:54:00 +0000 UTC
+IsCollection: true
+```
+
+### Requesting Checksums
+
+If the origin supports it, you can request checksum information for an object:
```
jhiemstrawisc, line 310: The next question a user will have is "how do I know if the Origin supports it?" I don't think there's a good way to know this from a user's perspective, so maybe you can follow up with "Origins that don't support checksums will return an error like <blah>" or whatever the actual truth happens to be (my bet is on silently ignoring checksums with little to no indication to the client user).

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+
+```bash copy
+pelican object stat --checksums md5 pelican://osg-htc.org/pelicanplatform/test/hello-world.txt
```
jhiemstrawisc, line 313: I'm confused about this command -- does it cause the client to verify the download against a specific checksum? Does it download the checksum as a separate file for use by something else?

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+
+This will transfer all objects from the remote `/pelicanplatform/test` collection into the local directory, preserving the directory structure.
+If the local directory already contains some of the objects at their full size, those objects will be skipped.
+
+### Sync from a Local Directory to a Remote Collection
+
+To upload local files to a remote collection:
+
+```bash copy
+pelican object sync /local/path/to/directory pelican://osg-htc.org/my-namespace/data -t /path/to/token
```
jhiemstrawisc, line 363: For the various token commands included here, do all of them work with device-code flow setups? If so, I think it's worth discussing both token and OAuth2 modes of interaction. I don't think that needs to be discussed in depth for every command, but maybe in depth in once place and then referenced elsewhere.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+
+This will upload objects from the local directory into the remote collection, creating any necessary paths in the namespace.
+A token is required because writing to a federation always requires authorization.
+
+<Callout type="warning">
+When uploading a single file, be careful with trailing slashes on the destination path.
+A trailing `/` signals that the destination is a collection, and the file should be placed inside it.
+However, if that collection does not already exist at the remote, the trailing `/` is ignored and the entire path (including what looks like a collection name) becomes the object's name.
+
+For example, syncing `myfile.txt` to `pelican://osg-htc.org/my-namespace/new-dir/` will place `myfile.txt` inside `new-dir` if that collection exists, but if `new-dir` does not exist, it will create an object literally named `new-dir/` instead.
```
jhiemstrawisc, line 374: Can you avoid referencing directories in the path prefix here, and rather than saying "inside", say "under"? I think it's important to seed these concepts of how object stores are different from filesystems wherever we can. "inside" is a directory concept, and doesn't extend to object stores like S3.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+To avoid this, either ensure the destination collection exists beforehand, or omit the trailing `/` so the destination is treated as an explicit object name.
+</Callout>
+
+### Sync with a Specific Cache
+
+Use the `-c` flag to prefer a specific cache for downloads:
```
jhiemstrawisc, line 380: This could use more exposition to explain that it has the explicit effect of priming the cache, and that the data need not be entirely housed by the cache for the sync to pick up date from the Origin.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -217,27 +486,53 @@ The `?recursive` and the `?directread` queries do **not** require any sort of va
...
 ### Global Flags:
 
-- **`-h` or `--help`:** Takes no argument and can be used with any Pelican sub command for more information about the sub command and additional supported flags.
+These flags are available on every `pelican` subcommand, including all `object` subcommands:
+
+- **`-h` or `--help`:** Displays usage information for the current subcommand, including all supported flags. Available on every Pelican command and subcommand.
 - **`-f` or `--federation`:** Takes a URL that indicates to Pelican which federation the request should be made to.
 - **`-d` or `--debug`:** Takes no argument, but runs Pelican in debug mode, which provides verbose output for debugging purposes.
 - **`--config`:** Takes a filepath and indicates to Pelican the location of the Pelican configuration file.
-- **`--json`:** Takes no argument and outputs results in JSON format.
```
jhiemstrawisc, line 228: This looks like it's technically still available for all `object` commands:
```
$ pelican object -h
<snipped>

Global Flags:

Use "pelican object [command] --help" for more information about a command.
```

I'm not sure what it actually _does_ for something like a `pelican object get`, so I'm not sure whether we want the docs here to be "technically correct" or "more targeted but potentially confusing".

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -217,27 +486,53 @@ The `?recursive` and the `?directread` queries do **not** require any sort of va
...
 - **`-l` or `--log`:** Takes a string that specifies a file location to output the pelican logs to rather than the stderr.
 - **`--version`:** Prints the version of Pelican and exits.
 
 ### Flags For `object get/put/copy`:
 
 - **`-c` or `--cache`:** Takes a cache URL and indicates to Pelican that only the specified cache should be used. When used, Pelican will not attempt to use other caches if the provided cache cannot provide the file.
 - **`--caches`:** Takes the path to a JSON file containing a list of caches. Similar to the `-c` flag, Pelican will attempt to use only these caches in the order they are listed.
-- **`-h` or `--help`:** Gives additional information on how to use the command as well as lists these flags with short descriptions for the `object copy` command.
 - **`--methods`:** Takes a comma separated list of methods to try for downloads/uploads, the default is just http.
```
jhiemstrawisc, line 507: I'm not sure what the `--method` flag is, but I don't see it for a `pelican object get`?
```
$ pelican object get -h

Flags:
      --async                   Run the transfer asynchronously through the client API server and return a job ID
  -c, --cache string            A comma-separated list of preferred caches to try for the transfer, where a "+" in the list indicates
                                the client should fallback to discovered caches if all preferred caches fail.
      --caches string           A JSON file containing the list of caches
      --direct                  Download directly from an origin, bypassing any caches (same as '?directread' query)
      --dry-run                 Show what would be downloaded without actually downloading
  -h, --help                    help for get
      --inplace                 Write files directly to destination (default: use temporary files)
      --pack string             Package transfer using remote packing functionality (same as '?pack=' query). Options: auto, tar, tar.gz, tar.xz, zip. Default: auto when flag is provided without an explicit value
  -r, --recursive               Recursively download a collection.  Forces methods to only be http to get the freshest collection contents
  -t, --token string            Token file to use for transfer
      --transfer-stats string   A path to a file to write transfer statistics to
      --wait                    When used with --async, wait for the job to complete before returning

Global Flags:
```

I know you didn't write the line, but it caught my eye and I'm not sure what it's supposed to be.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -217,27 +486,53 @@ The `?recursive` and the `?directread` queries do **not** require any sort of va
...
 - **`-d` or `--debug`:** Takes no argument, but runs Pelican in debug mode, which provides verbose output for debugging purposes.
 - **`--config`:** Takes a filepath and indicates to Pelican the location of the Pelican configuration file.
-- **`--json`:** Takes no argument and outputs results in JSON format.
 - **`-l` or `--log`:** Takes a string that specifies a file location to output the pelican logs to rather than the stderr.
 - **`--version`:** Prints the version of Pelican and exits.
 
 ### Flags For `object get/put/copy`:
 
 - **`-c` or `--cache`:** Takes a cache URL and indicates to Pelican that only the specified cache should be used. When used, Pelican will not attempt to use other caches if the provided cache cannot provide the file.
 - **`--caches`:** Takes the path to a JSON file containing a list of caches. Similar to the `-c` flag, Pelican will attempt to use only these caches in the order they are listed.
```
jhiemstrawisc, line 506: I don't think there's a `--caches` option for puts.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -217,27 +486,53 @@ The `?recursive` and the `?directread` queries do **not** require any sort of va
...
 
-- **`-h` or `--help`:** Takes no argument and can be used with any Pelican sub command for more information about the sub command and additional supported flags.
+These flags are available on every `pelican` subcommand, including all `object` subcommands:
+
+- **`-h` or `--help`:** Displays usage information for the current subcommand, including all supported flags. Available on every Pelican command and subcommand.
 - **`-f` or `--federation`:** Takes a URL that indicates to Pelican which federation the request should be made to.
 - **`-d` or `--debug`:** Takes no argument, but runs Pelican in debug mode, which provides verbose output for debugging purposes.
 - **`--config`:** Takes a filepath and indicates to Pelican the location of the Pelican configuration file.
-- **`--json`:** Takes no argument and outputs results in JSON format.
 - **`-l` or `--log`:** Takes a string that specifies a file location to output the pelican logs to rather than the stderr.
```
jhiemstrawisc, line 500: I think the log shorthand is a capital `-L`.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+```bash copy
+pelican object share pelican://osg-htc.org/my-namespace/data/shared-dataset
+```
+
+This outputs a URL with an embedded authorization token that can be shared with others.
+Recipients can use this URL directly with `pelican object get` to download the shared objects.
+
+### Generate a Read-Write Sharing URL
+
+To allow writes to the shared prefix, add the `--write` flag:
```
jhiemstrawisc, line 428: The title says "Read-Write" but if someone didn't just read two sections above, the description and the flag indicate a write-only sharing URL. Can you add a quick note that `--write` here implies read?

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+
+## Sharing Object Access
+
+The `object share` command generates a URL that grants access to a namespace prefix.
+This is useful for sharing access to objects without distributing long-lived tokens.
+
+<Callout type="info">
+Sharing is **prefix-based**, meaning all objects whose names match the given prefix will be accessible via the generated URL.
+For example, sharing `pelican://osg-htc.org/my-namespace/data/` grants access to all objects under that path.
+</Callout>
```
jhiemstrawisc, line 410: I much prefer the language in this section to the collections language in previous sections. Let's open a follow-up issue to figure out what to do about our two concepts of "collections" in the codebase.

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -164,6 +165,274 @@ and to do an `object put`:
...
+You can verify this by running `pelican object ls` against the collection first.
+</Callout>
+
+### How Sync Determines What to Transfer
+
+The `object sync` command uses object sizes to determine whether an object needs to be transferred:
+
+- If an object at the destination does not exist, it will be transferred.
+- If an object at the destination is the same size or larger than the corresponding source object, it will **not** be overwritten.
+- If an object at the destination is smaller than the source object, `sync` assumes a previous transfer was interrupted and appends the remaining bytes to complete the partial transfer.
```
jhiemstrawisc, line 339: Is this entirely true? I checked it out, and it seems like there are different modes:
https://github.com/PelicanPlatform/pelican/blob/c54914990adf9cdf8e14d2b3e7f804734e107423/client/handle_http.go#L441-L445

Strangely, I don't think these are settable via the CLI (not sure why we bother with them if that's the case). However, the client is hardcoded to use a `SyncSize` for gets and puts:
https://github.com/PelicanPlatform/pelican/blob/c54914990adf9cdf8e14d2b3e7f804734e107423/cmd/object_sync.go#L180
https://github.com/PelicanPlatform/pelican/blob/c54914990adf9cdf8e14d2b3e7f804734e107423/cmd/object_sync.go#L202

So in the functions that actually determine whether to skip or proceed with an action, we'll end up modifying an object whenever the sizes aren't exactly equal:
https://github.com/PelicanPlatform/pelican/blob/c54914990adf9cdf8e14d2b3e7f804734e107423/client/handle_http.go#L4135-L4137
https://github.com/PelicanPlatform/pelican/blob/c54914990adf9cdf8e14d2b3e7f804734e107423/client/handle_http.go#L4164-L4166

docs/app/getting-data-with-pelican/client/page.mdx
```diff
@@ -217,27 +486,53 @@ The `?recursive` and the `?directread` queries do **not** require any sort of va
...
+### Flags For `object stat`:
+
+- **`-j` or `--json`:** Print results in JSON format.
+- **`--checksums`:** Checksums to request from the server. Known values: `md5`, `crc32c`, `crc32`, `sha`.
+- **`-t` or `--token`:** Token file to use for accessing protected namespaces.
+
+### Flags For `object sync`:
+
+- **`-c` or `--cache`:** A comma-separated list of preferred caches, with `+` to allow fallback to discovered caches.
+- **`--inplace`:** Write files directly to the destination instead of using temporary files.
```
jhiemstrawisc, line 532: I wasn't aware of this flag so I looked into it, and I think the description should be expanded. It's inherently directional in that these temporary files are only something done on the local filesystem (i.e. a sync from remote to local). The description here also doesn't explain why you'd want to use it or what benefits it gives (atomic file renames on the local filesystem).

jhiemstrawisc (changes requested):

PR #3341, opened 2026-04-13
Minor security fixups

### LINKED ISSUE #3322: Harden security: address low severity findings
A recent security audit identified a handful of low severity issues that don't require immediate fixes but should be addressed in a normal development cycle. 

### Areas to address

- Open redirect
- Registry database layer: one query builds a SQL fragment via string formatting instead of parameterized placeholders. Just remove it since it is a dead code snippet.
- TO BE CONTINUED... (refer to the SWAMP report)

bbockelm:
Various and sundry low-priority issues identified by #3322.  I don't think any of these are currently exploitable or in production code (hence the public PR) but all seem worth fixing.

**IMPORTANT THING WORTH NOTING** for @brianhlin and @williamnswanson: this causes Gin to ignore proxy headers (`X-Forwarded-For` and friends) unless the configuration file explicitly denotes that this component is behind a proxy (otherwise it becomes relatively straightforward to bypassing the IP rate-limiter).  Hence, when put in place, you'll want to enable the corresponding configuration knob (`Server.TrustedProxies`) otherwise a director behind traefik would only recognize the traefik IP address as the source.  There's a `*` option that is equivalent to any IPv4/v6 address; probably sufficient for the sake of the setup on Tiger.

Fixes: #3322

### DIFF
--- a/origin_serve/handlers.go
+++ b/origin_serve/handlers.go
@@ -191,17 +191,23 @@ func extractTokens(r *http.Request) []string {
 
 // getActionFromMethod determines the token scope action from HTTP method.
 //
-// Scope hierarchy (least to most privileged):
+// Scopes by mutation effect:
 //   - read:   no mutation (GET, HEAD, OPTIONS, PROPFIND)
-//   - create: mutation without data loss (PUT, POST, MKCOL, LOCK, UNLOCK, PROPPATCH)
-//   - modify: mutation that may cause data loss (DELETE, COPY, MOVE, and any unknown method)
+//   - create: mutation without data loss -- adds new state but does not
+//     destroy existing data (PUT, POST, MKCOL, LOCK, UNLOCK, PROPPATCH,
+//     and COPY: COPY only adds a resource at the destination; RFC 4918's
+//     Overwrite header would make it destructive but Pelican does not
+//     currently honor it, so today's COPY is purely additive. Promote
+//     to modify if we ever wire up overwrite-aware COPY.)
+//   - modify: mutation that may cause data loss (DELETE, MOVE, and any
+//     unknown method)
 func getActionFromMethod(method string) token_scopes.TokenScope {
 	switch method {
 	case http.MethodGet, http.MethodHead, http.MethodOptions, "PROPFIND":
 		return token_scopes.Wlcg_Storage_Read
-	case http.MethodPut, http.MethodPost, "MKCOL", "LOCK", "UNLOCK", "PROPPATCH":
+	case http.MethodPut, http.MethodPost, "MKCOL", "LOCK", "UNLOCK", "PROPPATCH", "COPY":
 		return token_scopes.Wlcg_Storage_Create
-	case http.MethodDelete, "COPY", "MOVE":
+	case http.MethodDelete, "MOVE":
 		return token_scopes.Wlcg_Storage_Modify
 	default:
 		// Unknown methods default to the most restrictive scope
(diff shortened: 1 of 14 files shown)


origin_serve/handlers.go
```diff
@@ -188,19 +188,23 @@ func extractTokens(r *http.Request) []string {
...
+	case http.MethodPut, http.MethodPost, "MKCOL", "LOCK", "UNLOCK", "PROPPATCH":
 		return token_scopes.Wlcg_Storage_Create
-	case http.MethodDelete:
+	case http.MethodDelete, "COPY", "MOVE":
 		return token_scopes.Wlcg_Storage_Modify
-	case "PROPFIND":
-		return token_scopes.Wlcg_Storage_Read
 	default:
-		return token_scopes.Wlcg_Storage_Read
+		// Unknown methods default to the most restrictive scope
```
jhiemstrawisc, line 213: This seems wrong -- `storage.modify` is the most permissive scope, not the least.

origin_serve/handlers.go
```diff
@@ -188,19 +188,23 @@ func extractTokens(r *http.Request) []string {
...
 func getActionFromMethod(method string) token_scopes.TokenScope {
 	switch method {
-	case http.MethodGet, http.MethodHead:
+	case http.MethodGet, http.MethodHead, http.MethodOptions, "PROPFIND":
 		return token_scopes.Wlcg_Storage_Read
-	case http.MethodPut, http.MethodPost:
+	case http.MethodPut, http.MethodPost, "MKCOL", "LOCK", "UNLOCK", "PROPPATCH":
 		return token_scopes.Wlcg_Storage_Create
-	case http.MethodDelete:
+	case http.MethodDelete, "COPY", "MOVE":
```
jhiemstrawisc: I don't follow the choice to put `COPY` here -- I'd think it's purely additive. Can you elaborate with a comment?

web_ui/ui.go
```diff
@@ -69,6 +70,20 @@ var (
 
 const notFoundFilePath = "frontend/out/404/index.html"
 
+// isSafeRedirectURL checks whether a redirect target is a relative
+// path (no scheme or host), preventing open-redirect attacks.
+func isSafeRedirectURL(rawURL string) bool {
```
jhiemstrawisc, line 87: Can you add unit tests for this?

jhiemstrawisc (changes requested):

origin_serve/handlers.go
bbockelm: ?  Modify allows things like deleting files.... isn't it the least permissive?

origin_serve/handlers.go
bbockelm: Hm - that's a good point. There is an "overwrite mode" that we could implement... but currently don't.  Will fix.

bbockelm: Fixed in the latest rebase.

web_ui/ui.go
bbockelm: Done in the latest rebase.

jhiemstrawisc (approved):
LGTM. Merge when ready!

bbockelm:
@williamnswanson - please keep this one in your mental upgrade notes for 7.26.  Will require a director config change (or anything else running behind traefik).

Similarly, @jhiemstrawisc this should get into the release notes.  I think one cache runs like this potentially.

jhiemstrawisc:
Pinging @turetske, as she's release manager for `v7.26` and has been working on release notes for it.

PR #3343, opened 2026-04-13
Split client/server binaries

### LINKED ISSUE #1765: Don't build client in server binary or servers in client binary
Once #1755 is okay and merged, we'll have two binaries (pelican and pelican-server). This should allow us to start setting up custom build tags so that we don't build client code in the server binary and vice versa.

jhiemstrawisc:
Closes #1765 

This splits the client and server logic into two separate binaries. For each command we export, I decided whether I thought it belonged in the "server" binary (e.g. `pelican-server origin <command>`), the "client" binary (e.g. `pelican object <command>`) or both (e.g. `pelican config <command>`).

I accomplished the split using some light build tag refactoring in the `cmd/` package by leveraging the fact that go only builds dependencies transitively -- that is, if you're building the client and don't compile any of the server-related commands, you won't import any server-only packages and they're kept out of the binary.

Some of the commits involved splitting apart various other dependencies because the client was still transitively importing things like the Web UI, etc.

Using this approach, I shrunk the client binary by ~68%:
```
$ du -sh ./dist/pelican-server_linux_arm64_v8.0/pelican-server 
81M     ./dist/pelican-server_linux_arm64_v8.0/pelican-server

$ du -sh dist/pelican_linux_arm64_v8.0/pelican 
26M     dist/pelican_linux_arm64_v8.0/pelican
```

Here are the sub commands I assigned to each:
```
$ ./pelican-server -h
<snipped>
Available Commands:
  apikey      Manage API keys for server operations
  cache       Operate a Pelican cache service
  config      View and search for configuration parameters
  credentials Interact with the credential configuration file
  director    Launch a Pelican Director
  downtime    Manage server's own downtime periods
  generate    Generate credentials for Pelican server
  help        Help about any command
  key         Manage Pelican issuer keys
  namespace   Work with namespaces
  origin      Operate a Pelican origin service
  registry    Interact with a Pelican registry service
  server      Manage server operations
  token       Interact with tokens used to interact with objects in Pelican
  
$ ./pelican -h
<snipped>
Available Commands:
  client-agent Manage the Pelican client agent server
  config       View and search for configuration parameters
  credentials  Interact with the credential configuration file
  downtime     Manage server's own downtime periods
  help         Help about any command
  job          Manage asynchronous transfer jobs
  namespace    Work with namespaces
  object       Interact with objects in the federation
  plugin       Plugin management for HTCSS
  rclone       Commands for integrating Pelican with rclone
  token        Interact with tokens used to interact with objects in Pelican
```

To build each binary, I updated our goreleaser config to pass the `client` and `server` build tags, respectively. This also lets me get rid of the `lotman` tag, since that can be completely replaced by the `server` build tag. 

Lastly, I tried to convince myself the pelican package imports for the client looked reasonable, because that's where we care most about size:
```
$ go list -tags client -deps ./cmd/ | grep pelicanplatform/pelican
github.com/pelicanplatform/pelican/docs
github.com/pelicanplatform/pelican/byte_rate
github.com/pelicanplatform/pelican/param
github.com/pelicanplatform/pelican/logging
github.com/pelicanplatform/pelican/error_codes
github.com/pelicanplatform/pelican/pelican_url
github.com/pelicanplatform/pelican/utils
github.com/pelicanplatform/pelican/version
github.com/pelicanplatform/pelican/server_structs
github.com/pelicanplatform/pelican/config
github.com/pelicanplatform/pelican/oauth2
github.com/pelicanplatform/pelican/token_scopes
github.com/pelicanplatform/pelican/token
github.com/pelicanplatform/pelican/client
github.com/pelicanplatform/pelican/client_agent/types
github.com/pelicanplatform/pelican/client_agent
github.com/pelicanplatform/pelican/client_agent/apiclient
github.com/pelicanplatform/pelican/client_agent/store
github.com/pelicanplatform/pelican/cmd/config_printer
github.com/pelicanplatform/pelican/utils/registry_discovery
github.com/pelicanplatform/pelican/registry/registry_client
github.com/pelicanplatform/pelican/cmd
```

I think these are all reasonable -- the `server_structs` are used in the client to unmarshal Director responses, and the `registry_client` package is used for the `pelican namespace` commands.


jhiemstrawisc:
@turetske I know this is a big diff that touches a lot of files, but I don't think the majority of it is all that bad because the majority is made of build tag changes or moving functions from one place to another. I think reviewing commit by commit will greatly help with the review.

jhiemstrawisc:
Additionally pinging @brianhlin and @williamnswanson in case either of you want to weigh in on which commands are going to remain in each binary.

.goreleaser.in.yml
brianaydemir, line 103: If we intend to maintain hopes of running a Pelican service on macOS, we should configure GoReleaser to build `pelican-server` for macOS.

jhiemstrawisc:
Here's a diagram explaining the size breakdown per package (using the [gsa](https://github.com/Zxilly/go-size-analyzer) tool @brianaydemir linked me to). One note is that this is the result from building an "unstripped" binary that still has debug symbols, which were needed by the tool to work. In the final binary, the debug symbols won't be present, saving around 9MB of space (26MB instead of 35MB).

As I suspected, one of the larger imports in the client comes from the client agent code's use of sqlite (I think that's where it's coming from, at least). Moving that code into the server binary is a bit more nuanced because of its entanglement with `pelican object` via the `--async` flag. Not saying we can't do it in the future, but this PR is already getting long.

Screenshot 2026-04-14 at 14 59 52

.goreleaser.in.yml
jhiemstrawisc: Addressed in a later commit.

utils/registry_discovery/registry_discovery.go
```diff
@@ -16,7 +16,7 @@
  *
  ***************************************************************/
 
-package server_utils
+package registry_discovery
```
turetske: I don't love this name. Mostly because it makes my brain immediately go to "federation discovery", even though they are unrelated. Maybe something that makes it more clear this is issuer/jwks lookup. Or registry data lookup?

token/token_verify.go
```diff
@@ -399,7 +392,11 @@ func Verify(ctx *gin.Context, authOption AuthOption) (status int, verified bool,
 				return http.StatusOK, true, nil
 			}
 		case APITokenIssuer:
-			if err := checkApiTokenIssuer(token, authOption.Scopes, authOption.AllScopes); err != nil {
+			// This token issuer is ONLY available to server binaries, since it relies on the CheckApiTokenIssuerFunc
```
turetske, line 395: Can you create a follow up issue in order to disentangle this dependency more. I agree that for now, this fix is okay, but I'd like to do this properly. I'm mostly concerned with some coding agent seeing that verification exists and using it for some stupid reason.

cmd/main_test.go
```diff
@@ -1,6 +1,8 @@
+//go:build client && server
```
turetske: Why both? Would only build if both the client and server build tags are set at the same time, which is confusing.

cmd/downtime.go
```diff
@@ -1,6 +1,8 @@
+//go:build client || server
```
turetske: Why is downtime both client and server? Isn't downtime just a server function?

turetske: Also, this applies to downtime_delete, downtime_list, and downtime_create as well (downtime_update is just server)

turetske (changes requested):

brianhlin:
@jhiemstrawisc I think the tool split looks mostly reasonable to me but `downtime` in the client binary strikes me as out of place. I may be incorrectly assuming that the client binary is intended for consumers and producers of data in the federation (i.e., for data transfer) and not for administrators managing their Pelican services.

williamnswanson:
I think the `downtime` commands are functionally just API requests (though it's the same with the `server config` stuff for mutable config)? But I'd say the `downtime` sub-command should just be another section under the existing `server` sub-command.

brianhlin:
>  But I'd say the downtime sub-command should just be another section under the existing server sub-command.

Oh yeah, the `server` command, which looks a little odd with the split binaries: `pelican-server server...`

jhiemstrawisc:
Easy enough -- I can take the downtime commands out of the client binary. @williamnswanson I'm inclined to agree that the `downtime` commands should live under `server`, but I think that's out of scope here so I'm going to keep the current `pelican downtime <blah>` behavior.

utils/registry_discovery/registry_discovery.go
jhiemstrawisc: Renamed to `registry_jwks` to highlight that the package is related to JWKS operations specific to the Registry.

token/token_verify.go
jhiemstrawisc: I'll make a to-do to open the issue once we land this PR 🙂

cmd/downtime.go
jhiemstrawisc: The catch about `server && client` was a good catch -- I hadn't inspected this closely enough before. I was able to simplify some things after noticing that the affected file was using some of pelican's sub commands to test specific behavior related to the `--version` flag. The key observation was that we were doing `pelican --version`, `pelican origin --version` and `pelican origin get --version` (not sure where that last one came from 😅). This caused some entanglement because that's only available in the server binary, so I switched to use the `pelican config` command that's available in both. I'm not sure _how_, but this appears to have shrunk the server binary by an additional ~20MB so it was definitely worth it!

Lastly, I moved the downtime commands to be server-only.

turetske (approved):

PR #3346, opened 2026-04-14
Support multiple Globus exports

### LINKED ISSUE #3163: Globus Origins currently only support one export, but don't assert more aren't given
The XRootD Origin config template we construct for our Globus backend is capable of being handed a list of exports, but it only constructs configuration for whichever export is first in the list. See [here](https://github.com/PelicanPlatform/pelican/blob/365f04393a5b0153f397b587bf08948a26e1a29f/xrootd/resources/xrootd-origin.cfg#L117-L134).

I'll open another issue to support multi-export Globus Origins, but for now we need to fail on start if the Origin is configured with multiple exports. I'm not 100% sure the best place to make this assertion, but some options after light grepping include:
- [Initialization of Globus backend stuff]( https://github.com/PelicanPlatform/pelican/blob/365f04393a5b0153f397b587bf08948a26e1a29f/origin/globus.go#L97)
- I think I'd prefer [here](https://github.com/PelicanPlatform/pelican/blob/365f04393a5b0153f397b587bf08948a26e1a29f/server_utils/origin.go#L598), since it'd be nice for generation of the exports list to fail for multi-export Globus Origins.

### LINKED ISSUE #3164: Add multi-export support for Globus Origins (still single collection)
Right now, generation of the XRootD config for Globus Origins only handles a single export. See [here](https://github.com/PelicanPlatform/pelican/blob/365f04393a5b0153f397b587bf08948a26e1a29f/xrootd/resources/xrootd-origin.cfg#L117-L134).

Issue #3163 will temporarily disable the Origin from starting when multiple exports are provided.

In Pelican's worldview, an Origin should serve a single representation of some object store, and for Globus that means a single "Globus Collection". However, @williamnswanson has asked that we allow namespace mappings into what are effectively different mount points of that collection. For example, if a collection exposes a directory tree:
```
.
└── MyCollection/
    ├── Justin/
    ├── WilliamS/
    └── WilliamJ/
```
it may be desirable to create a namespace for each directory under that tree (for example, to create different auth policies), where each export is given a mount point via a `StoragePrefix`.

This mirrors what we do with the S3 backend, where the Origin can only point to one S3 service URL while being able to map different buckets to different federation prefixes.

Implementing this will require editing the aforementioned XRootD config template to range over exports. If the Globus OFS plugin doesn't support this, we'll want to take inspiration from the S3 plugin, which uses `s3.begin` and `s3.end` to delimit chunks of configuration for each export. See the S3 configuration in the [S3 OFS plugin README](https://github.com/PelicanPlatform/xrootd-s3-http) for more details

whwjiang:
Fixes 3164

Origin-side development to support multiple Globus exports of the same collection. After this commit, users will be able to export multiple paths within the same collection, as such:

```yaml
...
StorageType: "globus"
  SelfTest: false
  DirectorTest: false
  GlobusClientIDFile: "/workspaces/pelican/globus-client-id" # Where your Globus OAuth client ID is stored
  GlobusClientSecretFile: "/workspaces/pelican/globus-client-secret" # Where your Globus OAuth client secret is stored
  Exports:
    - FederationPrefix: /chtc/itb/globus
      # Possible capabilities: ["Reads", "PublicReads", "Writes", "DirectReads", "Listings"]
      # https://docs.pelicanplatform.org/federating-your-data#origin-and-namespace-capabilities
      Capabilities: ["PublicReads", "Reads", "DirectReads", "Writes", "Listings"]
      StoragePrefix: "/whjiang"
      GlobusCollectionID: "ccdab1cf-19d8-4cd8-9833-49a535e207fc" # The Globus collection UUID
      GlobusCollectionName: "Pelican Test Collection" # A human-readable name for this collection
    
    - FederationPrefix: /chtc/itb/globus-wns
      # Possible capabilities: ["Reads", "PublicReads", "Writes", "DirectReads", "Listings"]
      # https://docs.pelicanplatform.org/federating-your-data#origin-and-namespace-capabilities
      Capabilities: ["PublicReads", "Reads", "DirectReads", "Writes", "Listings"]
      StoragePrefix: "/wnswanson"
      GlobusCollectionID: "ccdab1cf-19d8-4cd8-9833-49a535e207fc" # The Globus collection UUID
      GlobusCollectionName: "Pelican Test Collection" # A human-readable name for this collection
...
```

In the origin front-end, the export will now look like this:

image

These changes rely on XRootD plugin changes to actually work.

Note that we will disallow different collections, for now, as it would involve a larger infrastructure/config parsing change.


h2zh:
@whwjiang could you rebase this branch to the latest main? hopefully most CI failures would be gone then.

origin/globus.go
```diff
@@ -75,7 +95,7 @@ const (
 )
 
 var (
-	// An in-memory map-struct to keep Globus collections information with key being the collection UUID.
+	// An in-memory map-struct to keep Globus collections information with key being the federation prefix.
```
h2zh, line 100: This PR refactors `globusExports` from being keyed by `CollectionID` to `FederationPrefix`, but the OAuth callback in globus_client.go:464 was missed. As a result, an error is thrown whenever user activates a Globus origin
```
Error contacting server: 400: Failed to update Globus collection ccdab1cf-19d8-4cd8-9833-49a535e207fc with name Pelican Test Collection: the Globus collection ccdab1cf-19d8-4cd8-9833-49a535e207fc with name Pelican Test Collection does not exist in Pelican
```

origin/globus_test.go
```diff
@@ -0,0 +1,50 @@
...
+	assert.Equal(t, []GlobusExportPathConfig{
+		{FederationPrefix: "/first/ns", StoragePrefix: "/foo"},
+		{FederationPrefix: "/second/ns", StoragePrefix: "/bar"},
+	}, config.Exports)
+}
+
+func TestGetGlobusBackendConfigRejectsMultipleCollections(t *testing.T) {
+	globusExports = map[string]*globusExport{
+		"abc123": {HttpsServer: "https://globus.example"},
+		"def456": {HttpsServer: "https://globus.example"},
```
h2zh, line 40: These two new tests should seed the map with FederationPrefix (not CollectionID)

origin/globus_test.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func TestGetGlobusBackendConfigSingleCollectionMultiExport(t *testing.T) {
+	globusExports = map[string]*globusExport{
+		"abc123": {
+			HttpsServer:       "https://globus.example",
+			TokenFile:         "/tokens/abc123.tok",
+			TransferTokenFile: "/tokens/abc123.transfer.tok",
+		},
+	}
```
h2zh, line 19: I guess the key should be a federation prefix like "/..." instead of "abc123". Also, could you add another export to this map? Since this test is called TestGetGlobusBackendConfigSingleCollectionMultiExport.

h2zh (changes requested):
This PR is a solid step forward toward multi-export Globus origin support! I'll do another round of review once the core functionality is working.

PR #3347, opened 2026-04-14
Fix DB backup

### LINKED ISSUE #3348: Fix DB backup
Integration team found several issues in the database backup feature:

1. MaxCount missing default
2. MaxCount=0 means backups are never cleaned up
3. SQLITE_BUSY error on `VACUUM INTO`
4. ITB-OSDF PelicanPlatform Origin (running v7.25.0-rc.1) has gotten it's sqlite database stuck locked. User can't even log into the WebUI (OIDC or password). There's a `pelican.sqlite-journal` file next to the database that google says is for atomic commits / transactions? Possibly related to a "SQL logic error: cannot start a transaction within a transaction" error been seen in the WebUI yesterday at about the same time.
5. `failed to upsert service name CHTC-ITB-PELICANPLATFORM-ORIGIN in local database: SQL logic error: cannot start a transaction within a transaction`

h2zh:
This PR fixes the DB backup issue found in 7.25 ITB testing phase.

### DIFF
--- a/client_agent/store/store.go
+++ b/client_agent/store/store.go
@@ -30,6 +30,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/pelicanplatform/pelican/client_agent/types"
+	dbutils "github.com/pelicanplatform/pelican/database/utils"
 )
 
 //go:embed migrations/*.sql
@@ -42,9 +43,7 @@ type Store struct {
 
 // NewStore creates a new Store instance and runs migrations
 func NewStore(dbPath string) (*Store, error) {
-	// glebarez/sqlite (modernc.org/sqlite) requires pragmas in `_pragma=name(value)`
-	// form; the mattn-style `_name=value` shorthand is silently ignored.
-	db, err := sql.Open("sqlite", dbPath+"?_pragma=journal_mode(WAL)&_pragma=busy_timeout(5000)")
+	db, err := sql.Open("sqlite", dbutils.SQLiteDSN(dbPath))
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to open database")
 	}
--- a/client_agent/store/store.go
+++ b/client_agent/store/store.go
@@ -42,9 +42,8 @@ type Store struct {
 
 // NewStore creates a new Store instance and runs migrations
 func NewStore(dbPath string) (*Store, error) {
-	// Open SQLite database. glebarez/sqlite (modernc.org/sqlite) requires pragmas
-	// in `_pragma=name(value)` form; the mattn-style `_name=value` shorthand is
-	// silently ignored, which silently leaves the DB in rollback-journal mode.
+	// glebarez/sqlite (modernc.org/sqlite) requires pragmas in `_pragma=name(value)`
+	// form; the mattn-style `_name=value` shorthand is silently ignored.
 	db, err := sql.Open("sqlite", dbPath+"?_pragma=journal_mode(WAL)&_pragma=busy_timeout(5000)")
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to open database")
--- a/client_agent/store/store.go
+++ b/client_agent/store/store.go
@@ -42,8 +42,10 @@ type Store struct {
 
 // NewStore creates a new Store instance and runs migrations
 func NewStore(dbPath string) (*Store, error) {
-	// Open SQLite database
-	db, err := sql.Open("sqlite", dbPath+"?_journal_mode=WAL&_busy_timeout=5000")
+	// Open SQLite database. glebarez/sqlite (modernc.org/sqlite) requires pragmas
+	// in `_pragma=name(value)` form; the mattn-style `_name=value` shorthand is
+	// silently ignored, which silently leaves the DB in rollback-journal mode.
+	db, err := sql.Open("sqlite", dbPath+"?_pragma=journal_mode(WAL)&_pragma=busy_timeout(5000)")
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to open database")
 	}
--- a/database/backup.go
+++ b/database/backup.go
@@ -54,6 +54,10 @@ import (
 	"github.com/pelicanplatform/pelican/version"
 )
 
+// backupMaxRetries is the number of times to retry a backup snapshot if it
+// fails with a transient error (e.g. SQLITE_BUSY).
+const backupMaxRetries = 3
+
 const (
 	backupFilePrefix = "pelican-db-backup-"
 	backupFileExt    = ".bak"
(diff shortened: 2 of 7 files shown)


bbockelm:
@h2zh - were you able to reproduce the SQL_BUSY issue?  How confident are you that a simple retry will solve it?

h2zh:
@bbockelm I'm not able to reproduce SQLITE_BUSY locally, probably because WAL mode (set in `InitSQLiteDB`) makes it hard to trigger outside of heavy production load. 

Building on your feedback, I realize we should not only solely on a "simple retry" alone. The new commit raises `busy_timeout` from 5s to 30s on the backup connection before running `VACUUM INTO`, giving it 6x more headroom to wait out long-running write transactions. The retry loop (3 attempts with backoff) is a second layer on top of that. Together, a backup would need to be blocked for over 90s of cumulative wall time to actually fail, which is far beyond any realistic transaction duration in Pelican.

database/backup.go
```diff
@@ -410,10 +414,53 @@ func createBackup(ctx context.Context) error {
...
+	defer sqlConn.Close()
+
+	// Raise busy_timeout to 30 s for the backup operation so VACUUM INTO can
+	// wait out long-running write transactions (the pool default is 5 s).
+	if _, err := sqlConn.ExecContext(ctx, "PRAGMA busy_timeout = 30000"); err != nil {
+		return errors.Wrap(err, "failed to set busy_timeout for backup")
+	}
+	defer func() {
+		// Reset to the pool's default before returning the connection.
+		if _, err := sqlConn.ExecContext(context.Background(), "PRAGMA busy_timeout = 5000"); err != nil {
```
jhiemstrawisc: Is this default of 5s defined somewhere else such that the two values could become out of sync? I'm wondering whether this should be a variable instead of a magic number.

database/backup.go
```diff
@@ -410,10 +414,53 @@ func createBackup(ctx context.Context) error {
...
+				attempt+1, backupMaxRetries, vacuumErr, backoff)
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(backoff):
+			}
+			// Clean up any partial file from the previous attempt.
+			os.Remove(vacuumPath)
+		}
+		_, vacuumErr = sqlConn.ExecContext(ctx, "VACUUM INTO '"+escapedPath+"'")
```
jhiemstrawisc: Is it true that this can only return a vacuum error? I'd imagine there are a host of other reasons the command execution might error, in which case you might end up retrying the backup in cases when the error is not transient. Is there a way to guard on `if errors.As(vacuumErr, <sqlite error type>) && <sqlite error type>.Err == <sqlite error type>.ErrBusy {// --> retry }`?

database/backup.go
```diff
@@ -410,10 +414,53 @@ func createBackup(ctx context.Context) error {
...
+
+	// Retry up to backupMaxRetries times with backoff to handle transient
+	// SQLITE_BUSY errors when concurrent writes hold locks beyond even the
+	// extended busy_timeout.
+	var vacuumErr error
+	for attempt := 0; attempt < backupMaxRetries; attempt++ {
+		if attempt > 0 {
+			backoff := time.Duration(attempt) * 5 * time.Second
+			log.Warnf("Database backup attempt %d/%d failed: %v; retrying in %s",
+				attempt+1, backupMaxRetries, vacuumErr, backoff)
```
jhiemstrawisc: Can you reformat the logic around this log statement so that we log the first failure as well? Right now we'll only see `2/3` and `3/3` because of the `if attempt > 0` guard.

jhiemstrawisc (dismissed):
A few minor nits, otherwise I think it looks okay. Admittedly, I don't know much about sqlite database management, so my assessment of "is this approach reasonable to solve the problem as described" involved asking Claude a lot of questions.

database/backup.go
h2zh: 5s is a common default for `busy_timeout`. Since we are pivoting to enforcing the write-ahead logging (WAL), which should resolve the long lasting lock on the DB by the `VACUUM INTO` reader. I plan to stick with the default instead of increasing it.
In rollback mode (the default journal mode Pelican unintentionally fallback to), every write takes an EXCLUSIVE lock on the whole DB file, which blocks readers too. That's why one slow VACUUM INTO freezes the entire web UI. In WAL mode, readers would sail past the backup.
See commit [f214d11](https://github.com/PelicanPlatform/pelican/pull/3347/commits/f214d11ec3e02f73f0387e86b1842be11d7b3499)

database/backup.go
h2zh: Good point. Since this retry logics is unnecessary in write-ahead logging (WAL) mode, I'll not address this comment.

database/backup.go
h2zh: That would be a nice improvement! Since this snippet is removed now, I won't take time to find this error type (maybe not in the sqlite package we use?)

database/utils/utils.go
```diff
@@ -62,6 +63,26 @@ func InitSQLiteDB(dbPath string) (*gorm.DB, error) {
...
+	// without COMMIT/ROLLBACK) poisons one connection; the next caller to pull that
+	// connection from the pool fails with "cannot start a transaction within a
+	// transaction" and, because the driver still holds the DB handle, the hot
+	// rollback journal never clears -- the whole DB appears stuck.
+	//
+	// Pinning MaxOpenConns to 1 serializes writes through a single connection, so
+	// a leaked transaction is immediately visible to its own caller instead of
+	// being deferred onto an unrelated code path, and the connection is reset
+	// cleanly by GORM's Transaction machinery on the next use. Readers are
+	// unaffected at our scale (this DB is metadata-sized) and _busy_timeout=5000
```
brianaydemir: I think we should replace `_busy_timeout=5000` with plain English, instead of trying to refer to the query parameter in the DSN.

database/utils/utils.go
```diff
@@ -33,7 +33,8 @@ func InitSQLiteDB(dbPath string) (*gorm.DB, error) {
 		dbPath += ".sqlite"
 	}
 
-	dbName := dbPath + "?_busy_timeout=5000&_journal_mode=WAL&_pragma=foreign_keys(1)"
+	// glebarez/sqlite wraps modernc.org/sqlite, whose DSN parser only honors `_pragma=name(value)` form.
```
brianaydemir: For the sake of consistency, I'd take the comment in `server_utils/db.go` and copy it here.

client_agent/store/store.go
```diff
@@ -42,8 +42,10 @@ type Store struct {
 
 // NewStore creates a new Store instance and runs migrations
 func NewStore(dbPath string) (*Store, error) {
-	// Open SQLite database
-	db, err := sql.Open("sqlite", dbPath+"?_journal_mode=WAL&_busy_timeout=5000")
+	// Open SQLite database. glebarez/sqlite (modernc.org/sqlite) requires pragmas
+	// in `_pragma=name(value)` form; the mattn-style `_name=value` shorthand is
+	// silently ignored, which silently leaves the DB in rollback-journal mode.
```
brianaydemir: For the sake of consistency, I'd take the comment in `server_utils/db.go` and copy it here.

brianaydemir (changes requested):

database/utils/utils.go
h2zh: Sounds ligit.

brianaydemir (approved):
For posterity:

The original issue surfaced, in part, due to the database being stored on a volume type where IO ops were known to suffer from high-latencies. It also doesn't help that the SQLite driver does not implement the API primitives that `database/sql` uses to reset connections before re-using them.

As a result, I think it's still possible for the connection pool to end up with a poisoned connection, but at least:

1. Reads shouldn't be affected, due to connections consistently using WAL, as originally intended.

2. Because there's only one open connection in the pool, if it happens to be poisoned, transactions will fail consistently instead of haphazardly — a boon for debugging.

turetske (approved):

bbockelm:
Two comments (not blocking the merge):

1.  We now have the DB connection parameters in 3 strings across the code base.  That seems like it'll bite us in the future.
2.  We assume going to one connection is OK because the load is minimal ... now.  I can see this being an unpleasant surprise in the future.

@h2zh - I don't want to change (2) currently because that's an explicit tradeoff.  (1) may be worth fixing if you have a few moments.

h2zh:
I handled (1) duplicated DSN and removed (2) max connection constraint in this PR, because the DSN string correction is a more fundamental fix, which makes max connection constraint unnecessary (unless the integration tests suggest otherwise). @bbockelm

PR #3350, opened 2026-04-14
Fix xrootd.stats metric parsing

### LINKED ISSUE #3349: xrdcl stats parser silently drops error, pause_duration, server_timeout, and client_timeout metrics
There's a small parsing bug in our XRootD Stats parser that prevents various metrics from making it to Prometheus.

The parser uses `strings.SplitN(key, "_", 4)` to split keys like `http_HEAD_200_server_timeout` into [http, HEAD, 200, server_timeout]. Because SplitN with n=4 stops after 3 splits, `parts[3]` retains its underscores intact. That is, we get "server_timeout", not "server". However, the switch cases [here](https://github.com/PelicanPlatform/pelican/blob/cd759bfb8f6861450ca3856eaa7ca042a481aa8d/metrics/xrootd_metrics.go#L2445-L2461) were written as if the underscores had been stripped.

jhiemstrawisc:
Closes #3349

Claude noticed a parsing bug while I was going over how to construct queries for the various metrics in `xrootd.stats`. This fix is pretty simple, and I glanced over the new tests Claude wrote (which all seem reasonable and well-written). The test I'm most happy to see is the parsing of the actual production cache stats file, which convinces me the bug is fixed. I verified by hand that the test fails in the previous case.

### DIFF
--- a/metrics/xrootd_metrics.go
+++ b/metrics/xrootd_metrics.go
@@ -2430,32 +2430,30 @@ func handleXrdcurlstatsPacket(stats []byte) error {
 
 		if strings.HasPrefix(key, "http_") {
 			parts := strings.SplitN(key, "_", 4)
-			if len(parts) == 3 { // http_VERB_field
+			if len(parts) == 3 { // http_VERB_field (e.g. http_HEAD_started)
 				verb, field := parts[1], parts[2]
 				switch field {
 				case "started":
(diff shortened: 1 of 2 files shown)


h2zh (approved):
LGTM. I can tell this PR manages to parse the payload correctly from the `/metrics` API endpoint.

PR #3357, opened 2026-04-15
Issue 3095

### LINKED ISSUE #3095: Origins disabling Director.Tests should not show up red
When origins disable director tests (which much occur for S3 and Globus) - we report the origin as being in a red state. This should not be the case. Maybe mark the origin test status as unknown because tests are disabled, but it shouldn't show up as critical to the admin or on the director status page.

However, I believe that caches should still be considered critical. We need to disincentivize allowing caches that join the federation to hide their status by disallowing the director to test them because otherwise we cannot tell if it's good cache and should rightfully deprioritize them.

turetske:
This will close #3095 

Origins will now no longer read as critical or be marked red in the UI when they have director health tests disabled.

Cache's however, will.

When testing this, in order to ensure no other errors are showing up, I recommend setting

```
  EnablePrometheus: false
```

If the origin does still how up red, look at the other health metrics to ensure there isn't a warning or other status beyond the director health tests causing issues using: 

```
curl -sk https://localhost:8446/api/v1.0/metrics/health | python3 -m json.tool
```

Also assigning Cannon because there are some frontend UI changes.

### DIFF
--- a/client_agent/store/store.go
+++ b/client_agent/store/store.go
@@ -95,6 +95,10 @@ func (s *Store) runMigrations() error {
 // Close closes the database connection
 func (s *Store) Close() error {
 	if s.db != nil {
+		// Checkpoint and truncate the WAL before closing so that the -wal and -shm
+		// sidecar files are removed. Without this, os.RemoveAll (e.g. t.TempDir cleanup)
+		// fails because those files are still present on disk.
+		_, _ = s.db.Exec("PRAGMA wal_checkpoint(TRUNCATE)")
 		return s.db.Close()
 	}
 	return nil
--- a/director/cache_ads.go
+++ b/director/cache_ads.go
@@ -358,8 +358,21 @@ func recordAd(ctx context.Context, sAd server_structs.ServerAd, namespaceAds *[]
 	}
 
 	if ad.DisableDirectorTest {
-		log.Debugf("%s server %s at %s has DisableDirectorTest set. Skip health test for this server.", ad.Type, ad.Name, ad.URL.String())
-		return
+		log.Debugf("%s.DirectorTest is set to false for %s server %s at %s. Skipping health tests.", ad.Type, ad.Type, ad.Name, ad.URL.String())
+		// Clean up any stale health test utilities from a previous registration
+		// where tests may have been enabled. This ensures advertisementToServerResponse()
+		// correctly reports HealthStatusDisabled instead of using a stale status.
+		func() {
+			healthTestUtilsMutex.Lock()
+			defer healthTestUtilsMutex.Unlock()
+			if existingUtil, ok := healthTestUtils[ad.URL.String()]; ok {
+				if existingUtil.Cancel != nil {
+					existingUtil.Cancel()
+				}
+				delete(healthTestUtils, ad.URL.String())
+			}
+		}()
+		return sAd
 	}
 
 	// Prepare and launch the director file transfer tests to the origins/caches if it's not from the topology AND it's not already been registered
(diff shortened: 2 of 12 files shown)


jhiemstrawisc (changes requested):
Apologies for the delay in providing a review, although we discussed some of this in person last week.

I think the approach here could be modified to simplify things and land on cleaner systems principles. I haven't verified all the details by doing my own source diving yet, so if I'm flat out wrong, we can take another stab.

If I recall correctly, the highlights a server's display card using the server's status, and when Director tests are disabled, some component of the server's status is placed in the `warning` or `error` state. This diff takes the approach of figuring out when these tests are disabled and ignoring the warning status, but I think the approach I'd prefer is to avoid assigning the warning status in the first place when the administrator explicitly turns the feature off. In that case, there's nothing to warn about because the admin made a choice and we're only doing what they asked.

I do partially agree with your instinct to penalize caches when they disable director tests since that might affect their reliability and thus the broader federation, but I think this needs to be a "policy" that can be configured by a federation operator, not hard coded into the redirection logic. So rather than putting the penalty in the EWMA calculation (which should be reserved for temporal load-based monitoring), you can create a new Director knob like `Director.RequireDirectorTests` that, when enabled, uses [cache predicate filtering](https://github.com/PelicanPlatform/pelican/blob/5ff1a8c6eadf7653a02f5bf3764bf0aaed6e1aa7/director/sort.go#L493-L505) to remove these caches from consideration.

Unfortunately, this method is a hammer -- it would be very difficult to penalize caches by a "specific amount" because even small changes in any weight value can produce large swings in sorting. The only way to accomplish something like that is by adjusting weights in the context of the weights' _distribution_ , which is a harder thing to do and a big part of the reason I've resisted some calls to just toss adjustment scalars into the mix.

This approach also raises the question of how a policy enacted at the Director is communicated back to the cache admins. If we filter someone's cache because they disable these tests, the Cache needs a way to ask about its status from the Director so we can display what's happening in its UI. This is generally true about all federation-level policies that might affect a server, so this in itself should be a carefully designed component. While the Director does send some information back to servers via its director test response, that can't really be used if the feature is turned off for the server.

Thoughts?

turetske:
> 
> Thoughts?

@jhiemstrawisc I agree with your notes. I think there are two options for the cache learning why it's being filtered.

1. Adding an API endpoint at the director that the cache can query to determine its status and why it's being filtered. This seems conceptually the correct answer, but it would require more work (auth, polling, etc.)
2. The cache reads the director's server listing and parses it. Pro: the end point already exists. The con here is that API is for the UI, not a server.

client_agent/store/store.go
```diff
@@ -95,6 +95,10 @@ func (s *Store) runMigrations() error {
 // Close closes the database connection
 func (s *Store) Close() error {
 	if s.db != nil {
+		// Checkpoint and truncate the WAL before closing so that the -wal and -shm
+		// sidecar files are removed. Without this, os.RemoveAll (e.g. t.TempDir cleanup)
+		// fails because those files are still present on disk.
+		_, _ = s.db.Exec("PRAGMA wal_checkpoint(TRUNCATE)")
```
jhiemstrawisc, line 101: I still have some more testing to do here -- but why is this included?

bbockelm: This was a bit of shotgun debugging -- unit tests were failing because the RemoveAll on Windows failed.  That occurs if a file handle is still open and I was seeing the WAL get the blame.

However, in subsequent work, I bet the actual issue was shutdown ordering - make sure the shutdown of the DB occurred before the directory cleanup.

I suspect this is now safe to remove.

PR #3358, opened 2026-04-16
Log the real advertisement response body

### LINKED ISSUE #3359: Fix error message in advertisement
Fix error message in advertisement to log the correct body.

h2zh:
Log the real advertisement response body, instead of the request body


jhiemstrawisc (approved):
Short and sweet, and a good catch!

PR #3365, opened 2026-04-17
Update Go version in generate target to 1.25

CannonLock:


h2zh (approved):
LGTM. The failed CI is caused by an outdated test object we no longer use for testing in newer versions.

PR #3367, opened 2026-04-17
Generate defaults from parameters.yaml & Cleanup config tool

### LINKED ISSUE #2451: `pelican config` commands don't effectively work with existing packaging
Pelican Version: 
All of them since we started distributing RPM packaging

Pelican Service:
- [ ] Client
- [ ] Plugin
- [X] Registry
- [X] Director
- [X] Origin
- [X] Cache

**Describe the bug**
Pelican client commands assume the default Pelican config lives at `/etc/pelican/pelican.yaml` but our RPM packaging sets up separate configurations per service (see the `{pelican,osdf}-*.yaml` files [here](https://github.com/PelicanPlatform/pelican/tree/464480154c80c76e829f8b43d8b6ffa55a92a236/systemd)).

This means that administrators have to run `pelican --config <service name> config dump` to get the configuration, which most folks won't get right on the first try since we obfuscate the config file used by each service through systemd. We had a support issue where we asked for the configuration and had to have many back and forths to figure out the appropriate invocation.

**To Reproduce**

1. Start a container, e.g. `podman run --rm -it hub.opensciencegrid.org/opensciencegrid/software-base:24-el9-release /bin/bash`
2. Install the OSDF origin package with `yum install -y osdf-origin`
3. Modify configuration in `/etc/pelican/osdf-origin.yaml`
4. Run `pelican config dump` and compare it vs `pelican --config /etc/pelican/osdf-origin.yaml config dump`

**Expected behavior**
I would expect Pelican client commands to use the config for the running service(s). Whether this mean unifying on a single config file entrypoint or separating out config commands (e.g., `pelican cache config dump`) depends on how we view Pelican configuration:

1. Pelican is a single service and caches, directors, origins, and registries are different daemons under this service, like HTCondor. This would mean that we have unified configuration and that all parameters work well in single vs multi daemon mode.
2. Pelican caches, directors, origins, and registries are separate services, analogous to the HTCondor vs HTCondor-CE split. This means completely independent configs and tooling to interact with one service vs another.

At this point, we should also be able to unify the config entrypoints for `osdf-origin` vs `pelican-origin` and just include things like the `Federation.DiscoveryUrl` config in the packaging of the former. This should ease some of the pain but ultimately we need to decide on route (1) or (2) for the long-term.

### LINKED ISSUE #2517: `pelican config get` does not return full paths
While reviewing another PR related to `pelican config get`, I noticed that invoking the command against params whose values should be full filepaths only give you the leaf of the path:
```
$ pelican config get issuerkeysdirectory
issuerkeysdirectory: "issuer-keys"
```

We should be returning the full path.

### LINKED ISSUE #2518: `pelican config get` not working against defaults defined via `defaults.yaml`
Some of our default configuration comes from `config/config.go` and some of it comes from some combination of `config/resources/{defaults,osdf}.yaml`.

It appears that the defaults defined in `defaults.yaml` don't display correctly through the `pelican config get` command.

For example, in a clean environment with no config file:
```
$ pelican config get metricauthorization 
monitoring.metricauthorization: false
```

However, this default is set to true in `defaults.yaml`. Note that the config dump that happens at server startup shows the correct value, so I don't think this is an issue with _setting_ the default, only reporting it.

### LINKED ISSUE #2594: Add ability to `pelican config` to indicate source of config param
Pelican Service:
- [X] Client
- [ ] Plugin
- [ ] Registry
- [ ] Director
- [ ] Origin
- [ ] Cache

When troubleshooting Pelican services, the `pelican config` tool is very useful. However with the various ways for setting a given config param, admins can easily get into a position where they know config is wrong but they don't know where they should change it (env var? web interface? one of many config.d dirs?).

An option to `pelican config get` that displays the source file / env var / default that defines a given parameter or set of parameters. 

**Additional context**
`condor_config_val -verbose` does something this:

```
$ condor_config_val -verbose VM_KILLING_TIMEOUT
VM_KILLING_TIMEOUT = 60
 # at: <Default>
 # raw: VM_KILLING_TIMEOUT = 60

$ condor_config_val -verbose UID_DOMAIN
UID_DOMAIN = chtc.wisc.edu
 # at: /etc/condor/condor_config.global, line 54
 # raw: UID_DOMAIN = chtc.wisc.edu

$ condor_config_val -verbose VM_KILLING_TIMEOUT UID_DOMAIN
VM_KILLING_TIMEOUT = 60
 # at: <Default>
 # raw: VM_KILLING_TIMEOUT = 60

UID_DOMAIN = chtc.wisc.edu
 # at: /etc/condor/condor_config.global, line 54
 # raw: UID_DOMAIN = chtc.wisc.edu

$ _condor_VM_KILLING_TIMEOUT=0 condor_config_val -verbose VM_KILLING_TIMEOUT
VM_KILLING_TIMEOUT = 0
 # at: <Environment>
 # raw: VM_KILLING_TIMEOUT = 0
```

### LINKED ISSUE #3033: `pelican-server config {summary,dump}` does not reflect the contents of the file specified by `Server.WebConfigFile`
## Summary

When Pelican's web interface is used to make configuration changes to the underlying service, those configuration changes are recorded in the file specified by `Server.WebConfigFile`.

When a user runs `pelican config {summary,dump}` or `pelican-server config {summary,dump}`, they expect to see the configuration that the corresponding `pelican`/`pelican-server` command would use when run with another sub-command.

Unfortunately, this is not the case. In particular, `pelican-server config` seems to ignore the contents of the file specified by `Server.WebConfigFile`. This makes it challenging to determine exactly what configuration a Pelican service might be using.

## Details

It would not be unreasonable to assume that if the user is invoking a binary named `pelican-server`, then they are interested in Pelican's server components and their configuration, not the client's.

It's also worth noting, just for context, that while we still do not yet have [a clean split between Pelican's client and server binaries](https://github.com/PelicanPlatform/pelican/issues/1765), at least Pelican's released container images all install and use only a binary named `pelican-server` (as of [#2725](https://github.com/PelicanPlatform/pelican/pull/2725)).

### LINKED ISSUE #3162: Incorrect log level derivations in `pelican{-server} config`
This comes from some futzing in the CHTC_PELICAN_CACHE, which was running `v7.23.0` at the time.

It seems like there's something strange going on with the log level hierarchy, wherein setting `Logging.Level: foo` should cause all unpinned log levels under `Logging.Cache` and `Logging.Origin` to inherit the value `foo`. For some reason I can't figure out, it doesn't seem like that inheritance is happening.

Here's the output from the `Logging` section of  `pelican-server config dump`:
```
Logging:
    Cache:
        Http: info
        Lotman: info
        Ofs: info
        Pfc: debug
        Pss: debug
        PssSetOpt: debug
        Scitokens: fatal
        Xrd: info
        Xrootd: info
    Client:
        ProgressInterval: 1m0s
    DisableProgressBars: false
    Level: debug
    LogLocation: /var/log/pelican/pelican.log
    Origin:
        Cms: info
        Http: info
        Ofs: info
        Oss: info
        Scitokens: fatal
        Xrd: info
        Xrootd: info
```

Note that `Logging.Cache.{Pfc,Pss,PssSetOpt}` were all pinned in config, but none of the other values were. Additionally, `Logging.{Cache,Origin}.Scitokens` is intentionally cranked down, so it's safe to ignore those right now. Despite `Logging.Level: debug`, none of the other knobs are inheriting the correct value.

More concerning, the output from `pelican-server config dump` disagrees with what I can see via the web UI where the values appear to have the correct inheritance:
Image

So which of these two is correct? To answer that I looked at the generated XRootD configuration and then compared it against the Pelican-->XRootD log mapping defined [here](https://github.com/PelicanPlatform/pelican/blob/365f04393a5b0153f397b587bf08948a26e1a29f/xrootd/xrootd_config.go#L1471):
```
$ cat /run/pelican/xrootd/cache/xrootd.cfg
<abridged>
pss.debug
pfc.trace debug
pss.setopt DebugLevel 4
pss.trace on
ofs.trace debug
xrd.trace debug
xrootd.trace debug emsg login stall redirect request stall
scitokens.trace debug info warning error
http.trace all
```

From this, the answer is that the Web UI is correct, while `pelican-server config dump` lies.

I'll note that I manually traced through the chained config files, and made sure `pelican-server config dump` was pointed at the right entrypoint. I also confirmed the web UI config file at `/etc/pelican/web-config.yaml` was empty, and saw no evidence this file location was overridden by other yaml config.

I attempted to poke at this locally by checking out the `v7.23.0` tag, but I couldn't reproduce.

### LINKED ISSUE #3190: `pelican config get` should not consult the Discovery URL by default
Getting the config should by default not require any running services such as the discovery metadata endpoint or the Pelican service on the host itself.

jhiemstrawisc:
Apologies for this large diff -- I was very motivated to get the config tool in the shape that I think it should be, and I had to pull on a few threads that expanded this PR beyond what I'd call a reasonable PR size. To assist the reviewer(s), I'm providing a design roadmap and the full set of manual testing I've done. I also recommend going commit-by-commit with this one, because the commits have little to no overlap.

This description and its contents were entirely written by me as a guarantee I vetted the things I'm claiming. I'll also point out that while I used Claude to generate a lot of the code, I spent considerable time with the actual diff and reviewed everything myself before opening this. That's not to say I didn't miss anything or make any bad choices!

## Design Principles in this PR
My initial goal with this PR was to fix a slew of issues related to the config tool because a) I suspected a general cleanup of the principles underpinning the tool would knock out lots of problems and b) fixing them one at a time would cause a lot of conflicts. The issues I targeted and believe I solved include:
- Closes #3190: The config tool shouldn't perform metadata discovery by default
- Closes #3162: The config tool gives incorrect information for sub-system logging levels
- Closes #3033: The config tool doesn't detect config provided via the Web UI
- Closes #2594: The config tool should tell callers the source of a param value (env var, which config file, etc)
- Closes #2518: Defaults defined in `config/resources/*.yaml` don't get picked up by the tool
- Closes #2451: The tool doesn't work natively with how we package RPM service configurations
- Closes #2517: The tool doesn't provide full paths when getting some configured values

The number and scope of these bugs _strongly_ indicated structural design issues with the tool itself. In particular, when I was assessing how to fix these, I determine I needed to accomplish the following:
1. Simplify config initialization entrypoints so that both the config tool and the actual initialization/startup sequence agreed on the state of things.
2. Get rid of the headache of having so many different places/ways to define config (holy cow I'm tired of seeing PRs to fix missing default initializations) -- the parameters file _must_ be programmatically turned into defaults!
3. Implement some type of provenance tracking on top of viper, since this is something the viper team seems unmotivated to do despite user interest. While we want to get rid of viper long term, I explicitly avoided that in this set of work.

### Defaults loading from `docs/parameters.yaml`
One core theme in this PR was having all configuration defaults resolved from `parameters.yaml` so that someone who creates a new config param can't forget to actually set the default in the code. I accomplished this by programmatically generating code for the default setters via parsing the parameters file.

The tricky part in the generator was dependency ordering -- some defaults inherit from other params, e.g. `Server.ExternalWebUrl` has a default of `https://${Server.Hostname}:${Server.WebPort}`. The first step in resolving these was to impose a structure on how we reference these nested params. You'll notice I chose a bash-like syntax, so to reference `Some.Param` in the default of `Other.Param`, you'd do:
```yaml
name: Other.Param
default: ${Some.Param}
``` 
to indicate that. Similarly for environment variables, you reference the env var using `$` and traditional env var capitalization:
```yaml
name: Other.Param
default: $ALL_CAPS_ENV_VAR
```

The next step is a topo sort for the actual defaults initialization: `Server.ExternalWebUrl` has a default of `https://${Server.Hostname}:${Server.WebPort}`, so `Server.Hostname` and `Server.WebPort` have to be set first. Defaults initialization can't be ordered solely by direct references, because a direct reference count doesn't account for transitive dependencies. For example, you could imagine a case where the default for `Some.Param` references only `${Server.ExternalWebUrl}`, which in turn requires two other defaults to be set first. For this, Claude decided to use Kahn's algorithm, which seems appropriate based on what I read about it. Here I mostly relied on the derived tests to make sure I thought things were reasonable. Note that this sort only happens at compile time, not run time. Also, this generated code is a large part of the diff. It's probably reasonable to check a few examples from it and then skip the other 900 lines!

Lastly, the parameters file has different types of defaults, e.g. `osdf_default`, `root_default`, `server_default`, e.g. These are called "tiers" in the new code, and they have an order precedence just in case we ever need to layer them.

### Provenance tracking
The source tracking code uses a global singleton that maps each lowercase viper key to a struct that holding metadata about where that config is defined. Overall it uses a diff-based approach for config files where we take a snapshot of the current config before and after merging a new layer of configuration, and we update the source for any keys that differ between the before/after snapshot. Defaults and environment variables use a simpler bulk approach at the start since these aren't layered in the same way.

## Bugs Fixed
I'll use this section to provide evidence the aforementioned bugs have been fixed.

### No metadata discovery by default
I start with a config pointing to my local metadata discovery endpoint:
```
Federation:
  DiscoveryUrl: https://21e78195fb87:4443
```
and run a `pelican config dump` (cutting out the extra stuff in the output):
```
$ pelican config dump
Federation:
    BrokerUrl: ""
    DirectorUrl: ""
    DiscoveryUrl: https://21e78195fb87:4443
    JwkUrl: ""
    RegistryUrl: ""
```
None of the other federation config params are filled in, and no debug logs about discovery appear, as expected.

I can then run again with the new `--with-discovery` and see everything resolved:
```
$ pelican config dump --with-discovery
Federation:
    BrokerUrl: ""
    DirectorUrl: https://21e78195fb87:8445
    DiscoveryUrl: https://21e78195fb87:4443
    JwkUrl: https://21e78195fb87:8445/.well-known/issuer.jwks
    RegistryUrl: https://21e78195fb87:8446

DEBUG[2026-04-17T18:18:55Z] Configured federation URL: https://21e78195fb87:4443 
DEBUG[2026-04-17T18:18:55Z] Attempting to discover federation services via URL: https://21e78195fb87:4443 
DEBUG[2026-04-17T18:18:55Z] Performing federation service discovery for against https://21e78195fb87:4443/.well-known/pelican-configuration 
DEBUG[2026-04-17T18:18:56Z] Federation service discovery resulted in director URL https://21e78195fb87:8445 
DEBUG[2026-04-17T18:18:56Z] Federation service discovery resulted in registry URL https://21e78195fb87:8446 
DEBUG[2026-04-17T18:18:56Z] Federation service discovery resulted in JWKS URL https://21e78195fb87:8445/.well-known/issuer.jwks 
DEBUG[2026-04-17T18:18:56Z] Federation service discovery resulted in broker URL  
DEBUG[2026-04-17T18:18:56Z] Setting global director url to https://21e78195fb87:8445 
DEBUG[2026-04-17T18:18:56Z] Setting global registry url to https://21e78195fb87:8446 
DEBUG[2026-04-17T18:18:56Z] Setting global jwks url to https://21e78195fb87:8445/.well-known/issuer.jwks 
```
Discovery was triggered by the flag

### Correct subsystem logging levels
This bug occurred because the subsystem logging level mapper that handled inheritance wasn't triggered by the config tool, such that setting `Logging.Level: debug` wouldn't set the subsystem levels to `debug` for the config tool.

Using this configuration:
```
Logging:
  Level: debug
  Origin:
    Scitokens: trace
```
I now get:
```
$ pelican config dump
Logging:
    Cache:
        Http: debug
        Lotman: debug
        Ofs: debug
        Pfc: debug
        Pss: debug
        PssSetOpt: debug
        Scitokens: debug
        Xrd: debug
        Xrootd: debug
    Client:
        ProgressInterval: 1m0s
    DisableProgressBars: false
    Level: debug
    LogLocation: ""
    Origin:
        Cms: debug
        Http: debug
        Ofs: debug
        Oss: debug
        Scitokens: trace
        Xrd: debug
        Xrootd: debug
```
Inheritance is correct in that `debug` maps downward except for the pinned `Logging.Origin.Scitokens` value of `trace`

## Web UI configuration isn't detected
This was also fixed by tightening up the entrypoints for config initialization.

To test, I set this in my `/etc/pelican/pelican.yaml`:
```
Logging:
  Level: debug
  Origin:
    Scitokens: info
```
and in my `/etc/pelican/web-config.yaml:
```
Logging:
  Level: trace
```

This specific config file setup was chosen because it's a nuanced challenge: the Web UI must be ingested during config initialization before the subsystem levels are applied, meaning the `Logging.Level` from the web config should override the base debug and be applied downward _except_ for the pinned scitokens config. Running the tool, I get the correct answer:
```
Logging:
    Cache:
        Http: trace
        Lotman: trace
        Ofs: trace
        Pfc: trace
        Pss: trace
        PssSetOpt: trace
        Scitokens: trace
        Xrd: trace
        Xrootd: trace
    Client:
        ProgressInterval: 1m0s
    DisableProgressBars: false
    Level: trace
    LogLocation: ""
    Origin:
        Cms: trace
        Http: trace
        Ofs: trace
        Oss: trace
        Scitokens: info
        Xrd: trace
        Xrootd: trace
```

### Defaults from the `config/resources/*.yaml` files are missing
This one was solved by... getting rid of these files. In the linked issue, the example given was that the tool showed a default of `false` for the `Monitoring.MetricAuthorization` param. This happened because the config initialization between the config tool and actual server startup weren't tightly coupled resulting in the config tool entirely missing these extra defaults files. Now everything comes from a single source, `parameters.yaml` and the tool gives the correct value:
```
$ pelican config get metricauthorization
monitoring.metricauthorization: true
```

### Config tool outputs incomplete paths
This gets into my choice of formalizing `${ConfigBase}` as a param. It turned out that the config tool could start without defining the entrypoint for config and other things, so params that were scoped under the config base didn't have the appropriate path prepended. The example from the linked issue showed no path associated with `IssuerKeysDirectory`. Fixed:
```
$ pelican config get issuerkeysdirectory
issuerkeysdirectory: "/etc/pelican/issuer-keys"
```

### Accurate `pelican config summary` output
Another great thing about this overall cleanup is that `pelican config summary`, which is supposed to give you only the param values that differ from their defaults, actually works as intended now. In the future, I'm hoping we can get away from people passing around `pelican config dump` outputs with all the stuff we don't care about, because `pelican config summary` now gives you only the knobs that aren't changed.

### Toggle for metadata discovery
Explained in a previous bugfix report -- the `--with-discovery` flag is a new option to toggle metadata discovery for the tool. Off by default.

### Parameter provenance tracking
Other than deriving defaults from parameters.yaml, this is the one I'm happiest with. Now, using the `--verbose` flag with `config get` will tell you exactly where that config value comes from:
```
# Defined from base config:
$ pelican config get --verbose logging.origin.scitokens
logging.origin.scitokens: "info"
    # source: /etc/pelican/pelican.yaml

# Defined via a `ConfigLocations` file pointed at by `/etc/pelican/pelican.yaml`:
$ pelican config get --verbose discoveryurl
federation.discoveryurl: "https://123.org"
    # source: /workspaces/pelican/config.d/10-file.yaml
    
# Defined in the web config overrides
$ pelican config get --verbose logging.level
logging.level: "trace"
    # source: web-config /etc/pelican/web-config.yaml

# Defined via env var
$ PELICAN_LOGGING_ORIGIN_CMS=trace pelican config get --verbose logging.origin.cms
logging.origin.cms: "trace"
    # source: env PELICAN_LOGGING_ORIGIN_CMS

# From a default, not defined anywhere
$ pelican config get --verbose logging.origin.http
logging.origin.http: "trace"
    # source: default
```

### Service flag
Lastly, I added a `--service` flag to the config tool. This is meant to help sys admins who may not know the canonical path to the right config entrypoint for their RPM-installed or systemd service. For example, to dump config for an RPM cache, rather than doing `pelican config dump --config /etc/pelican/pelican-cache.yaml` the admin can now do `pelican config dump --service cache`. Available services are `cache`, `origin`, `director` and `registry`, with stubs for `localcache` and `broker` (if those ever get packaged, or turned on by anyone in the first place). Note that the `--service` flag and the `--config` flag are mutually exclusive, and if you use `--service` but the service-specific config doesn't exist, the command falls back to `/etc/pelican/pelican.yaml` with a warning.


jhiemstrawisc:
Also pinging @williamnswanson and @brianhlin on this since the intent of this PR was to work on the various bits of feedback you've given about `pelican config <blah>`. For you two, I think the most useful sections of the PR description will be "Bugs Fixed" and "Features Added".

brianhlin:
Very exciting! Some thoughts:

> No metadata discovery by default

Does `--with-discovery` query the server's runtime config or does it just tell the client to do metadata discovery? As @brianaydemir noted in #3190, an admin wants to know:

1. What their config on disk says and
2. What is the running config? 

If `--with-discovery` is doing metadata discovery, it's possible that this will drift from (2).


> In the future, I'm hoping we can get away from people passing around pelican config dump outputs with all the stuff we don't care about, because pelican config summary now gives you only the knobs that aren't changed.

If it doesn't already, you should consider including the Pelican version number in the output so you know what set of default config params to compare to for support. And FWIW, I view this mainly as a nicety for developers -- HTCSS has the same feature and I always just ask for the full config dump since:

1.   I can never remember what the default param values are for a given version and
5.   It generally reduces the number of round trips for a support issue

> Service flag

Cool! I'm curious how it does this but we don't need to sort that out here

jhiemstrawisc:
@brianhlin 
> Does `--with-discovery` query the server's runtime config or does it just tell the client to do metadata discovery? As @brianaydemir noted in #3190, an admin wants to know:
> 
>     1. What their config on disk says and
> 
>     2. What is the running config?
> If `--with-discovery` is doing metadata discovery, it's possible that this will drift from (2).

Using `--with-discovery` gives you the "runtime" configuration. Metadata lookup happens and fills in the federation blanks, but if you've overridden one of the values (e.g. you set your own `Federation.DirectorUrl`), you'll see your configured value. For the purely "on-disk" configuration, omit the `--with-discovery` flag.

> If it doesn't already, you should consider including the Pelican version number in the output so you know what set of default config params to compare to for support.

You can get the version in the output by tacking on `--version`, but I can make that a default so you don't have to remember an extra flag. I may do that in a follow-up PR given this one's length.

> > Service flag
> 
> Cool! I'm curious how it does this but we don't need to sort that out here

I use the provided service to set the config entry point to `/etc/pelican/pelican-{service}.yaml`. I normalize for capitalization and throw an error on an unrecognized service. This is fragile, but it seems like our RPM configs are stable enough that this is an okay starting place.

jhiemstrawisc:
I think the currently-failing tests are going to require some nontrivial work to fix (all stemming from the "port 0" thing I hate in our codebase).

I've decided this has grown too large and complicated, so I'm de-scheduling this PR from 7.26, and I plan to break it up into smaller pieces moving forward. I'll target one PR for the defaults generation, and a second for the config tool changes.

brianhlin:
> I use the provided service to set the config entry point to /etc/pelican/pelican-{service}.yaml. I normalize for capitalization and throw an error on an unrecognized service. This is fragile, but it seems like our RPM configs are stable enough that this is an okay starting place.

@jhiemstrawisc do you also fall back to `/etc/pelican/pelican.yaml`? IIRC, that's the entrypoint used in the containers

jhiemstrawisc:
> @jhiemstrawisc do you also fall back to `/etc/pelican/pelican.yaml`? IIRC, that's the entrypoint used in the containers

@brianhlin Yep, if the service-specific file isn't there, I fall back to the default config.

jhiemstrawisc:
I'm closing this PR because I'm currently in the process of splitting it into its two constituent components -- centralized config generation from `parameters.yaml` and cleanups to the `pelican config` tool. This is already too large for a reasonable review, so I'm hoping that this split will make this more manageable.

jhiemstrawisc:
As I worked to split this into smaller PRs, I was facing a constant merge conflict rebase nightmare. I think I'm going to revive this one and see if I can prevent it from growing much more than it already is. I did test things fairly thoroughly and provided evidence of it, so I can quickly retest once I have the conflicts cleaned up again.

config/source_tracker.go
```diff
@@ -0,0 +1,204 @@
...
+func SnapshotViperKeys(v *viper.Viper) map[string]string {
+	return snapshotViperKeys(v)
+}
+
+// RecordConfigFileDiff compares a pre-merge snapshot (from snapshotViperKeys)
+// against the current state and records any new or changed keys as coming from
+// the given config file path.
+func (st *SourceTracker) RecordConfigFileDiff(before map[string]string, v *viper.Viper, filePath string, sourceType ConfigSourceType) {
+	for _, key := range v.AllKeys() {
+		currentVal := v.GetString(key)
```
turetske: Both snapshotViperKeys (L130) and this diff use v.GetString(key) to compare values. viper.GetString returns "" for any stringSlice / object / map-typed key regardless of content. So when a config file overrides such a key, the before/after snapshots are both "", the diff sees no change, and the key is never re-recorded — it stays tagged SourceDefault.

Two impacts:

1. Labeling (broad): pelican config get --verbose reports # source: default for any list/object param the user actually set in a file — directly undercutting the #2594 feature this PR ships.

2. Data loss (narrow but real): handleDeprecatedConfig decides userSetReplacement from the tracker. Two deprecated→replacement pairs have stringSlice targets:

```
Origin.ExportVolume (string) → Origin.ExportVolumes (stringSlice)
Registry.AdminUsers (stringSlice) → Server.UIAdminUsers (stringSlice)
```

When an admin has both the deprecated key and its new replacement set, the tracker wrongly reports the replacement as default, so the handler overwrites the user's explicit list with the deprecated value. Reproduced end-to-end against InitConfigInternal:

```
Origin:
  ExportVolumes:        # what the admin wants
    - /user/A:/A
    - /user/B:/B
  ExportVolume: /old:/old   # deprecated scalar
```
→ after init, param.Origin_ExportVolumes.GetStringSlice() == ["/old:/old"]. The admin's list is silently discarded — and the value chosen is the deprecated one, the opposite of the handler's intended "new key wins" precedence. (Setting only the new key is safe; the handler doesn't run unless viper.IsSet(deprecated).)

Suggested fix: compare a type-aware serialization of v.Get(key) rather than GetString. Minimal version:

```
// snapshotViperKeys
for _, key := range v.AllKeys() {
    snap[key] = fmt.Sprintf("%v", v.Get(key))
}
// RecordConfigFileDiff
currentVal := fmt.Sprintf("%v", v.Get(key))
```

%v is enough to detect change for the diff's purpose (it doesn't need to be canonical, just stable before-vs-after). Please add a regression test covering a slice override — see the comment on the integration test below.

config/source_tracker.go
```diff
@@ -0,0 +1,204 @@
...
+		if oldVal, existed := before[key]; !existed || oldVal != currentVal {
```
turetske: This records a source only when the value changed. The ConfigLocations loop in handleContinuedCfg merges multiple files, each with its own snapshot. If two files set the same key to the same value, the source stays pinned to the first file — contradicting the "last writer wins" policy documented at L50-L52.

The merged config value is still correct (viper handles that); only the reported source is wrong.It's the same root cause as the above error and would be covered naturally by the same fix + an integration test that loads two files touching one key.

config/source_tracker_test.go
```diff
@@ -0,0 +1,194 @@
...
+	// Non-PELICAN env vars should not be recorded.
+	t.Setenv("HOME", "/root")
+	_, ok = st.Get("home")
+	assert.False(t, ok)
+}
+
+// TestGlobalSourceTrackerIntegration exercises the global singleton through the
+// full config initialization sequence: RecordDefaultKeys → RecordConfigFileDiff →
+// RecordEnvVarSources, verifying that later stages correctly overwrite earlier ones.
+func TestGlobalSourceTrackerIntegration(t *testing.T) {
```
turetske: This test drives the stages manually (RecordDefaultKeys → RecordConfigFileDiff → RecordEnvVarSources) with synthetic keys (foo/bar) on a hand-built viper. It never runs the actual InitConfigInternal sequence, so the real wiring — and the GetString-on-slice behavior — is unasserted.

Could we add a test that writes a temp config file and runs InitConfigInternal, asserting the resulting sources for:

a scalar override → SourceConfigFile
a stringSlice override → SourceConfigFile (this is the case that catches C1)
an env-var override → SourceEnvVar
a same-key-same-value second file → still attributed to the second file (catches C2)
A GetStringSlice assertion on Origin.ExportVolumes with the deprecated Origin.ExportVolume also set would lock in the data-loss regression from C1 directly.

turetske (changes requested):

cmd/config_printer/config_get.go
```diff
@@ -146,6 +150,24 @@ func configGet(cmd *cobra.Command, args []string) {
 
 	for _, match := range matches {
 		fmt.Printf("%s: %s\n", match.HighlightedKey, match.HighlightedValue)
+		if verbose {
+			if src, ok := config.GetSourceTracker().Get(strings.ToLower(match.OriginalKey)); ok {
```
turetske, line 154: Do we want an else clause for this if? Are we sure this should print nothing if src is empty?

config/net_config.go
```diff
@@ -55,6 +55,17 @@ func UpdateConfigFromListener(ln net.Listener) {
...
+					if err := param.Director_AdvertiseUrl.Set(newUrlStr); err != nil {
+						log.WithError(err).Warn("Failed to update Director.AdvertiseUrl from listener")
+					}
+				} else if parsed, perr := url.Parse(curAdvUrl); perr == nil && parsed.Port() == "0" {
+					parsed.Host = parsed.Hostname() + ":" + strconv.Itoa(tcpAddr.Port)
+					if err := param.Director_AdvertiseUrl.Set(parsed.String()); err != nil {
+						log.WithError(err).Warn("Failed to update Director.AdvertiseUrl from listener")
+					}
+				}
 				if viper.GetString("Federation.RegistryUrl") == serverUrlStr {
```
turetske, line 69: I know you didn't introduce this, but I'm seeing it while looking above. Shouldn't this be a param.get rather than a viper call?

turetske (changes requested):
Two very small nitpicks

config/net_config.go
jhiemstrawisc: This is one of the last remaining places where we still need raw viper on a raw string. See https://github.com/PelicanPlatform/pelican/issues/3109

cmd/config_printer/config_get.go
jhiemstrawisc: Good catch -- I'll update this with an "else" that indicates the source is unknown.

turetske (approved):

PR #3368, opened 2026-04-20
Improve apikey cli

### LINKED ISSUE #3227: Touch up `pelican api-key` token creation API
Some notes from William's testing of the features introduced in #2958.

I’m able to generate something that the CLI tool says is a key, but there’s no documentation for how I use it.

Additionally, the CLI --help section doesn’t state that the --expiration and --server flags are required.

It would also be helpful if I could find an exhaustive list of possible scopes somewhere and if the CLI tool would give an example of a timestamp in the required RFC3339 format for non-Dew drinkers.

h2zh:
Closes #3227 

- Fix --expiration 'never' not working despite being documented as valid (fixes #3227)
- Accept ISO 8601 date-only format (e.g., 2025-12-31), interpreted as midnight UTC
- Improve --help text with expiration format docs, common scopes, and usage examples
- Update Grafana monitoring docs to recommend pelican apikey generate for token creation

### DIFF
--- a/cmd/api_key_generate.go
+++ b/cmd/api_key_generate.go
@@ -46,8 +46,27 @@ var (
 	apiKeyGenerateCmd = &cobra.Command{
 		Use:   "generate",
 		Short: "Generate a new API key for the server",
-		Long:  "Generate a new API key with specified scopes.",
-		RunE:  generateApiKey,
+		Long: `Generate a new API key with specified scopes.
+
+Common scopes for API keys:
+  monitoring.query    Query the server's Prometheus endpoints
+  monitoring.scrape   Scrape the server's /metrics endpoint
+
+The --expiration flag accepts the following formats:
+  - 'never'                       Token does not expire
+  - RFC3339 timestamp             e.g. 2025-12-31T23:59:59Z
+  - Date only (ISO 8601)          e.g. 2025-12-31 (interpreted as midnight UTC)
+
+Examples:
+  # Generate an API key that never expires
+  pelican apikey generate --server https://my-origin.com:8447 --scopes "monitoring.query" --expiration never
+
+  # Generate an API key expiring on a specific date
+  pelican apikey generate --server https://my-origin.com:8447 --scopes "monitoring.query,monitoring.scrape" --expiration 2025-12-31
+
+  # Generate an API key with a precise expiration
+  pelican apikey generate --server https://my-origin.com:8447 --scopes "monitoring.query" --expiration 2025-12-31T23:59:59Z`,
+		RunE: generateApiKey,
 	}
 
 	apiKeyScopes     string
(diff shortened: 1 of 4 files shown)


turetske (changes requested):
I believe you need to run go generate to ensure that the cli docs are updated properly. The validation action for it is failing.

h2zh:
@melody0123 This would give you more info on `apikey`. You can review this before https://github.com/PelicanPlatform/pelican/pull/3472

cmd/api_key_generate.go
```diff
@@ -46,8 +46,33 @@ var (
...
+  pelican apikey generate --server https://my-origin.com:8447 --scopes "monitoring.query" --expiration never
+
+  # Generate an API key expiring on a specific date
+  pelican apikey generate --server https://my-origin.com:8447 --scopes "monitoring.query,monitoring.scrape" --expiration 2025-12-31
+
+  # Generate an API key with a precise expiration in UTC
+  pelican apikey generate --server https://my-origin.com:8447 --scopes "monitoring.query" --expiration 2025-12-31T23:59:59Z
+
+  # Generate an API key with a timezone offset
+  pelican apikey generate --server https://my-origin.com:8447 --scopes "monitoring.query" --expiration 2025-12-31T18:59:59-05:00`,
```
melody0123, line 74: cobra.Command has a dedicated Example field for example usage. I would recommend put examples there.

cmd/api_key_generate.go
```diff
@@ -46,8 +46,33 @@ var (
...
+		RunE: generateApiKey,
```
melody0123, line 75: I would recommend to add SilenceUsage: true because otherwise a logic error (instead of misspelling command) will print usage message, which buries real error messages in noise

melody0123 (changes requested):
Thanks for the invitation to review the PR! The PR is generally great, with only 2 minor fixes:
1. Move examples from `cobra.Command.Long` to `cobra.Command.Example`
2. Set `cobra.Command.SilenceUsage` to `true`

cmd/api_key_generate.go
h2zh: Good call — moved the examples to the dedicated `Example` field (following the `pelican token create` idiom) and regenerated the CLI docs, so they now render under a proper "Examples" section. Done in 93f0b99a6.

cmd/api_key_generate.go
h2zh: Added `SilenceUsage: true` in 93f0b99a6 — agreed, runtime errors shouldn't be buried under the usage dump.

melody0123 (approved):
Great! All requests to change are fixed. One more advice for the future: After change is requested, do not do force push to override existing commits if possible. Otherwise:
1. Inline comments will miss anchor points.
2. It will be harder for the reviewer to see the change (and reviewer may suspect if anywhere else in addition to the requested change is also modified)

h2zh:
@melody0123 To pass the "Check Rebase on Main" CI, each PR branch has to rebase upon main. After this rebase, all new commits in this branch will get a new commit ID, although the content is unchanged. Then a force-push is unavoidable.

turetske (approved):

PR #3376, opened 2026-04-22
Install RPM dependencies from osg-testing with version pinning

### LINKED ISSUE #3300: Image builds: pull RPMs from osg-testing but pin versions
The Dockerfile directly pulls several dependency RPMs directly from kojihub2000 because we pin the version (using build arguments). This used to not be possible with the OSG repos because the RPMs in the `osg` (release) repos were sometimes too old for Pelican but `osg-testing` only contained the latest packages, meaning it couldn't be used for version pinning.

osg-testing was changed so that it contains old versions too, which means version pinning works again. In addition,
since packages only end up in osg-testing once they've passed OSG VMU tests, by using the osg-testing repo there is a certain guarantee of smoke-testing.

The syntax is `dnf install -y --enablerepo=osg-testing ${NAME}-${VERSION}-${RELEASE}` (instead of building up a Koji URL and installing that).

Relatedly, pin the version of `xrootd-multiuser` since we don't seem to be doing that.

matyasselmeci:
Replace direct Koji URL installs with pinned package installs from osg-testing.  This guarantees that the RPMs have at least gone through the OSG VMU smoke tests.  The RPMs are pinned to a specific Version by default, but not a specific Version-Release, though that can be changed via an ARG.

For consistency, I added `*_RELEASE` args to all of the dependencies that we can install via RPMs.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -39,7 +39,6 @@
 ARG ALMALINUX_TAG=9
 ARG BASE_OS=el9
 ARG OSG_SERIES=25
-ARG OSG_REPO=release
 
 # For compiling custom/one-off releases of Pelican.
 # (We cannot supply a default value because GoReleaser will use it as-is.)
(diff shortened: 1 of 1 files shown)


jhiemstrawisc (approved):
Tested the build and manually checked installations/versions for a handful of the packages. All looks good to me!

PR #3383, opened 2026-04-24
Build a container image for running the Pelican client

### LINKED ISSUE #3377: Create client image
It would be helpful to have a small Docker image containing, let's say, Alpine and the Pelican client, for use in CI flows using Kubernetes or initContainers, similar to how there is an image for curl. ([alpine/curl](https://hub.docker.com/r/alpine/curl)).

brianaydemir:
The basic idea here is to use one of Google's "distroless" images as the base for a Pelican client container image, because Google builds around glibc, unlike Alpine.

I've updated the `Dockerfile` with a client-specific target, and updated the GitHub Actions to build-and-push that image as needed.

---

Closes #3377.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -13,7 +13,7 @@ name: Build and Test
 #
 #   - Pushes to main: We push only pelican-test and pelican-dev.
 #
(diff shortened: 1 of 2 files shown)


matyasselmeci (approved):
Yep, LGTM.

PR #3384, opened 2026-04-26
Make shutdown routines more robust

### LINKED ISSUE #3363: Fix segfault in ssh origin backend
The SSH backend runs for a few minutes before segfaulting in [itb-ssh-origin](https://github.com/opensciencegrid/tiger-osg-config/tree/master/manifests/tiger/osdf-dev/itb-ssh-origin).

Selected summary of the logs:
```
Pelican origin starts, XRootD advertises, SSH helper launches

First SSH keepalive/pong succeeds

Flood of Poll retrieve error lines: DNS lookups for itb-ssh-origin.osdf-dev.chtc.io time out, both directly and through 10.96.0.10:53 (the cluster's CoreDNS). Roughly 2 minutes of this.

Failed to send ping to helper: EOF
```

(the container is sleeping; and the operation team was exec’ing in and running pelican-server origin serve by hand)

The documentation doesn’t say where to set the paths to export on the remote side, so the operation team guessed Origin.Exports.StorageLocation as if it were a POSIX mount.

See @matyasselmeci's comment in Jira Ticket OPS # 508 for the full logs and original comment.

bbockelm:
- Actually cancel to shutdown various tickers.
- Add nil-checks and make copies of pointers to prevent a panic.

This fixes a potential segfault when the helper is shutting down.

Fixes #3363

### DIFF
--- a/ssh_posixv2/helper.go
+++ b/ssh_posixv2/helper.go
@@ -174,14 +174,17 @@ func (c *SSHConnection) StartHelper(ctx context.Context, helperConfig *HelperCon
 		return errors.Wrap(err, "failed to write newline to helper stdin")
 	}
 
-	// Create errgroup for managing helper goroutines
-	egrp, egrpCtx := errgroup.WithContext(ctx)
+	// Create errgroup for managing helper goroutines. We layer a separate
+	// cancellable context on top of the caller's so StopHelper can actively
+	// wake ticker-driven goroutines (runPongMonitor / runStdinKeepalive)
+	// without waiting for their next tick. Without this, the SIGKILL
+	// fallback path nils helperIO while those goroutines are still ticking
+	// and they panic on the stale field deref.
+	cancelCtx, cancel := context.WithCancel(ctx)
+	egrp, egrpCtx := errgroup.WithContext(cancelCtx)
 	c.helperErrgroup = egrp
 	c.helperCtx = egrpCtx
-	c.helperCancel = func() {
-		// Signal shutdown via stdin before cancelling context
-		_ = c.sendShutdownMessage()
-	}
+	c.helperCancel = cancel
 
 	// Goroutine: Read helper stdout for pong responses
 	egrp.Go(func() error {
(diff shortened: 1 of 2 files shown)


ssh_posixv2/helper.go
```diff
@@ -420,19 +431,27 @@ func (c *SSHConnection) runStdinKeepalive(ctx context.Context) error {
 
 // sendPing sends a ping message to the helper via stdin
 func (c *SSHConnection) sendPing() error {
+	// Snapshot helperIO so a concurrent StopHelper that nils it (or its
+	// stdin) can't turn this into a nil deref. sendShutdownMessage uses
+	// the same pattern.
```
h2zh: `sendShutdownMessage` still reads `c.helperIO` directly without snapshotting, inconsistent with this comment.

ssh_posixv2/helper.go
```diff
@@ -174,14 +174,17 @@ func (c *SSHConnection) StartHelper(ctx context.Context, helperConfig *HelperCon
...
+	// and they panic on the stale field deref.
+	cancelCtx, cancel := context.WithCancel(ctx)
+	egrp, egrpCtx := errgroup.WithContext(cancelCtx)
 	c.helperErrgroup = egrp
 	c.helperCtx = egrpCtx
-	c.helperCancel = func() {
-		// Signal shutdown via stdin before cancelling context
-		_ = c.sendShutdownMessage()
-	}
+	c.helperCancel = cancel
```
h2zh, line 190: Minor observation: The docstring of `helperCancel` func in types.go ("cancels the helper context and triggers clean shutdown") should be updated to reflect that it only cancels the local context, no longer sends a shutdown message to the remote helper.
The "triggers clean shutdown" part is misleading because the actual shutdown message to the remote helper is now sent separately by `sendShutdownMessage` in `StopHelper`, not by this function.

h2zh (changes requested):
Generally looks good, just two observations:

h2zh:
Hey @turetske - If we are going unhide SSH backend docs in 7.26, this PR needs to be backported there

h2zh (approved):
LGTM. I also tested the graceful shutdown on a SSH backend manually.

PR #3385, opened 2026-04-27
Additional client/server split cleanup (docs, CLI references, systemd service files, etc.)

### LINKED ISSUE #3378: Update documentation in light of client/server binary split
With #3343 merged, now we need to comb through our docs to figure out how they need to be updated.

### LINKED ISSUE #3393: Cli Doc Generator produced invalid Markdown
Generator is dropping the following into Markdown resulting in the docs throwing error for invalid Markdown.

When the cli docs are being generated we will need to escape reserved characters so as to produce valid markdown.

Example:

Broken md:

`    --id <client-id> \`

Full File:

```
## pelican origin issuer client update

Update an existing OIDC client

### Synopsis

Update the configuration of an existing OIDC client.

Only the flags you provide are changed; omitted fields are left unchanged.

Example — add token-exchange grant and narrow scopes:
  pelican origin issuer client update --server https://my-origin:8447 \
    --id <client-id> \
    --grant-types "urn:ietf:params:oauth:grant-type:token-exchange,refresh_token" \
    --scopes "openid,storage.read:/"

```
pelican origin issuer client update [flags]
```

### Options

```
      --grant-types string   Comma-separated list of grant types
  -h, --help                 help for update
      --id string            Client ID to update (required)
      --scopes string        Comma-separated list of scopes
```

### Options inherited from parent commands

```
      --namespace string    Federation namespace prefix for the issuer (e.g. /data/analysis) (required)
      --server string       Web URL of the Pelican origin server (e.g. https://my-origin:8447)
      --token string        Path to a file containing an admin token (optional; generated automatically if omitted)
```

### SEE ALSO

* [pelican origin issuer client](/commands-reference/origin/issuer/client/)	 - Manage OIDC clients for the origin's embedded issuer
```

jhiemstrawisc:
This PR runs through a host of things that I initially missed when I was splitting apart the binaries.

While this is a large diff, the vast majority of that comes from splitting `docs/app/commands-reference` into `docs/app/commands-reference/pelican` and `docs/app/commands-reference/pelican-server`. This all happens via our built-in docs generator, so the reviewer(s) can probably gloss over the majority of those lines (see commit `c08caac`).

The only thing I'm aware of that needs updating that isn't included in this diff is the `<DownloadsComponent>` component for Nextra -- this lives in our separate docs repo, and it should be updated to provide a toggle for `client/server`. This is opened as a separate issue: https://github.com/PelicanPlatform/docs/issues/46

Closes #3378


turetske (changes requested):
There are still a few places where the docs reference 'pelican' where it should now be 'pelican-server'. For example, the warning in "Managing Server Downtime" says ```The pelican downtime CLI commands interact directly with your server’s admin API and require either local admin access or a valid admin token.``` where I think it now should say 'pelican-server downtime'. Unsure if there are others, I didn't go through the docs with a fine tooth comb, but that is one I noticed.

turetske (approved):
LGTM

PR #3387, opened 2026-04-27
Install additional utilities into all container images

### LINKED ISSUE #3386: Install additional utilities into container images
As a practical matter, it is quite convenient if a container image contains a variety of standard utilities, as it enables an operator who `exec`s into a running container to immediately start investigating and debugging the state of the running processes. And with respect to Pelican's server images, there is a difference between being [needlessly wasteful](https://github.com/PelicanPlatform/pelican/pull/3093) and including useful-but-not-strictly-necessary utilities.

Thus, given that Pelican makes use of an SQLite database, it would be convenient if the official container images included the SQLite CLI.

brianaydemir:
Closes #3386.

Refer to that issue for this PR's background and motivation.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -289,6 +289,7 @@ RUN --mount=type=bind,source=images/pelican.yaml,target=/context/pelican.yaml \
     procps \
     psmisc \
     python3-pip \
+    sqlite \
     tini \
     unzip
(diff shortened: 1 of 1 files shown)


bbockelm (approved):
No reason to overthink this one...

PR #3389, opened 2026-04-27
Fix refresh token expiry in embedded issuer custom grant flows

### LINKED ISSUE #3388: Embedded OAuth2 issuer: refresh tokens expire after 1 hour instead of 7 days

## Summary

In the embedded OAuth2 issuer (`oauth2/issuer`), refresh tokens issued via the device code flow and the RFC 8693 token exchange flow are stored with an `expires_at` of `now + 1h` instead of the hard-coded 7-day lifespan. This causes them to be deleted by the background cleanup goroutine after one hour, making all refresh tokens unusable after that point with an `invalid_grant` / "malformed or not valid" error.

## Root cause

Both `handleDeviceTokenExchange` (`handlers.go`) and `handleTokenExchange` (`admin_handlers.go`) build a `fosite.AccessRequest` manually and call the HMAC core strategy's `GenerateRefreshToken` directly, bypassing fosite's standard `NewAccessRequest`/`NewAccessResponse` pipeline. As a result, `Session.SetExpiresAt(fosite.RefreshToken, ...)` was never called before `CreateRefreshTokenSession`.

`createTokenSession` in `storage.go` falls back to `time.Now().Add(1h)` when the refresh token expiry on the session is zero. The cleanup goroutine (running every 10 minutes) then deletes the row once that hour elapses.

brianaydemir:

Fixes #3388.

## Changes

In `handleDeviceTokenExchange` (`handlers.go`) and `handleTokenExchange` (`admin_handlers.go`), call `ar.GetSession().SetExpiresAt(fosite.RefreshToken, time.Now().Add(provider.config.RefreshTokenLifespan))` before `GenerateRefreshToken` so that `createTokenSession` stores the correct expiry instead of falling back to `now + 1h`.

The standard `authorization_code` and `refresh_token` flows go through fosite's built-in pipeline and are not affected.

## Testing

`go test ./oauth2/issuer/...` passes.

### DIFF
--- a/oauth2/issuer/admin_handlers.go
+++ b/oauth2/issuer/admin_handlers.go
@@ -21,6 +21,7 @@ package issuer
 import (
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/ory/fosite"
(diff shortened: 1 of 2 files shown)


bbockelm (approved):
Fantastic catch!

brianaydemir:
@jhiemstrawisc @h2zh I created #3392 for patching this into v7.25.

PR #3391, opened 2026-04-28
Close race condition in broker callback

### LINKED ISSUE #3390: Another whack at the TestBrokerApi test timeouts
Same as #3066 / #3070 -- take another run at making TestBrokerApi's race condition go away.

bbockelm:
There was a slight logic bug in the last rendition of trying to fix the broker callback.  By setting `SetReadDeadline`, you preempt any in-progress readers -- but asynchronously.  It's possible that an in-progress reader still consumes a packet and causes the origin to miss the ClientHello.

This forces both (a) the read to be blocked and (b) all in-progress read syscalls to finish first.

As part of the debugging,
- Prevent the client from using HTTP/2 as that breaks hijacking.
- Log the HTTP server's errors via logrus, not stderr.

Fixes #3390

### DIFF
--- a/broker/client.go
+++ b/broker/client.go
@@ -70,6 +70,16 @@ type (
 		// This prevents the HTTP transport's background TLS reader from consuming
 		// data that we need after hijacking the connection.
 		blockReads atomic.Bool
+		// inFlight tracks Read calls currently inside the underlying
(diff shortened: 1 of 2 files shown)


bbockelm:
Ok, I'm just going to push this forward because it's (a) currently low impact and (b) a really tricky bit of code where I'm, unfortunately, the only expert...

PR #3395, opened 2026-04-28
Make embedded issuer token lifespans configurable; fix refresh token tombstone security bug

brianaydemir:

The four token lifespans used by the embedded OAuth2 issuer (access token, refresh token, authorization/device code, and ID token) were previously hardcoded. This change exposes them as operator-configurable parameters under the `Issuer.*` namespace, with defaults matching the prior hardcoded values so existing deployments are unaffected.

A security bug in the refresh token cleanup logic is also fixed. Rotated refresh tokens are kept in the database as tombstones so that any replay of an already-used token is detected and triggers revocation of the entire token family, forcing re-authentication. The cleanup job was incorrectly removing these tombstones before their natural expiry, which meant a replay during that window would go undetected — the token would simply appear not to exist rather than being flagged as reused. Tombstones are now retained until their natural expiry passes, with a separate conservative cleanup path for older rows that predate the current scheme.

Two smaller correctness fixes are also included: a token-exchange code path that was ignoring the newly-configurable access token lifespan, and a gap in the per-token-type expiry logic that caused certain session types to always receive a hardcoded one-hour database lifetime.

### DIFF
--- a/config/resources/defaults.yaml
+++ b/config/resources/defaults.yaml
@@ -174,6 +174,10 @@ OIDC:
   UserInfoEndpoint: "https://cilogon.org/oauth2/userinfo"
   Scopes: ["openid", "email", "profile"]
 Issuer:
+  AccessTokenLifetime: 1h
+  AuthorizationCodeLifetime: 10m
+  IDTokenLifetime: 1h
+  RefreshTokenLifetime: 168h
   TomcatLocation: /opt/tomcat
   ScitokensServerLocation: /opt/scitokens-server
   QDLLocation: /opt/qdl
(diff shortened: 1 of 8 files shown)


bbockelm (approved):
Pre-approving but requesting two sanity checks:
1. "Lifespan" seems a funny descriptor.  Do we use "lifetime" elsewhere?  Just seems ... odd.  Feel free to disagree with me, but may be worth a second eyeball.
2.  When we load values, can we have a sanity check threshold?  E.g., refuse to set a value below X and set it at the minimum (emitting a warning advising the system is doing so).  I think we have other examples of this occurring in the code.

bbockelm:
@brianaydemir - ping on this one...

brianaydemir:
@turetske I think this will need to be patched into 7.26 in order to get it into user's hands sooner rather than later.

PR #3396, opened 2026-04-28
param: reject bare numbers for duration config fields

### LINKED ISSUE #3370: Update Server.StartupTimeout units to seconds
On Pelican 7.24.2, I've set `Server.StartupTimeout: 600`, assuming that the units were in seconds with the default being listed as 10s (https://docs.pelicanplatform.org/parameters#Server-StartupTimeout). But it appears that the units are actually in nanoseconds:

```
Error: url https://mydirector.org/api/v1.0/health didn't respond with the expected status code 200 within the timeout of 600ns (Server.StartupTimeout): context deadline exceeded
```

brianaydemir:

When a user sets a duration parameter in `pelican.yaml` with a bare integer — e.g., `Server.StartupTimeout: 600` — the value was silently accepted and interpreted as 600 **nanoseconds** rather than 600 seconds. No warning was emitted, and the resulting behavior was nearly impossible to diagnose.

The root cause is that `mapstructure`'s `StringToTimeDurationHookFunc` only converts *string* values (e.g., `"10m"`) via `time.ParseDuration`. A bare YAML integer arrives as a Go `int`, bypasses the hook entirely, and is coerced directly to `time.Duration` — which is an `int64` counting nanoseconds — by `WeaklyTypedInput: true`.

This PR adds a hook that intercepts any attempt to decode a bare integer or float into a `time.Duration` field and returns a clear error directing the user to include a unit suffix (e.g., `"10m"`, `"600s"`). Programmatically-set `time.Duration` values — from `param.Set`, `viper.Set`, or `viper.SetDefault` — are explicitly allowed through, so no existing behavior changes for code that sets durations correctly. The hook composition is also centralized into a single helper to prevent the three decoder sites from drifting out of sync in the future.

### DIFF
--- a/param/param.go
+++ b/param/param.go
@@ -19,10 +19,12 @@
 package param
 
 import (
+	"fmt"
 	"reflect"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/mitchellh/mapstructure"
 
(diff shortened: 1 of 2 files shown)


bbockelm (approved):
YES.

Ready to go when @brianaydemir says it isn't a draft anymore.

PR #3397, opened 2026-04-29
Upgrade actions to versions that run on Node.js 24

brianaydemir:
Exactly what it says on the tin. Addresses warnings such as the following:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/cache@v4, docker/build-push-action@v5, docker/metadata-action@v5, docker/setup-buildx-action@v3. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026.

I am unable to address `test-summary/action` here because an updated version is not yet available.

There's also one `actions/cache@v4` warning I am unable to address because it is being triggered by the internals of `golanci/golangci-lint-action`.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -256,7 +256,7 @@ jobs:
 
       - name: Set up metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
         with:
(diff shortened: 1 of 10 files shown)


jhiemstrawisc (approved):
👍

PR #3398, opened 2026-04-29
Upgrade more actions to versions that run on Node.js 24

brianaydemir:
This is a follow up to #3397 wherein I read the output from our plethora of actions more carefully.

I'm also taking the opportunity to address the warning from the GoReleaser action by locking the version of GoReleaser used to a specific major version.

I'm also taking the opportunity to address the Minio warning in the macOS tests be simply no longer installing it — we actually haven't used it since 0a7acb1fda5545f607d69ffe414f4bf3f7da604e!

I'm also taking the opportunity to address the brew's warning about (re)installing Ninja in the macOS tests by also no longer installing it — [it's officially already there](https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md)!

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -467,7 +467,7 @@ jobs:
           password: ${{ secrets.PELICAN_HARBOR_ROBOT_PASSWORD }}
 
       - name: Download digests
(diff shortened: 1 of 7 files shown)


jhiemstrawisc (approved):
👍

PR #3399, opened 2026-04-29
Now seeds rand with time to ensure actual randomness

### LINKED ISSUE #3401: 20% re-request from the director happening 100% of the time
The director redirect logic says that we should re-request the information approximately 20% of the time if it's not stale. However it's happening 100% of the time. This is because the rand function being used isn't being seeded, so the value is the same for all runs (which in this case happens to be within the 20% range). 

The fix is to just seed math.rand() with the current time.

turetske:
Actually ensures that the random seed isn't the same value repeated over and over again. Which caused the randomness not to be random. (So ~20% requests for new director info was turning up as 100% before this change).

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -77,7 +77,8 @@ const (
 	Retryable     = 11
 )
 
-func shouldRequestDirectorDecision(pct int) bool {
+func shouldRequestDirectorDecision() bool {
+	pct := param.Plugin_DirectorDecisionPercentage.GetInt()
 	if pct <= 0 {
 		return false
 	}
@@ -500,7 +501,7 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
 
 			urlCopy := *(pUrl.GetRawUrl())
 			jobCtx := context.Background()
-			if shouldRequestDirectorDecision(param.Plugin_DirectorDecisionPercentage.GetInt()) {
+			if shouldRequestDirectorDecision() {
 				jobCtx = client.WithDirectorDebug(jobCtx)
 			}
 			tj, err = tc.NewTransferJob(jobCtx, &urlCopy, transfer.localFile, upload, recursive, client.WithAcquireToken(false), client.WithCaches(caches...))
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -77,6 +77,17 @@ const (
 	Retryable     = 11
 )
 
+func shouldRequestDirectorDecision(pct int) bool {
+	if pct <= 0 {
+		return false
+	}
+	if pct >= 100 {
+		return true
+	}
+	// We use [0,100) so pct=1 means ~1% and pct=99 means ~99%.
+	return rand.Intn(100) < pct
+}
+
 func init() {
 	rootCmd.AddCommand(rootPluginCmd)
 
@@ -489,7 +500,7 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
 
 			urlCopy := *(pUrl.GetRawUrl())
 			jobCtx := context.Background()
-			if pct := param.Plugin_DirectorDecisionPercentage.GetInt(); pct > 0 && rand.Intn(100) < pct {
+			if shouldRequestDirectorDecision(param.Plugin_DirectorDecisionPercentage.GetInt()) {
 				jobCtx = client.WithDirectorDebug(jobCtx)
 			}
 			tj, err = tc.NewTransferJob(jobCtx, &urlCopy, transfer.localFile, upload, recursive, client.WithAcquireToken(false), client.WithCaches(caches...))
--- a/cmd/plugin_test.go
+++ b/cmd/plugin_test.go
@@ -29,6 +29,8 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"math"
+	"math/rand"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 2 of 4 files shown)


cmd/plugin.go
```diff
@@ -98,6 +98,11 @@ func stashPluginMain(args []string) {
 		log.Warningln("Failed to set plugin mode:", err)
 	}
 
+	// The plugin is commonly invoked as a new process per transfer.  Without a seed,
+	// math/rand will generate the same "random" sequence every invocation, defeating
+	// percentage-based sampling (for example, Plugin.DirectorDecisionPercentage).
+	rand.Seed(time.Now().UnixNano())
+
```
jhiemstrawisc: This looks reasonable to me -- how did you test that it works? I'm not sure off the top of my head how to do it.

turetske: There isn't really a way without just running a large number of requests like integration does. I mostly just checked that before it was getting the same value, and now math.rand is getting a different value.

jhiemstrawisc: According to the docs for `math/rand`, the [`rand.Seed` API](https://pkg.go.dev/math/rand#Seed) has been deprecated since go 1.21, and is a no-op as of go 1.24. The bug was observed in `v7.25`, which is on go 1.25, so I don't think this is actually going to fix anything. More confusingly, the docs for `Seed` say "If Seed is not called, the generator is seeded randomly at program startup", so I'm not sure this is actually a seeding problem in the first place?

Since this almost escaped me, I think we should have a unit test that actually makes sure the `Plugin.DirectorDecisionPercentage` knob actually causes the request to fire with the debug header at the correct percentage.

jhiemstrawisc (changes requested):

brianaydemir:
@jhiemstrawisc @turetske — Copilot (or, rather, Copilot's Rubber Duck, because Copilot alone wasn't giving me convincing answers to my questions):

> **The actual root cause:** `config.InitClient()` (called at plugin startup) calls `initFilterLogging()`, which unconditionally sets logrus's global level to `TraceLevel` (line 186 of `logging.go`). After this, `log.IsLevelEnabled(log.DebugLevel)` **always returns `true`** — regardless of the configured "warn" level — because it checks the global logrus level, not the effective level maintained by the hook system.
>
> In `queryDirector`, the check `log.IsLevelEnabled(log.DebugLevel) || forceDebug` therefore fires on **every single request**, bypassing the 20% sampling gate entirely.
>
> **The correct fix** is one line in `client/director.go`:
> ```go
> // Before (broken — always true after initFilterLogging):
> if log.IsLevelEnabled(log.DebugLevel) || forceDebug {
>
> // After (correct — uses the hook-aware effective level):
> if config.GetEffectiveLogLevel() >= log.DebugLevel || forceDebug {
> ```
>
> PR #3399's `rand.Seed` fix is entirely irrelevant — the PRNG sampling gate is never even reached because the `log.IsLevelEnabled` check short-circuits it first.

In a one-off, quick test, this proposed fix does indeed seem to work.

jhiemstrawisc (changes requested):
I think I'm still going to stick by my previous comment -- can you add unit testing checking that the configured probability triggering `forceDebug` actually results in the requests firing within some delta of what's expected?

turetske:
> I think I'm still going to stick by my previous comment -- can you add unit testing checking that the configured probability triggering `forceDebug` actually results in the requests firing within some delta of what's expected?

Yep, I think I created the test, but forgot to commit it :D

cmd/plugin.go
```diff
@@ -489,7 +500,7 @@ func runPluginWorker(ctx context.Context, upload bool, workChan <-chan PluginTra
 
 			urlCopy := *(pUrl.GetRawUrl())
 			jobCtx := context.Background()
-			if pct := param.Plugin_DirectorDecisionPercentage.GetInt(); pct > 0 && rand.Intn(100) < pct {
+			if shouldRequestDirectorDecision(param.Plugin_DirectorDecisionPercentage.GetInt()) {
```
jhiemstrawisc: It feels strange to give this generic function a specific name, and then pass the part that makes the function specific to the Director as an input. Since the function name already implies it's going to use the `Plugin.DirectorDecisitionPercentage` param, can you move that inside the body?

cmd/plugin_test.go
```diff
@@ -71,6 +73,64 @@ var (
...
+// Test that the configured percentage for requesting the Director's decision
+// information results in a corresponding fraction of transfers including
+// DeveloperData.DirectorDecision.
+//
+// This exercises the real production path in runPluginWorker that conditionally
+// applies client.WithDirectorDebug(...) based on Plugin.DirectorDecisionPercentage.
+func TestShouldRequestDirectorDecisionProbability(t *testing.T) {
+	// shouldRequestDirectorDecision uses the package-global RNG. Seed it so the
+	// behavior is deterministic and stable across runs.
+	rand.Seed(1)
```
jhiemstrawisc: `Seed` is [deprecated](https://pkg.go.dev/math/rand#Seed), so we shouldn't be using it. Also, I don't think it's a good idea to seed tests that are meant to check probabilities.

cmd/plugin_test.go
```diff
@@ -71,6 +73,64 @@ var (
...
+			pct int
+			n   int
+		}
+		cases := []tc{
+			{pct: 1, n: 200000},
+			{pct: 10, n: 100000},
+			{pct: 33, n: 100000},
+			{pct: 50, n: 80000},
+			{pct: 90, n: 100000},
+			{pct: 99, n: 200000},
```
jhiemstrawisc: What's the reasoning for running each of these a different number of times?

jhiemstrawisc (changes requested):

jhiemstrawisc (approved):
I realized in my last review that the new unit tests still wouldn't have caught the original bug because they didn't exercise the logic in `queryDirector` where the bug existed. To cut down on the back-and-forth, I added this test myself and plan to merge once I see green tests.

I also tested that this new test is valid by reverting the config level fix and observing that the test fails in that case.

PR #3402, opened 2026-04-29
Hide some docs/CLI commands

### LINKED ISSUE #3400: Hide some features' documentation while they're known not to work
In testing `v7.25` ([ticket](https://opensciencegrid.atlassian.net/browse/OPS-508)), @williamnswanson identified that the rclone integration and the ssh Origin backend aren't ready for users to pick up and run with. Until we think these are ready, I've been asked to hide their documentation/commands/parameters.

jhiemstrawisc:
These are things that are currently in ongoing releases, but that have been having issues in testing. Until we get them over the line, hide their usage so users don't find them and expect that they work.

### DIFF
--- a/cmd/rclone.go
+++ b/cmd/rclone.go
@@ -26,8 +26,9 @@ import (
 
 var (
 	rcloneCmd = &cobra.Command{
-		Use:   "rclone",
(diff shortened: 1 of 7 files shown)


williamnswanson (approved):
LGTM

PR #3405, opened 2026-04-30
Skip activation flow when admins are pre-configured

### LINKED ISSUE #1747: Allow WebUI startup with OAuth2-only administrators
Pelican Service:
- [ ] Client
- [ ] Plugin
- [X] Registry
- [X] Director
- [X] Origin
- [X] Cache
- [X] Other (please give the detail)

In the OSDF, we're responsible for spinning up multiple Pelican services and we'd like to use the WebUI but we have to set up an admin password for each one. To be safe, we should have a different password for each service but that starts to become a management problem in of itself.

I'd like to be able to use the WebUI with only a list of OAuth2-based administrators and forego passwords altogether.

brianaydemir:
## Summary

htpasswd file present | Admins pre-configured | Activation flow
-- | -- | --
No | No | Required — activation code printed to stdout; server waits for login
Yes | No | Skipped — htpasswd already bootstrapped
No | Yes | Skipped — admin can log in via OAuth2
Yes | Yes | Skipped — both paths available

## Details

Closes #1747.

When `Server.UIAdminUsers` or `Server.AdminGroups` is set, at least one admin can already authenticate via OAuth2, so the one-time activation-code flow serves no purpose. This PR skips that flow in those cases:

- `handleWebUIAuth` no longer redirects to the initialization page, and visiting the initialization page redirects away to `/view/`.
- `waitUntilLogin` exits immediately, cleaning up any stale activation file from a prior run.
- `loginHandler` and `initLoginHandler` return HTTP 400, since password-based login is not meaningful in this mode.

A `sync.Once`-guarded helper, `hasConfiguredAdmins()`, reads the two config slices once at startup and caches the result. A reset hook is registered with `server_utils` so that `ResetTestState` clears the latch between tests.

### DIFF
--- a/web_ui/authentication.go
+++ b/web_ui/authentication.go
@@ -555,6 +555,13 @@ func DowntimeAuthHandler(ctx *gin.Context) {
 func loginHandler(ctx *gin.Context) {
 	db := authDB.Load()
 	if db == nil {
+		if hasConfiguredAdmins() {
+			ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+				Status: server_structs.RespFailed,
+				Msg:    "Password login is not available",
+			})
+			return
+		}
 		newPath := path.Join(ctx.Request.URL.Path, "..", "initLogin")
 		initUrl := ctx.Request.URL
 		initUrl.Path = newPath
@@ -634,6 +641,13 @@ func loginHandler(ctx *gin.Context) {
 
 // Handle initial code-based login for admin
 func initLoginHandler(ctx *gin.Context) {
+	if hasConfiguredAdmins() {
+		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    "Activation code login is not available",
+		})
+		return
+	}
 	db := authDB.Load()
 	if db != nil {
 		ctx.JSON(400,
--- a/web_ui/authentication.go
+++ b/web_ui/authentication.go
@@ -642,10 +642,11 @@ func loginHandler(ctx *gin.Context) {
 // Handle initial code-based login for admin
 func initLoginHandler(ctx *gin.Context) {
 	if hasConfiguredAdmins() {
-		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
-			Status: server_structs.RespFailed,
-			Msg:    "Activation code login is not available",
-		})
+		ctx.JSON(http.StatusBadRequest,
+			server_structs.SimpleApiResp{
+				Status: server_structs.RespFailed,
+				Msg:    "Code-based login is not available",
+			})
 		return
 	}
 	db := authDB.Load()
--- a/web_ui/ui.go
+++ b/web_ui/ui.go
@@ -35,6 +35,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -273,7 +274,6 @@ func handleWebUIRedirect(ctx *gin.Context) {
 
 func handleWebUIAuth(ctx *gin.Context) {
 	requestPath := ctx.Param("requestPath")
-	db := authDB.Load()
 	user, userId, groups, err := GetUserGroups(ctx)
 
 	// Skip auth check for static files other than html pages
@@ -282,9 +282,10 @@ func handleWebUIAuth(ctx *gin.Context) {
 		return
 	}
 
-	// Handle initialization. If db is nill, then redirect user to the initialization page
+	// Handle initialization.
+	// If the web UI is not yet initialized, redirect to the initialization page.
 	if strings.HasPrefix(requestPath, "/initialization") {
-		if db != nil { // Password initialized, redirect away from the init page
+		if shouldSkipActivationFlow() {
 			ctx.Redirect(http.StatusFound, "/view/")
 			ctx.Abort()
 			return
@@ -292,7 +293,7 @@ func handleWebUIAuth(ctx *gin.Context) {
 			ctx.Next()
 			return
 		}
-	} else if db == nil { // For all other paths, if the password is not initialized
+	} else if !shouldSkipActivationFlow() {
 		ctx.Redirect(http.StatusFound, "/view/initialization/code/")
 		ctx.Abort()
 		return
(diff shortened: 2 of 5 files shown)


web_ui/ui.go
```diff
@@ -721,10 +722,40 @@ func configureMetrics(engine *gin.Engine) error {
...
+	configuredAdminsVal  bool
+)
+
+func hasConfiguredAdmins() bool {
+	configuredAdminsOnce.Do(func() {
+		configuredAdminsVal = len(param.Server_UIAdminUsers.GetStringSlice()) > 0 ||
+			len(param.Server_AdminGroups.GetStringSlice()) > 0
+	})
+	return configuredAdminsVal
+}
```
h2zh, line 742: I think this `sync.Once` cache is unneccessary complexity and we should integrate the check into `shouldSkipActivationFlow`, because:

1. Pelican does not hot-reload config, which means there is no in-process viper state mutation that the sync.Once latch is defending against. Every config-touching action triggers a full process re-exec.
2. The cost of `param.*.GetStringSlice()` is cheap. `shouldSkipActivationFlow` is not being called frequently.
3. It cascadingly creates RegisterWebUIReset func and web_ui/reset_register.go file purely to support tests.

web_ui/ui.go
```diff
@@ -721,10 +722,40 @@ func configureMetrics(engine *gin.Engine) error {
...
+
+// shouldSkipActivationFlow reports whether the one-time activation-code
+// flow should be skipped. It returns true if the htpasswd file has been
+// bootstrapped or if at least one admin identity has been configured
+// (meaning an admin can log in without completing the activation flow).
+func shouldSkipActivationFlow() bool {
```
h2zh, line 748: If the user configures Server_UIAdminUsers / Server_AdminGroups but forget to configure OIDC.*, we should throw an error/warning and do not skip the activation flow. Because the configured admin user / group need to match their counterparts from OIDC.

web_ui/ui.go
```diff
@@ -721,10 +722,40 @@ func configureMetrics(engine *gin.Engine) error {
...
+	return authDB.Load() != nil || hasConfiguredAdmins()
```
h2zh, line 749: It would be nice to add a info/debug level log when the Activation Flow is skipped.

web_ui/authentication.go
```diff
@@ -555,6 +555,13 @@ func DowntimeAuthHandler(ctx *gin.Context) {
 func loginHandler(ctx *gin.Context) {
 	db := authDB.Load()
 	if db == nil {
+		if hasConfiguredAdmins() {
+			ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+				Status: server_structs.RespFailed,
+				Msg:    "Password login is not available",
+			})
+			return
```
h2zh, line 563: I got a backward compatibility concern: On a server with Server.UIAdminUsers set and an htpasswd file already exists on disk, password login still works against the htpasswd. So either change the error message, or move this check outside the `db == nil` block.

web_ui/ui.go
```diff
@@ -282,17 +282,18 @@ func handleWebUIAuth(ctx *gin.Context) {
 		return
 	}
 
-	// Handle initialization. If db is nill, then redirect user to the initialization page
+	// Handle initialization.
+	// If the web UI is not yet initialized, redirect to the initialization page.
```
h2zh: I think it is worth to mention the extra condition "and no pre-configured admin user/group"

web_ui/ui.go
```diff
@@ -721,10 +722,40 @@ func configureMetrics(engine *gin.Engine) error {
...
+
 // Send the one-time code for initial web UI login to stdout and periodically
 // re-generate one-time code if user hasn't finished setup
 func waitUntilLogin(ctx context.Context) error {
-	if authDB.Load() != nil {
+	activationFile := param.Server_UIActivationCodeFile.GetString()
+
+	if shouldSkipActivationFlow() {
+		// Remove any stale activation file left from a previous run.
+		_ = os.Remove(activationFile)
```
h2zh: Could you also add a guard here to make sure `activationFile` is not an empty string? Maybe also drop a debug level log if the removal fails.

h2zh (changes requested):
Thanks for working on this ticket! I can see there are many thoughtful ideas in this PR. It would be nice to add a 2×2 table to the PR description showing when activation is skipped, along the axes of UIAdminUsers/AdminGroups set (or not) and htpasswd file present (or not)

web_ui/ui.go
brianaydemir: > Every config-touching action triggers a full process re-exec.

Via `server set-logging-level`, it _is_ possible to update `Logging.Level` without triggering a re-exec of the process. So, the precedent for hot-reloading config has been set.

Thus, the `sync.Once` latch isn't to guard against current behavior. It's to guard against _future_ changes where someone accidentally makes the config parameters editable without an obvious re-exec of the process.

The intended design invariant is that the decision is fixed on "start up", which `sync.Once` makes clear, without needing to understand the rest of the code base.

web_ui/ui.go
brianaydemir: It's obvious from interacting with the web UI whether the activation flow was skipped; there's no need to look in the log file for that information.

I also don't think it's the sort of decision that deserves a permanent record in the log. Presumably, someone knew what they were doing when they configured the admin lists in the first place.

web_ui/ui.go
brianaydemir: I've configured live, working origins without having to override any of the `OIDC.*` config parameters. So, unless you're proposing to check that, say, the live OIDC client ID and client secret values "work", there's not much we can reasonably do here.

web_ui/authentication.go
brianaydemir: Unless I'm mistaken (or Copilot is mistaken…), when `db == nil`, there are no password-based admins configured via `htpasswd`, so if admins are configured via other means, it is correct to reject password-based logins.

web_ui/ui.go
brianaydemir: Added!

web_ui/ui.go
brianaydemir: Added!

web_ui/authentication.go
h2zh: Right, I was mistaken when writing this comment. Btw thanks for adding the table to the PR writeup!

h2zh (approved):
LGTM - Thanks for the thorough thoughts and clear explanation!

PR #3406, opened 2026-04-30
Ignore non-YAML files in ConfigLocations directories

### LINKED ISSUE #2703: Pelican reads backup files in config directory
Pelican Version: 7.19

Pelican Service:
- [ ] Client
- [ ] Plugin
- [x] Registry
- [x] Director
- [x] Origin
- [x] Cache

**Describe the bug**

When editing files in `/etc/pelican/config.d`, sometimes text editors will leave backup files that contain the old version of the config; these files apparently still get read by pelican, which results in confusion as to why the admin's changes aren't being read.  In Linux, these files typically have `~` at the end; on Windows, `.bak`.

In addition, package managers will make backups of config files that were modified by the admin, for example files ending in `.rpmsave` or `.rpmnew` for RPM.  These have the same issue.

**To Reproduce**
1. Open a Pelican config file in `vi`
2. Type `:set backup`
3. Make a change to the config
4. Save and quit (`:wq`)
5. Note the presence of a file ending in `~`
6. Run `pelican config ...` and note that the settings did not take into effect.

**Expected behavior**
At least files with these suffixes should be ignored:
`~`, `.bak`, `.rpmsave`, `.rpmnew`, `.dpkg-old`, `.dpkg-dist`, `.cfsave`

but an allowlist would be more robust than a denylist. Perhaps only parse files ending in `.yaml`, `.yml` (case-insensitive).

brianaydemir:

Pelican reads every file it finds in each `ConfigLocations` directory. This means editor backup files (e.g., `~`, `.bak`) and package-manager saves (e.g., `.rpmsave`, `.rpmnew`) are silently merged into the configuration, causing admin changes to appear to have no effect.

- Add `isSupportedConfigExt` in `config/config.go` — an allowlist predicate that accepts `.yaml` and `.yml` (case-insensitive). This is the single place to update if Pelican ever gains support for additional config formats.
- Update `handleContinuedCfg` to skip any file that does not pass the check.
- Add a `test-backup-files-ignored` subtest to `TestExtraCfg` that places every backup-suffix variant mentioned in the issue alongside a valid YAML file and asserts only the YAML value is loaded.

Fixes #2703.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -506,6 +506,38 @@ func TestExtraCfg(t *testing.T) {
 		assert.Equal(t, "bar", viper.GetString("otherVal"))
 	})
 
+	t.Run("test-backup-files-ignored", func(t *testing.T) {
+		// Verify that editor/package-manager backup files in a ConfigLocations
+		// directory are silently ignored. Only .yaml/.yml files should be read.
+		ResetConfig()
+		dir1 := t.TempDir()
+		setupConfigLocations(t, []string{dir1})
+
+		// The valid YAML file sets the expected value.
+		err := os.WriteFile(filepath.Join(dir1, "config.yaml"), []byte("TestVal: correct"), 0644)
+		require.NoError(t, err)
+
+		// Each backup variant carries a conflicting value that must NOT be loaded.
+		backupFiles := map[string]string{
+			"config.yaml~":       "TestVal: from-tilde",
+			"config.yaml.bak":    "TestVal: from-bak",
+			"config.rpmsave":     "TestVal: from-rpmsave",
+			"config.rpmnew":      "TestVal: from-rpmnew",
+			"config.dpkg-old":    "TestVal: from-dpkg-old",
+			"config.dpkg-dist":   "TestVal: from-dpkg-dist",
+			"config.cfsave":      "TestVal: from-cfsave",
+			"config.noextension": "TestVal: from-noext",
+		}
+		for name, content := range backupFiles {
+			err = os.WriteFile(filepath.Join(dir1, name), []byte(content), 0644)
+			require.NoError(t, err)
+		}
+
+		err = handleContinuedCfg()
+		assert.NoError(t, err)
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -506,6 +506,41 @@ func TestExtraCfg(t *testing.T) {
...
+		backupFiles := map[string]string{
+			"extra_config.yaml.swp":    "TestVal: from-swp",
+			"extra_config.yaml~":       "TestVal: from-tilde",
+			"extra_config.yaml.bak":    "TestVal: from-bak",
+			"extra_config.rpmsave":     "TestVal: from-rpmsave",
+			"extra_config.rpmnew":      "TestVal: from-rpmnew",
+			"extra_config.dpkg-old":    "TestVal: from-dpkg-old",
+			"extra_config.dpkg-dist":   "TestVal: from-dpkg-dist",
+			"extra_config.cfsave":      "TestVal: from-cfsave",
+			"extra_config.noextension": "TestVal: from-noext",
```
matyasselmeci:
```suggestion
			"extra_config.yaml.swp":        "TestVal: from-swp",
			"extra_config.yaml~":           "TestVal: from-tilde",
			"extra_config.yaml.bak":        "TestVal: from-bak",
			"extra_config.yaml.rpmsave":    "TestVal: from-rpmsave",
			"extra_config.yaml.rpmnew":     "TestVal: from-rpmnew",
			"extra_config.yaml.dpkg-old":   "TestVal: from-dpkg-old",
			"extra_config.yaml.dpkg-dist":  "TestVal: from-dpkg-dist",
			"extra_config.yaml.cfsave":     "TestVal: from-cfsave",
			"extra_config":                 "TestVal: from-noext",
```
.rpmsave, .rpmnew, etc. are appended to the extension, they don't replace it. Also, the file for the "no extension" test should probably have no extension :)

matyasselmeci (approved):
A suggested text fixup but otherwise LGTM.

PR #3408, opened 2026-04-30
Hide rclone and SSH origin from docs navigation while keeping pages reachable

### LINKED ISSUE #3400: Hide some features' documentation while they're known not to work

jhiemstrawisc:
The rclone integration and SSH origin backend are not yet functional. Their documentation and config parameters should not be discoverable by regular users, but testers working on bug fixes still need shareable links to the relevant pages.

To accomplish this, the commands are marked as hidden, which prevents our docs generator from producing their pages. However, the docs generation code has a map of which commands should result in generated-but-unlinked docs, and it uses this to un-hide those commands temporarily while their docs are generated. After generation, they're re-hidden and their `_meta.js` entries are skipped, preventing any links from appearing in the docs.

Closes #3400

### DIFF
--- a/cmd/doc_gen.go
+++ b/cmd/doc_gen.go
@@ -29,12 +29,37 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
 )
 
 // angleTagRe matches bare angle-bracket identifiers that MDX would interpret as
 // JSX tags (e.g. <client-id>, <path>, <pelican-url>).
 var angleTagRe = regexp.MustCompile(`<([a-zA-Z][a-zA-Z0-9_-]*)>`)
+
+// hiddenFromDocs lists command paths (relative to the docs root, using forward
+// slashes) whose generated documentation pages should remain on disk and
+// remain reachable by direct URL, but should not be discoverable via the
+// site's navigation. Use this for features that are not yet ready for general
+// use but where targeted users (e.g., bug-fix testers) still need a link to
+// the docs.
+//
+// Effects of inclusion:
+//   - The on-disk page.mdx files for the command (and its subcommands) are
+//     still generated, so links like /commands-reference/<cmd>/ continue to
+//     resolve and can be shared directly. This works even when the cobra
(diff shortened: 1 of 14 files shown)


jhiemstrawisc:
Some of this needs to be merged back into 7.25.x to hide the docs for features there -- however, there would be serious merge conflicts cherry picking this PR directly because of the binary split. This is why I separated the changes to docs generation and the output of that generation into two commits -- hopefully I can cherry pick only the generator changes and then run the generator on the `v7.25.x` branch once this is merged.

cmd/doc_gen.go
```diff
@@ -88,6 +113,13 @@ func generateCLIDocs(outputDir string) error {
 		return fmt.Sprintf("---\ntitle: %s\n---\n\n", title)
 	}
 
+	// Cobra's doc generator skips commands whose Hidden field is true (and,
+	// transitively, all of their subcommands). For features in hiddenFromDocs
+	// we still want the pages generated so testers can reach them via direct
+	// URLs, so temporarily un-hide them for the duration of generation.
+	restoreHidden := temporarilyUnhideForDocs(rootCmd)
```
turetske, line 120: Why not just "defer temporarilyUnhideForDocs(rootCmd)"?

turetske (changes requested):

cmd/doc_gen.go
jhiemstrawisc: This has to be written as two lines rather than `defer temporarilyUnhideForDocs(rootCmd)`. With defer, the call itself is postponed until `generateCLIDocs` returns, meaning the commands would still be hidden during generation. We need `temporarilyUnhideForDocs` to run immediately (to flip `Hidden=false` now), and only its returned restore function to be deferred (to flip Hidden back to true on exit). The key thing is that the function itself returns a function, and it's that returned function that is run deferred.

turetske (approved):

Issue #3361, opened 2026-04-16
Document running in unprivileged user container


bbockelm:
While we have caches running in unprivileged user container setups, the current documentation:

https://docs.pelicanplatform.org/install/docker#understanding-default-paths

always assumes you're running as root -- and what needs to be set is passed around from admin to admin like it's a state secret.  Let's surface these things so we can better help sites that prefer (or require) this mode.

Two things off the bat:
1.  You need to set `ConfigDir` manually, typically to `/etc/pelican`.  This is because the home directory isn't set for the `pelican` user inside the container.  @brianaydemir - should we just set the home dir to `/etc/pelican` as an "ease of use" type thing?
2. You need to set `RuntimeDir` to a directory the Pelican UID can run as.  Maybe also this can be automated in the container?

@williamnswanson / @jthiltges / @brianhlin - any other things come to mind?

brianaydemir:
Can someone share these state secrets with me? And what "unprivileged user container" means? I might be better able to test and devise a solution if I have the full picture.

For example:

1. Setting the `pelican` user's home directory does **not** help when the `pelican-server` binary computes `ConfigDir` as `${HOME}/.config/pelican` — there is no value of `${HOME}` that will get us to `/etc/pelican`.

2. In my quick test, `pelican-server` creates a suitable temporary `RuntimeDir` on its own, [as the documentation suggests should happen](https://docs.pelicanplatform.org/parameters#RuntimeDir).

frobnitzem:
When running the cache for the first time as an unprivileged user, I get the following error:

```
sh-5.1$ /entrypoint.sh pelican-server cache serve
Starting Pelican...
Running pelican-server with arguments: cache serve
[WARN  tini (57)] Tini is not running as PID 1 and isn't registered as a child subreaper.
Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.
INFO[2026-04-27T15:27:06Z] Version: 7.24.2                              
INFO[2026-04-27T15:27:06Z] Build Date: 2026-04-08T16:53:45Z             
INFO[2026-04-27T15:27:06Z] Build Commit: ba034ad3c09b5037a763fe824d04de163e792e1f 
[...]

PR #3410, opened 2026-04-30
Rename `ConfigDir`, cleanup `parameters.yaml`, fix `Cache.LowWaterMark` casing

jhiemstrawisc:
This is the first of several PRs that overhaul how Pelican loads and tracks its configuration. It's meant to be the first cleavage from #3409, which grew too large for a reasonable review, and it's intentionally scoped to the small, mechanical prep work so the meatier PRs that follow can stay focused.

Three things happen here:
1. Rename `ConfigDir` to `ConfigBase`: `ConfigDir` was a historical name suggesting it held only configuration files, but it also holds data (databases, credentials, monitoring). The `parameters.yaml` entry has been called `ConfigBase` for a while; this just makes the Go side match. The old manual `param.ConfigDir` variable is dropped in favor of the generated `param.ConfigBase`. I'm not worried about backwards compatibility because `PELICAN_CONFIGDIR` was internal and undocumented.
2. Fix `Cache.LowWatermark` capitalization: It was inconsistent with `Cache.HighWaterMark` and every other `WaterMark`-style param in that namespace. Renamed to `Cache.LowWaterMark`. This does NOT affect backwards compatibility because viper stores everything internally lowercased — both names resolve to `cache.lowwatermark` under the hood, and can be set with any casing via yaml.
3. Clean up `parameters.yaml`: Adds a header comment documenting the parameter naming rules, the `${Param.Name}` and `$ENV_VAR` reference syntax, and what will turn into five "default tiers" (`default`, `root_default`, `osdf_default`, `client_default`, `server_default`) with their semantics and precedence. Standardizes references to `${...}` syntax. Also removes three redundant `osdf_default: none` entries (`Federation.DirectorUrl`, `.RegistryUrl`, `.JwkUrl`) — `none` is already the universal default for those params, and `osdf_default` only needs to be set when the OSDF binary should differ.

Note that the header documents intent for the `${...}` / `$ENV_VAR` syntax and the default tiers, but none of that machinery is wired up in this PR. A future PR will handle generating actual defaults for all parameters directly from `parameters.yaml`.


brianaydemir:
> I'm not worried about backwards compatibility because PELICAN_CONFIGDIR was internal and undocumented.

But we've been telling people to set it: [#3361](https://github.com/PelicanPlatform/pelican/issues/3361#issuecomment-4328496710).

This change is worth noting in the final release notes.

turetske (changes requested):
LGTM, but needs a rebase and fix.

jhiemstrawisc:
I closed this because I'm going back to the uni-PR approach in #3367 so I can get everything over the finish line atomically.

PR #3416, opened 2026-05-01
Logging metadata param configuration

### LINKED ISSUE #3417: Create Params for the Logging Metadata Namespace
Create the initial parameters needed for logging metadata. Part 1 in a larger process that's being split into much smaller tasks.

turetske:
Add three new config parameters (`Logging.LogExports.Enabled`, `Logging.LogExports.AllowFederationAdmin`, `Registry.EnableAutoLoggingRegistration`), all defaulting to `false`; validate on startup and fail clearly if log export is enabled but no log file path is configured

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -434,6 +434,26 @@ default: error
 components: ["cache"]
 runtime_configurable: true
 ---
+name: Logging.LogExports.Enabled
+description: |+
+  When true, this server exports its logs via the Pelican logging namespace (`/pelican/logging/{Server.ID}`).
+  Clients with the appropriate token can retrieve log entries as virtual time-bucketed objects.
+  Requires `Logging.LogLocation` to be set to a real file path (not empty or `/dev/null`).
+  Defaults to false — log export is opt-in.
+type: bool
+default: false
+components: ["origin"]
+---
+name: Logging.LogExports.AllowFederationAdmin
+description: |+
+  When true, the federation issuer URL is added as an authorized issuer for this server's logging namespace,
+  allowing federation administrators to read logs from this server using a federation-issued token.
+  Has no effect when `Logging.LogExports.Enabled` is false.
+  Defaults to false.
+type: bool
+default: false
+components: ["origin"]
+---
 ############################
 # Federation-Level Configs #
 ############################
@@ -2932,6 +2952,18 @@ description: |+
 type: bool
 default: false
 osdf_default: true
+components: ["registry"]
+---
+name: Registry.EnableAutoLoggingRegistration
+description: |+
+  When true, the Registry automatically approves logging namespace registrations (`/pelican/logging/{Server.ID}`)
+  submitted by Origins or Caches whose main server registration is already approved.
+  If the server registration is still pending at the time the logging namespace is submitted,
+  approval is deferred until the server is approved.
+  This is a federation-level gate distinct from `Logging.LogExports.Enabled` on each individual server.
+  Defaults to false.
+type: bool
+default: false
 components: ["registry"]
 ---
 ############################
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -438,7 +438,8 @@ name: Logging.LogExports.Enabled
 description: |+
   When true, this server exports its logs via the Pelican logging namespace (`/pelican/logging/{Server.ID}`).
   Clients with the appropriate token can retrieve log entries as virtual time-bucketed objects.
-  Requires `Logging.LogLocation` to be set to a real file path (not empty or `/dev/null`).
+  Requires `${Logging.LogLocation}` to be set to a real file path (not empty or `/dev/null`).
+  `${IssuerKeysDirectory}` must also be configured so the server can auto-generate access tokens for the logging namespace.
   Defaults to false — log export is opt-in.
 type: bool
 default: false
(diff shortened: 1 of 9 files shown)


docs/parameters.yaml
```diff
@@ -434,6 +434,26 @@ default: error
...
 ---
+name: Logging.LogExports.Enabled
+description: |+
+  When true, this server exports its logs via the Pelican logging namespace (`/pelican/logging/{Server.ID}`).
+  Clients with the appropriate token can retrieve log entries as virtual time-bucketed objects.
+  Requires `Logging.LogLocation` to be set to a real file path (not empty or `/dev/null`).
+  Defaults to false — log export is opt-in.
+type: bool
+default: false
+components: ["origin"]
```
jhiemstrawisc, line 446: If this only affects Origins, I'd argue it should live under the `Origin.EnableLogExports` key. Isn't the idea that this will also apply to Caches?

docs/parameters.yaml
```diff
@@ -434,6 +434,26 @@ default: error
 components: ["cache"]
 runtime_configurable: true
 ---
+name: Logging.LogExports.Enabled
+description: |+
+  When true, this server exports its logs via the Pelican logging namespace (`/pelican/logging/{Server.ID}`).
+  Clients with the appropriate token can retrieve log entries as virtual time-bucketed objects.
+  Requires `Logging.LogLocation` to be set to a real file path (not empty or `/dev/null`).
```
jhiemstrawisc: In another vein of work, I'm in the process of standardizing how we reference other param variables in our parameters table. Can you change `Logging.LogLocation` to `${Logging.LogLocation}` and do similarly with the other references here?

server_utils/server_utils.go
```diff
@@ -817,3 +817,34 @@ func LaunchConcurrencyMonitoring(ctx context.Context, egrp *errgroup.Group, sTyp
 		return doLoadMonitoring(ctx)
 	})
 }
+
+// ValidateLogExportsConfig checks that the Logging.LogExports configuration is internally
```
jhiemstrawisc: I think config validation should happen during server or config initialization. I see this gets called in launchers, and only for Origins. Why do you think it should happen at this later stage?

server_utils/server_utils.go
```diff
@@ -817,3 +817,34 @@ func LaunchConcurrencyMonitoring(ctx context.Context, egrp *errgroup.Group, sTyp
...
+//     tokens for the logging namespace; a warning is emitted but startup is not blocked.
+func ValidateLogExportsConfig() error {
+	if !param.Logging_LogExports_Enabled.GetBool() {
+		return nil
+	}
+
+	logLocation := param.Logging_LogLocation.GetString()
+	if logLocation == "" || logLocation == "/dev/null" {
+		return errors.New(
+			"Logging.LogExports.Enabled is true but Logging.LogLocation is not set to a real file path; " +
```
jhiemstrawisc: Can you use the `param` package to reference param variables in logs? Also, I don't think `!= ""` is a sufficient check that the log location isn't a valid file path. Can you either parse it to check it's a valid filepath, or change the message to say something like "cannot be empty and cannot be /dev/null"?

server_utils/server_utils.go
```diff
@@ -817,3 +817,34 @@ func LaunchConcurrencyMonitoring(ctx context.Context, egrp *errgroup.Group, sTyp
...
+				"set Logging.LogLocation to enable log export",
+		)
+	}
+
+	if param.IssuerKeysDirectory.GetString() == "" {
+		log.Warn("Logging.LogExports.Enabled is true but IssuerKeysDirectory is not configured; " +
+			"the server will not be able to auto-generate access tokens for the logging namespace")
+	}
```
jhiemstrawisc: It feels like this should be documented alongside the params if it's recommended. FWIW, I suspect an Origin/Cache couldn't start without this, but it's okay to have a defensive check here.

server_utils/server_utils_test.go
```diff
@@ -362,3 +363,74 @@ func TestSetFedTok(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateLogExportsConfig(t *testing.T) {
+	t.Run("no-op when disabled", func(t *testing.T) {
```
jhiemstrawisc: Can you toss in a top-level `t.Cleanup` that resets test state so these don't leave configuration around after they finish?

server_utils/server_utils_test.go
```diff
@@ -362,3 +363,74 @@ func TestSetFedTok(t *testing.T) {
...
+		assert.Contains(t, err.Error(), "Logging.LogLocation")
+	})
+
+	t.Run("error when enabled but LogLocation is /dev/null", func(t *testing.T) {
+		t.Cleanup(ResetTestState)
+		require.NoError(t, param.Logging_LogExports_Enabled.Set(true))
+		require.NoError(t, param.Logging_LogLocation.Set("/dev/null"))
+		err := ValidateLogExportsConfig()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "Logging.LogLocation")
```
jhiemstrawisc: Can you use the param package for config params so we can avoid raw strings here and throughout the test?

jhiemstrawisc (changes requested):

docs/parameters.yaml
turetske: Yes, eventually, which is why this exists here. But I'm implementing just origins for now. I don't want to add "cache" as a component until I get to that point, but I can change it if we want.

jhiemstrawisc (approved):
LGTM! Excited to see where this goes.

PR #3418, opened 2026-05-01
Add CI check enforcing documentation updates on PRs

### LINKED ISSUE #3419: Add a CI check that forces us to assess whether a PR should have docs
The idea is a GHA that fails in the absence of the `no-docs-required` label unless suitable changes also exist in `docs/app`. At the very least, this should force us to think about whether a PR warrants documentation, and can help leave an audit trail if developers skirt around it by adding the `no-docs-required` label to things that clearly should have docs.

jhiemstrawisc:
The idea for this PR came out of a discussion with @williamnswanson. It introduces .github/workflows/enforce-docs-updates.yml, a GitHub Actions workflow that fails any PR against main which neither updates the Pelican docs nor carries a `no-docs-required` label. While nothing can truly _force_ that we developers write documentation, this at least creates some fiction when we don't (and leaves an audit trail for any of us that decide to blow off documentation writing)

This new check:
- Triggers on opened/synchronize/reopened/labeled/unlabeled events so the status is always present and re-runs immediately when the label is toggled
- Requires at least one change under `docs/app/`, excluding the auto-generated `docs/app/commands-reference/` directory and any `_meta.js` navigation files. While this isn't perfect (for example, it doesn't mean you've added anything useful or related to the feature you're working on), it's a starting point.

The `no-docs-required` label provides a low-friction opt-out for PRs that genuinely don't warrant documentation updates (small bug fixes, internal-only changes, CI tweaks, etc.).

### DIFF
--- a/.github/workflows/enforce-docs-updates.yml
+++ b/.github/workflows/enforce-docs-updates.yml
@@ -1,0 +1,89 @@
+name: Enforce Documentation Updates
+
+on:
+  pull_request:
+    branches: [main]
+    # Re-run on label add/remove so toggling `no-docs-required` flips the
+    # check immediately without needing a new push. `synchronize` covers
+    # subsequent pushes; `opened`/`reopened` cover PR creation/reopen.
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  check-docs-updates:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Short-circuit if `no-docs-required` label is present
+        id: check_label
+        run: |
+          set -euo pipefail
+          if jq -e '.pull_request.labels[]?.name | select(. == "no-docs-required")' \
+              "$GITHUB_EVENT_PATH" > /dev/null; then
+            echo "✅ `no-docs-required` label is set; skipping documentation check."
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Check out repository with full history
+        if: steps.check_label.outputs.skip != 'true'
+        uses: actions/checkout@v6
+        with:
+          # Need full history so we can diff against the PR base branch.
+          fetch-depth: 0
+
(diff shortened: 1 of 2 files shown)


jhiemstrawisc:
@brianaydemir you can check half of the action's failure mode by toggling the `no-docs-required` label -- the only thing you can't fully check in this PR alone is that a change in the documentation is sufficient to suppress the failure in the absence of the label.

jhiemstrawisc:
Linking https://github.com/PelicanPlatform/pelican/pull/3420 -- this PR (which _should not_ be merged) touches the docs but does not include the `no-docs-required` label. This demonstrates the other half of the action.

brianaydemir (changes requested):
Perhaps this is just me, but the `no-docs-required` label feels both misleading and confusing, because in the context of a label, "docs" is a concept, not a directory.

---

Conceptually, we have three kinds of "docs":

1. Release notes.
2. The manual at https://docs.pelicanplatform.org.
3. Other sundry files (`README.me`, `CONTRIBUTE.md`, etc.)

Taking this PR as an example:

1. It doesn't require a mention in release notes.
2. It doesn't require an update to the manual.
3. It _does_ require an update to other sundry files.

The PR's `no-docs-required` label clashes with (3).

---

If this were me, my design might be to require every PR include exactly one of

- `updates-manual`
- `skips-updating-manual`

The action then enforces:

1. That the developer documents their decision
2. That the decision aligns with the PR's modifications

This approach neatly avoids the scenario where the developer applies `no-docs-required` and the reviewer accepts it not because no updates to the manual are required, but because the updates are being done on another PR (or no one has time to make the updates, etc.).

And for someone who didn't read the fine-print (`CONTRIBUTORS.md`), an error message mentioning these labels makes it clear that you are expected to consider whether the manual needs updating in-tandem with your code changes.

---

In essence, the labels on a PR are best thought of as documenting what the PR actually does, or what it requires someone downstream to do. Enforcing the presence of 1-out-of-N labels then becomes a means of enforcing that developers and reviewers consider the question of what the PR _should_ be doing.

jhiemstrawisc:
Closing this since I don't think it's very high priority to re-design with my remaining time.

PR #3438, opened 2026-05-07
Downgrade POSIX permission validation failure to a warning

### LINKED ISSUE #3437: Handle the false positive failure for stat-unfriendly mounts in posix storage dir permissions check
Some filesystems synthesize ownership at stat(2) time, such as ResearchDrive origin. So a directory appears to be owned by whichever user inspects it. The startup check in validatePosixPermissions infers permissions from the directory's stat info, which produces false positives on these filesystems and prevents the origin from starting.

h2zh:
## Summary
- Change the per-export daemon-user permission check in `validateExtra` from a hard error (blocking origin startup) to a warning log, so origins on stat-unfriendly filesystems (FUSE, NFS, etc.) can still start. Because those filesystems don't store real uid/gid on disk; instead, they fabricate those values on the fly when system call `stat(2)` is called. As a result, the permission check in `validatePosixPermissions` can be a false positive: stat might report uid=0 (root) or some other synthesized value that doesn't match the XRootD daemon user, but the daemon can actually read/write the directory just fine at runtime.
- Refine the permission error message to include actionable context (daemon user, uid/gid, and the note about false positives on synthesized-ownership filesystems)
- Preserve the existing multiuser-mode skip path unchanged

Closes #3437.
Related prev. PR #3355.

### DIFF
--- a/server_utils/origin_posix.go
+++ b/server_utils/origin_posix.go
@@ -174,10 +174,12 @@ func (o *PosixOrigin) validateExtra(e *OriginExport, _ int) error {
 			"(files are accessed as the requesting user, not the XRootD daemon user)", e.FederationPrefix)
 	} else {
 		if err := o.validatePosixPermissions(e.StoragePrefix, e.Capabilities, e.FederationPrefix); err != nil {
-			log.Warningf("Permission validation failed on storage prefix %q for export %q. "+
-				"The origin will still start, but runtime errors may occur if the XRootD daemon user truly lacks access. "+
-				"This may also be a false positive on filesystems that synthesize ownership at system call stat(2) time (e.g. FUSE, NFS).",
-				e.StoragePrefix, e.FederationPrefix)
+			log.Warningf("Storage prefix %q may lack permissions for export %q: %v. "+
+				"The origin will still start, but runtime errors may occur "+
+				"if the XRootD daemon user truly lacks access. "+
+				"This may be a false positive on filesystems that don't store real uid/gid on disk "+
+				"and synthesize ownership at runtime (e.g. FUSE, NFS, etc.).",
+				e.StoragePrefix, e.FederationPrefix, err)
 		}
 	}
 
@@ -249,7 +251,7 @@ func (o *PosixOrigin) validatePosixPermissions(storagePath string, caps server_s
 		return errors.Wrapf(ErrInvalidOriginConfig,
 			"storage prefix %q for export %q requires %q permissions for the %q user (uid=%d, gid=%d) "+
 				"to support the %q capability, but the current permissions are %q (owner uid=%d, gid=%d). "+
-				"Please adjust the ownership or permissions of the directory so that the XRootD daemon user "+
+				"You may need to adjust the ownership or permissions of the directory so that the XRootD daemon user "+
 				"has the necessary access",
 			storagePath, federationPrefix, requiredPerms, username, uid, gid,
 			capability, mode.Perm().String(), dirUID, dirGID)
--- a/server_utils/origin_posix.go
+++ b/server_utils/origin_posix.go
@@ -174,7 +174,10 @@ func (o *PosixOrigin) validateExtra(e *OriginExport, _ int) error {
 			"(files are accessed as the requesting user, not the XRootD daemon user)", e.FederationPrefix)
 	} else {
 		if err := o.validatePosixPermissions(e.StoragePrefix, e.Capabilities, e.FederationPrefix); err != nil {
-			return err
+			log.Warningf("Permission validation failed on storage prefix %q for export %q. "+
+				"The origin will still start, but runtime errors may occur if the XRootD daemon user truly lacks access. "+
+				"This may also be a false positive on filesystems that synthesize ownership at system call stat(2) time (e.g. FUSE, NFS).",
+				e.StoragePrefix, e.FederationPrefix)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


docs/parameters.yaml
```diff
@@ -1135,6 +1135,21 @@ root_default: true
...
+  has the filesystem permissions required by each export's declared capabilities.
+
+  This validation is normally helpful: it surfaces ownership/permission misconfigurations
+  before they manifest as opaque runtime errors. However, on filesystems that synthesize
+  ownership at stat(2) time, a directory may appear to be owned by whichever user is
+  inspecting it, which can cause this stat-based check to produce false negatives or false
+  positives. Set this to true on such origins to bypass the check. The admin remains
+  responsible for ensuring the daemon user can read/write/list the underlying storage.
+type: bool
+default: false
```
jhiemstrawisc: If we know the validation feature causes problems, it's probably safer to make this true by default or to switch the param name to `Origin.EnablePosixPermissionsCheck`. In general, I think we should stick to verbs like "enable/disable" rather than "skip" in our params so that we have more consistent naming conventions.

docs/parameters.yaml
```diff
@@ -1135,6 +1135,21 @@ root_default: true
...
+name: Origin.SkipPosixPermissionsCheck
+description: |+
+  A bool indicating whether to skip the startup validation that the XRootD daemon user
+  has the filesystem permissions required by each export's declared capabilities.
+
+  This validation is normally helpful: it surfaces ownership/permission misconfigurations
+  before they manifest as opaque runtime errors. However, on filesystems that synthesize
+  ownership at stat(2) time, a directory may appear to be owned by whichever user is
+  inspecting it, which can cause this stat-based check to produce false negatives or false
+  positives. Set this to true on such origins to bypass the check. The admin remains
```
jhiemstrawisc: How can I as an Origin administrator know whether this applies to me? This points out it applies to some filesystems, but doesn't list any common ones an admin might actually use.

server_utils/origin_posix.go
```diff
@@ -172,6 +172,11 @@ func (o *PosixOrigin) validateExtra(e *OriginExport, _ int) error {
 	if param.Origin_Multiuser.GetBool() {
 		log.Infof("Skipping directory permission validation for export %q because multiuser mode is enabled "+
 			"(files are accessed as the requesting user, not the XRootD daemon user)", e.FederationPrefix)
+	} else if param.Origin_SkipPosixPermissionsCheck.GetBool() {
+		log.Warnf("Skipping directory permission validation for export %q because %s is enabled. "+
+			"The XRootD daemon user must still have appropriate permissions on the underlying storage; "+
```
jhiemstrawisc: Since we have to worry about non-xrootd Origins now, does this log message get triggered when running one of them? Does the filesystem validation happen outside of xrootd Origins?

docs/parameters.yaml
```diff
@@ -1135,6 +1135,21 @@ root_default: true
...
 components: ["origin"]
 ---
+name: Origin.SkipPosixPermissionsCheck
+description: |+
+  A bool indicating whether to skip the startup validation that the XRootD daemon user
+  has the filesystem permissions required by each export's declared capabilities.
+
+  This validation is normally helpful: it surfaces ownership/permission misconfigurations
+  before they manifest as opaque runtime errors. However, on filesystems that synthesize
+  ownership at stat(2) time, a directory may appear to be owned by whichever user is
```
jhiemstrawisc: Will the Origin admin know what "stat(2)" refers to, or is this too mountain dewy of a term?

jhiemstrawisc (changes requested):

server_utils/origin_posix.go
h2zh: For non-posix backends (S3, Globus, or other non-xrootd ones), this permission check is completely skipped, because `validateExtra` is a method with a pointer receiver, which is invoked only for posix origin.

server_utils/origin_posix.go
```diff
@@ -174,7 +174,10 @@ func (o *PosixOrigin) validateExtra(e *OriginExport, _ int) error {
 			"(files are accessed as the requesting user, not the XRootD daemon user)", e.FederationPrefix)
 	} else {
 		if err := o.validatePosixPermissions(e.StoragePrefix, e.Capabilities, e.FederationPrefix); err != nil {
-			return err
+			log.Warningf("Permission validation failed on storage prefix %q for export %q. "+
+				"The origin will still start, but runtime errors may occur if the XRootD daemon user truly lacks access. "+
```
bbockelm: Log the xrootd user name / UID / GID here, it may not be obvious.

server_utils/origin_posix.go
```diff
@@ -174,7 +174,10 @@ func (o *PosixOrigin) validateExtra(e *OriginExport, _ int) error {
...
+				"This may also be a false positive on filesystems that synthesize ownership at system call stat(2) time (e.g. FUSE, NFS).",
```
bbockelm: Since we have through the multiuser plugin the ability to switch to the user, should we do that and just use the access syscall instead?

docs/parameters.yaml
h2zh: Examples added.

server_utils/origin_posix.go
h2zh: Done. I added the error message to this log, which contains user name / UID / GID

server_utils/origin_posix.go
h2zh: This validation intentionally runs during Go-level config validation, before xrootd is initialized (GetOriginExports() is called at the top of OriginServe, well before any xrootd process is spawned). We can't call into xrootd plugins at this point because they don't exist yet.

Additionally, the xrootd-multiuser plugin is only loaded when Origin.Multiuser is enabled, which defaults to disable.

h2zh:
@jhiemstrawisc I decided to proceed with carefully refining the wording of the relevant error messages. Would love to hear your thoughts!

jhiemstrawisc (approved):
LGTM.

PR #3452, opened 2026-05-14
Fix HA director cascade-failure bugs in peer discovery and self-advertisement

### LINKED ISSUE #3449: Multi-director federation: a single director outage can permanently break advertisement on the surviving director
### Summary

In a federation with two directors, A and B, if A goes offline then B can enter a state in which:

- B's `/api/v1.0/director/directors` API returns only A (or only stale entries), even though B itself is healthy.
- Origins, caches, and B itself update their advertisement target list to match, and so direct all advertisements at the offline A.
- B's namespace store empties as ads expire, and clients querying B receive HTTP 404 "No sources found".
- The state is self-sustaining while A remains unreachable. Recovery requires A to come back online (so gossip can resume and refresh B's self-entry), a restart of B, or manual intervention.

This was observed in production on 2026-05-13 in the OSDF federation, ~45 minutes after one of the two directors went offline. In that deployment, A was the failed advertise-only endpoint and B was simultaneously the federation's primary `director_endpoint` (the URL clients use), which is why the cascade became externally visible as client 404s.

### How a director's `/directors` response is built

Each director maintains two collections of peer-director information:

- **`directorAds`** — a TTL cache, keyed by director name, of the peers currently considered live. Entries expire automatically.
- **`directorEndpoints`** — a slice of `DirectorAd`s, written exclusively by the periodic `doDiscovery` routine. It does **not** self-evict expired entries.

The `/directors` handler (`listDirectors`) returns `directorAds` if non-empty, and otherwise falls back to `directorEndpoints`.

`doDiscovery` rebuilds `directorEndpoints` once per advertisement interval by querying every configured "seed" URL's own `/directors` endpoint and merging the responses. The configured seed URLs themselves are never written into `directorEndpoints`; only what peers report is. A fallback to `fed.DirectorEndpoint` exists, but it fires only when the merged response is **completely empty**.

Origins and caches advertise to whatever `AdvertiseUrl`s appear in `directorEndpoints`; they do not use the configured `director_endpoint` as a direct advertisement target.

#### Flaw 1 — A director's self-entry in `directorAds` depends on a peer round-trip

[`sendMyAd`](https://github.com/PelicanPlatform/pelican/blob/b3aecd47/director/director_advertise.go#L499-L534) forwards a freshly-built `DirectorAd` for the local director to every non-self peer in `directorAds`, but does **not** insert it into the local cache. B's self-entry is refreshed only when a peer receives the ad and gossips it back via `/registerDirector`.

When A is the only peer and A is unreachable, the round-trip is broken. B's self-entry counts down to expiry with no way to renew it. Once it expires, B's `directorAds` (and therefore B's `/directors`) contains only A. We call this the **`[A only]` window**.

#### Flaw 2 — `doDiscovery` discards seeds that peers fail to confirm

[The `doDiscovery` query loop](https://github.com/PelicanPlatform/pelican/blob/b3aecd47/server_utils/director_discovery.go#L120-L160) merges `/directors` responses into `endpointMap`. Configured seeds — even those known to be reachable, like B itself — are never inserted directly.

During the `[A only]` window, with OSDF's seed configuration (no static `Server.DirectorUrls` overrides), every component (origins, caches, and B itself) runs `doDiscovery` and arrives at `directorEndpoints = [A only]`:

- The query to A times out (A is offline).
- The query to B succeeds, but returns only `[A]`.
- The "completely empty" fallback to `fed.DirectorEndpoint` does not fire, because `endpointMap` is non-empty.

All subsequent advertisements target A and fail. Components stop advertising to any director. B's namespace store empties. Clients receive 404s.

#### Flaw 3 — Expired entries can circulate indefinitely once `directorAds` is empty

Eventually A's entry also expires from B's `directorAds`, leaving it empty. [`listDirectors`](https://github.com/PelicanPlatform/pelican/blob/b3aecd47/director/director_advertise.go#L137-L145) then falls back to `directorEndpoints`, which has no expiration filter. [`doDiscovery`](https://github.com/PelicanPlatform/pelican/blob/b3aecd47/server_utils/director_discovery.go#L151-L159) likewise accepts entries from peer responses without checking `Expiration`. Together these form a closed loop: B serves stale entries from `directorEndpoints`; `doDiscovery` reads them back; `directorEndpoints` is rewritten with the same stale data. Meanwhile `updateInternalDirectorCache` [rejects entries with `adTTL <= 0`](https://github.com/PelicanPlatform/pelican/blob/b3aecd47/director/director_advertise.go#L557-L558), so `directorAds` can never repopulate from this stale data.

The exact stale snapshot that locks in depends on timing:

- If B's self-entry expired before A's (the production scenario), the snapshot is `[A only]`, and the client-visible 404 cascade results.

- If both expire near-simultaneously, the snapshot is `[A_stale, B_stale]`. Origins and caches keep B in their target list and continue to advertise successfully, so clients are unaffected — but B's `/directors` still serves expired data forever.

### Proposed fixes

Each fix targets one of the three flaws and is independently useful; together they harden the system in depth.

#### Fix 1 — `doDiscovery`: keep configured seeds as synthetic fallback entries

After the `doDiscovery` query loop, add any valid configured seed URL that no peer reported as a synthetic entry in `endpointMap` (with `AdvertiseUrl` set and `Name` empty). A scheme/host validity check is required: `fed.DirectorEndpoint` is unconditionally inserted into the seed set even when federation discovery is unconfigured (yielding `""`), so invalid URLs must be filtered before synthetic entries are created.

This restores the missing director — `B` in the cascade above — as a viable advertisement target for origins and caches. They can then continue to advertise successfully to B even while A is offline, eliminating the client-visible 404s.

`advertiseInternal` only requires `AdvertiseUrl` from a `DirectorAd`, so synthetic entries are sufficient for origin/cache advertisement. `updateInternalDirectorCache` rejects entries with an empty `Name`, however, so a synthetic entry alone does not repair B's own `directorAds` — Fix 2 is required for that.

#### Fix 2 — `sendMyAd`: insert the local director's own ad into `directorAds`

Have `sendMyAd` insert the freshly-built ad into the local `directorAds` (under the write lock) before forwarding it to peers. This makes the self-entry's lifetime independent of any peer round-trip, so the `[A only]` window can never open in the first place. B's `/directors` always includes B itself, regardless of peer health.

The redundant outbound forward this triggers is harmless: receiving directors de-duplicate by ad freshness.

#### Fix 3 — `doDiscovery`: filter peer responses by `Expiration` (defense-in-depth)

In the `doDiscovery` inner loop, skip entries whose `Expiration` is in the past. With Fix 2, `directorAds` should never become empty on a healthy director, so this path is unreachable in normal operation. It is a safety net for edge cases such as a director restarting after being offline long enough for all entries to have expired.

### Suggested acceptance tests

- After `sendMyAd`, the local director's ad is present in `directorAds` without any peer round-trip.
- After A goes offline and B's prior self-entry would have expired, `listDirectors` on B still returns B's own ad.
- `doDiscovery`, given configured seeds `[A, B]` where B's `/directors` returns only `[A]` and A is unreachable, produces a result that includes both A (from B's response) and a synthetic B (from the seed).
- `doDiscovery` does not synthesize an entry for an empty or relative `fed.DirectorEndpoint`.
- `doDiscovery` discards entries from peer responses whose `Expiration` is in the past.
- Origin/cache advertisement succeeds (to at least one director) when one of two configured directors is unreachable, without requiring a restart.

---

With a tip of the hat to Copilot.

brianaydemir:
Fixes #3449.

## What broke and why

In a two-director federation (A and B), if A goes offline, B can enter a self-sustaining broken state: B's `/directors` API returns only A, origins and caches redirect all advertisements to the offline A, B's namespace store empties, and clients receive HTTP 404 "No sources found." Recovery requires A to come back online, a restart of B, or manual intervention. This was observed in the OSDF production federation on 2026-05-13.

Three independent design flaws compose into the cascade.

**Flaw 1 — B's self-entry in `directorAds` depends on a peer round-trip.**
`sendMyAd` forwards B's ad to peers but never inserts it into the local `directorAds`. B's self-entry is refreshed only when a peer gossips it back via `/registerDirector`. With A offline, that round-trip is broken. B's self-entry expires, and B's `/directors` starts returning only A.

**Flaw 2 — `doDiscovery` discards seeds that peers fail to confirm.**
The discovery loop merges peer `/directors` responses into `endpointMap`, but never inserts the configured seed URLs themselves. When B's `/directors` returns `[A only]`, `doDiscovery` produces `directorEndpoints = [A only]`. Origins, caches, and B itself all stop advertising to B.

**Flaw 3 — Expired entries circulate indefinitely.**
`doDiscovery` accepts entries from peer responses without checking `Expiration`. `listDirectors` falls back to `directorEndpoints` (which has no expiry filter) when `directorAds` is empty. Together, these form a closed loop where stale entries in `directorEndpoints` are fed back through `doDiscovery` and written back unchanged, with no way to break the cycle.

## Fixes

**Fix 1** (`doDiscovery`) — After the query loop, insert any successfully-contacted seed URL that no peer reported as a synthetic entry in `endpointMap`. This restores B as a viable advertisement target for origins and caches. Only seeds that returned HTTP 200 with valid JSON are synthesized; offline seeds are not.

**Fix 2** (`sendMyAd`) — Self-insert the freshly-built ad into the local `directorAds` under the write lock before forwarding to peers. B's self-entry is now independent of any peer round-trip, so the `[A only]` window cannot open in the first place.

**Fix 3** (`doDiscovery`) — Skip entries in peer responses whose `Expiration` is in the past. This breaks the stale-gossip loop and serves as a defense-in-depth safety net for edge cases such as a director restarting after a long outage.

## Tests

- `TestSendMyAdInsertsIntoDirectorAds` — Fix 2: self-entry present in `directorAds` immediately after `sendMyAd`.
- `TestListDirectorsReturnsSelfAfterPeerExpires` — Fix 2: self-entry still present after all peer ads expire.
- `TestDoDiscoveryAddsSyntheticSeedMissingFromPeerResponse` — Fix 1: online seed omitted from all peer responses appears as a synthetic entry.
- `TestDoDiscoveryNoSyntheticEntryForOfflineSeed` — Fix 1: genuinely offline seed is not synthesized.
- `TestDoDiscoveryFiltersExpiredPeerEntries` — Fix 3: peer-reported entries with a past `Expiration` are excluded.
- `TestDirectorShutdown` (rewritten) — end-to-end: a peer whose listener is closed (connection refused) disappears from `directorEndpoints` within one ad lifetime.

---

With a tip of the hat to Copilot.

### DIFF
--- a/director/director_advertise.go
+++ b/director/director_advertise.go
@@ -1,6 +1,6 @@
 /***************************************************************
  *
- * Copyright (C) 2025, Pelican Project, Morgridge Institute for Research
+ * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License.  You may
@@ -517,6 +517,13 @@ func sendMyAd(ctx context.Context) {
 		AdvertiseUrl: adUrl,
 	}
 	directorAd.Initialize(name)
+
+	// Insert the current director's ad into directorAds so that its
+	// lifetime is independent of any gossip round-trip through a peer.
+	egrp := ctx.Value(config.EgrpKey).(*errgroup.Group)
+	directorAdMutex.Lock()
+	updateInternalDirectorCache(ctx, egrp, directorAd)
+	directorAdMutex.Unlock()
 
 	directorAdMutex.RLock()
 	defer directorAdMutex.RUnlock()
--- a/director/director_advertise_self_test.go
+++ b/director/director_advertise_self_test.go
@@ -1,0 +1,100 @@
+/***************************************************************
+ *
+ * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License.  You may
+ * obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ***************************************************************/
+
+package director
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/pelicanplatform/pelican/config"
+	"github.com/pelicanplatform/pelican/param"
+	"github.com/pelicanplatform/pelican/test_utils"
+)
+
+// TestSendMyAdInsertsIntoDirectorAds verifies that sendMyAd inserts the
+// local director's own entry into directorAds without requiring a gossip
+// round-trip from a peer.
+func TestSendMyAdInsertsIntoDirectorAds(t *testing.T) {
+	config.ResetConfig()
+	t.Cleanup(config.ResetConfig)
+
+	const selfName = "test-self-director"
+	require.NoError(t, param.Server_ExternalWebUrl.Set("http://self.example.com"))
+
+	setupForwardingState(t)
+	directorName = selfName
+
+	ResetState()
+	t.Cleanup(ResetState)
+
+	ctx, cancel, _ := test_utils.TestContext(context.Background(), t)
+	defer cancel()
+
+	sendMyAd(ctx)
+
+	item := directorAds.Get(selfName)
+	require.NotNil(t, item, "directorAds must contain the self-entry after sendMyAd")
+	require.NotNil(t, item.Value())
+	require.NotNil(t, item.Value().ad)
+	assert.Equal(t, selfName, item.Value().ad.Name)
+	assert.Equal(t, "http://self.example.com", item.Value().ad.AdvertiseUrl)
+}
+
+// TestListDirectorsReturnsSelfAfterPeerExpires verifies the key
+// recoverability invariant: after all peer ads expire, the next sendMyAd
+// call ensures the self-entry is present in directorAds so that
+// listDirectors returns it.
+func TestListDirectorsReturnsSelfAfterPeerExpires(t *testing.T) {
+	config.ResetConfig()
+	t.Cleanup(config.ResetConfig)
+
+	const selfName = "test-self-director"
+	require.NoError(t, param.Server_ExternalWebUrl.Set("http://self.example.com"))
+
+	setupForwardingState(t)
+	directorName = selfName
+
+	ResetState()
+	t.Cleanup(ResetState)
+
+	// Register a peer with a very short TTL to simulate it expiring.
+	makeTestDirector(t, "peer-director", 1*time.Millisecond)
+
+	// Wait for the peer entry to be lazily evicted via Items() — no time.Sleep.
+	require.Eventually(t, func() bool {
+		return len(directorAds.Items()) == 0
+	}, 1*time.Second, 5*time.Millisecond, "peer ad should expire")
+
+	ctx, cancel, _ := test_utils.TestContext(context.Background(), t)
+	defer cancel()
+
+	sendMyAd(ctx)
+
+	items := directorAds.Items()
+	require.Len(t, items, 1, "directorAds must contain exactly the self-entry")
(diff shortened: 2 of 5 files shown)


jhiemstrawisc (approved):
I skimmed the diff and this seems like a reasonable fix to the problem you presented. I also threw this at an AI to assist with the review and gives it the thumbs up.

Thanks!

jhiemstrawisc:
Do we have specific patch targets for this? 7.26 I assume, but should this be patched into a new 7.25 RC?

brianaydemir:
@jhiemstrawisc @turetske @h2zh This needs to be patched in to 7.25 and 7.26.

Issue #3451, opened 2026-05-14
Default values for Storage Monitoring Thresholds should be higher than the default value of Cache.HighWaterMark


williamnswanson:
Currently the Default values for Cache.HighWaterMark and Cache.LowWaterMark cause the Cache's storage space to oscillate between 90% and 95% usage, but the default values for Monitoring.StorageWarningThreshold and Monitoring.StorageCriticalThreshold will put a cache into Warning / Critical Status when the filesystem for any monitored director (including the Cache's StorageLocation) reaches 80% or 90% (respectively).

Since this causes the Degraded status before the Cache reaches the high watermark (and therefore before any cache purging is ever done), normal usage of a default-configuration cache will lead to an unrecoverable state of non-direction.

We should modify the default values of the relevant config knobs, raising the Storage thresholds ()

John Thiltges' initial suggestion:

- xrootd low=85%, high=89% (to keep it under 90% due to issues with fragmentation, slow writes in nearly-full filesystems )
- storage warning=92%, critical=97% (these should be abnormal conditions)

Additionally the Config knobs for Storage Monitoring are marked as "hidden", if there isn't a pressing reason to the contrary, can these be made public / un-hidden? [link](https://github.com/PelicanPlatform/pelican/blob/b3aecd47546ffe40fcb379452af5c2728aa2467b/docs/parameters.yaml#L4268-L4294
)
-  Monitoring.StorageHealthCheckInterval
-  Monitoring.StorageWarningThreshold
-  Monitoring.StorageCriticalThreshold

jhiemstrawisc:
@h2zh and I discussed this -- we'll address this issue in two parts:
1. We'll unhide documentation for the requested parameters, and we'll patch this docs-only change into an RC for `v7.25.0`. This RC won't require any testing because it is docs only. We'll also patch it into the `v7.26.` branch.
2. Separately, we'll adjust the default parameters per John's suggestions. This will be patched into a release candidate for `v7.25.1` so as not to block the `v7.25.0` release. It will also be patched into the `v7.26` branch.

jhiemstrawisc:
This wasn't properly linked to #3454, so it wasn't closed automatically when it should have been. Closing now!

PR #3453, opened 2026-05-15
Unhide `Monitoring.StorageXXX` params

jhiemstrawisc:
The default values of these params are problematic because they cause the cache to enter a storage-related "critical" status before the cache hits its high watermark. Because of the critical status, the cache is filtered from the federation and never hits its HWM, and is thus never cleaned to a point that would take it out of that state.

This is the first of a set of patches to fix the issue -- in this pass we only unhide the params so we can advise cache admins how to mitigate the problem via configuration. We'll adjust the defaults in a different patch (so as not to hold up the current release candidate that's already 3 months late).

@h2zh this should be included in the `v7.25.x` branch before release of `v7.25.0` -- it should not require testing because it is docs-only
@turetske, this should also be patched into `v7.26.x`

See issue #3451

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -4271,7 +4271,6 @@ description: |+
   The interval at which the server checks filesystem storage consumption for health monitoring.
 type: duration
 default: 5m
-hidden: true
 components: ["origin", "cache", "director", "registry", "broker", "localcache"]
 ---
 name: Monitoring.StorageWarningThreshold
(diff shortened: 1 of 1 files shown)


brianaydemir (approved):
LGTM.

h2zh:
This PR is cherry-picked into branch `v7.25.x` and will be released in v7.25.0 (as long as the v7.25.0-rc.4 works well in UWDF)

PR #3456, opened 2026-05-15
Avoid duplicate OIDC metadata endpoints registration

### LINKED ISSUE #3455: Fix duplicate OIDC endpoint registration
OIDC endpoint gets registered twice during the origin startup.

The bug was introduced in https://github.com/PelicanPlatform/pelican/commit/8019724f4129d6a815dd3287da66d4f81104c7da and will cause the origin panic.

h2zh:
- Remove duplicate OIDC metadata endpoints registration
- Also drop ReadOnlyMiddleware from the OIDC route group, because it's a functional no-op for these routes.
  - ReadOnlyMiddleware only rejects requests when both `Server.WebReadOnly=true` AND the method is outside {GET, HEAD, OPTIONS, TRACE, PROPFIND}.
  - RegisterOIDCAPI registers `/.well-known/openid-configuration` and JWKS. They are all GETs. They're already in the safe-methods list, so the middleware can never fire on these paths.

Closes #3455

### DIFF
--- a/launchers/origin_serve.go
+++ b/launchers/origin_serve.go
@@ -147,13 +147,6 @@ func OriginServe(ctx context.Context, engine *gin.Engine, egrp *errgroup.Group, 
 		}
 	}
 
-	// Register OIDC metadata endpoints
-	// Director also registers this metadata URL; avoid registering twice.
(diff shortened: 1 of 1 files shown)


CannonLock (approved):
LGTM

PR #3457, opened 2026-05-15
Clean up director test files at caches

### LINKED ISSUE #3065: Clean up Director tests at Cache
There are a few ongoing issues with Director tests at the cache. First we noticed that unless the cache's high water mark is hit, Director tests aren't cleaned up (sometimes leading to hundreds of thousands of files in a single directory). This in turn starts affecting the filesystem, which bogs down the Cache.

We need two main fixes:
1. Director tests should be cleaned up periodically by the Pelican process so we're not relying on XRootD's purge process. This unfortunately means we need to deal with these files being owned by XRootD, but being modifiable by Pelican (a pitfall with `dropprivs`).
2. Director test files are already uniquely timestamped, but we need to nest them in a daily directory. That is, every day we should use a new directory to make sure that no directory contains more than a single day's worth of files.

h2zh:
Closes #3065.

This PR introduces two parallel cleanup mechanisms for the director-test files that accumulate at caches, plus per-director and daily-nested path organization to bound single-directory growth and eliminate cross-director conflicts under a shared cache.

## Background

Each director's cache health test writes a small file under the cache's monitoring namespace. Previously these files lived flat under `/pelican/monitoring/` and were only swept by a per-minute pelican-side `os.Remove` scan — which became a no-op some time. Over time, single directories accumulated tens of thousands of stale test files.

### Per-director, daily-nested layout

Test files are now written under a per-director, per-day subdirectory:

`/pelican/monitoring/directorTest/<director-hostname>/YYYY-MM-DD/director-test-<RFC3339>.txt`

The `<director-hostname>` segment (from Director's `Server.Hostname`) keeps multiple directors from colliding under a shared cache (see *Multi-director coordination* below). The per-day segment bounds single-directory growth and lets older day-directories be removed wholesale.

### Two cleanup types

This PR introduces and clearly separates two cleanup mechanisms:

**1. Director-initiated cleanup.** After every successful cache health test, the director calls a new cache endpoint to evict the previous cycle's test file:

```
POST /api/v1.0/cache/evictTestFile
Authorization: Bearer <federation-issued token>
Body: {"path": "/pelican/monitoring/directorTest/<hostname>/YYYY-MM-DD/director-test-X.txt"}
```

The handler verifies a federation-scoped token, matches the request path against a strict regex (rejecting traversal), and then mints a locally-trusted token to call its own xrdhttp-pelican evict API. **A 200 OK from this path does *not* mean the file is unlinked from disk — it means xrootd's PFC has accepted the file as a purge candidate. The actual physical removal happens during PFC's periodic purge cycle (`pfc.diskusage … purgeinterval …`), which only acts when the configured disk-usage thresholds are crossed.** During that purge cycle, both the data block and the meta block (`.txt.cinfo`) are removed as a unit.

Logs you may see in Pelican Cache
```
scitokens_Access: Grant authorization based on scopes for operation=del, path=/pelican/monitoring/directorTest/.../director-test-2026-06-17T15:07:09Z.txt
pelican_evict: Evicted path /pelican/monitoring/directorTest/.../director-test-2026-06-17T15:07:09Z.txt
Marked director test file as PFC purge candidate via local plugin: .../director-test-2026-06-17T15:07:09Z.txt
Served Request  ... resource=/api/v1.0/cache/evictTestFile status=200
```

**2. Cache-initiated cleanup.** The cache also runs a daily scan as a safety net for stragglers the director couldn't evict (e.g., when a director restarts mid-cycle, an XRootD error):

- For each per-director subtree, removes day-directories older than today wholesale and trims today's directory to the latest test file + `.cinfo` file pair. This rule applies *per director*, so multiple directors retain their own latest pair independently.
- Sweeps legacy flat files from the pre-PR layout.

Under `Server.DropPrivileges`, this scan routes deletions through the same evict API used by the director, since the pelican process cannot unlink xrootd-owned files directly. Under normal privileges it uses `os.Remove` (the content blocks in `data/`/`meta/` are still reclaimed later by PFC's purge).

## Cleanup matrix

| Cleanup type | Trigger | Normal mode (`DropPrivileges=false`) | `Server.DropPrivileges=true` |
|---|---|---|---|
| **Director-initiated** | After each successful director test (~15 s) | ✅ via evict API (marks as PFC purge candidate) | ✅ via evict API (marks as PFC purge candidate) |
| **Cache-initiated** | Periodic scan inside the cache every day | ✅ via `os.Remove` (immediate symlink unlink; content blocks reclaimed by PFC) | ✅ via evict API (marks as PFC purge candidate) |

Both cleanup paths run in both modes. 

## Multi-director coordination

When multiple directors probe the same cache, each director writes into its own `directorTest/<hostname>/…` subtree and only asks the cache to evict files *it* created. The cache-initiated cleanup applies the "keep most recent test file + `.cinfo` file" rule independently per subtree, so no director can prune another's in-flight test file.

| Scenario | New behavior |
|---|---|
| Two directors writing simultaneously | Separate subtrees per hostname; each director's most-recent pair preserved |
| One director restarts mid-cycle | Orphan stays in that director's subtree; cleaned up next Cache-initiated pass |

Hostname is sourced from `Server.Hostname`; the director errors out if it's unset.

## Implementation notes

- New `buildCacheNsAdsWithIssuer` helper ensures the cache's generated `scitokens-cache-generated.cfg` includes the monitoring issuer **at startup**, eliminating a ~3-minute window where every locally-minted evict token was rejected with 403 until the runtime maintenance loop wrote the issuer entry.
- The evict token uses `config.GetServerIssuerURL()` (respecting `Server.IssuerUrl` precedence), `AddAudienceAny()` (the cache's scitokens.cfg has no `audience_json`, so the WLCG "any" audience is required), and a `storage.modify` scope on the *day-directory* path stripped of the issuer's `base_path` prefix (xrootd's scitokens plugin treats scope paths as relative to `base_path`).

## Known limitations

- **PFC defers physical removal.** As called out above, evict API 200 OK means "purge candidate." Per-test-file cost is small (~134 B per cycle) and PFC's high-water purge reclaims it under disk pressure.
- **Empty day-directories** are left behind under `DropPrivileges` mode (xrdhttp-pelican's evict API is file-only). XRootD's own housekeeping reaps these.


director/cache_monitor_test.go
```diff
@@ -0,0 +1,47 @@
+/***************************************************************
+ *
+ * Copyright (C) 2024, Pelican Project, Morgridge Institute for Research
```
jhiemstrawisc: It's 2026

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+func HandleDirectorEvictRequest(ctx *gin.Context) {
+	status, ok, err := token.Verify(ctx, token.AuthOption{
+		Sources: []token.TokenSource{token.Header},
+		Issuers: []token.TokenIssuer{token.FederationIssuer},
+		Scopes:  []token_scopes.TokenScope{token_scopes.Pelican_DirectorTestReport},
+	})
+	if !ok || err != nil {
+		ctx.JSON(status, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    fmt.Sprint("Failed to verify the token: ", err),
```
jhiemstrawisc: This message won't return correctly if `!ok` with a nil error.

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+	if err := ctx.ShouldBindJSON(&reqBody); err != nil || reqBody.Path == "" {
+		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    "Missing or invalid 'path' in request body",
+		})
+		return
+	}
+
+	// Validate the path is under the monitoring namespace to prevent arbitrary evictions
+	if !strings.HasPrefix(reqBody.Path, server_utils.MonitoringBaseNs+"/") {
```
jhiemstrawisc: What about `/pelican/monitoring/../../some/other/path`? I think you should a stronger path parser here.

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+
+// evictViaLocalPlugin calls the cache's own xrdhttp-pelican evict API with a locally-minted
+// token. This mirrors how the cache self-test (xrootd/self_monitor.go generateFileTestScitoken)
+// accesses its own XRootD port using the server's own issuer, which is already trusted by the
+// cache's scitokens configuration.
+func evictViaLocalPlugin(ctx context.Context, testFilePath string) error {
+	issuerUrl := param.Server_ExternalWebUrl.GetString()
+	if issuerUrl == "" {
+		return errors.New("Server_ExternalWebUrl is empty; cannot mint evict token")
+	}
```
jhiemstrawisc: This is an anti-pattern to avoid -- use `config.GetServerIssuerUrl()`. Also, please avoid raw param strings like this in error messages.

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+
+	tokenCfg := token.NewWLCGToken()
+	tokenCfg.Lifetime = time.Minute
+	tokenCfg.Issuer = issuerUrl
+	tokenCfg.Subject = "cache"
+	tokenCfg.AddResourceScopes(token_scopes.NewResourceScope(token_scopes.Wlcg_Storage_Modify, "/"))
+	tokenCfg.AddAudienceAny()
```
jhiemstrawisc, line 412: This one probably _shouldn't_ have the any audience, since it's meant to work at exactly one service.

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+	issuerUrl := param.Server_ExternalWebUrl.GetString()
+	if issuerUrl == "" {
+		return errors.New("Server_ExternalWebUrl is empty; cannot mint evict token")
+	}
+
+	tokenCfg := token.NewWLCGToken()
+	tokenCfg.Lifetime = time.Minute
+	tokenCfg.Issuer = issuerUrl
+	tokenCfg.Subject = "cache"
+	tokenCfg.AddResourceScopes(token_scopes.NewResourceScope(token_scopes.Wlcg_Storage_Modify, "/"))
```
jhiemstrawisc: Can this be scoped to the directory or file that needs to be removed?

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+	tokenCfg.AddAudienceAny()
+
+	tok, err := tokenCfg.CreateToken()
+	if err != nil {
+		return errors.Wrap(err, "failed to create evict token")
+	}
+
+	cacheUrl := param.Cache_Url.GetString()
+	if cacheUrl == "" {
+		return errors.New("Cache.Url is empty; cannot call evict API")
```
jhiemstrawisc: Raw string param.

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+	}
+
+	evictUrl, err := url.Parse(cacheUrl)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse Cache.Url")
```
jhiemstrawisc: Raw string param.

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+	}
+	evictUrl.Path = "/pelican/api/v1.0/evict"
+	q := evictUrl.Query()
+	q.Set("path", testFilePath)
+	q.Set("authz", "Bearer "+tok)
+	evictUrl.RawQuery = q.Encode()
+
+	client := http.Client{Transport: config.GetTransport()}
```
jhiemstrawisc: Should grab the client from config.

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, evictUrl.String(), nil)
```
jhiemstrawisc, line 430: Why not make this more restful with a `DELETE` to `/pelican/api/v1.0/evic/<path>`?

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to send evict request")
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	}
+	body, _ := io.ReadAll(resp.Body)
+	return fmt.Errorf("evict API returned status %d: %s", resp.StatusCode, string(body))
```
jhiemstrawisc: Can you quote the formatted body so it's apparent to the message consumer if it's empty?

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+}
+
+// Periodically scan the directorTest directory to clean up test files.
+// Handles both legacy flat files and daily-nested subdirectories (YYYY-MM-DD/).
+//
+// This serves as a backup cleanup mechanism. The primary cleanup is performed by the
+// director, which asks the cache to evict old test files via the cache web API endpoint
+// POST /api/v1.0/cache/evictTestFile (see HandleDirectorEvictRequest). This local cleanup
+// catches files that the director failed to evict (e.g., due to network issues or director restarts).
```
jhiemstrawisc, line 454: If drop privileges is on and the Director fails to clean something up because it restarts, what then? Why not have the Cache routine manage its own evictions via the local evict plugin?

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+//
+// Note: when Server.DropPrivileges is true, this function is a no-op because the Pelican
+// process cannot delete files owned by the xrootd user. In that case, the director-side
+// evict API is the only cleanup path.
+func LaunchDirectorTestFileCleanup(ctx context.Context) {
+	// Skip if drop privileges is enabled, because Director test files are owned by the xrootd user.
+	// The unprivileged pelican user does not have permission to remove them.
```
jhiemstrawisc: But it _could_ evict using its own evict API, no?

cache/cache_api.go
```diff
@@ -51,57 +61,243 @@ func CheckCacheSentinelLocation() error {
...
+		return errors.New("Cache.Url is empty; cannot call evict API")
+	}
+
+	evictUrl, err := url.Parse(cacheUrl)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse Cache.Url")
+	}
+	evictUrl.Path = "/pelican/api/v1.0/evict"
+	q := evictUrl.Query()
+	q.Set("path", testFilePath)
```
jhiemstrawisc, line 425: What does this do with the directories after we move onto a new daily directory? Do they get cleaned up?

jhiemstrawisc (changes requested):
Other than the inline comments, my biggest question is how this works in a multi-director setup where each director is probing/testing the cache. Since this seems to keep the 2 most recent test files, what happens if three directors and the cleanup routine all fire very close together? Maybe the test file names need some kind of identifier for the Director that created the file?

cache/cache_api.go
h2zh: Added a regex to enforce the path constraint.

cache/cache_api.go
h2zh: Sharp catch!

cache/cache_api.go
h2zh: I'm afraid this is unlikely to happen. xrdhttp-pelican currently only supports `GET /evict?path=...`. Switching to DELETE requires changes to the xrdhttp-pelican plugin and a release there.

cache/cache_api.go
h2zh: Right, thanks for the spark!

cache/cache_api.go
h2zh: Thanks for the suggestion! See the new table in the PR description that demonstrates the revised approaches in different scenarios.

cache/cache_api.go
h2zh: I made some attempts but eventually go back to use the WLCG "any" audience. The cache's scitokens.cfg is generated without a [Global] audience_json entry (WriteCacheScitokensConfig does not populate cfg.Global.Audience the way WriteOriginScitokensConfig does), so the scitokens plugin rejects tokens with any specific audience. Switching to a scoped audience would require extending WriteCacheScitokensConfig to add a cache-side audience first, which goes beyond this PR's scope and might cause unexpected behavior.

cache/cache_api.go
h2zh: See the table above.

h2zh:
@jhiemstrawisc Thanks a lot for your inputs! I pushed a few new commits (after '[Address PR reviews](https://github.com/PelicanPlatform/pelican/pull/3457/commits/ff57db19598f9c2a17b6880c2bd91739b4ccff96)') and revamped the PR description to reflect the revised logics.

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+	if err != nil {
+		return err
+	}
+	return evictViaLocalPlugin(ctx, nsPath)
+}
+
+// removeTestDir deletes every file in dirPath. With privileges, it uses os.RemoveAll
+// and removes the directory itself. Under DropPrivileges it evicts each file via the
+// API and leaves the now-empty directory in place — xrootd reaps empty dirs on its own.
+func removeTestDir(ctx context.Context, dirPath string) error {
```
jhiemstrawisc: The function name "removeTestDir" is misleading if some cases don't result in the actual directory being removed. Maybe you can use `cleanTestDir` since the directory is always cleaned but only sometimes removed?

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+		return os.RemoveAll(dirPath)
+	}
+	entries, err := os.ReadDir(dirPath)
+	if err != nil {
+		return err
+	}
+	var firstErr error
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue // daily directories are flat; ignore unexpected nesting
```
jhiemstrawisc: Is this the right call? If we expect a flat directory but we somehow encounter a nested directory, what does that imply? My guess is it means something really strange happened, and we should still be responsible for cleaning up the bad state we left. If you disagree, can you leave an explicit comment outlining why you make this choice?

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+		}
+		if err := removeTestFile(ctx, filepath.Join(dirPath, entry.Name())); err != nil && firstErr == nil {
```
jhiemstrawisc: My concern with this iterative approach comes from the worst case where we're cleaning a day old director test directory that for some reason has all its original test files. In that case, there are `60 * 60 * 24 / 15 =  5760` files to clean. If we call the evict API because we're in drop privs mode, will we end up making 5760 different HTTP requests? Is it possible to batch the evictions? If you're confident we don't end up in a bad state like this, can you leave a comment that explains consequences to a future developer who may grab the function without thinking through potential consequences?

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+			firstErr = err
+		}
+	}
+	return firstErr
```
jhiemstrawisc, line 151: Why only the first error and not an aggregation of errors we might encounter?

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+}
+
+// fsToNamespacePath strips the cache namespace-location prefix from an absolute
+// filesystem path, returning the namespace-relative path the evict API expects.
+func fsToNamespacePath(fsPath string) (string, error) {
+	nsLoc := strings.TrimRight(param.Cache_NamespaceLocation.GetString(), "/")
+	if nsLoc == "" {
+		return "", errors.New("cache namespace location is not configured")
```
jhiemstrawisc: Unless it's configured as just `/`, which could be the case. Also, rather than referirng to `cache namespace location`, can you use the actual param name so the log viewer immediately knows which knob they need to investigate?

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+	}
+	if !strings.HasPrefix(fsPath, nsLoc+"/") {
+		return "", fmt.Errorf("path %q is not under the cache namespace location", fsPath)
```
jhiemstrawisc: Can you also include the cache namespace location and knob name in the error message?

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+	}
+	return strings.TrimPrefix(fsPath, nsLoc), nil
+}
+
+// cleanupDirectorTestFiles removes old director test files from the directorTest directory.
+// It handles two layouts:
+//   - Legacy flat files: directorTest/director-test-*.txt (pre-PR, swept entirely).
+//   - Per-director daily-nested: directorTest/<id>/YYYY-MM-DD/director-test-*.txt
+//     (current). Per-director subtrees are recursed into; within each, day-dirs older
+//     than today are removed wholesale and today's dir is trimmed to the latest 2 files.
```
jhiemstrawisc, line 177: What happens right at midnight? Can we wind up in a situation where "yesterday's" directory is still being used for ongoing director tests because we're within a few seconds of midnight, but this fires and tries to clean those files? If so, an easy way to get around it would be scheduling this to fire sufficiently far from midnight, since presumably we only need to check for "yesterday's" directory once a day.

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+//
+// Under Server.DropPrivileges, removals are routed through the cache's evict API
+// (see removeTestFile / removeTestDir) since the pelican process cannot directly
+// delete xrootd-owned files.
+func cleanupDirectorTestFiles(ctx context.Context, dirTestPath string) error {
```
jhiemstrawisc, line 184: Can you document what forms you expect the `dirTestPath` input to take? The docstring mentions `directorTest/director-test-*.txt` and `directorTest/<id>/YYYY-MM-DD/director-test-*.txt` which make me think you're expecting a full filepath as input, but the function errors out on non-directories.

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+				log.WithError(err).Warnf("Failed to clean up today's director test directory: %s", dateDir)
+			}
+		}
+	}
+	return nil
+}
+
+// cleanupOldFilesInDir removes all but the keepCount most recent files in a directory.
+// Files are sorted by name (which includes an RFC3339 timestamp), so the last entries
+// are the most recent.
```
jhiemstrawisc: Objects in the cache are typically represented with two files, the data file and the `.cinfo` file, e.g. `T09.txt` and `T09.txt.cinfo`. Consider the case where something strange happens to the cache resulting in one of these being orphaned, or firing while one is still being created. You might end up with a sorted list of files like:
```
T10.txt
T9.txt.cinfo
T9.txt
```

I don't think the "keep 2" rule as implemented will end up cleaning the right files because it has no knowledge of the pairing between a cinfo file and a data file.

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+	return nil
+}
+
+// HandleDirectorEvictRequest handles eviction requests from the director. When the director
+// completes a successful cache health test, it asks the cache to evict the previous test file
+// via this endpoint. The cache then calls its own xrdhttp-pelican evict API using a locally-
+// minted token, matching the self-test authorization pattern (see xrootd/self_monitor.go).
+//
+// This avoids modifying the cache's scitokens configuration to trust the federation issuer,
+// since the cache already trusts its own issuer for the /pelican/monitoring namespace.
```
jhiemstrawisc: I'd argue the cache _should_ trust the federation issuer for monitoring namespaces as a matter of principle, but we don't need to pull on that thread here. Maybe the comment can be modified to remove the suggestion that this is the way we want things to be.

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+			Status: server_structs.RespFailed,
+			Msg:    "Eviction failed: " + err.Error(),
+		})
+		return
+	}
+
+	log.Debugf("Successfully evicted director test file via local plugin: %s", reqBody.Path)
+	ctx.JSON(http.StatusOK, server_structs.SimpleApiResp{
+		Status: server_structs.RespOK,
+		Msg:    "Eviction successful",
```
jhiemstrawisc: Maybe instead of "Eviction successful" you can say "Eviction successfully scheduled"

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+		return errors.Wrap(err, "cannot mint evict token: failed to determine server issuer URL")
+	}
+	cacheUrl := param.Cache_Url.GetString()
+	if cacheUrl == "" {
+		return errors.Errorf("%s is empty; cannot call evict API", param.Cache_Url.GetName())
+	}
+	tokenCfg := token.NewWLCGToken()
+	tokenCfg.Lifetime = time.Minute
+	tokenCfg.Issuer = issuerUrl
+	tokenCfg.Subject = "cache"
```
jhiemstrawisc: Can you use the cache's URL as the subject so that our subject's are globally unique?

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return errors.Wrap(err, "failed to send evict request")
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	}
+	body, _ := io.ReadAll(resp.Body)
```
jhiemstrawisc, line 444: Why throw this error on the floor?

director/cache_monitor_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/pelicanplatform/pelican/server_utils"
+)
+
+// TestCacheTestFilePath verifies that runCacheTest constructs the correct
+// per-director daily-nested test file path format:
+// /pelican/monitoring/directorTest/<director-id>/YYYY-MM-DD/director-test-<RFC3339>.txt
+func TestCacheTestFilePath(t *testing.T) {
```
jhiemstrawisc: This test is tautological and does not cover any source code.

cache/cache_cleanup_test.go
```diff
@@ -0,0 +1,242 @@
...
+			dayDir := filepath.Join(directorSubtree, day)
+			require.NoError(t, os.MkdirAll(dayDir, 0755))
+			createTestFile(t, dayDir, "director-test-"+day+"T10:00:00Z.txt")
+			createTestFile(t, dayDir, "director-test-"+day+"T10:00:00Z.txt.cinfo")
+		}
+
+		// Create today's directory with multiple files inside the per-director subtree
+		todayDir := filepath.Join(directorSubtree, todayStr)
+		require.NoError(t, os.MkdirAll(todayDir, 0755))
+		createTestFile(t, todayDir, "director-test-"+todayStr+"T08:00:00Z.txt")
```
jhiemstrawisc, line 78: This should also be extended to handle the case where an orphaned file pair exists at the top of the sort.

cache/cache_api.go
```diff
@@ -51,57 +61,342 @@ func CheckCacheSentinelLocation() error {
...
+//
+//	<MonitoringBaseNs>/<DirectorTestDir>/<director-id>/YYYY-MM-DD/director-test-<suffix>
+//
+// The director-id is a non-empty path segment (typically the director's hostname) that
+// keeps multiple directors from colliding under a shared cache. The suffix is left
+// unconstrained so that both ".txt" and ".cinfo" files are accepted. Matching against
+// a strict pattern (rather than a prefix check) rejects path-traversal attempts like
+// "/pelican/monitoring/../../etc/passwd", double slashes, and any filename outside the
+// director-test naming convention.
+var directorTestFilePattern = regexp.MustCompile(
```
jhiemstrawisc, line 85: This regex holds critical security implications. Can you make sure it's tested?

director/cache_monitor.go
```diff
@@ -29,44 +31,123 @@ import (
...
+	if err != nil {
+		log.Warningf("Failed to get federation info for evict token: %v", err)
+		return
+	}
+
+	testTokenCfg := token.NewWLCGToken()
+	testTokenCfg.Lifetime = time.Minute
+	testTokenCfg.Issuer = fedInfo.DiscoveryEndpoint
+	testTokenCfg.AddAudiences(cacheWebUrl)
+	testTokenCfg.Subject = "director"
```
jhiemstrawisc, line 114: This should use the Director hostname so we can tell which Director issued the token.

director/cache_monitor.go
```diff
@@ -29,44 +31,123 @@ import (
...
+		log.Warningf("Failed to send evict request to cache %s: %v", cacheWebUrl, err)
+		return
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		log.Debugf("Successfully evicted test file %s from cache %s", testFilePath, cacheWebUrl)
 	} else {
-		return fmt.Errorf("cache response file does not match expectation. Expected:%s, Got:%s", server_utils.DirectorTestBody, strBody)
+		body, _ := io.ReadAll(resp.Body)
```
jhiemstrawisc, line 150: Why throw this error on the floor?

xrootd/authorization.go
```diff
@@ -1128,34 +1128,47 @@ func makeSciTokensCfg() (cfg ScitokensCfg, err error) {
 	return cfg, nil
 }
 
+// BuildCacheNamespaceAds returns the cache's director-advertised namespace ads, augmented
+// with a monitoring-namespace issuer when Cache.SelfTest is enabled. The monitoring issuer
+// authorizes locally-minted tokens the cache uses for self-tests and for evicting director
+// test files via the local xrdhttp-pelican evict API. Centralizing this here ensures both
+// the startup write and the runtime EmitScitokensConfig path produce the same scitokens.cfg.
+func BuildCacheNamespaceAds(cacheServer *cache.CacheServer) ([]server_structs.NamespaceAdV2, error) {
```
jhiemstrawisc: It's confusing to have both `BuildCacheNamespaceAds` and `cacheServer.GetNamespaceAds` which do slightly different things. Can you rename one of these to more immediately distinguish between the two?

jhiemstrawisc (changes requested):

cache/cache_api.go
h2zh: Right, I agree with your instinct! If something weird happens and there are nested dirs in the new day-directory, we should clean them up too.

cache/cache_api.go
h2zh: Updated the docstring of `cleanTestDir` to discuss this worse case scenario.

cache/cache_api.go
h2zh: Failures are highly correlated, so the first one is representative. Especially in the worst case scenario you mentioned above - we don't want to see the same error repeat 5760 times. Besides, this logic can ensure the loop not to be aborted if the error happens on one subdir.

cache/cache_api.go
h2zh: Done. Good catch here!

cache/cache_api.go
h2zh: Same as above.

cache/cache_api.go
h2zh: I appreciate the suggestion! Since now the director's evict API is the primary cleanup, this sweep only needs to be a backup: run it once at startup, then every 24h. This could greatly reduce the memory usage and logging entries, comparing to the 15s interval previously.

Besides, no admin start a cache at exact midnight. Even they do, there is no casualty. Let's say the director test at 2026-05-16 23:59:59 creates and reads back its file within a single GET in `runCacheTest`; by the time the file exists on disk it has already served its purpose, and the director evicts it on the next cycle anyway. The sweep at 2026-05-17 00:00:00 deleting 2026-05-16/ early is functionally the same as that eviction.

cache/cache_api.go
h2zh: Updated the docstring to clarify `dirTestPath`

cache/cache_api.go
h2zh: I modifed this function to group `.txt` with its `.txt.info`. See commit 69bc8dc2bea0ce79a37d4948dbee138095552b99 

Also added regression tests.

cache/cache_api.go
h2zh: I removed these two lines for now

cache/cache_api.go
h2zh: Done

cache/cache_api.go
h2zh: Done

director/cache_monitor_test.go
h2zh: Thanks for bringing it out. Removed.

cache/cache_cleanup_test.go
h2zh: Updated tests in commit 69bc8dc2bea0ce79a37d4948dbee138095552b99 to cover the orphaned files

cache/cache_api.go
h2zh: Good point - done.

director/cache_monitor.go
h2zh: This is the non-OK branch, where the only purpose of reading the body is to enrich a warning log with the server's response text.

On a read error, io.ReadAll returns whatever bytes it managed to read, so string(body) still gets a partial or empty string. Then user can still see what's happening in the log line.

director/cache_monitor.go
h2zh: This token is federation-signed (`Issuer = fedInfo.DiscoveryEndpoint`), not bound to a specific Director. So putting a hostname in sub is not accurate and misleading.

h2zh: I pick up this convention from another place using the federation issuer: director/monitor.go#L56

xrootd/authorization.go
h2zh: Renamed func BuildCacheNamespaceAds to buildCacheNsAdsWithIssuer

cache/cache_api.go
h2zh: The body is only being read to make a nicer error message, and we're already returning an error anyway. Omit the error here can avoid cluttering the message.

director/cache_monitor.go
jhiemstrawisc: I'd still argue subjects in tokens should be globally unique because the token should be enough to tell where the token originated. I won't push back too hard at this point, though 😉

jhiemstrawisc (approved):
I poked this again with Claude, and it didn't produce much that actually concerned me. Really nice work with this!

PR #3460, opened 2026-05-18
Logging metadata registry integration

### LINKED ISSUE #3430: Registry Integration for Logging Metadata Namespaces for Origins
Registry Integration for the logging metadata namespace: implementation for origins first

* On Origin startup with log export enabled
  -- auto-register /pelican/logging/{Server.ID} in the Registry using the server's existing key; 
  -- auto-approve the registration if the Origin is already approved, or approve it once the Origin is approved later
* Add server name search to the Registry API so the CLI can resolve a human-friendly server name to a Server ID
* Operators must ensure /pelican/logging is never registered as a standalone namespace prefix — doing so would block all logging namespace registrations

turetske:
Origins with log exports enabled auto-register a `/pelican/logging/{ServerID}` namespace on startup; the registry can auto-approve it when the origin is already approved

Logging namespace registrations are tagged in the registry DB so they can be identified without parsing the prefix string

The name filter on the registry servers API now does an exact name lookup

Closes #3430

### DIFF
--- a/config/config_validation.go
+++ b/config/config_validation.go
@@ -109,20 +109,29 @@ func validateConfigKeys() []string {
 // for all server types. It is a no-op if Logging.LogExports.Enabled is false.
 //
 // Rules enforced:
+//   - Xrootd.Sitename must be set (it is used as the logging namespace path segment).
 //   - Logging.LogLocation must be set to an absolute file path other than "/dev/null",
 //     because the virtual-object handler reads that file to assemble log responses.
 //   - If IssuerKeysDirectory is empty the server will not be able to auto-generate access
 //     tokens for the logging namespace; a warning is emitted but startup is not blocked.
 func ValidateLogExportsConfig() error {
-	if !param.Logging_LogExports_Enabled.GetBool() {
+	if !param.Logging_EnableLogExports.GetBool() {
 		return nil
+	}
+
+	if param.Xrootd_Sitename.GetString() == "" {
+		return fmt.Errorf(
+			"%s is true but %s must be set — it is used as the logging namespace path segment (/pelican/logging/{sitename})",
+			param.Logging_EnableLogExports.GetName(),
+			param.Xrootd_Sitename.GetName(),
+		)
 	}
 
 	logLocation := param.Logging_LogLocation.GetString()
 	if !filepath.IsAbs(logLocation) || logLocation == "/dev/null" {
 		return fmt.Errorf(
 			"%s is true but %s must be set to an absolute file path (not \"/dev/null\") to enable log export",
-			param.Logging_LogExports_Enabled.GetName(),
+			param.Logging_EnableLogExports.GetName(),
 			param.Logging_LogLocation.GetName(),
 		)
 	}
@@ -130,7 +139,7 @@ func ValidateLogExportsConfig() error {
 	if param.IssuerKeysDirectory.GetString() == "" {
 		log.Warningf("%s is true but %s is not configured; "+
 			"the server will not be able to auto-generate access tokens for the logging namespace",
-			param.Logging_LogExports_Enabled.GetName(),
+			param.Logging_EnableLogExports.GetName(),
 			param.IssuerKeysDirectory.GetName())
 	}
 
--- a/launcher_utils/register_namespace.go
+++ b/launcher_utils/register_namespace.go
@@ -265,11 +265,15 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
 	return nil
 }
 
-// Register the namespace. If failed, retry every 10s (default)
-// RegisterLoggingNamespaceWithRetry registers the logging namespace /pelican/logging/{Server.ID}
-// for the origin server. It is a no-op when Logging.LogExports.Enabled is false.
-// Must be called after the origin prefix has been successfully registered so that GetServerMetadata
-// can retrieve the assigned Server.ID from the registry.
+// RegisterLoggingNamespaceWithRetry registers the logging namespace
+// /pelican/logging/{Server.ID} for the origin server.
+// It is a no-op when Logging.LogExports.Enabled is false.
+//
+// Must be called after the origin server prefix has been registered so that
+// the server record (and its Server.ID) exists in the registry.  If the
+// origin prefix registration is still retrying in the background, the server
+// ID will not yet be available; in that case this function spawns a goroutine
+// in egrp that polls until the ID appears, then calls RegisterNamespaceWithRetry.
 func RegisterLoggingNamespaceWithRetry(ctx context.Context, egrp *errgroup.Group) error {
 	if !param.Logging_LogExports_Enabled.GetBool() {
 		return nil
@@ -278,15 +282,50 @@ func RegisterLoggingNamespaceWithRetry(ctx context.Context, egrp *errgroup.Group
 	if err != nil {
 		return errors.Wrap(err, "failed to get server metadata for logging namespace registration")
 	}
+	// metadata.ID is the 7-character server ID assigned by the registry the
+	// first time the origin's namespace prefix is successfully registered.  It
+	// is empty when the registry has never seen this origin — for example
+	// because the initial namespace registration failed and is still retrying
+	// in the background.  In that case GetServerMetadata returns an empty
+	// ServerRegistration because no server record exists yet.  Spawn a goroutine
+	// that polls until the ID becomes available, then hands off to
+	// RegisterNamespaceWithRetry.
 	if metadata.ID == "" {
-		log.Warning("Server ID is empty; cannot register logging namespace. Log export will be unavailable until the server is restarted.")
+		log.Warning("Server ID is empty; logging namespace registration will be retried once the origin prefix is registered.")
+		retryInterval := param.Server_RegistrationRetryInterval.GetDuration()
+		if retryInterval == 0 {
+			retryInterval = 10 * time.Second
+		}
+		egrp.Go(func() error {
+			ticker := time.NewTicker(retryInterval)
+			defer ticker.Stop()
+			for {
+				select {
+				case <-ticker.C:
+					md, err := server_utils.GetServerMetadata(ctx, server_structs.OriginType)
+					if err != nil {
+						log.Warningf("Failed to get server metadata while retrying logging namespace registration: %v", err)
+						continue
+					}
+					if md.ID == "" {
+						continue
+					}
+					loggingPrefix := server_structs.LoggingNamespaceForServer(md.ID)
+					log.Debugf("Registering logging namespace %s", loggingPrefix)
+					return RegisterNamespaceWithRetry(ctx, egrp, loggingPrefix)
+				case <-ctx.Done():
+					return nil
+				}
+			}
+		})
 		return nil
 	}
-	loggingPrefix := server_structs.LoggingNamespacePrefix + "/" + metadata.ID
+	loggingPrefix := server_structs.LoggingNamespaceForServer(metadata.ID)
 	log.Debugf("Registering logging namespace %s", loggingPrefix)
 	return RegisterNamespaceWithRetry(ctx, egrp, loggingPrefix)
 }
 
+// RegisterNamespaceWithRetry registers the namespace. If failed, retries every 10s (default)
 func RegisterNamespaceWithRetry(ctx context.Context, egrp *errgroup.Group, prefix string) error {
 	retryInterval := param.Server_RegistrationRetryInterval.GetDuration()
 	if retryInterval == 0 {
--- a/launcher_utils/register_namespace.go
+++ b/launcher_utils/register_namespace.go
@@ -40,7 +40,6 @@ import (
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/registry/registry_client"
 	"github.com/pelicanplatform/pelican/server_structs"
-	"github.com/pelicanplatform/pelican/server_utils"
 )
 
 type (
@@ -266,61 +265,26 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
 }
 
 // RegisterLoggingNamespaceWithRetry registers the logging namespace
-// /pelican/logging/{Server.ID} for the origin server.
-// It is a no-op when Logging.LogExports.Enabled is false.
+// /pelican/logging/{sitename} for the origin server, where sitename is
+// the value of Xrootd.Sitename (the human-readable unique server name).
+// It is a no-op when Logging.EnableLogExports is false.
 //
-// Must be called after the origin server prefix has been registered so that
-// the server record (and its Server.ID) exists in the registry.  If the
-// origin prefix registration is still retrying in the background, the server
-// ID will not yet be available; in that case this function spawns a goroutine
-// in egrp that polls until the ID appears, then calls RegisterNamespaceWithRetry.
+// The logging namespace prefix is derived entirely from local configuration
+// and does not require a registry lookup, so this can be called at any point
+// after server configuration is initialised.  The logging namespace will be
+// submitted as pending if the origin is still awaiting admin approval;
+// Registry.EnableAutoLoggingRegistration controls whether the registry
+// auto-approves it once the associated origin is approved.
 func RegisterLoggingNamespaceWithRetry(ctx context.Context, egrp *errgroup.Group) error {
-	if !param.Logging_LogExports_Enabled.GetBool() {
+	if !param.Logging_EnableLogExports.GetBool() {
 		return nil
 	}
-	metadata, err := server_utils.GetServerMetadata(ctx, server_structs.OriginType)
-	if err != nil {
-		return errors.Wrap(err, "failed to get server metadata for logging namespace registration")
+	sitename := param.Xrootd_Sitename.GetString()
+	if sitename == "" {
+		return errors.Errorf("cannot register logging namespace: %s must be set when %s is true",
+			param.Xrootd_Sitename.GetName(), param.Logging_EnableLogExports.GetName())
 	}
-	// metadata.ID is the 7-character server ID assigned by the registry the
-	// first time the origin's namespace prefix is successfully registered.  It
-	// is empty when the registry has never seen this origin — for example
-	// because the initial namespace registration failed and is still retrying
-	// in the background.  In that case GetServerMetadata returns an empty
-	// ServerRegistration because no server record exists yet.  Spawn a goroutine
-	// that polls until the ID becomes available, then hands off to
-	// RegisterNamespaceWithRetry.
-	if metadata.ID == "" {
-		log.Warning("Server ID is empty; logging namespace registration will be retried once the origin prefix is registered.")
-		retryInterval := param.Server_RegistrationRetryInterval.GetDuration()
-		if retryInterval == 0 {
-			retryInterval = 10 * time.Second
-		}
-		egrp.Go(func() error {
-			ticker := time.NewTicker(retryInterval)
-			defer ticker.Stop()
-			for {
-				select {
-				case <-ticker.C:
-					md, err := server_utils.GetServerMetadata(ctx, server_structs.OriginType)
-					if err != nil {
-						log.Warningf("Failed to get server metadata while retrying logging namespace registration: %v", err)
-						continue
-					}
-					if md.ID == "" {
-						continue
-					}
-					loggingPrefix := server_structs.LoggingNamespaceForServer(md.ID)
-					log.Debugf("Registering logging namespace %s", loggingPrefix)
-					return RegisterNamespaceWithRetry(ctx, egrp, loggingPrefix)
-				case <-ctx.Done():
-					return nil
-				}
-			}
-		})
-		return nil
-	}
-	loggingPrefix := server_structs.LoggingNamespaceForServer(metadata.ID)
+	loggingPrefix := server_structs.LoggingNamespaceForServer(sitename)
 	log.Debugf("Registering logging namespace %s", loggingPrefix)
 	return RegisterNamespaceWithRetry(ctx, egrp, loggingPrefix)
 }
--- a/launcher_utils/register_namespace.go
+++ b/launcher_utils/register_namespace.go
@@ -40,6 +40,7 @@ import (
 	"github.com/pelicanplatform/pelican/param"
 	"github.com/pelicanplatform/pelican/registry/registry_client"
 	"github.com/pelicanplatform/pelican/server_structs"
+	"github.com/pelicanplatform/pelican/server_utils"
 )
 
 type (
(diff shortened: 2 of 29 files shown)


launcher_utils/register_namespace.go
```diff
@@ -265,6 +266,27 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
...
+	if metadata.ID == "" {
+		log.Warning("Server ID is empty; cannot register logging namespace. Log export will be unavailable until the server is restarted.")
+		return nil
+	}
+	loggingPrefix := server_structs.LoggingNamespacePrefix + "/" + metadata.ID
+	log.Debugf("Registering logging namespace %s", loggingPrefix)
+	return RegisterNamespaceWithRetry(ctx, egrp, loggingPrefix)
+}
+
 func RegisterNamespaceWithRetry(ctx context.Context, egrp *errgroup.Group, prefix string) error {
```
jhiemstrawisc, line 293: Can you restore the docstring for this function? Your added function separated them.

launcher_utils/register_namespace.go
```diff
@@ -265,6 +266,27 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
 }
 
 // Register the namespace. If failed, retry every 10s (default)
+// RegisterLoggingNamespaceWithRetry registers the logging namespace /pelican/logging/{Server.ID}
+// for the origin server. It is a no-op when Logging.LogExports.Enabled is false.
+// Must be called after the origin prefix has been successfully registered so that GetServerMetadata
+// can retrieve the assigned Server.ID from the registry.
```
jhiemstrawisc: @h2zh can you weigh in on this statement? I thought the Origin knew its own "ID" before it tried to register.

launcher_utils/register_namespace.go
```diff
@@ -265,6 +266,27 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
...
+		return nil
+	}
+	metadata, err := server_utils.GetServerMetadata(ctx, server_structs.OriginType)
+	if err != nil {
+		return errors.Wrap(err, "failed to get server metadata for logging namespace registration")
+	}
+	if metadata.ID == "" {
+		log.Warning("Server ID is empty; cannot register logging namespace. Log export will be unavailable until the server is restarted.")
+		return nil
+	}
```
jhiemstrawisc: Can you leave a comment explaining:
1. Why the metadata ID will be blank until the server is restarted
2. Why this is better as a logged warning instead of an error

Also, if a server reboot is necessary at this point, why not initiate one?

launcher_utils/register_namespace_test.go
```diff
@@ -331,3 +332,23 @@ func TestMultiKeysRegistration(t *testing.T) {
 	assert.True(t, isRegistered)
 	assert.Equal(t, svr.URL+"/api/v1.0/registry", registerURL)
 }
+func TestRegisterLoggingNamespaceWithRetry_NoopWhenDisabled(t *testing.T) {
+	t.Cleanup(test_utils.SetupTestLogging(t))
+	t.Cleanup(func() {
+		server_utils.ResetTestState()
+	})
```
jhiemstrawisc: Minor nit, but I think this can be simplified to `t.Cleanup(server_utils.ReserTestState)`.

registry/registry.go
```diff
@@ -444,6 +445,37 @@ func keySignChallengeCommit(ctx *gin.Context, data *registrationData) (bool, map
 		ns.CustomFields["server_id"] = data.ServerID // Pass the server ID to the CustomFields map, to fit in the registration workflow
 	}
 
+	// Tag logging namespace registrations so the auto-approval logic and API filters
+	// can identify them without parsing the prefix string.
+	if strings.HasPrefix(data.Prefix, server_structs.LoggingNamespacePrefix+"/") {
```
jhiemstrawisc: It feels fragile and error prone to be using `server_structs.LoggingNamespacePrefix+"/"` all over the place. Maybe you can create a helper function or include the trailing `/` in the prefix itself?

This section is also missing testing, but it seems pretty important to test.

registry/registry.go
```diff
@@ -444,6 +445,37 @@ func keySignChallengeCommit(ctx *gin.Context, data *registrationData) (bool, map
...
+							ns.AdminMetadata.ApproverID = "system"
+							ns.AdminMetadata.ApprovedAt = time.Now()
+							log.Debugf("Auto-approving logging namespace %s because origin server %s is approved", data.Prefix, serverID)
+							break
+						}
+					}
+				}
+			}
+		}
+	}
```
jhiemstrawisc, line 607: 5 levels of nesting in a function that's now over 200 lines long -- can you consider breaking some of this out into a helper function? I won't push back if you don't think that's feasible for this, but this gets hard to read.

It also feels like there are lots of strange cases in this that are glossed over and not logged. For example, what would cause `serverID == ""` at this level? Why would `serverReg, err := getServerByID(serverID)` return a nil error but result in nil `serverReg`? Are these expected cases? If not, should we be logging when they occur

registry/registry_db_server_test.go
```diff
@@ -688,3 +688,72 @@ func TestServerWithMultipleServices(t *testing.T) {
 		assert.True(t, server2.IsCache, "Server2 should be marked as cache")
 	})
 }
+
+// TestLoggingNamespaceAutoApprovalCondition verifies that getServerByID correctly returns
+// the registration status of the parent origin, which is the key input to the auto-approval
+// logic in keySignChallengeCommit.
+func TestLoggingNamespaceAutoApprovalCondition(t *testing.T) {
```
jhiemstrawisc: It doesn't look like this tests any of your new code -- what is this meant to test?

registry/registry_ui.go
```diff
@@ -910,7 +911,26 @@ func deleteNamespace(ctx *gin.Context) {
...
+				Msg:    "Failed to list servers"})
+			return
+		}
+		if server != nil {
+			servers = []server_structs.ServerRegistration{*server}
+		}
+	} else {
+		servers, err = listServers()
+	}
 	if err != nil {
```
jhiemstrawisc: This `err` can only be non-nil if the `else` branch of the preceding block is executed. It feels like it should be scoped there.

registry/registry_ui.go
```diff
@@ -910,7 +911,26 @@ func deleteNamespace(ctx *gin.Context) {
 }
 
 func listServersHandler(ctx *gin.Context) {
-	servers, err := listServers()
+	nameQuery := ctx.Query("name")
```
jhiemstrawisc, line 922: Does this new query param need to be documented in our API docs?

launcher_utils/register_namespace_test.go
```diff
@@ -331,3 +332,23 @@ func TestMultiKeysRegistration(t *testing.T) {
 	assert.True(t, isRegistered)
 	assert.Equal(t, svr.URL+"/api/v1.0/registry", registerURL)
 }
+func TestRegisterLoggingNamespaceWithRetry_NoopWhenDisabled(t *testing.T) {
```
jhiemstrawisc, line 334: This tests the no-op route. Can you add a test for the other case where this actually runs?

server_structs/registry.go
```diff
@@ -136,6 +150,20 @@ func (Registration) TableName() string {
 	return "registrations"
 }
 
+// IsLoggingNamespace reports whether reg is a logging namespace registration,
+// i.e. one whose CustomFields map contains {"type": "logging"}.
+func (reg *Registration) IsLoggingNamespace() bool {
```
jhiemstrawisc, line 156: Can you supply a unit test?

jhiemstrawisc (changes requested):
I'm also going to request a full review from @h2zh since it starts touching registry logic.

jhiemstrawisc:
I'll also point out the PR description confused me a bit -- I think it was written for a previous PR, as this one doesn't introduce the new parameters.

turetske:
> I'll also point out the PR description confused me a bit -- I think it was written for a previous PR, as this one doesn't introduce the new parameters.

Fun copy paste error on my part :(

launcher_utils/register_namespace.go
turetske: It's not that it doesn't know it, it's more that it need to make sure that it must be called after the origin server prefix has been registered so that the server record (and its Server.ID) exists in the registry.  The successful retrieval is a check to that it's been registered properly, not that it need the ID. I'll clarify the issue.

launcher_utils/register_namespace.go
turetske: I guess it depends. If the origin is running it seemed disruptive for a non-critical ability. But since this is at startup/registration, the chances of a restart causing major problems is minor.

However, if for some reason regular registration is failing, then there is a reason for it and forcing a restart could get us into a restart loop rather than waiting for the registration with retry to succeed. I think the thing I actually need to fix is why it isn't spawning a retry of its own, especially given the name.

launcher_utils/register_namespace.go
```diff
@@ -264,7 +265,67 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
...
+// /pelican/logging/{Server.ID} for the origin server.
+// It is a no-op when Logging.LogExports.Enabled is false.
+//
+// Must be called after the origin server prefix has been registered so that
+// the server record (and its Server.ID) exists in the registry.  If the
+// origin prefix registration is still retrying in the background, the server
+// ID will not yet be available; in that case this function spawns a goroutine
+// in egrp that polls until the ID appears, then calls RegisterNamespaceWithRetry.
+func RegisterLoggingNamespaceWithRetry(ctx context.Context, egrp *errgroup.Group) error {
+	if !param.Logging_LogExports_Enabled.GetBool() {
```
jhiemstrawisc: I can't remember if this was defined in a previous PR, but I'm only now noticing that this naming scheme doesn't really fit with any other "Enable" params. Not that our parameters are totally coherent, but I think this should be `Logging.EnableLogExports` to fit with the few dozen other examples we already have.

launcher_utils/register_namespace.go
```diff
@@ -264,7 +265,67 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
...
+	// metadata.ID is the 7-character server ID assigned by the registry the
+	// first time the origin's namespace prefix is successfully registered.  It
+	// is empty when the registry has never seen this origin — for example
+	// because the initial namespace registration failed and is still retrying
+	// in the background.  In that case GetServerMetadata returns an empty
+	// ServerRegistration because no server record exists yet.  Spawn a goroutine
+	// that polls until the ID becomes available, then hands off to
+	// RegisterNamespaceWithRetry.
+	if metadata.ID == "" {
+		log.Warning("Server ID is empty; logging namespace registration will be retried once the origin prefix is registered.")
```
jhiemstrawisc: Is this tied to registration or approval of a pending registration?

launcher_utils/register_namespace_test.go
```diff
@@ -331,3 +332,94 @@ func TestMultiKeysRegistration(t *testing.T) {
...
+	ctx, cancel, egrp := test_utils.TestContext(context.Background(), t)
+	defer func() {
+		cancel()
+		require.NoError(t, egrp.Wait())
+	}()
+
+	// Logging.LogExports.Enabled is false by default.
+	// The function must return nil immediately without touching the network or
+	// adding any goroutines to the errgroup.
+	goroutinesBefore := runtime.NumGoroutine()
```
jhiemstrawisc: I'm worried about this approach because we run our tests with `-p=4`, meaning you can't really account for goroutines because some might be running in parallel tests.

launcher_utils/register_namespace.go
```diff
@@ -264,7 +265,67 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
...
+func RegisterLoggingNamespaceWithRetry(ctx context.Context, egrp *errgroup.Group) error {
+	if !param.Logging_LogExports_Enabled.GetBool() {
+		return nil
+	}
+	metadata, err := server_utils.GetServerMetadata(ctx, server_structs.OriginType)
+	if err != nil {
+		return errors.Wrap(err, "failed to get server metadata for logging namespace registration")
+	}
+	// metadata.ID is the 7-character server ID assigned by the registry the
+	// first time the origin's namespace prefix is successfully registered.  It
```
jhiemstrawisc: I had a chat with @h2zh about how these server IDs are assigned and I think this comment is incorrect -- the server (origin,cache) generates these IDs itself and then sends them to the Registry. If all you need is the ID that will be associated with the server, maybe @h2zh can suggest a method for a local lookup that doesn't require polling the Registry like this.

launcher_utils/register_namespace_test.go
```diff
@@ -331,3 +332,94 @@ func TestMultiKeysRegistration(t *testing.T) {
...
+	require.NoError(t, database.InitServerDatabase(server_structs.RegistryType))
+
+	gin.SetMode(gin.TestMode)
+	engine := gin.Default()
+	registry.RegisterRegistryAPI(engine.Group("/"))
+	svr := httptest.NewServer(engine)
+	defer svr.CloseClientConnections()
+	defer svr.Close()
+
+	require.NoError(t, param.Set(param.Federation_RegistryUrl, svr.URL))
```
jhiemstrawisc, line 384: The mocked federation root from earlier won't have the correct registry URL. Is that a problem?

launcher_utils/register_namespace_test.go
```diff
@@ -331,3 +332,94 @@ func TestMultiKeysRegistration(t *testing.T) {
...
+	originPrefix := server_structs.GetOriginNs(originHost)
+	require.NoError(t, RegisterNamespaceWithRetry(ctx, egrp, originPrefix))
+
+	// Enable log exports and register the logging namespace.
+	require.NoError(t, param.Logging_LogExports_Enabled.Set(true))
+	require.NoError(t, RegisterLoggingNamespaceWithRetry(ctx, egrp))
+
+	// Get the server ID by querying the registry.
+	tr := config.GetTransport()
+	httpClient := &http.Client{Transport: tr}
```
jhiemstrawisc: Can you use `config.GetClient()` when creating HTTP clients?

registry/registry.go
```diff
@@ -264,6 +265,41 @@ func keySignChallengeInit(data *registrationData) (map[string]interface{}, error
 	return res, nil
 }
 
+// applyLoggingNamespaceAutoApproval auto-approves ns if the origin server
+// identified by serverID already has an approved registration.  It mutates
+// ns.AdminMetadata in place on success.  The serverID must be a non-empty
+// 7-character ID from the servers table (the suffix after LoggingNamespacePrefix).
+func applyLoggingNamespaceAutoApproval(serverID string, ns *server_structs.Registration) {
```
jhiemstrawisc: Two notes here on this.

First, I'm not sure I was unable to get the auto approval to happen in testing. My test procedure was:
- start a Director and Registry (with the auto-approval knob set to true)
- start an Origin with log exports enabled
- visit the Registry and observe two data namespaces, the server namespace, and the logging namespace
- at this point I approved everything but the logging namespace and waited for a minute or two
- the logging namespace never left the "pending registration"

When you were developing, how did you test it? Am I missing something with my procedure?

Second, I used Claude to specifically probe this endpoint because it made me nervous to auto approve something as sensitive and security-critical as a namespace. I strongly encourage you to do the same with something equivalent to an Opus model (I haven't found Sonnet to be sufficient for this kind of thing). I won't include the full trace of the exploit it found here, but this is the synopsis:

## Vulnerability: Logging-namespace squatting + key substitution under any registered server ID
TL;DR: With `Registry.EnableAutoLoggingRegistration=true`, any anonymous client that can reach the registry's registration endpoint can permanently claim `/pelican/logging/<victimServerID>` with their own public key and have it auto-approved as if the legitimate origin had registered it. Even without auto-approval enabled, the same client can squat the namespace and block the legitimate origin from ever registering it. Server IDs are publicly visible in director advertisements, so target selection is trivial.

Why this matters in Pelican's trust model
The registry's core security guarantee is (prefix → public key) ownership. Any service that verifies a WLCG/Scitokens2 token for a namespace fetches the issuer keys from the registry by prefix. Whoever the registry says owns the prefix can mint tokens that the entire federation will trust.

This PR plants a namespace family (/pelican/logging/*) that:

- bypasses `verifyServerOwnership` (origin/cache only),
- bypasses key-chaining (no hierarchical relationship to origin/cache prefixes),
- bypasses `RequireOriginApproval/RequireCacheApproval` (those gate origin/cache namespaces only),
- and is auto-approved based purely on the existence of an approved registration for a server ID the attacker can read out of any director advertisement.

The concrete capability the attacker gains today is "mint federation-trusted tokens scoped to /pelican/logging/<victimID>." How damaging that is depends on what the in-progress log-export consumer trusts — but the registry must establish the binding correctly now, because consumers will read it as ground truth later.

## Recommended mitigations
In order of importance:

1. Bind the logging registration to a key the federation already trusts for that server. In `keySignChallengeCommit`, when the prefix is a logging namespace, look up the server by ID, and require that `data.Pubkey` matches one of the registered keys for that server's existing approved origin/cache registration (use the same matchKeys helper used by key chaining). Reject with a permission error otherwise. This single check defeats both squatting and key substitution.
2. Require auto-approval to also confirm the key match, not just "any registration approved." Even if mitigation 1 is in place, `applyLoggingNamespaceAutoApproval` should re-verify the (server, key) pair rather than approving on server-ID lookup alone.
3. Tighten `LoggingNamespaceServerID` to reject IDs containing / (or, better, enforce the actual server-ID shape — the 7-char alphanumeric scheme used elsewhere).

launcher_utils/register_namespace.go
```diff
@@ -264,7 +265,67 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
...
+// Must be called after the origin server prefix has been registered so that
+// the server record (and its Server.ID) exists in the registry.  If the
+// origin prefix registration is still retrying in the background, the server
+// ID will not yet be available; in that case this function spawns a goroutine
+// in egrp that polls until the ID appears, then calls RegisterNamespaceWithRetry.
+func RegisterLoggingNamespaceWithRetry(ctx context.Context, egrp *errgroup.Group) error {
+	if !param.Logging_LogExports_Enabled.GetBool() {
+		return nil
+	}
+	metadata, err := server_utils.GetServerMetadata(ctx, server_structs.OriginType)
```
jhiemstrawisc: You should double check, but Claude flagged this one for me -- if the namespace isn't yet registered, this returns an error and you exit right after without entering the retry loop. Even if that's not true, can you add a test case that covers running `RegisterLoggingNamespaceWithRetry` before the underlying server's namespace has been registered to make sure?

server_structs/registry_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	t.Run("LoggingTypeValue", func(t *testing.T) {
+		reg := &Registration{CustomFields: map[string]interface{}{RegistrationTypeKey: LoggingRegistrationType}}
+		assert.True(t, reg.IsLoggingNamespace())
+	})
+}
+
+func TestLoggingNamespaceServerID(t *testing.T) {
+	t.Run("ValidPath", func(t *testing.T) {
+		id, ok := LoggingNamespaceServerID("/pelican/logging/abc1234")
```
jhiemstrawisc: Can you add a test for further nesting, e.g. `/pelican/logging/abc1234/subpath`?

launcher_utils/register_namespace.go
```diff
@@ -264,7 +265,67 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
 	return nil
 }
 
-// Register the namespace. If failed, retry every 10s (default)
+// RegisterLoggingNamespaceWithRetry registers the logging namespace
```
jhiemstrawisc, line 267: Right now registered logging namespaces show up in the general namespaces table at the registry:

Image

I don't yet have a strong opinion for how to "correctly" display this data, but I don't think it should be lumped into this table. Maybe we need a new display pane for these registration?

I'm not sure whether you're planning to tackle this in a future PR, but this PR seems scoped to "registry" stuff, so I think it'd be appropriate to knock it out here.

jhiemstrawisc (changes requested):

launcher_utils/register_namespace.go
turetske: I think you re correct, but also after talking with Brian, I'm switching to using hostname.

launcher_utils/register_namespace.go
turetske: There is a higher lever retry loop. So it does, but I'm adding a test to confirm.

jhiemstrawisc: Can you point me to the higher-level retry loop? I may be missing it, but to me it looks like this function is called directly in `‎launchers/origin_serve.go`, and I don't see a retry baked in there.

launcher_utils/register_namespace.go
turetske: I'm going to hide it for now and create a different step to deal with UI issues

config/config_validation.go
```diff
@@ -109,28 +109,37 @@ func validateConfigKeys() []string {
...
 //     because the virtual-object handler reads that file to assemble log responses.
 //   - If IssuerKeysDirectory is empty the server will not be able to auto-generate access
 //     tokens for the logging namespace; a warning is emitted but startup is not blocked.
 func ValidateLogExportsConfig() error {
-	if !param.Logging_LogExports_Enabled.GetBool() {
+	if !param.Logging_EnableLogExports.GetBool() {
 		return nil
 	}
 
+	if param.Xrootd_Sitename.GetString() == "" {
```
jhiemstrawisc, line 122: I'm still not convinced site name is the right thing to use here, and I'd like @h2zh and @bbockelm to weigh in.

In my view, we shouldn't be associating the registration of the logging namespace with something users are allowed to change whenever they feel like updating their configuration. I think this is the same reason we're moving toward an immutable server ID as the thing we register under the hood when we set up Origins/Caches. My recollection is that Brian said he wanted to get rid of the "sitename" concept in the coming year.

turetske: Agreed, but he wanted there to be the idea of the id (hostname) and human readable name (our current site name) and the namespace should be under the site name. My impression is we'd be changing site name, not removing it, so I made it that until the change gets implemented since the change is outside the scope of this work.

launcher_utils/register_namespace.go
h2zh: In the latest diff, seems like we're moving away from using server id.

config/config_validation.go
h2zh: One side effect of using sitename in the prefix is that when the origin admin changes to a new sitename, ` /pelican/logging/{old_sitename}` will be an orphan.

turetske: Fair. But will this change when the sitename/id paradigm is updated? The idea is that it should be human readable. We can also make it when the origin admin changes it, the old site name gets removed from the registry? Or maybe we want to keep it so no one else can pretend to be it?

h2zh: What sitename/id paradigm change do you refer to? AFAIK, there's no plan to update it. The only future change that may be relevant is rebranding sitename to name (UI only).

Since the goal is making this namespace human-readable, I second the sitename is our only go-to. 

Besides, I agree that the orphan row can prevent someone else from rebinding `/pelican/logging/{old_sitename}`.

h2zh (approved):
LGTM after a rebase

PR #3461, opened 2026-05-19
Allow Clients, Caches and Origins to declare their own locations

### LINKED ISSUE #2709: Let Clients/Servers declare their own location
This was originally proposed by @jthiltges in [this discussion](https://github.com/orgs/PelicanPlatform/discussions/2538). The idea is that both clients/servers can specify their own location when interacting with the Director's server sorting algorithms.

I think this helps us in two important ways:
- Clients/servers can fix their own GeoIP lookup failures without waiting for a federation administrator to make configuration changes
- Federation administrators can debug Director sorting more easily by pretending to be from any location

jhiemstrawisc:
This PR adds a new top-level param "GeoLocation" that, when configured, is used by clients, caches and origins to declare their own physical locations to a federation. This is picked up the Director and used for routing/matchmaking.

I decided this should be a top-level knob because I assume that multiple services reading from the same yaml configuration file are co-located.

When configured, the declared location from a client or server takes precedence over central services ip-based overrides. I chose this precedence because I viewed location information as part of how users of a federation declare they want to interact with that federation -- if a fed admin disagrees with a server's choice of lat/long, it should place that service into downtime until the service admin takes corrective action.

I tested this in my local federation by starting up two caches with different settings for `GeoLocation` alongside a client with it's own `GeoLocation` override. When I pulled an object in debug mode, the Director's redirect JSON confirmed the "declared" lat/long choices were used all around:
```
{
  "clientInfo": {
    "Coordinate": {
      "Lat": 51.500786,
      "Long": -0.124681,
      "AccuracyRadius": 0,
      "Source": "declared",
      "FromTTLCache": false
    },
    "ipAddr": "172.17.0.3"
  },
  "serversInfo": {
    "https://0ee241b2e244:8881": {
      "Coordinate": {
        "Lat": 55.75212,
        "Long": 37.61766,
        "AccuracyRadius": 0,
        "Source": "declared",
        "FromTTLCache": false
      },
      "RedirectWeights": {
        "distanceWeight": 0.004575618912398561,
        "ioLoadWeight": 1,
        "statusWeight": 0.3975739370572201,
        "availabilityWeight": 0.5
      }
    },
    "https://0ee241b2e244:8891": {
      "Coordinate": {
        "Lat": 48.8583,
        "Long": 2.2945,
        "AccuracyRadius": 0,
        "Source": "declared",
        "FromTTLCache": false
      },
      "RedirectWeights": {
        "distanceWeight": 0.4801634804408707,
        "ioLoadWeight": 1,
        "statusWeight": 0.5,
        "availabilityWeight": 0.5
      }
    }
  },
  "directorSortMethod": "adaptive"
}
```

Closes #2709

### DIFF
--- a/client/director.go
+++ b/client/director.go
@@ -121,6 +121,10 @@ func queryDirector(ctx context.Context, verb string, pUrl *pelican_url.PelicanUR
 		// cannot.
 		req.Header.Set("User-Agent", getUserAgent(""))
 
+		if token != "" {
+			req.Header.Set("Authorization", "Bearer "+token)
+		}
+
 		// If the client has a declared GeoLocation, send it as X-Pelican-Coordinate
 		// so the director can use it for client-server matchmaking.
 		if geoStr := param.GeoLocation.GetString(); geoStr != "" {
--- a/client/director.go
+++ b/client/director.go
@@ -21,6 +21,7 @@ package client
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"math/rand"
 	"net"
@@ -120,8 +121,15 @@ func queryDirector(ctx context.Context, verb string, pUrl *pelican_url.PelicanUR
 		// cannot.
 		req.Header.Set("User-Agent", getUserAgent(""))
 
-		if token != "" {
-			req.Header.Set("Authorization", "Bearer "+token)
+		// If the client has a declared GeoLocation, send it as X-Pelican-Coordinate
+		// so the director can use it for client-server matchmaking.
+		if geoStr := param.GeoLocation.GetString(); geoStr != "" {
+			if lat, long, geoErr := utils.ParseCoordinateStr(geoStr); geoErr == nil {
+				req.Header.Set(string(server_structs.XPelicanCoordinateHeaderName),
+					fmt.Sprintf("lat=%g,long=%g", lat, long))
+			} else {
+				log.Warningf("Ignoring invalid %s value %q: %v", param.GeoLocation.GetName(), geoStr, geoErr)
+			}
 		}
 
 		forceDebug, _ := ctx.Value(directorDebugCtxKey{}).(bool)
@@ -352,7 +360,7 @@ func getDirectorInfoForPath(ctx context.Context, pUrl *pelican_url.PelicanURL, h
 // servers.
 func ParseDirectorInfo(dirResp *http.Response) (server_structs.DirectorResponse, error) {
 	var xPelNs server_structs.XPelNs
-	if err := (&xPelNs).ParseRawResponse(dirResp); err != nil {
+	if err := (&xPelNs).ParseRawHeader(&dirResp.Header); err != nil {
 		// Only suppress the specific "header not present" error.  If the header
 		// exists but is malformed, return an error so the caller knows something
 		// is wrong rather than silently continuing with default values.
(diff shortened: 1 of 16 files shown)


h2zh (changes requested):
GeoLocation is designed as a server operator knob (trusted config) but the client-facing implementation treats an HTTP header from anonymous traffic with the same trust level. Maybe we should not allow client self-identify their location.

Also, I think the server-provided location should be validated too, since our core algorithm heavily relies on it. A simple deviation check between the server-provided and maxmind-calculated locations would be acceptable. Another more complicated solution would be add the server-provided location to registration flow and store it in Registry DB, which require federation admin's approval. What do you think?

jhiemstrawisc:
@h2zh, I have a few thoughts about this:
1. I'm not worried about anonymous clients being able to declare their own location via a header. The only effect that this header can have on a Director is the ordering of caches/origins that is returned, and maybe some monitoring statistics about where our users are (and I'm not worried about that either, because we already don't know where a large chunk of the IPs geolocate). In fact, any client could already go to the Director and get a list of all the caches/origins and choose whichever they want to talk to. Is there a specific security concern you have?
2. I agree that it's not a bad idea to validate the server-provided geolocation information, and I also agree that this information would be good to store in the Registry. For now, I'm going to hold off on the Registry integration and leave this as part of the server ads, since we've already talked about future work to have the Registry reconstruct server/namespace ads for the Director.

As for the geolocation validation, lets say a server gives a location that's different from what the Director's MaxMind instance thinks. We should use the server-provided location, but then the question becomes "how and to whom do we report this difference"?

Because the geolocation is currently delivered via server ad, it'd be very difficult to punch a message path back to the server to let it know. In fact, the server probably doesn't even care what MaxMind says, because it's already telling us what location it wants us to use. To me this means the reporting should be Director only, but the question is then "what is a federation admin supposed to do with it"? I don't think there's very much the _can_ do with it because again, we're just noting a difference. We're ultimately going to use whatever the server provided, so I don't think this warrants much more than a debug log.

Thoughts?

h2zh:
We talked about this PR offline today. Since the custom location doesn't provide extra power to the user (client already can talk directly to a Cache), I think the location customization is a nice feature to have. The next step we agree on is calculating the location discrepancy between the server-claimed one and the Director-calculated one. Drop a line in the logs if the discrepancy is greater than a certain value, and create a ticket for the next step.

client/director.go
```diff
@@ -120,8 +121,15 @@ func queryDirector(ctx context.Context, verb string, pUrl *pelican_url.PelicanUR
 		// cannot.
 		req.Header.Set("User-Agent", getUserAgent(""))
 
-		if token != "" {
-			req.Header.Set("Authorization", "Bearer "+token)
```
h2zh, line 124: Are these two lines removed by accident?

h2zh (changes requested):
THAT close to a LGTM...

client/director.go
jhiemstrawisc: Whoa, good catch! Not sure what got into Claude 😅

jhiemstrawisc: Okay, fixed and pushed.

director/sort_utilities.go
```diff
@@ -377,17 +404,66 @@ func getServerCoordinate(sAd server_structs.ServerAd) (coord server_structs.Coor
 	return
 }
 
+// logServerLocationDiscrepancy compares a server-declared coordinate against the coordinate the
+// director would have resolved from GeoIP, and logs (at debug level) when they differ by more
+// than locationDiscrepancyThresholdKm. This is purely observational -- it does not change the
+// coordinate used for sorting/redirects.
+//
+// TODO: surface this discrepancy in the director web UI as a follow-up (see issue-2709 next step).
+func logServerLocationDiscrepancy(sAd server_structs.ServerAd) {
```
h2zh, line 414: Idea: computing the discrepeancy for every ad every minute is still kinda resource-consuming. After all, the logging message is on the debug level. So it would be better to make this entire function to debug level:
```
if !log.IsLevelEnabled(log.DebugLevel) { return }
```

h2zh (approved):
I manually tested it and it looked good. A one-line change for your consideration!

director/sort_utilities.go
jhiemstrawisc: I think this is a great suggestion, I'll do it really quickly before I merge. (I'll even be really careful that I did it correctly 😛)

jhiemstrawisc:
The failing MacOS test is a known (and pesky) flaky test, so I'm going to call this good.

PR #3462, opened 2026-05-19
Allow Registry startup when institutions API is down

### LINKED ISSUE #3458: Registry fails to startup if institutional ID API is down
When the institutional ID API is down, the registry fails to start.

There shouldn't be a hard requirement here.  In today's OSDF outage, we eventually just had to replace the API with something that returned `[]` to get things going again.

h2zh:
Closes #3458 

This PR allows Registry to start when institutions API is down. Instead of throwing an error on startup, it will startup normally, only the features related to institution create/update/list will fail:

1. Registration form fails to load. `getNamespaceRegFields` gets a 500 error with "Couldn't fetch registration fields"
2. Namespace create/update fails. `validateInstitution` gets a 400 with `Validation for Institution failed: Error fetching institutions from TTL cache`
3. Institution dropdown endpoint (GET /api/v1.0/registry_ui/institutions) fails to get data. `listInstitutions` gets 500 error.

### DIFF
--- a/registry/custom_reg_fields.go
+++ b/registry/custom_reg_fields.go
@@ -248,10 +248,19 @@ func InitInstConfig(ctx context.Context, egrp *errgroup.Group) error {
 		instCacheTTL := param.Registry_InstitutionsUrlReloadMinutes.GetDuration()
 		institutions, err := getCachedOptions(param.Registry_InstitutionsUrl.GetString(), instCacheTTL)
 		if err != nil {
-			return err
+			// Don't block startup on an unavailable institution API.
+			// OptionsUrl is still set above, so once that API recovers, subsequent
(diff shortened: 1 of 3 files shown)


jhiemstrawisc (approved):
Looks good to me. I'll note that when I started a registry with a bad InstitutionsUrl, the page to create a new namespace shows up blank except for an error and a "submit" button:

Image

I think that's probably fine because we do typically expect a functioning institutions API for creating new namespaces. If you want to update the code to provide a better experience on that page, feel free. Otherwise I'm okay approving as is because this addresses the failure to start.

h2zh:
Thanks for the feedback! I think this unpleasant UI behavior could be handled in #3471 (See [this](https://github.com/PelicanPlatform/pelican/issues/3471#issuecomment-4559316062) comment)

PR #3464, opened 2026-05-20
Get ready for breaking lotman and xrootd-lotman changes

jhiemstrawisc:
This commit bumps a few dependencies that contain breaking changes while I get ready to roll out the consumers of those changes. To avoid a period during which the lotman tests always fail, I've also compiled the test files out in a way that's easy to restore once they're ready.


jhiemstrawisc:
FYI, I did test that the container builds successfully with this command:
```
$ docker build -t jhiemstra/pelican-test:updated-lotman-deps -f images/Dockerfile .
```

turetske (approved):

PR #3465, opened 2026-05-20
Kingfisher reservations

### LINKED ISSUE #1847: Create mechanism so Pelican caches can manage lot expiration/deletion
Cache administrators can configure various lot objects through the Pelican configuration layer, but there's currently no auto-management of expired/deletable lots. This means that when a lot hits its expiration or deletion timestamp, its data is treated as transient and purgeable until the administrator manually deletes the lot and re-creates it with new timestamps. Worse yet, Pelican doesn't offer very helpful tools for doing this unless the admin sends a hand-crafted JSON along with a valid token to the Lotman UI endpoint (that's off by default).

It seems reasonable that the cache service should be able to manage expired lots on its own. One issue here is that lots using less than their dedicated storage are completely protected from being wiped unless they're allowed to expire (although the cache does fall back to LRU in the event that Lotman doesn't provide it with enough stuff to delete in the purge cycle). This means lots may slowly accumulate protected data, even if that data isn't actively being used. Because of this, I think we need to think more about how lots should be allowed to expire, which would trigger the cleanup.

### LINKED ISSUE #1884: Lotman APIs are not included in the Pelican API docs
The Lotman APIs can be complicated to work with (and are still largely toothless), but they'll never be used if they aren't documented. I need to add them to our swagger docs so they show up in our docs website.

### LINKED ISSUE #1953: Lotman should dynamically adjust lots whenever cache rediscovers namespaces
Caches must periodically poll the Director to get namespace information, which caches use to keep a tidy auth/scitokens config. However, as namespaces come and go, Lotman should also be adjusting lots to make sure we don't wind up with unusable lot partitions.

Getting this right will be tricky, because lotman is fed usage updates as diffs by XRootD. This means that the disappearance of a namespace would currently _require_ re-allocation of its lot's disk usage to another lot -- most likely the default lot, but that's not a guarantee. Otherwise, the incoming diffs wouldn't correctly adjust the values Lotman stores.

In addition to usage re-allocation, we should also handle adjustment of lot management policy attributes. Keep the cache full, keep the cache happy!

### LINKED ISSUE #3421: Plumb lotman reservation API into Caches
Plumb the lotman API into caches so that lot reservations can be made. This will also help me solve some of the other problems surrounding namespaces and renewal of lots.

jhiemstrawisc:
# Kingfisher: Per-Namespace Reservations for Pelican Caches

## Summary

This PR turns Pelican's Lotman integration from a flat, top-level-prefix-only accounting layer into a full **reservation system** for cache storage. Caches now build a nested lot tree from the federation's namespace hierarchy, continuously renew per-namespace lots so coverage never lapses, expose a stable REST surface (`/api/v1.0/lots/*`) for namespace owners to inspect and adjust their own reservations, and rely on Lotman's `strict_hierarchy` enforcement so that every reservation Pelican commits to is provably consistent with every other reservation already in the database.

The work is the Pelican-side counterpart to two upstream changes:
- **lotman ≥ v0.1.0** added the primitives this PR relies on: `strict_hierarchy` admission control with three reservation axioms, explicit per-parent `parent_attributions` contracts, sentinel values for unbounded (`-1`) and zero-capacity (`0`) axes, non-expiring lot timestamps (sentinel `0`), and the `lotman_get_available_capacity` advisory query.
- **xrootd-lotman ≥ v0.1.0** (the cache's purge plugin): this PR ensures the lot graph it consumes is always populated, hierarchically consistent, and continuously refreshed.

## Motivation

Before this PR, Pelican only minted one flat Lotman lot per *top-level* federation namespace, because Lotman lacked the hierarchy primitives to safely model nested paths. That meant:

- Sub-namespaces (e.g. `/foo/bar` under `/foo`) were invisible to Lotman, so the cache could not enforce per-sub-namespace quotas, give them dedicated capacity, or report on their usage separately from their parent.
- Lots had a single, fixed lifetime — there was no scheduler keeping coverage alive — so once a lot's expiration passed, every file attributed to it became, in effect, default-lot data and was first in line for eviction. Per-namespace fairness silently degraded over time.
- There was no public, durable HTTP API for creating or inspecting reservations. The `lotman_ui.go` surface was tied to the admin cookie path and leaked the internal `Lot` struct as its wire format.
- A misconfiguration or a bug could create a lot tree that violated quota math (e.g. children whose dedicated shares exceeded the parent's budget). Lotman happily accepted it; there was no admission control.

The Kingfisher cache deployment exposed each of these as a concrete operational problem. This PR addresses all of them in one coherent change set, and pins the cache to the matching Lotman release.

### 1. Strict-hierarchy admission control

At Lotman init time the cache sets:

- `strict_hierarchy = true` — every `add_lot` and `update_lot` is validated against three axioms (Σ child reservations ≤ parent; peak concurrent attributed usage across siblings ≤ parent; child's `[creation_time, expiration_time)` ⊂ each parent's interval) and rolled back on failure.
- `contraction_policy = always` — quotas may not be silently shrunk under reservations that already depend on them.
- `admin_override = false` by default, with deliberate, scoped opt-ins where Pelican itself must contract (see §4).

The cache then mints two structural lots:

- **`root`** — given the cache's full disk as its `dedicated_GB`. It is the ownership anchor for every other lot and is never itself a purge candidate.
- **`default`** — zero capacity on every axis. Any byte that lands here is immediately over-quota and is the first thing the xrootd-lotman purge plugin evicts. This makes the default lot the natural catch-all for unclaimed paths.

Both carry the non-expiring sentinel (all three timestamps `0`), so they cover all time.

### 2. Nested lot tree from the namespace hierarchy

`buildLotTree` walks the full director-provided namespace ad list and builds a parent/child tree by **path-prefix containment**: `/foo/bar`'s parent is the lot for `/foo` if such a namespace exists, otherwise `root`. Lots are emitted in pre-order so Lotman's "parent must exist before child" precondition is always satisfied.

`allocateQuotas` then walks the tree and distributes each of the three independent axes (`dedicated_GB`, `opportunistic_GB`, `max_num_objects`) using a recursive **(N+1) rule**:

- **Top level:** N direct children of `root` split `root`'s budget evenly: `share = root / N`. Root retains no extra reserve because root itself never accepts new sibling reservations.
- **Deeper levels:** each parent with N children divides by **N+1**: every child gets `parent / (N+1)`, and the parent retains one share as its own unallocated reserve. That reserve is precisely what lets a namespace owner *later* attach a reservation to a deeper sub-path without forcing a global re-allocation of the whole subtree.
- Sentinel values are **never divided**: `-1` (unbounded) and `0` (zero capacity) propagate verbatim.

Each child's `ParentAttributions[parent]` is set equal to its own quota, which is the minimal write that satisfies axiom 1 with strict slack.

### 3. Renewal scheduler with epoch-aware quota stamping

Lots are immutable once minted. To keep every advertised namespace continuously covered without ever silently contracting a reservation, the cache runs two background loops:

#### Renewal loop (`LaunchRenewalRoutine`, every `Lotman.RenewalCheckInterval`, default 1h)

For each advertised namespace path P, processed shortest-path-first so parents are planned before their children, the planner walks the post-tick timeline (existing ∪ already-planned successors) and **fills every uncovered region** inside `[now, now+SchedulingHorizon)` — not just the first. Multi-fill matters because a cold start or a missed tick can leave several holes; filling only the nearest would require multiple wakeups to catch up.

For each gap the planner mints a successor whose window is:

```
creation_time   = max(gapStart, predecessor.expiration_time)
expiration_time = min(creation_time + DefaultLotExpirationLifetime,
                      gapEnd,
                      effectiveParent.expiration_time)
deletion_time   = min(creation_time + DefaultLotDeletionLifetime,
                      effectiveParent.deletion_time)
                  capped overall at MaxLotLifetime
```

Successors whose `creation_time` would fall beyond the horizon are **deferred** to a future tick, keeping each scheduling decision tied to a meaningful look-ahead window. Gaps narrower than `Lotman.MinFillerWidth` (default 0 — fill everything) are skipped with an audit reason emitted.

#### Epoch-aware quota stamping

Because a successor's window may overlap *multiple* "epochs" — intervals during which the set of active sibling lots is constant — naïvely stamping a quota using "right now's" sibling count would let a future sibling appear later in the lot's lifetime and force an effective contraction. `allocateEpochAwareQuotas` partitions each parent's timeline into maximal intervals where the active-child set is constant and, for each epoch, computes:

```
residual_e = parent_dedicated_e - Σ existing_sibling.dedicated_GB active in e
divisor_e  = nNewActive_e + 1     (top level)
           = nNewActive_e + 2     (deeper levels)   # +1 for self, +1 for parent reserve
share_e    = residual_e / divisor_e
```

The stamped quota is the **minimum** `share_e` across all epochs the new lot spans, guaranteeing axiom 1 holds for the entire lifetime of the reservation. This is the same fair-share math as `lot_tree.go::distribute`, just re-expressed for incremental additions.

#### Lot GC loop

Removes any lot row whose `deletion_time + Lotman.LotRecordRetention` is in the past (default retention: 60 days). Underlying cached *data* is reclaimed independently by the xrootd-lotman purge plugin at `deletion_time`; the GC loop is concerned only with database hygiene and leaves a forensics window for inspection.

#### Age-based purge wiring

Because the renewal loop guarantees every cached file is always inside an active lot, watermark-driven eviction would *never* fire on a cache that stays below `Cache.HighWaterMark` — cold files would accumulate forever. To bound on-disk file age independently of fill level, `Lotman.MaxLotLifetime` is now propagated to xrootd's `pfc.diskusage … purgecoldfiles <age>` directive whenever `Cache.EnableLotman` is true. Because the value goes directly into xrootd's accepted range, values outside `[1h, 360d]` are rejected at startup.

### 4. REST API

The legacy `lotman_ui.go` (tightly coupled to the admin cookie path, exposing the raw internal `Lot` struct as its wire format) is replaced with a public REST surface designed around a **hotel-reservation metaphor**: a namespace owner POSTs a `path + quota` request to the cache and receives back a `reservation_id`. Endpoints:

| Method | Path | Purpose |
|---|---|---|
| `GET`    | `/api/v1.0/lots`                          | List all lot names |
| `POST`   | `/api/v1.0/lots`                          | Create a new reservation |
| `GET`    | `/api/v1.0/lots/{lotName}`                | Inspect a reservation |
| `PATCH`  | `/api/v1.0/lots/{lotName}`                | Modify MPA and/or parent attributions |
| `DELETE` | `/api/v1.0/lots/{lotName}`                | Delete the lot record |
| `POST`   | `/api/v1.0/lots/{lotName}/reclaim`        | Reclaim (mark deleted; evict data) |
| `GET`    | `/api/v1.0/lots/{lotName}/children`       | Enumerate child lots |
| `GET`    | `/api/v1.0/lots/{lotName}/usage`          | Per-axis current usage |
| `GET`    | `/api/v1.0/lots/{lotName}/policy`         | Binding (ancestor) policy per axis |
| `GET`    | `/api/v1.0/lots/by-path`                  | All lots covering a path |
| `GET`    | `/api/v1.0/lots/by-path/capacity`         | Advisory available capacity for a path |

Design invariants:

- **Server-minted UUIDs.** Clients do not need to choose or know identifiers in advance.
- **Server-stamped reclaim timestamps.** `reclaimed_at_ms` is set exclusively from the server clock; clients cannot back- or future-date the audit trail.
- **Independent auth paths.** Cookie-based admin auth and bearer-token (JWT) auth are handled by separate code paths; the new `getBearerToken` deliberately does *not* fall back to the `login` cookie the way `token.GetAuthzEscaped` did. This closes an issuer-confusion vector in which a rejected admin session could be silently re-interpreted as a federation credential.
- **Namespace-aware ownership.** Whether a lot is created via cookie or via JWT, its `owner` is resolved the same way: by querying the director for the *issuer* of the target namespace. Admin-created and programmatically-created lots are therefore governed by the same authorization principal.
- **MPA temporal invariants enforced at the API layer.** PATCH requests are validated against the *merged* (existing ∪ patched) MPA before reaching Lotman, so `creation < expiration ≤ deletion` is always preserved regardless of which subset of fields the caller supplies.
- **Stable public response shape.** A dedicated `Reservation` projection type is returned on every endpoint; the internal `Lot` struct is never serialized. `AvailableCapacityResponse` is similarly re-projected field-by-field so the external JSON contract is immune to upstream Lotman ABI churn.
- A new token scope, `lot.reclaim`, is added to `docs/scopes.yaml` and regenerated through `go generate` rather than hand-edited into `token_scopes.go`.

### 5. Production-hardening fixes

Four independent failure modes surfaced during initial Kingfisher deployment and are fixed here:

1. **Concurrent SQLite access.** Lotman funnels every operation through a single shared SQLite handle. Pelican's renewal-tick, GC-tick, and HTTP handler goroutines were calling in concurrently and producing intermittent `bad parameter or other API misuse` errors. `callerMutex` is now held across *every* Go entrypoint that touches the C library, including all read-paths (`GetLot`, `GetLotsFromDir`, `ListAllLots`, `GetChildrenNames`, `GetParentNames`, `GetOwners`, `IsRoot`, `LotExists`, `GetPolicyAttributes`, `GetLotDirs`, `GetLotUsage`, `GetAvailableCapacity`, `GetAuthorizedCallers`, …).
2. **Admin-driven MPA contraction blocked.** When an operator lowers `Cache.FilesMaxSize` or `Cache.HighWaterMark`, the root and default lots' `dedicated_GB` must shrink too — which `contraction_policy=always` refuses. The new `UpdateLotAsAdmin` flips `admin_override=true` only for the duration of the MPA write, restoring it on return. If the root lot's contraction still fails because existing descendants hold reservations that exceed the new budget (strict-hierarchy violation), `reclaimDescendantsForContraction` removes those children so the renewal tick can replan their coverage against the reduced budget.
3. **Init-on-restart overlapping windows.** Without filtering, restarting the cache would re-mint a fresh lot for every advertised path while the *old* lot was still alive in the database, producing overlapping windows that tripped axiom 1. `filterAdsAlreadyScheduled` now skips any path whose namespace already has an active non-sentinel lot in the init planning window; those paths fall to the renewal tick, whose job it is to mint non-overlapping successors.
4. **`topLevelSiblings` treated `/` as a real ancestor.** The synthetic `/` root was incorrectly participating in ancestor-containment checks, hiding every real sibling. `/` is now explicitly skipped.

The same commit makes `Lotman.GarbageCollectionInterval` configurable (default 1h) and fixes a misconfiguration in the xrootd `purgecoldfiles` wiring.

## New Configuration Surface

| Parameter | Default | Purpose |
|---|---|---|
| `Lotman.DefaultLotExpirationLifetime` | `24h` (was `2016h`) | Window of each lot's `[creation, expiration)` |
| `Lotman.DefaultLotDeletionLifetime`   | `48h` (was `4032h`) | Window of each lot's `[creation, deletion)` |
| `Lotman.MaxLotLifetime`               | `168h` | Hard cap on `expiration − creation`; also fed to xrootd `purgecoldfiles` (range `[1h, 360d]`) |
| `Lotman.RenewalCheckInterval`         | `1h`   | Tick of the renewal scheduler |
| `Lotman.SchedulingHorizon`            | `48h`  | How far ahead each tick fills coverage |
| `Lotman.MinFillerWidth`               | `0s`   | Smallest gap the planner will fill (0 = fill everything) |
| `Lotman.LotRecordRetention`           | `1440h` (60d) | How long a deleted lot's *row* stays around for forensics |
| `Lotman.GarbageCollectionInterval`    | `1h`   | Hidden; tick of the GC loop |

The previous lifetime defaults (`2016h` / `4032h`) reflected a world where lots were not renewed; with continuous renewal the system can afford much shorter windows and benefits from them (faster recovery from misconfiguration, finer-grained re-planning).

## Implications

With this PR in place:

- Every advertised federation namespace, at any depth, has a continuously refreshed, hierarchy-consistent reservation in Lotman. Per-namespace quota enforcement and per-namespace usage reporting become first-class.
- The cache can never commit to a lot tree that violates the reservation axioms; Lotman is the authority and Pelican is one of several principals subject to it.
- Namespace owners can read, create, and adjust reservations against their paths through a documented, versioned REST API rather than through a server-internal admin UI.
- The xrootd-lotman purge plugin gets a richer, hierarchical view to drive its `del / exp / opp / ded` policy ladder; per-sub-namespace fairness is enforced rather than aspirational.
- Bounded on-disk file age is no longer dependent on reaching the high-watermark, which removes a class of "cold cache, infinitely old files" deployments.

### 1. Sub-namespace data may be purged when the parent self-overruns

The xrootd-lotman purge plugin, by design, asks Lotman for the *paths* attached to lots that are over a given policy threshold (`del`, `exp`, `opp`, `ded`) and then performs a recursive purge of each returned path. Because Pelican attaches each namespace lot to its namespace prefix with `recursive=true`, the plugin will only ever be handed the parent's path — e.g. `/foo` — even when the reason the parent is "over quota" is that its **own self-usage** (data attributed to `/foo` directly, not to any child) exceeds its dedicated share, while child `/foo/bar` is comfortably inside its own reservation.

The recursive purge under `/foo` does not (and from inside the plugin, *cannot*) distinguish those two cases: it walks the filesystem under `/foo` and evicts files there, including files that physically live under `/foo/bar` and therefore "belong to" the child's still-valid reservation. The child's accounting will subsequently re-converge, but in the interim the child's reserved bytes can be evicted to make room for a parent that overcommitted its own share.

This is a property of Lotman's path-attribution model combined with the purge plugin's directory-walking semantics, not a Pelican bug per se. Fixing it would require the plugin to be told *which axis* and *which lot* a returned path belongs to, and to skip subtree entries that are themselves attributed to a different, in-compliance lot — i.e. a more granular file-level filter at purge time. For now, operators should be aware that a parent namespace that consistently runs over its self-share can collaterally evict cached data belonging to compliant child namespaces under it.

### 2. Reservations are owner-only; this is not a fair-share marketplace

The reservation API authorizes a request against the *namespace issuer* for the target path: only the owner of `/foo` (as advertised by the director) can create or modify lots whose paths sit under `/foo`. This is sufficient for the use case Kingfisher was built for — namespace owners pinning a known amount of capacity inside their own slice of the federation — and it composes naturally with the existing token-issuer trust model.

It is **not**, however, a mechanism that lets *anyone* with a federation token slice and reserve cache capacity. Allowing arbitrary callers to carve out reservations would require the cache to reason about *cross-tenant fair share*: how much of the cache any one caller is allowed to pin, how competing reservations are reconciled when total demand exceeds capacity, how to evict speculative reservations when a higher-priority owner appears, how to prevent denial-of-service through reservation spam, and how to expose to operators a coherent picture of "who has claimed what." Those questions are substantially harder than the per-owner case this PR solves and are deliberately deferred. The current system can be extended in that direction (the parent-reserve share built into the (N+1) rule is precisely the headroom a future fair-share allocator would draw from), but the policy work required to do it safely is out of scope here.

Closes #1847
Closes #1884
Closes #1953
Closes #3421


jhiemstrawisc:
I did a last round of manual testing to make sure that newly-discovered namespaces are eventually given a share of the cache via a reservation after the scheduling window finds a duration in which there's unallocated space. Here was my test procedure:

1. Start federation central services (director, registry) -- nothing special here.
2. Start an Origin that advertises two namespaces; `/prefix1` and `/prefix2`
3. Start a lotman-enabled cache with relatively short scheduling params conducive to a quick test:
```
Lotman:
  MaxLotLifetime: 1h
  DefaultLotExpirationLifetime: 1m
  DefaultLotDeletionLifetime: 2m
  RenewalCheckInterval: 20s
  SchedulingHorizon: 40s
  LotRecordRetention: 1m
  GarbageCollectionInterval: 10s
Cache:
  EnableLotman: true
  HighWaterMark: 5g
  LowWaterMark: 4g
  FilesMaxSize: 20m
  FilesNominalSize: 15m
  FilesBaseSize: 10m
```
4. In the lot database, I immediately observed scheduled lots for `/prefix1` and `prefix2` that are each given 10MB of dedicated space.
5. Start a second Origin that advertises `/prefix3` and `/prefix1/subprefix`. At this point, I turned to something else to burn a few minutes. When I came back I observed that the new top-level lots being minted were for `/prefix1`, `prefix2` and `prefix3` and each was getting a dedicated allocation of 6.66MB. A new sublot existed for `/prefix1/subprefix` and it had the expected 3.333MB share of its parent's quota.

jhiemstrawisc:
I'm not going to foist this massive diff on anyone -- it's mostly self-contained, and has a minimal interface with existing code. I've also reviewed the majority of this code (which is why I've been working on it for over a month) and have done extensive hand testing.

PR #3467, opened 2026-05-20
Updated the error message for three '///' when using a pelican protocol

### LINKED ISSUE #3255: Improve error message for wrong number of slashes
**Proposal**

Change the error message from 

```
Error code 1000: pelican URL 'pelican:///<director-address>/<namespace>/<object>' is invalid because it has no host
```

to 

```
Error code 1000: pelican URL 'pelican:///<director-address>/<namespace>/<object>' is invalid because it has no host; did you mean 'pelican://<director-address>/...'?
```

**Background**

A lot of users are used to the three slashes in the protocol (`osdf:///`) and it throws them off when they switch to the direct pelican protocol (`pelican://<host>/`). They mistakenly include a third slash between the protocol and the director address (`pelican:///<host>/`) which results in an error. 

And it doesn't trip up only users. When a user hit this earlier today, I ended up going down a rabbit hole about the namespace because it didn't seem like it was listed in the Director. But the user (on their own) realized the discrepancy between `//` and `///` and fixing that addressed the problem.

turetske:
close #3255

### DIFF
--- a/pelican_url/parse.go
+++ b/pelican_url/parse.go
@@ -181,9 +181,21 @@ func validateOsdfStashUrl(parsedUrl *url.URL) error {
 
 func validatePelicanUrl(parsedUrl *url.URL) error {
 	if parsedUrl.Host == "" {
-		return error_codes.NewParameterError(
-			fmt.Errorf("pelican URL '%s' is invalid because it has no host", parsedUrl.String()),
-		)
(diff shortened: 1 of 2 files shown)


jhiemstrawisc (approved):
Short and sweet, and when tested it solves the problem. Thanks!

PR #3470, opened 2026-05-21
Clean up `pelican token` tooling

### LINKED ISSUE #2719: `pelican token create` command should warn users if flags don't match configured namespace perms
If a namespace doesn't allow writes but someone is trying to create a token with the `--write` flag, the command should detect this and warn the user that the namespace doesn't support the intended operation.

### LINKED ISSUE #2941: `pelican token create` emits misleading warnings and errors concerning issuers
## Narrative

I'm a naïve user. I just want a token:

```
$ pelican token create osdf://ospool/ap40/data --read
WARNING[2026-01-05T08:38:39-06:00] Unable to get JWKS from issuer URL https://osg-htc.org/ospool: Error getting JWKS URL from issuer URL: failed to unmarshal openid-configuration for issuer https://osg-htc.org/ospool: json: cannot unmarshal array into Go value of type string; skipping 
Error: unable to determine issuer for resource osdf://ospool/ap40/data; you may need to re-run with '--issuer <issuer URL>' to specify an issuer: none of the issuers discovered at the director match your signing key; issuers that were checked: https://osg-htc.org/ospool
Usage:
  pelican token create <pelican-url> [flags]

Examples:
To create a read/write token for /some/namespace/path in OSDF: pelican token create --read --write pelican://osg-htc.org/some/namespace/path

Flags:
  -a, --audience string         Specify the token's 'audience/aud' claim. If not provided, the equivalent 'any' audience for the selected profile will be used (e.g. 'https://wlcg.cern.ch/jwt/v1/any' for the 'wlcg' profile).
  -h, --help                    help for create
  -i, --issuer string           Set the token's 'issuer/iss' claim. If not provided, the issuer will be discovered via the Director.
  -l, --lifetime int            Set the token's lifetime in seconds. (default 1200)
  -m, --modify                  Indicate the requested token should provide the ability to modify/delete the specified resource.
  -k, --private-key string      Path to the private key used to sign the token. If not provided, Pelican will look for the private key in the default location pointed to by the 'IssuerKeysDirectory' config parameter.
  -p, --profile string          Create a token with a specific JWT profile. Accepted values are scitokens2 and wlcg. (default "wlcg")
      --raw-claim stringArray   Set claims to be added to the token. Format: <claim_key>=<claim_value>. 
      --raw-scope stringArray   Set non-typical values for the token's 'scope' claim. Scopes should be space-separated, e.g. 'storage.read:/ storage.create:/'.
  -r, --read                    Indicate the requested token should provide the ability to read the specified resource.
      --scope-path string       Specify the path to use when creating the token's scopes. This should generally be the object path without the namespace prefix.
  -s, --stage                   Indicate the requested token should provide the ability to stage the specified resource.
      --subject string          Set token's 'subject/sub' claim. If not provided, the current user will be used as the default subject.
  -w, --write                   Indicate the requested token should provide the ability to create/write the specified resource. Does not grant the ability to overwrite/modify existing resources.

Global Flags:

ERROR[2026-01-05T08:38:39-06:00] Fatal error occurred at the start of the program. Cleanup started: unable to determine issuer for resource osdf://ospool/ap40/data; you may need to re-run with '--issuer <issuer URL>' to specify an issuer: none of the issuers discovered at the director match your signing key; issuers that were checked: https://osg-htc.org/ospool
```

**This impressive amount of text completely obscures my actual mistake:** "none of the issuers discovered at the director match your signing key".

- "unable to determine issuer for resource osdf://ospool/ap40/data" is emitted twice, and it reads like a lie both times, by the client's own admission: "issuers that were checked: https://osg-htc.org/ospool".

- The initial warning that "Unable to get JWKS from issuer URL https://osg-htc.org/ospool" is actually a straight up bug in the client.

## Details

Pelican version:

```
Version: 7.22.0
Build Date: 2025-12-19T19:00:13Z
Build Commit: 074a9766f8c06e281c4fbfeb98a205f021288b56
Built By: goreleaser
```

### Problem the First

The "Unable to get JWKS from issuer URL" warning comes from

https://github.com/PelicanPlatform/pelican/blob/c8edb77b4da955a53743c453943479c78f6972d2/cmd/token.go#L157-L163

Tracing backwards through the rest of the warning message and the code, we eventually arrive at

https://github.com/PelicanPlatform/pelican/blob/c8edb77b4da955a53743c453943479c78f6972d2/server_utils/registry.go#L98-L102

The type of `openIDCfgMap` is simply incorrect. A quick check of https://osg-htc.org/ospool/.well-known/openid-configuration (or https://cilogon.org/.well-known/openid-configuration, or what have you) reveals that some of the string keys can indeed have array values:

```
{  
   "issuer":"https://osg-htc.org/ospool",
   "jwks_uri":"https://osg-htc.org/ospool/oauth2/certs",
   "token_endpoint": "https://osdf-ospool-issuer.osgdev.chtc.io/scitokens-server/token",
   "userinfo_endpoint": "https://osdf-ospool-issuer.osgdev.chtc.io/scitokens-server/userinfo",
   "registration_endpoint": "https://osdf-ospool-issuer.osgdev.chtc.io/scitokens-server/oidc-cm",
   "device_authorization_endpoint": "https://osdf-ospool-issuer.osgdev.chtc.io/scitokens-server/device_authorization",
   "authorization_endpoint": "https://osdf-ospool-issuer.osgdev.chtc.io/scitokens-server/authorize",
   "token_endpoint_auth_methods_supported": [
      "client_secret_post",
      "client_secret_basic"
   ],
   "grant_types_supported": [
      "authorization_code",
      "refresh_token",
      "urn:ietf:params:oauth:grant-type:device_code"
   ]
}
```

### Problem the Second

The "unable to determine issuer for resource" error comes from

https://github.com/PelicanPlatform/pelican/blob/c8edb77b4da955a53743c453943479c78f6972d2/cmd/token.go#L298-L304

which would seem to take as through the flow in the previous problem, ending at

https://github.com/PelicanPlatform/pelican/blob/c8edb77b4da955a53743c453943479c78f6972d2/cmd/token.go#L184-L185

### LINKED ISSUE #3233: `pelican token create` outputs bad error when namespace caps don't match token caps
I tried doing `pelican token create --write` on a namespace that didn't support writes (or any other protected option). The error I got:
```
DEBUG[2026-03-13T15:30:07Z] Fatal error occurred at the start of the program. Cleanup started: unable to determine issuer for resource pelican://21e78195fb87:4443/my-prefix; you may need to re-run with '--issuer <issuer URL>' to specify an issuer: no issuers found for /my-prefix in the Director response 
```

Because the namespace doesn't support protected operations, there _is_ no issuer (so this message is technically true). Instead of the current error message, we could query the Director for the namespace's capabilities and provide a more meaningful message.

jhiemstrawisc:
In addition to some general cleanup, this PR does the following:
- We now warn the user when they're trying to create a token with a scope that doesn't appear to be supported as determined by the namespace ad.
- We fix the issue where failure to match a KID against a supplied list of issuers manifested as "no issuer found". Now we more clearly say "there were issuers, but none of them matched the signing key you're using"
- When the namespace doesn't require token auth for reads and doesn't support writes, it may not advertise an issuer. Rather than reporting we couldn't find the issuer, we now report that the user likely doesn't need to create a token in the first place.

Here's a description of some of the hand testing I ran:
1. Set up a namespace that does not require any auth -- previously caused an error about missing issuers, now emits a warning that the namespace likely doesn't need a token:
```
$ pelican token create --read pelican://`hostname`:4443/my-prefix2/foo.txt
WARNING[2026-05-21T16:52:02Z] Director indicates that the resource at pelican://0ee241b2e244:4443/my-prefix2/foo.txt is publicly readable so a token is not actually required to read it, but the --read flag was provided; adding read scope to token anyway 
Error: unable to determine issuer for resource pelican://0ee241b2e244:4443/my-prefix2/foo.txt. This is likely because the Director reports that this namespace does not require token issuance for reads and does not support writes. Are you sure you need a token? You may need to re-run with '--issuer <issuer URL>' to specify an issuer: no issuers found for /my-prefix2 in the Director response
```
Note that in this example, an error is returned with no token because the tool lacks sufficient information to produce the token.
2. This is already demonstrated by the previous example, but the tool now emits a warning when requested scopes don't align with the namespace capabilities for the specified resource. To test, I set up a namespace with protected reads but no writes:
```
$ pelican token create --write --modify pelican://`hostname`:4443/my-prefix2/foo.txt
WARNING[2026-05-21T16:55:35Z] Director indicates that the resource at pelican://0ee241b2e244:4443/my-prefix2/foo.txt does not support write/modify operations, but --write flag was provided; adding write scope to token anyway 
WARNING[2026-05-21T16:55:35Z] Director indicates that the resource at pelican://0ee241b2e244:4443/my-prefix2/foo.txt does not support write/modify operations, but --modify flag was provided; adding modify scope to token anyway 
WARNING[2026-05-21T16:55:35Z] Unable to get current user, using 'pelican_client' as default token subject 
<the token>
```

Lastly, the setup to hand test fixes for #2941 were a bit more involved than I had energy for. For that I rely on the unit tests, which are evidence enough for me that the bug is fixed.

Closes #3233
Closes #2719
Closes #2941

### DIFF
--- a/cmd/token.go
+++ b/cmd/token.go
@@ -120,50 +120,54 @@ These two flags are mutually exclusive.`,
 	}
 )
 
-var (
-	readFlag   bool
-	writeFlag  bool
-	modifyFlag bool
-)
-
 func addScopeFlags(cmd *cobra.Command) {
-	cmd.Flags().BoolVarP(&readFlag, "read", "r", false, "Indicate the requested token should provide the ability to read the specified resource.")
-	cmd.Flags().BoolVarP(&writeFlag, "write", "w", false, "Indicate the requested token should provide the ability to create/write the specified resource. "+
+	cmd.Flags().BoolP("read", "r", false, "Indicate the requested token should provide the ability to read the specified resource.")
+	cmd.Flags().BoolP("write", "w", false, "Indicate the requested token should provide the ability to create/write the specified resource. "+
 		"Does not grant the ability to overwrite/modify existing resources.")
-	cmd.Flags().BoolVarP(&modifyFlag, "modify", "m", false, "Indicate the requested token should provide the ability to modify/delete the specified resource.")
+	cmd.Flags().BoolP("modify", "m", false, "Indicate the requested token should provide the ability to modify/delete the specified resource.")
+}
+
+// addTokenCreateFlags registers every flag that the `token create` command and
+// its createToken handler read. It is the single source of truth for that flag
+// set, shared by init() and tests so the two cannot drift apart.
+func addTokenCreateFlags(cmd *cobra.Command) {
+	// Token capabilities
+	addScopeFlags(cmd)
+
+	cmd.Flags().BoolP("stage", "s", false, "Indicate the requested token should provide the ability to stage the specified resource.")
+	cmd.Flags().String("scope-path", "", "Specify the path to use when creating the token's scopes. This should generally be "+
+		"the object path without the namespace prefix.")
+
+	// Additional token fields
+	cmd.Flags().StringP("audience", "a", "", "Specify the token's 'audience/aud' claim. If not provided, the equivalent 'any' audience "+
+		"for the selected profile will be used (e.g. 'https://wlcg.cern.ch/jwt/v1/any' for the 'wlcg' profile).")
+	cmd.Flags().IntP("lifetime", "l", 1200, "Set the token's lifetime in seconds.")
+	cmd.Flags().String("expiration", "", "Set the token's expiration as an absolute RFC3339 timestamp (e.g., 2026-12-31T23:59:59Z). Mutually exclusive with --lifetime.")
+	cmd.MarkFlagsMutuallyExclusive("lifetime", "expiration")
+	cmd.Flags().String("subject", "", "Set token's 'subject/sub' claim. If not provided, the current user will be used as the default subject.")
+	cmd.Flags().StringP("issuer", "i", "", "Set the token's 'issuer/iss' claim. If not provided, the issuer will be discovered via the Director.")
+	cmd.Flags().StringArray("raw-claim", []string{}, "Set claims to be added to the token. Format: <claim_key>=<claim_value>. ")
+	cmd.Flags().StringArray("raw-scope", []string{}, "Set non-typical values for the token's 'scope' claim. Scopes should be space-separated, e.g. "+
+		"'storage.read:/ storage.create:/'.")
+	cmd.Flags().StringP("profile", "p", "wlcg", "Create a token with a specific JWT profile. Accepted values are scitokens2 and wlcg.")
+	cmd.Flags().StringP("private-key", "k", "", fmt.Sprintf("Path to the private key used to sign the token. If not provided, Pelican will look for "+
+		"the private key in the default location pointed to by the '%s' config parameter.", param.IssuerKeysDirectory))
 }
 
 func init() {
 	rootCmd.AddCommand(tokenCmd)
-	tokenCmd.AddCommand(tokenCreateCmd)
-	tokenCmd.AddCommand(tokenFetchCmd)
 
-	// Token capabilities
-	addScopeFlags(tokenCreateCmd)
+	// Token create command setup
+	tokenCmd.AddCommand(tokenCreateCmd)
+	addTokenCreateFlags(tokenCreateCmd)
+
+	// Token fetch command setup
+	tokenCmd.AddCommand(tokenFetchCmd)
 	addScopeFlags(tokenFetchCmd)
 
 	// Token fetch requires exactly one of read, write or modify
 	tokenFetchCmd.MarkFlagsMutuallyExclusive("read", "write", "modify")
 	tokenFetchCmd.MarkFlagsOneRequired("read", "write", "modify")
-
-	tokenCreateCmd.Flags().BoolP("stage", "s", false, "Indicate the requested token should provide the ability to stage the specified resource.")
-	tokenCreateCmd.Flags().String("scope-path", "", "Specify the path to use when creating the token's scopes. This should generally be "+
-		"the object path without the namespace prefix.")
-
-	// Additional token fields
-	tokenCreateCmd.Flags().StringP("audience", "a", "", "Specify the token's 'audience/aud' claim. If not provided, the equivalent 'any' audience "+
-		"for the selected profile will be used (e.g. 'https://wlcg.cern.ch/jwt/v1/any' for the 'wlcg' profile).")
-	tokenCreateCmd.Flags().IntP("lifetime", "l", 1200, "Set the token's lifetime in seconds.")
-	tokenCreateCmd.Flags().String("expiration", "", "Set the token's expiration as an absolute RFC3339 timestamp (e.g., 2026-12-31T23:59:59Z). Mutually exclusive with --lifetime.")
-	tokenCreateCmd.MarkFlagsMutuallyExclusive("lifetime", "expiration")
-	tokenCreateCmd.Flags().String("subject", "", "Set token's 'subject/sub' claim. If not provided, the current user will be used as the default subject.")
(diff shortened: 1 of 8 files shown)


cmd/token.go
```diff
@@ -308,6 +369,9 @@ func createToken(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	// Grab the namespace ad from the Director -- we'll use this later to validate token scopes.
+	nsAd, err := getNsAd(directorInfo)
```
turetske: This err is never checked/handled (I'm actually surprised it passed go linting because of that)

cmd/token.go
```diff
@@ -338,25 +402,78 @@ func createToken(cmd *cobra.Command, args []string) error {
...
+			log.Warningf("Director indicates that the resource at %s is publicly readable so a token is not actually required to read it, but the --read flag was provided; adding read scope to token anyway", rawUrl)
+		} else if !nsAd.Caps.Reads {
+			log.Warningf("Director indicates that the resource at %s does not support read operations, but --read flag was provided; adding read scope to token anyway", rawUrl)
+		}
+		scopes = append(scopes, tokenProfile.ReadScope(sPath))
+	}
+	if write {
+		if !nsAd.Caps.Writes {
+			log.Warningf("Director indicates that the resource at %s does not support write/modify operations, but --write flag was provided; adding write scope to token anyway", rawUrl)
+		 }
```
turetske: Small white space error here, the bracket isn't lining up with the 'if'

cmd/token.go
```diff
@@ -237,6 +238,66 @@ func getIssuer(directorInfo server_structs.DirectorResponse, kidSet map[string]s
 		combineUrls(directorInfo.XPelAuthHdr.Issuers))
 }
 
+// Given a Director's redirect response, re-query the Director's UI endpoint to get the full namespace ad information.
+// This can be used later to determine what capabilities the namespace supports in comparison with the requested
+// token scopes.
+func getNsAd(directorInfo server_structs.DirectorResponse) (server_structs.NamespaceAdV2Response, error) {
```
turetske, line 248: Is this the right approach? If we know the redirect response, should we be getting all the ads (via "/api/v1.0/director_ui/namespaces") when we know the specific namespace we are looking for?

cmd/token.go
```diff
@@ -237,6 +238,66 @@ func getIssuer(directorInfo server_structs.DirectorResponse, kidSet map[string]s
...
+
+	return server_structs.NamespaceAdV2Response{}, errors.Errorf(
+		"no namespace advertisement found for namespace %s",
+		namespace,
+	)
+}
+
+
 // Create a token using the provided flags/args
 func createToken(cmd *cobra.Command, args []string) error {
```
turetske, line 305: There are no unit tests for this function, can you add them?

turetske (changes requested):

cmd/token.go
jhiemstrawisc: Unfortunately, it's the only approach we can take right now -- the redirect response doesn't contain all the information we need, and there is no API at the Director to get only a single namespace. If I add one, then new clients that use it won't work with old Directors.

cmd/token.go
jhiemstrawisc: Good catch -- addressed in the latest commit!

cmd/token.go
turetske: I feel like it's something we should ad. But maybe as a separate step (add it to the director, then once it's been in the director for a bit, shift the clients over. With the understanding that directors are generally kept up to date.

jhiemstrawisc: Sure, I can open a follow-up ticket for it once this is merged.

turetske (approved):

jhiemstrawisc:
The currently-failing test is related to infrastructure outages that are causing a missing test object from the OSDF. I think this is safe enough to merge as is.

PR #3472, opened 2026-05-22
Document how to scrape a Pelican service from an external Prometheus instance

### LINKED ISSUE #3447: Document how to scrape a Pelican service with Prometheus
We have documentation for pointing Grafana at Prometheus (though this needs to be updated to use an API key): https://docs.pelicanplatform.org/monitoring-pelican-services/grafana#configure-authentication. But we don't have the documentation for how to set up a separate Prometheus instance to scrape a Pelican service's Prometheus that's behind authN

1. Document how to generate an API key, especially with the appropriate `--scope` and `--server` values
2. Include an eample of the Prometheus scrape config

h2zh:
Including generating an API key with the monitoring.scrape scope and the corresponding Prometheus scrape_configs entry

Closes [#3447](https://github.com/PelicanPlatform/pelican/issues/3447)

### DIFF
--- a/docs/app/monitoring-pelican-services/prometheus/page.mdx
+++ b/docs/app/monitoring-pelican-services/prometheus/page.mdx
@@ -16,6 +16,52 @@ Pelican included metrics from built-in [gin](https://gin-gonic.com/) web server,
 
 Pelican also has a set of built-in metrics to monitor Pelican server's status, listed below.
 
+## Scrape Pelican Metrics with an External Prometheus
+
+If you operate a separate Prometheus instance and want it to scrape metrics from one or more Pelican services (for example, to collect federation-wide metrics into a central monitoring stack), you can point that Prometheus at the Pelican `/metrics` endpoint. Because `/metrics` is authenticated by default, you must issue a Pelican API key for the scraping job and configure Prometheus to send it as a bearer credential.
+
+### Generate an API key
+
+Use `pelican apikey generate` on the Pelican server you want to scrape. The key must include the `monitoring.scrape` scope, which grants access to the `/metrics` endpoint. (The separate `monitoring.query` scope is only needed when issuing PromQL queries against `/api/v1.0/prometheus`; it does not grant access to `/metrics`.)
+
+```bash copy
+pelican apikey generate \
+  --scopes "monitoring.scrape" \
+  --expiration "2038-01-19T03:14:07Z" \
+  --server https://<server-web-url>
+```
+
+* `--scopes`: must include `monitoring.scrape` for scraping `/metrics`. To allow the same key to also run PromQL queries, use `--scopes "monitoring.scrape,monitoring.query"`.
+* `--expiration`: RFC3339 timestamp at which the key stops being valid.
+* `--server`: must match the Pelican server's `Server.ExternalWebUrl`. When that value is not set explicitly, it falls back to `https://<Server.Hostname>[:<WebPort>]`, with the default `:443` port stripped. If the server is reachable by multiple hostnames (for example, both an A record and a CNAME), use the hostname the server is actually configured with; a mismatch will be rejected.
+
+The command prints the API key to standard output. Save it to a file that the scraping Prometheus instance can read (for example, `/etc/prometheus/secrets/pelican-apikey`) and restrict the file's permissions; treat it as a credential.
+
+> **Note**: An API key is validated against the database of the server that issued it. If the configured hostname is a CNAME that may fail over to a different deployment, the key will not be accepted by the failover target. Re-issue the key against the failover server, or generate it on each backing deployment ahead of time.
+
+### Configure the scrape job
+
+Add a job similar to the following to your Prometheus configuration:
+
+```yaml filename="prometheus.yml"
+scrape_configs:
+  - job_name: pelican
+    metrics_path: /metrics
+    scheme: https
+    static_configs:
+      - targets: ['<server-hostname>:<server-web-port>']
+    authorization:
+      type: Bearer
+      credentials_file: /etc/prometheus/secrets/pelican-apikey
+```
+
+* `metrics_path` is `/metrics`, which is also the Prometheus default.
+* `scheme` must be `https`; Pelican does not serve metrics over plain HTTP.
+* `targets` should be the same `host[:port]` you used for `--server` (without the `https://` scheme). For a server reachable on the default HTTPS port, the `:443` suffix can be omitted.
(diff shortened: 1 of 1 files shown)


h2zh:
This PR will help you understand the apikey feature in Pelican, and Prometheus in general. @melody0123 

Prerequisite: https://github.com/PelicanPlatform/pelican/pull/3368

docs/app/monitoring-pelican-services/prometheus/page.mdx
```diff
@@ -16,6 +16,52 @@ Pelican included metrics from built-in [gin](https://gin-gonic.com/) web server,
 
 Pelican also has a set of built-in metrics to monitor Pelican server's status, listed below.
 
+## Scrape Pelican Metrics with an External Prometheus
+
+If you operate a separate Prometheus instance and want it to scrape metrics from one or more Pelican services (for example, to collect federation-wide metrics into a central monitoring stack), you can point that Prometheus at the Pelican `/metrics` endpoint. Because `/metrics` is authenticated by default, you must issue a Pelican API key for the scraping job and configure Prometheus to send it as a bearer credential.
```
melody0123, line 21: I notice that Pelican server has its own built-in Prometheus server (https://docs.pelicanplatform.org/parameters#Monitoring-EnablePrometheus). I think it makes sense to introduce the best practice on deciding when to use built-in and when to use an external one.

> `/metrics` is authenticated by default
Wording is a little confusing. My suggestion is "`/metrics` needs authentication by default".

docs/app/monitoring-pelican-services/prometheus/page.mdx
```diff
@@ -16,6 +16,52 @@ Pelican included metrics from built-in [gin](https://gin-gonic.com/) web server,
...
+pelican apikey generate \
+  --scopes "monitoring.scrape" \
+  --expiration "2038-01-19T03:14:07Z" \
+  --server https://<server-web-url>
+```
+
+* `--scopes`: must include `monitoring.scrape` for scraping `/metrics`. To allow the same key to also run PromQL queries, use `--scopes "monitoring.scrape,monitoring.query"`.
+* `--expiration`: RFC3339 timestamp at which the key stops being valid.
+* `--server`: must match the Pelican server's `Server.ExternalWebUrl`. When that value is not set explicitly, it falls back to `https://<Server.Hostname>[:<WebPort>]`, with the default `:443` port stripped. If the server is reachable by multiple hostnames (for example, both an A record and a CNAME), use the hostname the server is actually configured with; a mismatch will be rejected.
+
```
melody0123, line 37: It seems that `--config` global flag will also affect the behavior of `apikey generate` command, as mentioned in the help message:

> Global Flags:

If server is launched with a config file located at a non-default place, is it necessary to explicitly specify the config file location?

docs/app/monitoring-pelican-services/prometheus/page.mdx
```diff
@@ -16,6 +16,52 @@ Pelican included metrics from built-in [gin](https://gin-gonic.com/) web server,
...
+```bash copy
+pelican apikey generate \
+  --scopes "monitoring.scrape" \
+  --expiration "2038-01-19T03:14:07Z" \
+  --server https://<server-web-url>
+```
+
+* `--scopes`: must include `monitoring.scrape` for scraping `/metrics`. To allow the same key to also run PromQL queries, use `--scopes "monitoring.scrape,monitoring.query"`.
+* `--expiration`: RFC3339 timestamp at which the key stops being valid.
+* `--server`: must match the Pelican server's `Server.ExternalWebUrl`. When that value is not set explicitly, it falls back to `https://<Server.Hostname>[:<WebPort>]`, with the default `:443` port stripped. If the server is reachable by multiple hostnames (for example, both an A record and a CNAME), use the hostname the server is actually configured with; a mismatch will be rejected.
```
melody0123, line 36: "use the hostname the server is actually configured with" needs more explanation. A simple example helps to clarify a lot. For example: aaa.com is the CNAME for bbb.com, and bbb.com is the A record of the server, which hostname should be used to issue the token?

docs/app/monitoring-pelican-services/prometheus/page.mdx
```diff
@@ -16,6 +16,52 @@ Pelican included metrics from built-in [gin](https://gin-gonic.com/) web server,
...
+      credentials_file: /etc/prometheus/secrets/pelican-apikey
+```
+
+* `metrics_path` is `/metrics`, which is also the Prometheus default.
+* `scheme` must be `https`; Pelican does not serve metrics over plain HTTP.
+* `targets` should be the same `host[:port]` you used for `--server` (without the `https://` scheme). For a server reachable on the default HTTPS port, the `:443` suffix can be omitted.
+* `authorization.credentials_file` should point to the file containing the API key value, with no `Bearer ` prefix or trailing newline. To inline the secret instead of reading it from a file, use `credentials: <api-key>` in place of `credentials_file`. (Older Prometheus syntax exposes the equivalent `bearer_token` / `bearer_token_file` fields; the `authorization` block is the current preferred form.)
+
+After reloading Prometheus, the new target should appear as `UP` on the Prometheus `/targets` page. If the target instead reports a `403`, double-check that the API key was issued with the `monitoring.scrape` scope and that the `--server` value used when generating the key matched the Pelican server's `Server.ExternalWebUrl`.
+
```
melody0123, line 64: It's worthwhile to write a command for user to launch the Prometheus with the said config given above for the sake of completeness

docs/app/monitoring-pelican-services/prometheus/page.mdx
```diff
@@ -16,6 +16,52 @@ Pelican included metrics from built-in [gin](https://gin-gonic.com/) web server,
...
+  - job_name: pelican
+    metrics_path: /metrics
+    scheme: https
+    static_configs:
+      - targets: ['<server-hostname>:<server-web-port>']
+    authorization:
+      type: Bearer
+      credentials_file: /etc/prometheus/secrets/pelican-apikey
+```
+
```
melody0123, line 57: I admit that this is a user-facing doc and usually server admins will config proper TLS certificate. However, it's not uncommon that we want to temporarily run servers with TLS off for testing purposes. It's worthwhile to include a convenient note for users to disable server certificate verification: `insecure_skip_verify`

melody0123 (changes requested):
I followed the doc and was able to set up an external Prometheus instance to scrape metrics from a Pelican director. I added my inline comments for the author to address. Overall this is a great doc for an entry-level system admin to set up the Prometheus for Pelican servers.

PR #3477, opened 2026-05-26
Improve computation of the ETag and checksums

### LINKED ISSUE #3476: Address POSIXv2 testing items / Improve ETag and checksum calculations
When @matyasselmeci was reviewing/testing the POSIXv2 backend, we noticed that the `ETag` contents looked funny: all the test files had the same value.  Appeared to be a bug.

After some digging, it appears there _isn't_ a bug but a limitation of the current approach: the `ETag` only takes into account the mtime and size of the file -- identical for all the test files.  To improve things, let's mix in the dev / inode IDs into the ETag so unique files will have unique ETags.

Additionally, Matyas found 
- checksum calculations defaulted to MD5; we use crc32c otherwise.
- A few cases where we check exactly for "posix"; instead of having an increasingly-long set of expressions to detect POSIX-like backends, let's just create a POSIX-like function.

bbockelm:
When Matyas was reviewing the POSIXv2 backend, we saw scary-looking behavior around the ETag and checksums:

1.  The checksums defaulted to MD5 whereas we use crc32c by default elsewhere.
2.  The ETags all looked the same.  This appears because ETag is based only on mtime / size which were the same for all our test files.

This defaults the checksum to crc32c and mixes in the dev / inode into the ETag computation.

Fixes #3476

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -3326,6 +3326,13 @@ func parseDigestHeader(header http.Header, fields log.Fields) (result []Checksum
 	}
 	for _, val := range header.Values("Digest") {
 		for _, entry := range strings.Split(val, ",") {
+			// Per RFC 7230 §7, list elements may be separated by a comma
+			// followed by optional whitespace (e.g. "md5=...., crc32c=....").
+			// Trim it so the algorithm name doesn't carry a leading space.
+			entry = strings.TrimSpace(entry)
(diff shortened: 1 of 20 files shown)


bbockelm:
Punching this forward as we're short on reviewer time this month ... mostly isolated to the POSIX V2 and cache V2 code.

PR #3482, opened 2026-05-27
Update Actions workflows to run all relevant unit tests

### LINKED ISSUE #3483: Actions workflows do not run all relevant tests
## Summary

The macOS and Windows CI workflows run `go test` without any build tags, silently skipping all tests gated on the `client` or `server` tags. 65 tests are never exercised on those platforms (40 require `client`, 20 require `server`, 5 require either).

## Current behavior

| Platform | Build tags    | Test invocations | Unique tests covered |
|----------|---------------|-----------------|----------------------|
| Linux    | `client` + `server` (two jobs) | 1,286 + 1,266 | 1,306 |
| macOS    | *(none)*      | 1,241           | 1,241                |
| Windows  | *(none)*      | 1,241           | 1,241                |

*Test counts were obtained by running `go test -list '.*' ./...` with the relevant `-tags` flags and collecting the output. "Unique tests covered" is the size of the union of tests across all jobs for that platform.*

## Proposed fix

- **macOS**: Add a `pelican-server` / `-tags server` matrix entry alongside  the existing entry, now using `-tags client`. Mirrors Linux's Go test  tag matrix. XRootD is already installed by `github_scripts/osx_install.sh`.
- **Windows**: Change the single job from no tags to `-tags client`.  No server leg is needed because `pelican-server` does not target Windows.

---

_([Brian A](https://github.com/brianaydemir): With a tip of the hat to Copilot.)_

brianaydemir:
Closes #3483.

Inspired by noticing that Copilot's attempt to update the testing instructions in `AGENTS.md` didn't actually sound right…

### DIFF
--- a/.github/workflows/test-macos.yml
+++ b/.github/workflows/test-macos.yml
@@ -33,7 +33,10 @@ jobs:
         include:
           - binary_name: pelican
             coverprofile: coverage.out
-            tags: ""
+            tags: client
+          - binary_name: pelican-server
+            coverprofile: coverage-server.out
+            tags: server
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -33,7 +33,7 @@ jobs:
         include:
           - binary_name: pelican
             coverprofile: coverage.out
-            tags: ""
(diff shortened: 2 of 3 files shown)


jhiemstrawisc (approved):
Good catch.

PR #3484, opened 2026-05-29
Prevent registry-fetch failure from aborting origin/cache advertisement

### LINKED ISSUE #3450: Hard dependency between {origin,cache} and registry
In [server_utils/sitename.go:111-120](https://github.com/PelicanPlatform/pelican/blob/main/server_utils/sitename.go#L111-L120) - the `err` variable is set, is logged/ignored, but not cleared!

This propagates up propagating up through [launcher_utils/advertise.go:120-123](https://github.com/PelicanPlatform/pelican/blob/main/launcher_utils/advertise.go#L120-L123) and aborting advertisement to all directors for that round.

The immediate fix is a one-line `err = nil` after the fallback succeeds.

*HOWEVER*, there's a problem with the whole setup: we calculate the metadata and store it to the database every advertise cycle.  We never _use_ the stored metadata.

The whole concept is that the service knows its name and tells it to the registry.  The ability to ask the registry was a one-time fallback.

h2zh:
This PR clears the `err` after logging the error message, so the registry-fetch failure no longer propagates up and aborts the advertise cycle.

There is a unresolved discussion in the GitHub issue. Regardless of the outcome, the changes in this PR remain necessary.

### DIFF
--- a/server_utils/sitename.go
+++ b/server_utils/sitename.go
@@ -111,12 +111,14 @@ func GetServerMetadata(ctx context.Context, server server_structs.ServerType) (m
 		metadata, err = getServerMetadataFromReg(ctx, originPrefix)
 		if err != nil {
 			log.Errorf("Failed to get metadata from the registry for the origin. Will fallback to using %s: %v", param.Xrootd_Sitename.GetName(), err)
+			err = nil
 		}
 	} else if server.IsEnabled(server_structs.CacheType) {
 		cachePrefix := server_structs.GetCacheNs(param.Xrootd_Sitename.GetString())
 		metadata, err = getServerMetadataFromReg(ctx, cachePrefix)
 		if err != nil {
 			log.Errorf("Failed to get metadata from the registry for the cache. Will fallback to use %s: %v", param.Xrootd_Sitename.GetName(), err)
+			err = nil
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


bbockelm:
What's the GitHub issue?

h2zh:
@bbockelm The GitHub issue is #3450. Based on the discussion there, I'd need to prioritize the name set by the server admin via config parameter, and use the name fetched from Registry as the fallback. I'll push a new commit for this change.

server_utils/sitename.go
```diff
@@ -89,7 +89,9 @@ func getServerMetadataFromReg(ctx context.Context, prefix string) (server server
 // under /origins; otherwise, we look it up under /caches.
 func GetServerMetadata(ctx context.Context, server server_structs.ServerType) (metadata server_structs.ServerRegistration, err error) {
 
-	// Fetch server metadata from the registry, if not, fall back server name to Xrootd.Sitename.
+	// Fetch server metadata from the registry. A failed lookup is logged and
+	// swallowedy. After the fetch, a non-empty Xrootd.Sitename takes priority
```
turetske:
```suggestion
	// swallowed. After the fetch, a non-empty Xrootd.Sitename takes priority
```

turetske (approved):
One minor grammar nitpick.

PR #3486, opened 2026-05-29
Ensure safe UI rendering by using unique key

### LINKED ISSUE #2772: Director Web UI duplicates server boxes
There was an issue with a similar symptom in the past, so I'm not sure whether this is a regression or a new bug.

When I entered `osg-stash` into a v7.20.2 Director's "Caches" search bar, the `ComputeCedar` cache started multiplying

Image

Notice that the Director correctly sees only 2 items, but displays many more.

Not sure whether this is more in Cannon's or Howard's court, so feel free to re-assign as needed.

h2zh:
Closes #2772 

A server from the Topology should not have the same name as another server. Even they do, this PR ensures the UI still get a unique key, which is composed by name and URL, so the UI can be rendered correctly.

Sidenote: The in-memory `serverAds` is keyed on `ad.URL.String()`, so the URL must be unique.

I also communicated this matter in #osdf-ops channel in [OSG slack](https://opensciencegrid.slack.com/archives/C05T5G2RUJ3/p1780071746233819).

### DIFF
--- a/web_ui/frontend/app/director/components/DirectorCardList.tsx
+++ b/web_ui/frontend/app/director/components/DirectorCardList.tsx
@@ -147,7 +147,7 @@ export function DirectorCardList({ data, cardProps }: DirectorCardListProps) {
         data={filteredData}
         Card={DirectorCard}
         cardProps={cardProps}
-        keyGetter={(o) => o.server.name}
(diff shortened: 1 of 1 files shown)


CannonLock (approved):
LGTM

PR #3488, opened 2026-05-29
Upgrade XRootD dependency to 5.9.5 (#3487)

### LINKED ISSUE #3487: Update xrootd to 5.9.5 in containers
XRootD 5.9.5 is out and we should update our containers to use it. This change should be backported to 7.26 and 7.25 as well.

matyasselmeci:


brianaydemir (approved):
LGTM — well, except for the fact that `XROOTD_VER` is buried in the middle of the `Dockerfile`, but that's a question best answered some other day.

PR #3490, opened 2026-05-30
Fix (some) data races found by `-race`

### LINKED ISSUE #3489: Data races found by the Go race detector (not necessarily exhaustive)
Running the test suite with `-race` surfaces a set of data races across several packages. Each is described below with the shared state that triggers the detector.

## 1. Federation discovery globals (`config/config.go`)

**Shared state:** `fedDiscoveryOnce *sync.Once` and `globalFedErr error`.

`GetFederation` reads `fedDiscoveryOnce` and calls `.Do()`; `SetFederation`, `InitClient`, `ResetConfig`, and `UpdateConfigFromListener` assign new values to the same pointer and error variable directly. Because the assignments are not synchronized with concurrent reads, the race detector fires on both the pointer write and the error write.

## 2. XRootD metric accumulators (`metrics/xrootd_metrics.go`)

**Shared state:** `lastOssStats OSSStatsGs` and `lastStats SummaryStat`.

`handleOSSStats` and `HandleSummaryPacket` update package-level delta accumulators: each reads the previous value, computes a delta, writes the new value, and emits a metric. Multiple goroutines processing packets concurrently perform these read-modify-write sequences on the same variables without any lock.

## 3. Origin handler registration flag (`origin_serve/handlers.go`)

**Shared state:** `handlersRegistered bool`.

`RegisterHandlers` reads `handlersRegistered` as an early-return guard and writes it to `true` after registration completes. Tests call `RegisterHandlers` concurrently (or reset and re-register without synchronization), so the check-and-set is not atomic.

## 4. Partially initialized director ad cache entries (`director/director_advertise.go`)

**Shared state:** `directorInfo.ad`, `directorInfo.cancel`, and `directorInfo.forwardAdChan`.

`updateInternalDirectorCache` allocates an empty `directorInfo`, publishes it into the shared TTL cache with `GetOrSet`, and then populates its fields. Concurrent readers of the cache (e.g. `sendMyAd`) could observe a partially-initialized entry where `ad`, `cancel`, and `forwardAdChan` were still zero values.

## 5. HTTP transport cache (`config/transport.go`)

**Shared state:** `onceTransport sync.Once` and the `transport`, `basicTransport`, `client`, etc. pointer variables.

`GetTransport` and other accessors call `onceTransport.Do(setupTransport)`. `ResetConfig` writes `onceTransport = sync.Once{}` and zeroes the pointer variables directly, without any lock. `sync.Once` serializes concurrent initializations through it, but does not protect against a concurrent assignment to the `Once` variable itself or to the transport pointers.

## 6. Launcher goroutine captures outer `err` (`launchers/launcher.go`)

**Shared state:** the `err` variable declared in the `LaunchModules` scope.

Several goroutines spawned by `LaunchModules` assign to the same captured `err` from the enclosing function. The parent may also read `err` after launching those goroutines, so concurrent writes and reads of the same variable occur without synchronization.

## 7. Director TTL cache iteration during eviction (`director/sort.go`)

**Shared state:** the internal list of the `serverAds` ttlcache.

`getAdsForPath` iterated the live cache with `Range`. The ttlcache eviction goroutine concurrently removes expired entries from the same internal list. The repository already avoids this pattern elsewhere by taking an `Items()` snapshot first.

## 8. ObjectStat request handler override (`director/stat.go`)

**Shared state:** `stat.ReqHandler`.

`queryServersForObject` read `stat.ReqHandler` inside goroutines it launched. Tests replace `stat.ReqHandler` between subtests. Goroutines from a prior subtest that have not yet finished can therefore race with the next subtest writing a new handler value.

## 9. Shared JWK key metadata mutation (`client/acquire_token.go`)

**Shared state:** a cached `jwk.Key` object.

Concurrent calls to `generateToken` each called `key.Set("kid", keyId)` on the same cached key object before signing. The JWK library mutates an internal metadata map when `Set` is called, so concurrent callers race on that map.

## 10. Advertisement `IOLoad` read outside lock (`director/monitor.go`)

**Shared state:** `Advertisement.ServerAd.IOLoad`.

`LaunchPeriodicDirectorTest` copies `adItem.Value().ServerAd` directly. That struct copy reads `IOLoad` without the `Advertisement`'s `RWMutex`, while `SetIOLoad` holds the same mutex when writing the field.

## 11. `*gin.Context` passed to storage and outbound HTTP calls (`oauth2/issuer/`, `web_ui/`)

**Shared state:** internal fields of a pooled `*gin.Context`.

`*gin.Context` implements `context.Context`, so it compiles wherever a `context.Context` is accepted. But gin reads and writes its own context fields (response writer, key/value store, etc.) throughout the request lifecycle. Storage calls and outbound HTTP requests that received a `*gin.Context` directly held a reference to that mutable object. The `database/sql` package spawns an `awaitDone` goroutine that reads the context after the handler returns; gin may have already recycled the object by then.

## 12. Server identity ad cache reset (`server_utils/sitename.go`)

**Shared state:** `baseAdOnce sync.Once`, `baseAd server_structs.ServerBaseAd`, and `baseAdErr error`.

`IsDirectorAdFromSelf` initializes the cache lazily through `baseAdOnce`. `ResetTestState` writes new values to all three variables directly. The same `sync.Once`-vs-assignment race as in the transport cache above: the `Once` serializes concurrent initializations but not a concurrent reset.

## 13. Shared `rand.Rand` in PelicanFS stress test (`client/pelican_fs_test.go`)

**Shared state:** a single `*math/rand.Rand` created before the stress-test goroutines are launched.

`math/rand.Rand` is not safe for concurrent use. All worker goroutines called `Intn` on the same instance, which races on the RNG's internal state.

## 14. Test log hook formatting races with cleanup (`test_utils/utils.go`)

**Shared state:** `logrus.StandardLogger().ReportCaller` and the test hook.

`SetupTestLogging` registers a `t.Cleanup` that restores the global logrus logger state. If a goroutine is still formatting or firing a log entry through the test hook when cleanup runs, it reads `entry.Logger.ReportCaller` while cleanup writes it.

------

_(Brian A: With a tip of the hat to Copilot. Inspired by the mess of failures after [the nightly unit test runs started using `-race`](https://github.com/PelicanPlatform/pelican/pull/3266).)_

brianaydemir:
Fixes the races described in #3489.

### `sync.Once` + reset races (`config`, `server_utils`)

Three caches follow the same broken pattern: a `sync.Once` guards lazy initialization, but test-reset paths overwrite the `Once` variable and the cached values with bare assignments, racing with concurrent readers.

`sync.Once` serializes concurrent calls *through* it, but provides no protection against a concurrent assignment to the `Once` variable itself. The fix in all three cases is the same: introduce a `sync.Mutex` that is held by both the init path and the reset path, making them mutually exclusive. The `Once` continues to do its job (run setup at most once per cycle); the mutex closes the gap.

Helper functions (`initTransport`/`resetTransport`, `resetFederationDiscoveryState`/`clearGlobalFederationState`) encapsulate the lock scope so call sites stay linear. The lock is *not* held across the entire `Do` callback — only across the `Do` call and the variable assignments — which keeps the critical sections short.

For the federation discovery case, `globalFedInfo` was already an `atomic.Pointer` and remains so; only `fedDiscoveryOnce` and `globalFedErr` needed the new mutex.

### Director ad cache initialization (`director/director_advertise.go`)

`updateInternalDirectorCache` was publishing a zero-value `*directorInfo` into the TTL cache via `GetOrSet`, then populating its fields afterward. Any reader that obtained the entry between `GetOrSet` and field population would see nil `forwardAdChan`, nil `cancel`, and an empty ad.

The fix constructs a fully initialized `directorInfo` — including `context.WithCancel`, `forwardAdChan`, and the ad itself — before calling `GetOrSet`. If `GetOrSet` reports that an existing entry won the race, the newly created cancel function is called immediately to release the unused context.

### `Advertisement.IOLoad` read outside lock (`director/monitor.go`)

`Advertisement` already has an `RWMutex` used by `SetIOLoad`/`GetIOLoad`. Copying the embedded `ServerAd` struct directly bypassed that lock because Go struct copies are not atomic. The fix adds `GetServerAd()` which holds the read lock for the duration of the copy, mirroring the existing `GetIOLoad` pattern.

### `*gin.Context` escaping to storage and I/O callers (`oauth2/issuer`, `web_ui`)

`*gin.Context` implements `context.Context`, so it compiles where a `context.Context` is expected. It is, however, mutable request state that gin reads and writes throughout the request lifecycle; passing it to storage or outbound-HTTP calls gives those callees a reference to that state, which races with gin's own concurrent modifications.

The fix passes `ctx.Request.Context()` instead — the underlying `net/http` request context, whose lifetime is tied to the HTTP request and which is safe to share with any goroutine. A small unexported helper `requestBoundContext` is introduced in the `oauth2/issuer` package to make the right choice explicit and to document the reason at the definition site. `mirrorDowntimeToRegistry` in `web_ui` follows the same pattern inline (`reqCtx := ctx.Request.Context()`) rather than calling the helper, because the helper is unexported and in a different package.

### JWK key cloning (`client/acquire_token.go`)

The cached `jwk.Key` was treated as effectively immutable for the purposes of signing, but `key.Set("kid", ...)` mutates an internal map. The fix calls `key.Clone()` before setting the key ID, leaving the cached key untouched.

### Director TTL cache snapshot (`director/sort.go`)

The repository already contains a comment in `director_api.go` noting that `Range` races with the eviction goroutine and that `Items()` should be used instead. This change applies the same fix to `getAdsForPath` in `sort.go`.

### Narrow-scope fixes

- **XRootD metric accumulators (`metrics/xrootd_metrics.go`):** Two separate mutexes, one per accumulator family, rather than one coarse lock. This avoids serializing unrelated metric families while keeping each read-modify-write sequence atomic.

- **Origin handler registration (`origin_serve/handlers.go`):** A single mutex around the check-and-set of `handlersRegistered`. The registration path can return errors and tests need to reset the flag, so a mutex is less surprising than a `sync.Once` or `atomic.Bool`.

- **Launcher error capture (`launchers/launcher.go`):** Goroutine-local error variables instead of capturing the outer `err`. The existing error channel is unchanged; only the intermediate local assignments are made goroutine-local.

- **PelicanFS stress test RNG (`client/pelican_fs_test.go`):** Per-goroutine `rand.Rand` instances. No production code is affected.

- **Test log hook (`test_utils/utils.go`):** The cleanup hook checks `entry.Caller` directly rather than `entry.Logger.ReportCaller`. The global logger flag is only needed to decide whether to populate `entry.Caller` in the first place; by the time the hook fires, `entry.Caller` is already set or nil.

---

_(Brian A: With a tip of the hat to Copilot.)_

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2276,8 +2276,8 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
 	if err := GenerateSessionSecret(); err != nil {
 		return err
 	}
-	// After we know we have the certs we need, call setupTransport (which uses those certs for its TLSConfig)
-	setupTransport()
+	// After we know we have the certs we need, call initTransport (which uses those certs for its TLSConfig)
+	initTransport()
 
 	// Setup CSRF middleware. To use it, you need to add this middleware to your chain
 	// of http handlers by calling config.GetCSRFHandler()
@@ -2432,9 +2432,7 @@ func ResetConfig() {
 	logging.ResetLogFlush()
 
 	// Clear cached transport object
-	onceTransport = sync.Once{}
-	transport = nil
-	basicTransport = nil
+	resetTransport()
 
 	// Clear cached SSRF transport object
 	ResetSSRFTransportForTest()
--- a/config/config.go
+++ b/config/config.go
@@ -1,6 +1,6 @@
 /***************************************************************
 *
-* Copyright (C) 2025, Pelican Project, Morgridge Institute for Research
+* Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
 *
 * Licensed under the Apache License, Version 2.0 (the "License"); you
 * may not use this file except in compliance with the License.  You may
@@ -173,6 +173,7 @@ var (
 	// until it's first needed, avoiding a web lookup for invoking configuration
 	// Note the 'once' object is a pointer so we can reset the client multiple
 	// times during unit tests
+	fedDiscoveryMu   sync.Mutex
 	fedDiscoveryOnce *sync.Once
 	globalFedInfo    atomic.Pointer[pelican_url.FederationDiscovery]
 	globalFedErr     error
@@ -219,6 +220,22 @@ var (
 	// Production code should NOT modify this value.
 	sysConfigLocation string = filepath.Join("/etc", "pelican")
 )
+
+func resetFederationDiscoveryState() {
+	fedDiscoveryMu.Lock()
+	defer fedDiscoveryMu.Unlock()
+
+	fedDiscoveryOnce = &sync.Once{}
+}
+
+func clearGlobalFederationState() {
+	fedDiscoveryMu.Lock()
+	defer fedDiscoveryMu.Unlock()
+
+	fedDiscoveryOnce = &sync.Once{}
+	globalFedInfo.Store(&pelican_url.FederationDiscovery{})
+	globalFedErr = nil
+}
 
 func init() {
 	en := en.New()
@@ -611,9 +628,9 @@ func discoverFederationImpl(ctx context.Context) (fedInfo pelican_url.Federation
 }
 
 // Reset the fedDiscoveryOnce to update federation metadata values for GetFederation().
-// Should only used for unit tests
+// Should only be used for unit tests.
 func ResetFederationForTest() {
-	fedDiscoveryOnce = &sync.Once{}
+	resetFederationDiscoveryState()
 }
 
 // Retrieve the federation service information from the configuration.
@@ -623,6 +640,9 @@ func ResetFederationForTest() {
 // If invoked before things are configured, it must be done from a single-threaded
 // context.
 func GetFederation(ctx context.Context) (pelican_url.FederationDiscovery, error) {
+	fedDiscoveryMu.Lock()
+	defer fedDiscoveryMu.Unlock()
+
 	if fedDiscoveryOnce == nil {
 		fedDiscoveryOnce = &sync.Once{}
 	}
@@ -655,6 +675,9 @@ func SetFederation(fd pelican_url.FederationDiscovery) {
 	}); err != nil {
 		log.WithError(err).Warn("Failed to update federation configuration")
 	}
+
+	fedDiscoveryMu.Lock()
+	defer fedDiscoveryMu.Unlock()
 
 	globalFedInfo.Store(&fd)
 	globalFedErr = nil
@@ -2268,7 +2291,7 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
 
 	// Sets (or resets) the federation info. Unlike in clients, we do this at startup
 	// instead of deferring it.
-	fedDiscoveryOnce = &sync.Once{}
+	resetFederationDiscoveryState()
 	if _, err := GetFederation(ctx); err != nil {
 		return err
 	}
@@ -2363,7 +2386,7 @@ func InitClient() error {
 	}
 
 	// Set (or reset) the deferred federation lookup
-	fedDiscoveryOnce = &sync.Once{}
+	resetFederationDiscoveryState()
 
 	// Set up the log filter mechanisms, e.g., for sensitive secrets
 	initFilterLogging()
@@ -2423,9 +2446,7 @@ func ResetConfig() {
 	ClearServerAds()
 
 	// Reset federation metadata
-	fedDiscoveryOnce = &sync.Once{}
-	globalFedInfo.Store(&pelican_url.FederationDiscovery{})
-	globalFedErr = nil
+	clearGlobalFederationState()
 
 	warnDeprecatedOnce = sync.Once{}
 	warnDebugOnce = sync.Once{}
--- a/config/config.go
+++ b/config/config.go
@@ -164,13 +164,13 @@ var (
 	preCleanupFuncs map[string]func()
 
 	// Global discovery info.  Using the "once" allows us to delay discovery
-	// until it's first needed, avoiding a web lookup for invoking configuration
-	// Note the 'once' object is a pointer so we can reset the client multiple
-	// times during unit tests
-	fedDiscoveryMu   sync.Mutex
-	fedDiscoveryOnce *sync.Once
-	globalFedInfo    atomic.Pointer[pelican_url.FederationDiscovery]
-	globalFedErr     error
+	// until it's first needed, avoiding a web lookup for invoking configuration.
+	// Both the once and the discovery result are held in atomic pointers so
+	// they can be published, read, and reset without a mutex: swapping in a
+	// fresh sync.Once (during unit tests, or after SetFederation) is a single
+	// atomic store, and readers snapshot the result via a single atomic load.
+	fedDiscoveryOnce atomic.Pointer[sync.Once]
+	globalFedInfo    atomic.Pointer[fedDiscoveryResult]
 
 	// Global struct validator
 	validate *validator.Validate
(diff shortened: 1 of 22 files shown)


config/config.go
```diff
@@ -173,6 +173,7 @@ var (
 	// until it's first needed, avoiding a web lookup for invoking configuration
 	// Note the 'once' object is a pointer so we can reset the client multiple
 	// times during unit tests
+	fedDiscoveryMu   sync.Mutex
 	fedDiscoveryOnce *sync.Once
 	globalFedInfo    atomic.Pointer[pelican_url.FederationDiscovery]
```
h2zh: The new mutex makes the atomic redundant. Now that every access to globalFedInfo happens under `fedDiscoveryMu`, a plain field would be equally safe. The atomic is leftover from when readers were lock-free; it's harmless but no longer load-bearing.

h2zh (commented):
I'm going to incremently review all these 14 commits one by one, and you could address these comments either all at once or in the opposite way. First of all, this PR needs to rebase to main because of the param generation overhaul. Here is the comment for the first commit (214f55b4561a4e4c4531ed92bd796c0227c77380)

config/transport.go
```diff
@@ -75,58 +74,75 @@ func GetTransport() *http.Transport {
...
+// initTransport runs setupTransport at most once per init cycle, under
+// transportMu. The mutex is necessary because sync.Once serializes
+// concurrent calls *through* it, but does not protect against a concurrent
+// resetTransport overwriting the onceTransport variable itself or the
+// transport pointers.
+func initTransport() {
+	transportMu.Lock()
+	defer transportMu.Unlock()
+	onceTransport.Do(setupTransport)
+}
```
h2zh, line 126: InitClient (config/config.go:2380) still calls setupTransport() directly, bypassing this new mutex.

config/config.go
```diff
@@ -2432,9 +2432,7 @@ func ResetConfig() {
 	logging.ResetLogFlush()
 
 	// Clear cached transport object
-	onceTransport = sync.Once{}
-	transport = nil
-	basicTransport = nil
+	resetTransport()
 
 	// Clear cached SSRF transport object
 	ResetSSRFTransportForTest()
```
h2zh, line 2518: SSRF transport singleton is similar to `transport`, which has the same race condition problem. Would you like to handle it in a new commit here or in a future PR?

h2zh (changes requested):
The following two comments are also for the `sync.Once` reset race condition. (commit ef2c087a884c9815b70300823dc65b99f2c209d1)

The commit message of ef2c087a asserts the mutex prevents reset-vs-getter-read races, and the `resetTransport` comment (config/transport.go:128-132) names "concurrent reads in the getters" as the hazard being guarded against. However the getters (transport.go:67-115) read the pointer outside `transportMu`, so neither is actually achieved. I ran race condition detector at test-linux.yml:95-107 locally and the result confirms that.

config/config.go
h2zh: Besides, this mutex is held during the entire `fedDiscoveryMu`, which includes a HTTP call in `discoverFederationImpl`. If this networking call hangs, the lock will hold for a long time. A context with timeout could mitigate this problem, however, most invocations of `GetFederation(ctx)` doesn't have a timeout in the codebase. 
A possible fix could be refactoring `GetFederation(ctx)` function: Split the slow part (discovery HTTP call) from the part that needs the lock (publishing the three variables). Snapshot the Once under the lock, release, run discovery lock-free, then re-acquire only to store results.

director/monitor.go
```diff
@@ -176,7 +176,7 @@ func LaunchPeriodicDirectorTest(ctx context.Context, serverUrlStr string) {
 			log.Infof("The Director doesn't have any advertisements for server with URL %s. Stopping director tests.", serverUrlStr)
 			return false
 		}
-		currentServerAd := adItem.Value().ServerAd
+		currentServerAd := adItem.Value().GetServerAd()
```
h2zh, line 184: director/director_api.go:106 also needs this fix. Same for director_api.go:322, :369, :376.

server_structs/director.go
```diff
@@ -607,6 +607,12 @@ func (ad *Advertisement) GetIOLoad() float64 {
 	return ad.IOLoad
 }
 
+func (ad *Advertisement) GetServerAd() ServerAd {
```
h2zh, line 697: Could you add a docstring here so human and agent knows this is the required function to get an ad? And `adItem.Value().ServerAd` is prohibited.

director/monitor.go
h2zh: Also director/cache_ads.go:282-297

metrics/xrootd_metrics.go
```diff
@@ -990,7 +990,9 @@ var (
 		Help: "Statistics about connection calls.",
 	}, []string{"type"})
 
+	lastStatsMu      sync.Mutex
 	lastStats        SummaryStat
+	lastOssStatsMu   sync.Mutex
 	lastOssStats     OSSStatsGs
 	lastS3CacheStats OssS3CacheGs
```
h2zh: Why here we don't add a lock for `lastS3CacheStats` in `handleS3CacheStats`?
That handler is reached through the same dispatch path as the one (`handleOSSStats`) this commit shows: handlePacket → handleOSSPacket → {handleOSSStats | handleS3CacheStats}. `lastS3CacheStats` has identical exposure to `lastOssStats` but was left unguarded

h2zh (changes requested):
When I run the nightly race condition detector at test-linux.yml:95-107 locally against this PR, four other data races surface. Could you take a look?
This is the last piece of the first-round review. When you address these comments, could you make new commits instead of overwrite the existing ones? In this way, I can view the diff clearer.

config/config.go
bbockelm: Maybe simpler - replace the sync pointer with an atomic sync pointer?

That said, the definition of Reset is that it should only be done when no goroutine is running that could read the once.  If we are seeing races then it strongly implies there are goroutines leaking between test runs: that would be the underlying issue!

bbockelm: I take that back - I see where we write to this during startup.

An atomic pointer would do!

That said, if we are worried about long-held mutexes, since the mutex lifetime is guaranteed to be the same as the Once blocking, we've not made the problem any worse or better.

PR #3504, opened 2026-06-03
Register all issuer public keys at startup, not just the active one

### LINKED ISSUE #3499: Issuer key auto-update fails to register keys present before startup
The integration team found two gaps in the issuer-key auto-update mechanism that leave the registry's stored JWKS out of sync with the keys an origin actually holds:

1. Keys present at startup are never advertised. If a second key file already exists in the issuer-keys directory before the Pelican process starts, only the first (active) key gets registered. The second key is never pushed to the registry.
2. #3505

h2zh:
An origin/cache only ever registered its active (lexicographically-first/smallest) issuer key. Additional keys already present in the issuer-keys directory before startup were never recorded by the registry, because initial registration sends a single key and the key-change-driven update only fires when a *new* key (a key with different key id) is later detected.

This PR records the full issuer keyset across three paths:

1. Initial registration: `NamespaceRegister` now advertises the full set from `GetIssuerPublicJWKS` so the Registry records every local issuer key in one registration. The signing key is forced to index 0, since `GetIssuerPublicJWKS` returns keys in non-deterministic map order and the registry verifies proof-of-possession against the first key only (`registry.validateJwks`). `validateJwks` keeps checking index-0 key for backward compatibility.
2. Startup match check: `registerNamespacePrep` now checks every key in the keyset and only fails if none match, fixing the spurious "already registered under a different key" error when the active key differs from the registered one. (The key used to register this namespace is no longer the active signing key, but still in the keys directory).
3. Restart reconcile: once a namespace is registered, initial registration is skipped, so keys added while the server was down were never pushed. `RegisterNamespaceWithRetry` now reconciles by pushing the full keyset via the update endpoint (a no-op when unchanged).

### DIFF
--- a/cmd/registry_client.go
+++ b/cmd/registry_client.go
@@ -92,24 +92,8 @@ func registerANamespace(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	publicKey, err := config.GetIssuerPublicJWKS()
-	if err != nil {
-		log.Error("Error: Failed to retrieve public key: ", err)
-		os.Exit(1)
-	}
-
-	/*
-	 * TODO: For now, we only allow namespace registration to occur with a single key, but
-	 *       at some point we should expose an API for adding additional pubkeys to each
-	 *       namespace. There is a similar TODO listed in registry.go, as the choices made
-	 *       there mirror the choices made here.
-	 * To enforce that we're only trying to register one key, we check the length here
-	 */
-	if publicKey.Len() > 1 {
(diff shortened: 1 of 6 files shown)


jhiemstrawisc (changes requested):
This drops the "TODO" in `registry/registry_validation.go`, but leaves the similar "TODO" in `cmd/registry_client.go` (line 101-107).

Can you bring these two areas into agreement, or update the comment to reflect why you want to leave the registry client as is?

Otherwise, I think this looks okay!

turetske (approved):
LGTM!

PR #3507, opened 2026-06-04
Remove v1 ads

### LINKED ISSUE #3506: Remove V1 Advertisements
V1 Server Advertisements are only used in Pelican v7.8 servers and earlier. No server in the OSDF federation still uses it. Excise the logic that enforces backwards compatibility for the sake of these origins and only use the V2 advertisements.

turetske:
Removes the V1 server advertisement code.

close #3506


cache/advertise.go
```diff
@@ -174,34 +174,14 @@ func (server *CacheServer) GetNamespaceAdsFromDirector() error {
...
-			}
-		} else {
-			return errors.Wrap(err, "Failed to make request")
-		}
-	} else {
-		err = json.Unmarshal(respData, &respNS)
-		if err != nil {
-			return errors.Wrapf(err, "Failed to marshal response in to JSON: %v", err)
-		}
+		return errors.Wrap(err, "Failed to make request")
```
jhiemstrawisc: Can you lowercase the error messages throughout the code you touched?

director/director.go
```diff
@@ -1192,24 +1192,15 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
 		}
 	}
 
-	ad := server_structs.OriginAdvertiseV1{}
 	adV2 := server_structs.OriginAdvertiseV2{}
```
jhiemstrawisc: Any reason not to switch the code so everything refers simply to "advertisement" and "ad" instead of keeping the "V2" stuff? I'm on the fence because "V2" is a clear artifact that something came before (breadcrumbs for future devs), but I'm also pretty certain the V1 stuff is sufficiently dead that there's no reason we'll ever have to do that archaeology.

director/director.go
```diff
@@ -1192,24 +1192,15 @@ func registerServerAd(engineCtx context.Context, ctx *gin.Context, sType server_
...
-			return
-		}
-	} else {
-		// If the OriginAdvertisement is a V1 type, convert to a V2 type
-		adV2 = server_structs.ConvertOriginAdV1ToV2(ad)
+		log.Debugln("Failed to parse ad of type", sType.String(), "due to error:", err)
+		ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+			Status: server_structs.RespFailed,
+			Msg:    fmt.Sprintf("Invalid %s registration", sType),
+		})
```
jhiemstrawisc: A few questions about observability of this error case:
1. Does the logged debug line include enough information that a Director administrator can quickly tell what service is producing this error? That's important so they know who to talk to next for debugging.
2. Why not return the full error to the service so they can quickly pass along the error message to the Director administrator instead of making the administrator toggle debug logs and then hunt for the log line (assuming 1 is answered affirmatively)?

director/director.go
```diff
@@ -1844,7 +1827,6 @@ func RegisterDirectorAPI(ctx context.Context, router *gin.RouterGroup) {
 		directorAPIV1.POST("/registerOrigin", serverAdMetricMiddleware, func(gctx *gin.Context) { registerServerAd(ctx, gctx, server_structs.OriginType) })
 		directorAPIV1.POST("/registerCache", serverAdMetricMiddleware, func(gctx *gin.Context) { registerServerAd(ctx, gctx, server_structs.CacheType) })
 		directorAPIV1.GET("/getFedToken", getFedToken)
-		directorAPIV1.GET("/listNamespaces", listNamespacesV1)
```
jhiemstrawisc, line 1852: Maybe instead of completely removing this, it should return a deprecation notice? I'm also on the fence here, where I think it's nice to leave breadcrumbs, but this stuff is also so old that I can't imagine anyone ever hitting this API.

jhiemstrawisc (changes requested):
Are there any Swagger docs that should be updated as part of this?

director/director.go
turetske: I want to be consistent, are we removing V1/V2 references or not. This contradicts your earlier suggestion of renaming the V2 stuff.

jhiemstrawisc: Yep, I completely agree that consistency is good. My earlier suggestion was more of a question about what you think and a statement that I'm on the fence but don't know which side of that fence I prefer. If you choose to keep things as is, I'm okay with that as long as we've talked about it and articulated the decision.

jhiemstrawisc: After a few more minutes of thought, I think it'd be better to leave this API plumbed, but to respond with a deprecation alert ([SO suggests HTTP 299](https://stackoverflow.com/questions/13884141/convention-for-http-response-header-to-notify-clients-of-deprecated-api)). This also prevents a developer from accidentally binding to the path in the future.

Keeping a public-facing but disabled API can be treated separately than our internally-named variables, where I think it makes more sense to switch to regular "ad" and "advertisement".

Thoughts?

turetske: I agree with this compromise. I'll implement it.

h2zh (approved):
This PR generally looks good. Could you also modify pelican-swagger.yaml:311 and :333?

PR #3512, opened 2026-06-08
Add built-in log rotation for file logging

### LINKED ISSUE #1999: Add mechanism to reopen logfile
Pelican Service:
- [ ] Client
- [ ] Plugin
- [x] Registry
- [x] Director
- [x] Origin
- [x] Cache

Currently log rotation (using logrotate) requires the `copytruncate` method, wherein a rotation is performed by copying the log file, then truncating the original.  The downside to this is that if the log file suddenly grows to be at least half of the log volume, you can no longer rotate it (because you don't have enough room to make a copy).  Since you can't rotate it, the original will just keep getting bigger until it fills up the volume.

A better way to do it is for logrotate to rename the file, then tell the daemon to close the file and open a new logfile (with the old name).  This is usually done with Unix signals, typically SIGHUP or SIGUSR1.

Add a mechanism by which logrotate can tell Pelican to close and reopen the log file.

### LINKED ISSUE #3514: Build in log rotation
We originally considered log rotation a "system issue" - we create the logging, the OS is responsible for managing the volume (rotation, compress, delete old, etc).

That's a reasonable principle on paper but has turned out to be impractical in practice.

If you're running under systemd in Linux, it works well.  If you're running in a container ... well, then the work is punted back to the user to set up the log rotation.  Worse, if you are utilizing the client agent, then you're an unprivileged user and don't have any OS functionality (or container functionality) to lean on.

Let's take up the mantle of log rotation.

bbockelm:
Picked at a long-running irritation over the weekend: lack of log rotation. What
started there grew into the surrounding pieces an operator needs before rotated
logs are actually useful — somewhere to read them from without shelling in, and
a way to hand out that access without handing out administration.

Four things ship here.

## 1. Built-in, asynchronous log writing and rotation

When `Logging.LogLocation` points at a regular file, Pelican now manages its own
log writing and rotation, so it "just works" on hosts with no external
log-rotation service — notably the user-level client agent.

A dedicated, errgroup-managed goroutine drains an in-memory batch to the log file
(flushing on a byte threshold or every `Logging.Rotation.FlushInterval`),
decoupling logging call sites from disk I/O. The buffer is bounded: once it
fills, `Write` applies backpressure so callers cannot outrun a slow or stalled
log device. On shutdown the writer flips to synchronous mode so late lines — from
signal handling, a panic, or `log.Fatal` — are written directly. A failed write is
fatal: the goroutine surfaces it so the errgroup cancels shutdown. Pelican should
not keep serving data it can no longer account for.

Rotation applies to regular files only, never to stderr, `/dev/stdout`, pipes or
devices:

- `Logging.Rotation.Frequency`: daily/hourly/none, on the calendar boundary in
  local time. Rotated files are named for the period they cover (e.g.
  `pelican.log.2026-06-08`) so an administrator can find logs by date.
- `Logging.Rotation.MaxSize`: size-triggered rotation, combinable with time.
- `Logging.Rotation.MaxRetentionSize` / `MaxRetentionPeriod`: independent size
  and age budgets, replacing a file-count cap.
- `Logging.Rotation.Disable` / `DisableCompress`: zero-value-friendly toggles.

Rotation holds an `os.Root` handle on the log directory, so it keeps working
after a privilege drop, and it survives restarts: a stale active file is rotated
under the period it was last written in. When rotation is left to an external
tool, the writer notices that the path no longer refers to the file it holds open
and reopens — otherwise `logrotate`'s rename would leave Pelican writing to an
unlinked inode forever.

Failing to rotate is tolerated: it is usually transient, and it is not a reason to
stop serving data. It is retried, and the growth it causes is bounded by
`MaxRetentionSize` — past that the log would fill the filesystem, which is fatal.

## 2. An in-memory log buffer and a browser log viewer

`config.LogRingBuffer` keeps a bounded window of recent lines in memory
(`Logging.Buffer.MaxSize`, `Logging.Buffer.BatchLines`), batched and LZ4-compressed,
so an administrator can read recent logs from the web UI instead of shelling into
the host. A new page at **Settings → Logs** tails it live, filters by level and
text, virtualizes its row list so a busy server cannot freeze the tab, and offers
the whole buffer as a gzip download.

The buffer holds what the log file holds, including the censoring: lines are
passed through the bearer-token redaction before being stored, so the buffer
never becomes a source of credentials the log file was careful not to keep.

Reads report what they could not show. The buffer evicts, so a client that falls
behind has lines disappear from under its cursor; because cursors are opaque, it
cannot work that out for itself. Responses carry a `dropped` count and the viewer
marks the break, rather than presenting a truncated history as a complete one.

## 3. A new user-grantable scope: `pelican.log_read`

**This adds a permission.** `pelican.log_read` is `userGrantable`, and it admits
its holder to the log viewer and the two new endpoints below without making them
an administrator. Granting it is equivalent to granting sight of everything the
server logs, which is exactly why it is separate: reading logs is a common
triage need, and it should not require handing over the server.

Only a full `server.admin` can grant it. `server.admin` carries it implicitly.
The scope is resolved from the user's grants when a request arrives, so it
governs what an account may reach rather than what a particular credential may
do.

New endpoints, both behind that gate:

- `GET /api/v1.0/logs/tail` — a page of buffered lines, addressed by an opaque
  cursor, forwards for live tail or backwards to scroll into history.
- `GET /api/v1.0/logs/download` — the whole buffer, gzipped.

Because the scope-only route into `/settings/` is new, the settings shell and the
subsystem sidebars now admit scope holders and hide the sub-pages they cannot
reach, and `/settings/` sends them to the first page they can.

## 4. Non-Linux runtime path defaults

`Origin.RunLocation`, `Cache.RunLocation`, `Cache.StorageLocation`,
`Cache.DataLocations`, `Cache.MetaLocations`, `Cache.NamespaceLocation` and
`LocalCache.RunLocation` get sane defaults on macOS and Windows, where the Linux
`/run` layout does not exist and the `AF_UNIX` path budget is tight enough to
matter. Configured values are respected; only the defaults move.

## Default changes worth noting

`Logging.Rotation.MaxSize` drops from 1GB to 100MB and
`Logging.Rotation.MaxRetentionSize` from 10GB to 1GB. The previous pair reserved
roughly 11GB of disk by default, which is a lot to claim without being asked —
particularly for the client agent, which writes under the user's own directory.

## Testing

Go tests cover the writer's rotation, retention, compression, shutdown and
failure paths (including concurrent writers under `-race`), the ring buffer's
cursor arithmetic, eviction accounting and redaction, and the authorization gate
on the new endpoints. Playwright tests cover the viewer end to end — live tail,
error and recovery, virtualization, scroll-back, dropped-line marking, download,
and who is and is not offered the page — on both the origin and cache suites.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -42,7 +42,12 @@ var cliExecErrorHook func(err error)
 //go:generate go run -tags server . generate-docs docs/app/commands-reference/pelican-server
 func main() {
 	logging.SetupLogBuffering()
+	// On a panic (or any unwind), flip the log writer to synchronous mode first
+	// so buffered lines are flushed and late lines reach the log file, then run
+	// the normal flush. Defers run LIFO, so EnterSyncMode (registered last) runs
+	// before FlushLogs.
 	defer logging.FlushLogs(false)
+	defer logging.EnterSyncMode()
 	if len(os.Args) > 1 && os.Args[1] == "generate-docs" {
 		outputDir := "docs/app/commands-reference"
 		if len(os.Args) > 2 {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -55,14 +55,24 @@ func main() {
 		}
 		err := generateCLIDocs(outputDir)
 		if err != nil {
-			os.Exit(1)
+			exitWithFlush(1)
 		}
 		return
 	}
 	err := handleCLI(os.Args)
 	if err != nil {
-		os.Exit(1)
+		exitWithFlush(1)
 	}
+}
+
+// exitWithFlush terminates the process after draining the log writer. The
+// deferred flush in main covers a normal return and a panic unwind, but
+// os.Exit runs no defers, so every explicit exit has to drain for itself or
+// lose whatever is still buffered -- including the line explaining the exit.
+func exitWithFlush(code int) {
+	logging.EnterSyncMode()
+	logging.FlushLogs(false)
+	os.Exit(code)
 }
 
 func handleCLI(args []string) error {
@@ -94,7 +104,7 @@ func handleCLI(args []string) error {
 		cliExecErrorHook(err)
 	}
 	if err != nil {
-		os.Exit(1)
+		exitWithFlush(1)
 	}
 	return nil
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -109,13 +109,10 @@ var egrpPostHandler func(error) (bool, error)
 func Execute() error {
 	egrp, egrpCtx := errgroup.WithContext(context.Background())
 	ctx := context.WithValue(egrpCtx, config.EgrpKey, egrp)
-	// Register the errgroup with the logging subsystem so the
-	// asynchronous log writer stops when the context is cancelled.
+	// Register the errgroup with the logging subsystem so the asynchronous log
+	// writer (and its compression worker) run under the errgroup and stop when
+	// the context is cancelled.
 	logging.SetErrgroup(ctx, egrp)
-	// Wait for any in-flight background log rotation to finish before the
-	// process exits, so a graceful shutdown does not abandon a compression
-	// mid-write.
-	defer logging.AwaitCompression()
 	exeErr := rootCmd.ExecuteContext(ctx)
 	if exeErr != nil {
 		log.Debugln("Fatal error occurred at the start of the program. Cleanup started:", exeErr)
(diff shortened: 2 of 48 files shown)


bbockelm:
@CannonLock - are you happy with the things you fixed?  I manually reviewed + went through with Fable ... I think all is good.  Ready to merge?

CannonLock (approved):
Yes, LGTM!

PR #3513, opened 2026-06-11
Stop iterating the live serverAds cache under its own lock

jhiemstrawisc:
getAdsForPath walked serverAds with ttlcache's Range on every redirect. Range releases and re-acquires the cache's internal read lock around each element and leaks that lock if an entry is evicted or deleted mid-walk -- routine under registration churn plus TTL expiry. One leaked reader is fatal: the write-preferring RWMutex then blocks every reader and writer on the cache. And because updateDowntimeFromRegistry held the global filteredServersMutex while reading serverAds, that stall propagated into the registration path and piled up goroutines until the director was OOM-killed.

Fix the class of bug, not just the one call site:

- Make getServerAdsSnapshot the single idiom for walking the cache, backed by the atomically-locked Items(). Callers iterate their own copy with no cache lock held, so a slow or misbehaving cache can't wedge them. Both the helper and the serverAds declaration document why Range must never be used, so a future performance tweak doesn't reintroduce it. Items() stays only where the keys or Item wrappers are actually needed.

- Never hold filteredServersMutex across a serverAds access -- the two subsystems have independent locks and nesting them is an ABBA hazard. updateDowntimeFromRegistry now snapshots before locking.

The underlying Range lock-leak is an upstream ttlcache bug worth reporting separately; this change makes the director robust regardless.

### DIFF
--- a/director/cache_ads.go
+++ b/director/cache_ads.go
@@ -56,6 +56,9 @@ const (
 
 var (
 	// The in-memory cache of xrootd server advertisement, with the key being ServerAd.URL.String()
+	//
+	// To iterate this cache, use getServerAdsSnapshot (or Items when you also need the keys or
+	// Item wrappers). Do not use Range, which is not safe to iterate concurrently with cache writes.
 	serverAds = ttlcache.New(ttlcache.WithTTL[string, *server_structs.Advertisement](param.Director_AdvertisementTTL.GetDuration()),
 		ttlcache.WithDisableTouchOnHit[string, *server_structs.Advertisement]())
 	// The map holds servers that are disabled, with the key being the ServerAd.Name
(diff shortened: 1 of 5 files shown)


jhiemstrawisc:
@h2zh and @turetske once this is approved/merged, it should be backported into 7.25 and 7.26. We _suspect_ this is the cause of the goroutine buildups in the 7.25 ITB/OSDF Directors.

h2zh (changes requested):
This is a clean fix. It aligns with what I found in the ITB Director core dump (lock-coupling defect of serverAds) and furthur pinpoint the root cause.

Good to merge after the failed tests are resolved. We discussed this offline, one is caused by Harbor outage, but `TestUpdateConfig_InvalidatesTokenCache` seems to fail because of the code change in this PR. I try to run tests in `/local_cache/cache_authz_test.go` locally. This test succeeds when I run it in the main branch, but fails in this branch.

jhiemstrawisc:
FYI, I rolled back the ttlcache library bump -- it caused failing tests, and after I started pulling on the thread I decided not to keep pulling. The fix here avoids the library's buggy `Range()` function.

h2zh (approved):
LGTM!

h2zh:
This PR is also cherry-picked into branch `v7.25.x` and then rolled out a new tag `v7.25.1-rc.6`.

PR #3517, opened 2026-06-15
Lotman V2

### LINKED ISSUE #3538: Lotman V2: native Go lot engine + lot-aware V2 cache
Replace the cgo-based libLotMan integration with a native Go lot-management engine and wire it into the V2 (persistent) cache so eviction, accounting, and quota enforcement are lot-aware.

(This is a placeholder issue corresponding to PR #3517)

bbockelm:
This started as an exploration of how we could integrate Lotman into "Cache V2" but stumbled into a relatively clever idea: flip the Go / C++ relationship upside down in the lotman library.

The core of this is a machine translation of the Lotman C++ code into Golang (and a few cleanups where I couldn't help myself: the SQLite tables aren't ported over identically but the general concepts).  That exposes a standalone shared library that `xrootd-lotman` can link against _and_ a Golang-based module that can be used directly in `pelican-server`.  The latter results in the opportunity to completely drop the `purego` bindings and return `pelican-server` into a standalone binary.

Because it was cheap, this also starts building out a CLI and integration tests for Lotman itself: there's one end-to-end test that tries to put all this together.

Pretty nifty stuff!


.devcontainer/devcontainer-lock.json
```diff
@@ -0,0 +1,9 @@
+{
```
bbockelm: This and some other files in the diff appear to be a mistake from `git add -A`?

.github/workflows/test-macos.yml
```diff
@@ -125,12 +125,23 @@ jobs:
...
+          # ~30% of federation launches with "XrootdOpen: Unable to connect
+          # socket to localhost:<port>; invalid argument" -> "Unable to setup
+          # primary monitor collector" -> "xroot protocol initialization
+          # failed", which is fatal to xrootd startup and takes most fed tests
+          # down with it. It is not FD exhaustion: raising RLIMIT_NOFILE and
+          # kern.maxfiles 13-17x changed nothing, nor did widening the ephemeral
+          # port range (peak TIME_WAIT sampled at 1), nor giving the collector a
+          # unique OS-assigned port. -p=1 and -p=2 are both fully green; -p=4 is
+          # not. The trigger is concurrency in the macOS loopback stack under CPU
+          # contention. Linux is unaffected and keeps its default.
```
bbockelm: This over-explains.  Trim it down: just mention it appears that MacOS runners are generally exhausted of resources when run at `-p=2` and that we already checked for leaks.

PR #3518, opened 2026-06-15
Default cache file-usage purge bands as percentages of disk

### LINKED ISSUE #3519: Finish steps needed for "on-by-default" lotman-enabled cache
Right now, enabling Lotman requires hand configuring the `Cache.FilesXXXSize` parameters.

It should be easy enough to automate this if they can accept percentage values similar to the way `Cache.HighWaterMark` and `Cache.LowWaterMark` work.

Once that's done, Lotman should be happy to start without any custom configuration other than `Cache.EnableLotman`, which may become the default in the future.

jhiemstrawisc:
Enabling Lotman is gated on also hand-sizing pfc.diskusage's file-usage band (Cache.FilesBaseSize/NominalSize/MaxSize), which xrootd requires to sit below the low watermark. No single absolute-size default can satisfy that on caches of differing disk sizes, so let these accept integer percentages (as the watermarks already do) and default them to percentages of total disk. Percentages scale with any cache and stay ordered below a percentage low watermark, which is the missing prerequisite for turning Lotman on without extra configuration.

Scope the defaults to when Lotman is enabled, and only when the operator set none of the three, so a plain cache's purge behaviour is unchanged and the existing all-or-none rule is preserved. Because the file sizes and the low watermark can now both be percentages, validate the max-below-low-watermark ordering up front when the two are expressed the same way and defer the mixed percentage/absolute case to xrootd, which alone knows the real disk total.

Also correct the documented low/high watermark defaults, which had drifted from the shipped 85/89 values.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2067,12 +2067,14 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
 				param.Cache_FilesBaseSize.GetString(), param.Cache_FilesNominalSize.GetString(), param.Cache_FilesMaxSize.GetString())
 		}
 
-		// xrootd additionally requires the max file-usage size to sit below the low
-		// watermark: the file-usage purge band must live inside the headroom the disk
-		// watermark leaves for metadata and any other tenant on the disk (the cache's
-		// data is necessarily a subset of total disk usage). When FilesMaxSize and the
-		// low watermark are expressed the same way we verify it here against a common
-		// basis; the mixed case again needs the disk total, so we defer it to xrootd.
+		// xrootd additionally requires FilesMaxSize to be below LowWaterark.
+		// FilesMaxSize bounds the cache's own data while LowWatermark bounds total
+		// disk usage, and the cache's data is necessarily a subset of total disk
+		// usage; keeping FilesMaxSize below LowWatermark leaves headroom under
+		// LowWatermark for metadata and any other tenant sharing the disk. When
+		// FilesMaxSize and LowWatermark are expressed the same way we verify it here
+		// against a common basis; the mixed case needs the disk total, so we defer
+		// it to xrootd.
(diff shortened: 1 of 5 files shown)


jhiemstrawisc:
Closes #3519

jhiemstrawisc:
@turetske and @h2zh this is one of those loose ends I'm trying to close. I'm pinging you for review in case you have some time and interest to look at this before I merge it later this week.

I've already tested manually that this is sufficient to set `Cache.EnableLotman`, and all the configuration for Lotman is automatically bootstrapped for a successful startup.

config/config.go
```diff
@@ -2021,27 +2038,52 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
...
+				param.Cache_FilesBaseSize.GetName(), param.Cache_FilesNominalSize.GetName(), param.Cache_FilesMaxSize.GetName(),
+				param.Cache_FilesBaseSize.GetString(), param.Cache_FilesNominalSize.GetString(), param.Cache_FilesMaxSize.GetString())
+		}
+
+		// xrootd additionally requires the max file-usage size to sit below the low
+		// watermark: the file-usage purge band must live inside the headroom the disk
+		// watermark leaves for metadata and any other tenant on the disk (the cache's
+		// data is necessarily a subset of total disk usage). When FilesMaxSize and the
+		// low watermark are expressed the same way we verify it here against a common
+		// basis; the mixed case again needs the disk total, so we defer it to xrootd.
```
h2zh: Sorry for being nitpicking here: this paragraph is kinda hard to read. Could you use the standardized word (i.e the param name) to refer all these concepts? For example, "the max file-usage size" should be replaced by 'FilesMaxSize". What do you think?

h2zh (approved):
I read through the code and they all look good. Just a small tiny trivial comment.

config/config.go
jhiemstrawisc: Done :)

PR #3525, opened 2026-06-19
Various and Sundry Cache V2 improvements

### LINKED ISSUE #3537: Cache V2: Various updates prior to release
The V2 (persistent) cache is missing several data-integrity and observability
capabilities that operators and testing need:

1. **No control over the periodic data-integrity scan.** Sites whose storage already guarantees at-rest integrity (e.g. ZFS with scrubbing) don't want the cache re-reading and re-checksumming every object on every cycle, but they do want an initial baseline checksum recorded (and compared against the origin's when available).

2. **No way to deliberately inject corruption** to exercise the cache's integrity-detection paths. This makes it hard to validate that corruption is actually detected/invalidated.

3. **No XRootD monitoring from the V2 cache.** The POSIXv2 origin emits XRootD-style monitoring packets and the XRootD-based cache does so via its XRootD process, but the V2 cache emits none.

Separately, `Cache.EnableSiteLocalMode` does not behave correctly with the V2 cache: a site-local cache fetched directly from origins.

bbockelm:
This branch has a grab-bag of improvements in the (unreleased) Cache V2.  Highlights:

- Site-local mode now works for cache V2.
- Addition of a new "verify once" data-scan mode.  If you have ZFS data scrubbing enabled, there's no point in periodically recalculating the checksum: do it once and it should be fine subsequently.
- Add a "chaos monkey" API that will corrupt cached files in admin-specified ways.
- Add XRootD-compatible monitoring packets so we don't lose monitoring info when we start updating services.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2660,6 +2660,22 @@ type: bool
 default: true
 components: ["cache"]
 ---
+name: Cache.EnableChaosAPI
+description: |+
+  Enable the V2 cache's chaos/fault-injection admin API.
+
+  When true, the cache server registers an admin-authenticated endpoint
+  (POST /api/v1.0/cache/introspect/chaos) that deliberately corrupts or truncates
+  the on-disk data of a cached object. This is used by `pelican cache chaos` to
+  exercise the cache's integrity-detection paths.
+
+  This is a destructive testing feature and is disabled by default; it should
+  only be enabled in test/staging environments.
+type: bool
+default: false
+components: ["cache"]
+hidden: true
+---
 name: Cache.DataScanMode
 description: |+
   Controls how the V2 (persistent) cache's periodic data-integrity scan re-verifies on-disk object data.
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -2689,9 +2689,30 @@ description: |+
       already guarantees at-rest integrity (for example ZFS with scrubbing), where repeatedly re-reading every object is
       wasteful but recording an initial baseline checksum is still desired.
 
+  Important: "once" mode trusts the underlying storage to detect bitrot after the initial check. The cache cannot verify
+  that the storage actually scrubs, so it emits a loud startup warning and exports the `pelican_cache_data_scan_mode_once`
+  metric (set to 1) whenever this mode is active. To retain a floor of at-rest detection even in this mode, the scan still
+  re-verifies a small fraction of already-checked objects each cycle (see ${Cache.DataScanResampleInterval}).
+
+  Note that "once" mode cannot detect corruption that was present at ingest (before the first scan): the first scan records
+  whatever is on disk as the baseline. Detecting bad-at-ingest data requires a full-file or origin checksum on the
+  download/completion path, which is independent of this setting.
+
   Unrecognized values are treated as "all".
 type: string
 default: all
+components: ["cache"]
+---
+name: Cache.DataScanResampleInterval
+description: |+
+  When ${Cache.DataScanMode} is "once", the data-integrity scan still re-verifies roughly 1 in N already-checked objects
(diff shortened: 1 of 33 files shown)


docs/parameters.yaml
```diff
@@ -2660,6 +2660,40 @@ type: bool
...
+  exercise the cache's integrity-detection paths.
+
+  This is a destructive testing feature and is disabled by default; it should
+  only be enabled in test/staging environments.
+type: bool
+default: false
+components: ["cache"]
+hidden: true
+---
+name: Cache.DataScanMode
```
couvares, line 2691: My one real worry in here: `once` mode assumes a fact Pelican can't see (that the underlying storage  scrubs), and nothing warns the admin/operator when it isn't true.  If someone sets `once` on a non-ZFS f/s (or even on ZFS with no scrub scheduled), subsequent bitrot never gets caught.
  The more serious case is the no-origin-checksum path: if `len(meta.Checksums) == 0` we compute the on-disk hash, record it as the baseline, mark it verified, and move on, with nothing to compare against.  That permanently blesses a corrupt-at-ingest file which is *exactly* what bit CIT last week, no?  So `once` is weakest in the case we care most about.

Could we at least (a) log a loud WARN and export a "scan mode = once" metric so a misconfigured cache is visible,  and (b) keep a low-rate random re-sample even in `once` mode, so the floor is *some* at-rest detection rather than zero?  Other ideas?

docs/parameters.yaml
```diff
@@ -2660,6 +2660,40 @@ type: bool
 default: true
 components: ["cache"]
 ---
+name: Cache.EnableChaosAPI
```
couvares, line 2675: Defaults look right.  Off, hidden, admin-auth, staging-only in the docs.

Two small things:  (1) Can enabling this trigger a persistent WARN and a visible "chaos enabled" status somewhere?  It's a deliberate data-corruption feature, and the obvious failure mode is a staging config accidentally making it into production.  (2) Does the tool reproduce the *actual* CIT failure (the ingest race: concurrent pulls, a timeout, partial-then-completed), or  only corrupt already-cached objects at rest?  If just the latter, it's exercising the path that didn't break.

couvares (commented):
This is great, but just to be clear, seems like the "detect and work around" half of we discused yesterday, not the "prevent" half, correct?  The defaults look right — disabled by default, the param's hidden, the endpoint needs admin auth, and the docs say staging-only.

The actual CIT issue was a bad file ingest, and nothing did a full-file or origin check before serving it; the scan only catches that  after a job has already read corruption.  Is the completion-path fix coming separately?  (Don't mark an object  serveable until a full-file checksum clears, against the origin's when we have one.)  That's the piece that actually  keeps corrupt data from getting to jobs.

IMHO the most important bit is the origin checksum.  Verify-once and self-heal only work if the cache can get truth from there.  What's our actual coverage across OSDF origins today?  (I.e., the standing "Pelican object checksum" item.)


Two inline comments are below on the new knobs...

bbockelm:
> Is the completion-path fix coming separately? (Don't mark an object serveable until a full-file checksum clears, against the origin's when we have one.) That's the piece that actually keeps corrupt data from getting to jobs.

That was already done as part of https://github.com/PelicanPlatform/pelican/pull/3477 (provided you're doing a full-object download; a partial object doesn't have the same protection).  When doing a full GET, the cache object will never be served if it doesn't match the origin's checksum.

This PR takes care of (eventually) the partial-object case: once the object copy is complete, it's eligible for a complete scan

docs/parameters.yaml
bbockelm: Done!

docs/parameters.yaml
bbockelm: Done!

PR #3528, opened 2026-06-23
Touch up usage of client-configured caches

### LINKED ISSUE #3526: PreferredCaches configuration leads to duplicate end points
The configuration to prefer a given cache, followed by the director's choices:
```
PreferredCaches:
        - https://osdfcache.ligo.caltech.edu:8443/
        - +
 ```

Can lead to duplicates in the list of endpoints to query:
```
DEBUG[2026-06-23T11:56:58-07:00] Downloading object from osdf:///igwn/ligo/frames/O4/hoft_C00/H1/H-H1_HOFT_C00-144/H-H1_HOFT_C00-1443082240-4096.gwf to /dev/null
DEBUG[2026-06-23T11:56:58-07:00] Will query the following endpoints for availability: osdfcache.ligo.caltech.edu:8443, osdfcache.ligo.caltech.edu:8443, osg-sunnyvale-stashcache.nrp.internet2.edu:8443
```

So, if something goes wrong on `osdfcache.ligo.caltech.edu:8443` we only actually get 2 queries and, presumably, transfer attempts, in that case.

jhiemstrawisc:
This PR does two things:
1. When a user supplies the client with a cache and also requests to fall back to Director cache choices, we now skip over Director-supplied caches if they match one provided by the user.
2. The "try 3 times" logic now only applies to Director-supplied caches; if the user gives a list of 10 servers to try, we'll iterate through them all until we succeed or have exhausted the list. If they provide 10 servers and also choose to fall back to the Director, we'll first try the user-preferred choices and then up to 3 suggestions from the Director.

Note that users can provide the same server arbitrarily many times in their client requests, and the client should _not_ collapse them down to a single request. This comes as a request from @williamnswanson, who sometimes wants to hit the same service multiple times during his testing workflows.

Closes #3526

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -2294,19 +2294,28 @@ func generateTransferDetails(remoteOServer string, opts transferDetailsOptions) 
 // the same cache more than once to force repeated attempts); those duplicates are honored here rather than collapsed.
 // nPreferred indicates how many entries at the start of sortedObjectServers came from the user's PreferredCaches
 // configuration; those entries will have Preferred=true in the returned details.
-func getObjectServersToTry(sortedObjectServers []string, job *TransferJob, oServersToTry int, packOption string, nPreferred int) (transfers []transferAttemptDetails) {
+//
+// directorServersToTry caps only the director-discovered servers. Every user-supplied preferred cache is always tried:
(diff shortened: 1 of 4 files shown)


h2zh (approved):
LGTM

PR #3530, opened 2026-06-24
Fix all bugs

jhiemstrawisc:
After some careful analysis, I determined that the overwhelming majority of defects in Pelican originated from the presence of source code.

By removing the source code, this PR eliminates the root cause rather than merely treating symptoms.

**Warning**: Not backwards compatible... Or forward compatible, for that matter.

(In case it's not abundantly obvious, this is a joke and SHOULD NOT BE MERGED)


bbockelm:
😆 at 358k lines removed, this must be the most efficient PR yet!

PR #3532, opened 2026-06-30
Add Web Client to Origin

### LINKED ISSUE #3533: Add Web Client To Origin
Add the web client to the Origin so that you can view and interact with objects in the backing store via the Origin.

CannonLock:
Most of the actual work was done in the npm packages that are imported. I had to rework some things to allow working without a director as the middleman. Additionally used elements specific to the Origin to make authorization exchanges handled in the background which is okay because this page requires login!

- Automatically get token using current authentication session
- All access is Origin local, federation agnostic

Makes a small change to allow federation in a box to provide the correct openid metadata.


CannonLock:
I mistakenly just recorded the window, but near the end I drag and drop a slew of files into the Origin.

https://github.com/user-attachments/assets/4b106138-2bc2-4c7c-9daa-5f1ea8ba64b3

CannonLock:
Flaky tests are the only ones failing: 

https://pelicanplatform.org/developer-zone/test-failures/Z2l0aHViLmNvbS9wZWxpY2FucGxhdGZvcm0vcGVsaWNhbi9kaXJlY3RvciBUZXN0Rm9yd2FyZERpcmVjdG9y/
https://pelicanplatform.org/developer-zone/test-failures/Z2l0aHViLmNvbS9wZWxpY2FucGxhdGZvcm0vcGVsaWNhbi9lMmVfZmVkX3Rlc3RzIFRlc3RDYWNoZUNvbnRyb2xfRXZpY3Rpb25VbmRlclByZXNzdXJl/
https://pelicanplatform.org/developer-zone/workflows/228169648/

web_ui/frontend/e2e/origin/pages/ObjectBrowserPage.ts
```diff
@@ -0,0 +1,166 @@
...
+  constructor(page: Page) {
+    this.page = page;
+    this.heading = page.getByRole('heading', { name: 'Object Browser' });
+    // MUI's Select renders a combobox; its label isn't wired for getByLabel.
+    this.namespaceSelect = page.getByRole('combobox');
+    this.fileInput = page.locator('input[type="file"]');
+    this.loginButton = page.getByRole('button', { name: 'Login' });
+    this.emptyCollection = page.getByText('You are in an empty collection.');
+    this.collectionHeader = page.getByRole('columnheader', {
+      name: 'Collection',
```
h2zh, line 64: Optional: Could you add a tooltip to explain the difference between namespace and collection? This might need to change the imported object viewer component?

server_utils/oidc.go
```diff
@@ -119,14 +119,25 @@ func createOidcConfigExporter(isDirector bool) func(ctx *gin.Context) {
 		// namespace for backward compatibility; clients should use the
 		// per-namespace discovery endpoints at /api/v1.0/issuer/ns/<prefix>/.well-known/openid-configuration.
 		if param.Origin_EnableIssuer.GetBool() {
+			// The embedded/OA4MP issuer routes are mounted on the bare web
```
h2zh: Just wanna double check, have you tested this PR on an origin with OA4MP issuer?

server_utils/oidc.go
```diff
@@ -119,14 +119,25 @@ func createOidcConfigExporter(isDirector bool) func(ctx *gin.Context) {
...
+			// root (e.g. /api/v1.0/issuer/ns/<prefix>), and the issuer stamps
+			// tokens with iss = <ExternalWebUrl>/api/v1.0/issuer/ns/<prefix>.
+			// issuerStr (from GetServerIssuerURL) may carry the co-located
+			// "/api/v1.0/origin" sub-path used for the origin's base issuer
+			// identity; building the issuer endpoints on top of it would
+			// double the path and advertise an issuer that matches neither
+			// the route mount nor the token's iss claim.  Always base the
+			// issuer endpoints (and jwks) on the bare external web URL.
+			issuerBaseUrl := param.Server_ExternalWebUrl.GetString()
```
h2zh: Changing `issuerStr` to `issuerBaseUrl` can avoid double path in fed-in-a-box, which means the function that computes `issuerStr` - `GetServerIssuerURL` - is not helpful here. Could you add a few unit tests for this change? Maybe try to set `Server.IssuerUrl`, `Server.IssuerHostname` to see what's happening.

h2zh (changes requested):

h2zh (changes requested):
Another note: origin's web client requires the origin admin sets up `Issuer_RedirectUris` to `https://<host>/view/origin/client/` so `launchers/origin_serve.go:357-364` can seeds the public client correctly. Otherwise, login silently fails. I think there should be a note somewhere to remind the origin admin, or we could derive `Issuer_RedirectUris` automatically from `Server_ExternalWebUrl`?

web_ui/frontend/e2e/origin/pages/ObjectBrowserPage.ts
CannonLock: Yeah, let me make that a web-client issue.

server_utils/oidc.go
CannonLock: Tests added.

server_utils/oidc.go
CannonLock: It won't but I just added a UI element to indicate why. I don't see a reason to spend the time making it work when we are removing in a month.

launchers/origin_serve.go
```diff
@@ -363,6 +384,19 @@ func configureEmbeddedIssuer(ctx context.Context, egrp *errgroup.Group, engine *
...
 
+	// Redirect URIs allowed for the pre-allocated public client used by
+	// browser-based (PKCE) logins. Start from the admin-configured list
+	// and, when the origin's web UI is enabled, auto-add its
+	// object-browser page so a missing Issuer.RedirectUris entry doesn't
+	// silently break login. Computed once and shared by every per-export
+	// provider below.
+	publicClientID := param.Issuer_PublicClientID.GetString()
+	redirectURIs := param.Issuer_RedirectUris.GetStringSlice()
+	if uiRedirect := webUIClientRedirectURI(); uiRedirect != "" && !slices.Contains(redirectURIs, uiRedirect) {
```
h2zh, line 452: On an existing origin (public client already seeded in a prior run, persisted in pelican.sqlite), `EnsurePublicClient` returns early and never reconciles redirect URIs. So this auto-derived /view/origin/client/ is never added, and browser login keeps failing silently. It only helps origins with a fresh client/DB.

```
v7.26 origin (embedded issuer)  →  pelican-public-client row already exists, redirect_uris=[]
        ↓ upgrade to a version featuring this branch
EnsurePublicClient() → GetClient() succeeds → returns early → auto-derived /view/origin/client/ never written
        ↓
Object Browser PKCE login → fosite rejects redirect_uri → silent login failure
```

I think we can merge/reconcile the auto-derived URI into the pre-seeded public client on startup (union with existing, so it doesn't tamper the URIs added by admin at runtime via `PUT /admin/clients/:id`), or at minimum document that upgrading admins must clear the client `/` use `PUT /admin/clients/:id`.

h2zh (changes requested):
Thanks for addressing all these comments above! One last crucial thing I caught:

launchers/origin_serve.go
CannonLock: I sorted this out, added a test, then tested locally by spinning up a origin with a preexisting client.

h2zh (approved):
LGTM. Thanks for the iterations!

PR #3535, opened 2026-07-01
Classify result-level errors in transfer ads (fix missing ErrorType)

### LINKED ISSUE #3534: Properly Process errors that occur after a transfer
When an error occurs after a transfer that has otherwise succeeded (eg, a checksum error which occurs after all the bytes have successfully transferred), Pelican doesn't currently process the errors to populate the DebugError codes, etc.

This should be fixed by appending the error into the transfer error list and making sure it's created properly when it occurs.

turetske:
A checksum mismatch fails a transfer *after* all download attempts have succeeded: the error is set on the top-level result.Error, and no individual attempt.Error is populated. addDataToClassAd only built the structured TransferErrorData (which carries ErrorType/PelicanErrorCode) from per-attempt errors or zero-attempt early failures, so such a result was reported as TransferSuccess=false with a TransferError message but no error code or ErrorType.

Add a branch that classifies result.Error when no per-attempt error already captured the failure, guarding against double-counting so a normal per-attempt failure still yields a single TransferErrorData entry.

closes #3534

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -1026,6 +1026,25 @@ func addDataToClassAd(resultAd *classad.ClassAd, result *client.TransferResults,
 		transferErrorData = append(transferErrorData, transferError)
 	}
 
+	// Handle result-level (post-transfer) errors that aren't attached to any individual
+	// attempt, e.g. a checksum mismatch: the download attempts all succeeded, so the
+	// error lives only on result.Error and neither the per-attempt loop nor the early-
+	// failure block above captured it. Without this, such a result would be reported as
+	// TransferSuccess=false with no ErrorType/PelicanErrorCode. Only add it when nothing
+	// was already recorded so a normal per-attempt failure still yields a single entry.
+	if result != nil && result.Error != nil && len(transferErrorData) == 0 {
+		adErr = developerData.Set("TransferError1", result.Error.Error())
+		if adErr != nil {
+			log.Errorf("Failed to set TransferError1: %s", adErr)
+		}
+		adErr = developerData.Set("IsRetryable1", client.IsRetryable(result.Error))
+		if adErr != nil {
+			log.Errorf("Failed to set IsRetryable1: %s", adErr)
+		}
+		transferError := createTransferError(result.Error)
+		transferErrorData = append(transferErrorData, transferError)
+	}
+
 	// Add checksum information if provided
 	if len(clientChecksums) > 0 {
 		clientChecksumsData := classad.New()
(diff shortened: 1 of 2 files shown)


cmd/plugin.go
```diff
@@ -1026,6 +1026,25 @@ func addDataToClassAd(resultAd *classad.ClassAd, result *client.TransferResults,
 		transferErrorData = append(transferErrorData, transferError)
 	}
 
+	// Handle result-level (post-transfer) errors that aren't attached to any individual
+	// attempt, e.g. a checksum mismatch: the download attempts all succeeded, so the
+	// error lives only on result.Error and neither the per-attempt loop nor the early-
+	// failure block above captured it. Without this, such a result would be reported as
+	// TransferSuccess=false with no ErrorType/PelicanErrorCode. Only add it when nothing
+	// was already recorded so a normal per-attempt failure still yields a single entry.
+	if result != nil && result.Error != nil && len(transferErrorData) == 0 {
```
h2zh: The `len(transferErrorData) == 0` guard assumes "nothing recorded yet" means "the error is result-level". But in a common OSDF pattern: first cache down, failover succeeds, then the checksum fails. Gate 1 (see it in the diagram) records the stale connection error, and Gate 3 stays quiet about the error that actually killed the transfer:

image

h2zh (changes requested):
I didn't work on HTCondor's Pelican plugin before, so I dug into it with Claude's help to understand the entire workflow. Here's what I found with it. Also, those failed CIs may be solved by a rebase.

cmd/plugin_test.go
```diff
@@ -1264,6 +1264,102 @@ func TestCreateTransferError(t *testing.T) {
...
+		errorDataList, ok := classad.GetAs[[]*classad.ClassAd](resultAd, "TransferErrorData")
+		require.True(t, ok)
+		require.Len(t, errorDataList, 1)
+	})
+
+	// Failover then post-transfer failure: an early attempt fails (e.g. a stale cache
+	// connection), a later attempt succeeds, and only then does the checksum fail. The
+	// download-level error must not stand in for the fatal result-level error; both are
+	// recorded, in the order they occurred (attempt errors first, result-level error last).
+	t.Run("FailoverThenChecksumMismatch", func(t *testing.T) {
```
h2zh, line 1326: It would be nice to have another test for ≥2 failed attempts (e.g. cacheA fails, cacheB fails, originC succeeds, checksum mismatches)

h2zh (approved):
It would be nice to squeeze in another unit tests before merging

PR #3542, opened 2026-07-04
Add V2 origin POSC + object-metadata tracking and publisher

### LINKED ISSUE #3543: Metadata and close behavior
We want to implement functionality around metadata and data handling, especially around closes:

1.  Port POSC functionality to POSIXv2.
2. Allow arbitrary metadata to be uploaded along with user data.  This could be sent out to a separate metadata catalog.
3. Also track object-level metadata (think: etags and whatnot).

bbockelm:
## Summary

Three related capabilities on the V2 (Go-native, webdav-backed) origin, all opt-in per-export:

**1. POSC (persist-on-successful-close).** Uploads stage to a hidden per-user temp file and only rename into place when `Close()` succeeds. A mid-upload disconnect no longer leaves a truncated object visible at its final name. Background expiry reaps stale temp files; a keepalive thread refreshes mtime on active ones. Mirrors the C++ POSC reference in xrootd-s3-http.

**2. Object-commit webhook publisher.** After every successful commit the origin POSTs a GitHub-style JSON event to a configured endpoint, authenticated with a JWT bearing a new `pelican.metadata:/prefix` scope. Two modes:
- *Transactional* — `Close()` blocks on the first publish attempt; failure rolls back the object.
- *Eventually-consistent* — durable enqueue via a shared write-behind SQLite batcher, background worker with backoff + jitter, deletion-aware retry skip.

Clients attach custom metadata via a new `X-Pelican-Object-Metadata` header parsed as RFC 9651 Structured Fields; the origin inlines those fields into the webhook JSON. An optional opaque body may ride alongside via `multipart/form-data` and is forwarded to the receiver as `multipart/related`.

**3. Local object-metadata tracking.** Per-namespace `Origin.Metadata.TrackAccess` records every object lifecycle event (commit / delete / rename / external observe / external modify / external delete / access) in a three-table SQLite layout (live + checksums + append-only history). `Stat` interposes at the `aferoFileSystem` layer with a `ttlcache`-fronted etag comparison to detect out-of-band changes; PROPFIND with `Depth != 0` short-circuits so 100k-file listings stay as cheap as today. Origin-supplied `EtagPolicy` (`uuid` or `sha256`) fills in when the backend declines. Background pruner trims aged history and aged soft-deleted live rows per configurable retention.

**Plumbing:**
- Write-behind SQLite batcher coalesces durable + best-effort writes (commit, observation, atime debounce, publish-queue insert) into single transactions. Concurrent-durable batching, per-flush timeout, `FlushNow` helper, blocking back-pressure with a wait-time histogram.
- In-memory atime debouncer with periodic flush.
- `sync.Pool` of sha256 hashers when `EtagPolicy=sha256`.
- `_txlock=immediate` on the shared SQLite DSN so no read-then-write tx can hit `SQLITE_BUSY_SNAPSHOT` under batcher concurrency.

**Operator surface:**
- Admin HTTP under `/api/v1.0/origin_ui/{metadata_queue,object_metadata}/*` (list / lookup / history / health / retry / delete).
- `pelican origin object-metadata {list,get,history}` CLI verbs.
- `pelican object put --metadata-file <json>` and `--metadata-body <file>` client flags + `WithObjectMetadata{,File}` transfer options on the SDK.
- ~20 new Prometheus instruments under `pelican_origin_{metadata,posc,object_metadata}_*`: publish attempts/latency/queue depth/health, POSC active uploads, batcher flush/size/age/enqueue-wait, observation cache hit rate, external-event counts, pruner deletions, atime debouncer depth.

**Schema (new Goose migrations):**
- `metadata_publish_queue` + optional `metadata_body` blob column
- `object_metadata` + `object_checksums` + `object_metadata_history`, with a partial-unique index on live rows so soft-delete + re-upload works cleanly.

**Deferred (documented in the design doc):**
- Web UI consumer for the admin endpoints
- Startup-time reconciliation of filesystem ↔ queue
- No `fsync` between rename and 2xx reply (matches C++ POSC and POSIX defaults)

Design doc, progress notes, and pre-PR gap list are under `docs/`.

## Test plan

- [ ] `go test ./origin_serve/... ./client/... ./cmd/...` passes
- [ ] End-to-end federation test with `pelican object put --metadata-file` lands the expected JSON at a receiver
- [ ] Transactional mode: a failing endpoint rolls back the object (no visible file, no queue row)
- [ ] Eventually-consistent mode: origin restart with rows in the queue resumes publishing after backoff
- [ ] `PROPFIND` on a 100k-file directory is not measurably slower with tracking on
- [ ] Pruner respects `Origin.Metadata.History.RetentionDays = 0` (never prune) and per-export overrides
- [ ] Admin retry / delete endpoints require the same auth as `pelican downtime`


bbockelm:
@brianaydemir - worked with Claude to expand test coverage. I _think_ I might have captured your issue.  Can you check?

origin_serve/tpc_handler.go
brianaydemir: **What:** In `handleCopyTPC`, the destination file is closed unconditionally (`origin_serve/tpc_handler.go:364`) *before* the copy-error and size-mismatch checks (`:368`, `:380`). Because the destination FS is the POSC / close-notify wrapper carrying the metadata close hook, `destFile.Close()` commits (rename into place) and fires the commit hook — publishing the `object.committed` webhook and recording a `commit` (with the captured source ETag) in the tracking DB — even when the transfer stalled, was too slow, was cancelled, or returned fewer bytes than promised. Only afterward does the handler `fs.RemoveAll(destPath)` to delete the failed object.

**Why it matters:** In transactional publish mode the webhook is delivered synchronously during Close(), so an external metadata service is told a partial/never-completed object was committed and then the object is deleted out from under it. In all modes the tracking history gains a bogus `commit` event (and source_etag) for a copy that failed, followed by a delete — state that misrepresents what actually happened.

**Fix:** Do not commit/fire the close hook on a failed TPC. Detect copy failure before Close (e.g. abort the staged file so POSC discards it without renaming or hooking), or gate the hook on `copyErr == nil` and the size check passing.

origin_serve/metadata_controller.go
```diff
@@ -0,0 +1,835 @@
...
+	}
+	rel := strings.TrimLeft(exportRelative, "/")
+	// If the close hook somehow received an already-rooted path,
+	// don't double-prefix.
+	if rel == "" {
+		return ns
+	}
+	if strings.HasPrefix("/"+rel, ns+"/") || "/"+rel == ns {
+		return path.Clean("/" + rel)
+	}
```
brianaydemir: **What:** `joinFederationPath` (`origin_serve/metadata_controller.go:704`) has a "don't double-prefix" guard that returns the input unchanged when the export-relative path already starts with `namespace + "/"`. But the close hook always receives an *export-relative* path (the webdav handler strips its Prefix), so a legitimate object whose first path component happens to match the namespace's trailing segment is misclassified as "already federation-rooted." E.g. namespace `/exp` + export-relative `/exp/data/x.bin` (a client-created subdir named `exp`) yields `/exp/data/x.bin` instead of the correct `/exp/exp/data/x.bin`. The unit test `TestJoinFederationPath` (`metadata_test.go:812`) locks in this incorrect behavior.

**Why it matters:** The published event's `object.path` and the tracking DB's `object_path` are recorded with the namespace prefix dropped, so a receiver that fetches the advertised path gets the wrong object or a 404, and the eventual-mode existence check Stats the wrong path. Any export that organizes objects under a top-level directory named like the namespace is affected.

**Fix:** Drop the ambiguous already-rooted heuristic; always join `namespace + exportRelative` since the close hook contract guarantees an export-relative input.

origin_serve/metadata_event.go
```diff
@@ -0,0 +1,192 @@
...
+	obj := make(map[string]any, len(e.CustomFields)+4)
+	obj["path"] = e.ObjectPath
+	obj["size"] = e.ObjectSize
+	obj["etag"] = e.ETag
+	obj["created_at"] = e.ObjectCreated.UTC().Format(time.RFC3339Nano)
+	for k, v := range e.CustomFields {
+		// Inline custom fields. Reserved keys cannot collide because
+		// they are stripped at parse time.
+		obj[k] = v
+	}
```
brianaydemir: **What:** `ObjectCommitEvent.MarshalJSON` first writes the trusted, origin-computed object fields (`path`, `size`, `etag`, `created_at`) and *then* overlays client-supplied custom fields on top of the same map (`origin_serve/metadata_event.go:154-163`). Any custom field whose key equals a reserved name silently overwrites the authoritative value in the webhook body. The sole safeguard is the parse-time reserved-key strip in `ParseObjectMetadataHeader` (`origin_serve/object_metadata_header.go:71-84`, list at `:47`).

**Why it matters:** The integrity of the webhook's authoritative identity fields (which receivers use for dedup / provenance) rests on a single, easily-broken layer. If a new reserved field is added to the event but not to `ReservedCustomFieldKeys`, or a future code path populates `CustomFields` from a source other than the header parser, a client could spoof `object.path`/`etag`/`size` seen by the metadata consumer.

**Fix:** Make the marshal step authoritative — inline custom fields first and write the reserved keys last (or skip any custom key that collides), so origin-computed values always win regardless of upstream stripping.

origin_serve/object_metadata.go
```diff
@@ -0,0 +1,630 @@
...
+				eventID, in.Namespace, in.ObjectPath, string(ObjectEventExternalModify), now,
+				in.Size, in.ETag, string(in.EtagSource), in.BackendMtime, in.Actor, extraJSON,
+			},
+		},
+		{
+			SQL: `UPDATE object_metadata
+				SET size = ?, etag = ?, etag_source = ?, backend_mtime = ?, last_modified = ?
+				WHERE namespace = ? AND object_path = ? AND deleted_at IS NULL`,
+			Args: []any{in.Size, in.ETag, string(in.EtagSource), in.BackendMtime, now, in.Namespace, in.ObjectPath},
+		},
```
brianaydemir, line 439: **What:** `RecordExternalChange` snapshots the new state and updates the live row, but its `UPDATE` sets only `size, etag, etag_source, backend_mtime, last_modified` (`object_metadata.go:376-380`) — it does **not** clear `source_etag`. This contradicts the documented contract on the column, which states it is "Cleared by any subsequent write that invalidates the object (RecordCommit, RecordExternalChange)" (`object_metadata.go:103-104`). `RecordCommit` does clear it; `RecordExternalChange` does not.

**Why it matters:** After an object produced by a TPC (which stores the upstream `source_etag`) is modified out-of-band, a Stat observes a different ETag and fires `RecordExternalChange`; the object's content/ETag change but `source_etag` stays at the now-obsolete upstream value. That stale value is served verbatim on PROPFIND as `{pelican}source-etag` (`propfind_pelican_props.go:73-84`), whose stated purpose is letting a sync client skip re-fetch on an ETag match — so a client can wrongly skip syncing an object that has actually changed.

**Fix:** Add `source_etag = NULL` to the `RecordExternalChange` UPDATE set-list so an out-of-band modification clears the no-longer-applicable source ETag, matching the documented contract.

origin_serve/metadata_event.go
bbockelm: Fixed!

origin_serve/tpc_handler.go
bbockelm: Done!

origin_serve/metadata_controller.go
bbockelm: Done!

origin_serve/object_metadata.go
bbockelm: Done!

bbockelm:
@brianaydemir - thanks for all the feedback!  Being that the Mac OS X builder is not working right now ... let's go for it!

bbockelm:
@brianaydemir - actually, the Mac build came back around 20 seconds after I suggested we just skip it.

Great timing too, as it revealed some failures.

For future knowledge: the Mac OS X GitHub runners have a ulimit of 256 file descriptors.  We're at the point if there's a single FD pair leaked per federation test run (2 FDs), around 50 unit test failures will be triggered.

It's esoteric and weird error messages ("Invalid Argument" creating a UDP socket in xrootd or somesuch) but it's certainly becoming our FD leak canary.

bbockelm:
@brianaydemir - I know you mentioned you had some error handling cases written up (but not posted) but mentioned it was in reasonable shape otherwise.  I'd like to get this in so we can help our NDP friends ... can you please post the review as a separate issue?

PR #3547, opened 2026-07-07
Add action that provides easy coverage metrics

### LINKED ISSUE #3546: Add first class test coverage action
Needed for easier reporting for metrics.

CannonLock:
Intended to make all metrics the need reporting easily accessible with http.

### DIFF
--- a/.github/workflows/coverage-scheduled.yml
+++ b/.github/workflows/coverage-scheduled.yml
@@ -1,0 +1,136 @@
+---
+name: Code Coverage [on schedule]
+
+# Runs the Go test suite once a day purely to measure code coverage, and
+# publishes the result as a JSON artifact ("code-coverage-json"). This is a
+# standalone workflow (rather than a hook into test-linux.yml) so that it runs
+# only "go test" with coverage instrumentation -- no GoReleaser build or
+# end-to-end scripts, none of which contribute to the coverage profile.
+#
+# Like the scheduled test workflow, it runs in the most recent published test
+# container image, since there is no per-run image build to depend on.
+
+on:
+  schedule:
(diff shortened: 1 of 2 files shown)


CannonLock:
Proof this works!

https://github.com/CannonLock/pelican/actions/runs/28877290503

.github/workflows/coverage-scheduled.yml
```diff
@@ -0,0 +1,128 @@
...
+          make web-build
+          echo "::endgroup::"
+          go test \
+            -p=4 \
+            -timeout=15m \
+            -coverpkg=./... \
+            -covermode=count \
+            -coverprofile=${{ matrix.coverprofile }} \
+            -tags=${{ matrix.tags }} \
+            ./...
```
h2zh, line 76: `go test` returns non-zero if any test fails, which makes GitHub Actions skip the two steps below it by default. Since this workflow's whole purpose is a reliable daily coverage number, one flaky test wipes out that day's data entirely.

h2zh (approved):
LGTM after fixing one bug. 

Btw, you mention this PR makes the coverage metrics "accessible with http", so I assume your next step would be pull these two metrics in https://pelicanplatform.org/developer-zone/ ?

.github/workflows/coverage-scheduled.yml
CannonLock: Thank you!

CannonLock: @h2zh Yes, exactly. For all the metrics we had scripts to generate I am going to make them so I can just remake them every morning in the developer zone website.

PR #3553, opened 2026-07-08
Issuer cors addition

### LINKED ISSUE #3554: Add CORS Headers for Embedded Issuer
Add CORS headers to the embedded issuer in the same capacity that they are in the OA4MP issuer.

CannonLock:
Add CORS headers to the embedded issuer in the same capacity that they are in the OA4MP issuer.

General practice is that I should be able to hit the endpoints and metadata that are attached from anywhere in the web so that I can complete the OAuth flows prescribed within.


h2zh (approved):
LGTM. No issues spotted and local testing passes.

PR #3557, opened 2026-07-10
Mitigate frontend build OOM (exit 137) in constrained containers

### LINKED ISSUE #3556: Frequent out-of-memory kill during the goreleaser run in dev container
This is a lingering problem for Pelican developers. Frequent OOM-killed (exit 137) during the goreleaser run on memory-constrained local dev containers slows down the development pace.

h2zh:
The Next.js webUI frontend build gets OOM-killed (exit 137) on memory-constrained local dev containers. To mitigate this, this PR lowers memory demand by:

- Pin Next.js tracing root (next.config.js)
Set outputFileTracingRoot to the frontend app directory. Next.js otherwise picks a tracing root by walking up for lockfiles; with stray package-lock.json stubs and a monorepo-like layout it can select a root high in the tree and then trace/scan an enormous file set, ballooning memory. Pinning the root stops the walk.

I also tried a few other approaches (e.g. Cap V8's old-space heap by adding `WEB_BUILD_NODE_OPTIONS ?= --max-old-space-size=2048` in `Makefile`). This one plays the most crucial role. It's reliable for local build, Github CI and the release build run.

Closes #3556

### DIFF
--- a/web_ui/frontend/next.config.js
+++ b/web_ui/frontend/next.config.js
@@ -1,9 +1,14 @@
 /** @type {import('next').NextConfig} */
+const path = require('path');
+
 let nextConfig = {
   output: 'export',
   basePath: '/view',
   trailingSlash: true,
   images: { unoptimized: true },
+  // Pin tracing root to this app so Next.js does not walk the monorepo
+  // (wrong root causes huge memory use and OOM / exit 137 in dev containers).
+  outputFileTracingRoot: path.join(__dirname),
 };
 
 const withBundleAnalyzer = require('@next/bundle-analyzer')({


CannonLock:
@h2zh LGTM. The fact we hit the prerequisite for this being a issue is surprising to me but I don't think this will hurt anything.

> with stray package-lock.json stubs and a monorepo-like layout it can select a root high in the tree and then trace/scan an enormous file set, ballooning memory. Pinning the root stops the walk.

CannonLock (approved):

h2zh:
@CannonLock  The culprit is an environmental artifact file, not something in the repo. The offending lockfile is at `/workspaces/dev-container-root/package-lock.json`, above the pelican repo `/workspaces/dev-container-root/pelican`, auto-created by a stray npm invocation. I confirmed the mechanism against Next 15.5.19's `findRootDir()`: it selects the outermost lockfile up the tree, not the nearest, so any stub above the app relocates the tracing root. In my dev container, it's a 4.9 GB directory that Next then walks, which is the OOM. Similar OOM also happens on Emma and Justin's dev containers when they enable `make web-build`.

On a clean checkout there's no lockfile above `web_ui/frontend`, so Next infers that dir as the root on its own and the pin in this PR is a no-op. But in reality the unsolicited lockfile exists in almost everyone's container, which makes this PR necessary.

PR #3559, opened 2026-07-13
Update to node 26

### LINKED ISSUE #3560: Upgrade Node to New Version
Upgrade Node to latest version.

CannonLock:
Should be a easy review. I updated the build to use node 26. Very minimal changes in the package-lock.json.

Did have to add a new package to the build images which is a dependency of node >= 24. So that the tests pass here I rigged in some code to download that new dep before each of the actions. That can be taken out in the next release, I can make a ticket for that if this is approved.

Sorts out this issue in a roundabout way. 

https://github.com/PelicanPlatform/pelican/actions/runs/29164730320/job/86575788515?pr=2970

Instead of capping the npm version being downloaded lets just update node to match the latest version of npm.

### DIFF
--- a/.github/workflows/coverage-scheduled.yml
+++ b/.github/workflows/coverage-scheduled.yml
@@ -35,6 +35,13 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
+      # actions/setup-node downloads its own Node.js binary, which (as of
+      # Node 24) dynamically links libatomic.so.1. The AlmaLinux-based container
+      # image may not ship it yet, so install it before setup-node runs Node.
+      # TODO: remove once the pelican-test image includes libatomic.
+      - name: Install libatomic (Node.js >= 24 runtime dependency)
+        run: dnf install -y libatomic
+
       - name: Set up Node.js
         uses: actions/setup-node@v6
         with:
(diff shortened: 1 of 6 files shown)


CannonLock:
Broken test is flaky: https://pelicanplatform.org/developer-zone/test-failures/Z2l0aHViLmNvbS9wZWxpY2FucGxhdGZvcm0vcGVsaWNhbi9kaXJlY3RvciBUZXN0RXhwaXJhdGlvbkRpcmVjdG9y/

.github/workflows/coverage-scheduled.yml
```diff
@@ -35,6 +35,13 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
+      # actions/setup-node downloads its own Node.js binary, which (as of
+      # Node 24) dynamically links libatomic.so.1. The AlmaLinux-based container
+      # image may not ship it yet, so install it before setup-node runs Node.
+      # TODO: remove once the pelican-test image includes libatomic.
```
h2zh, line 41: Maybe this should get @matyasselmeci 's attention?

CannonLock: @h2zh I would think not as we only use node for the build step which is completed before handing off for RPMs etc.

h2zh (approved):
LGTM

PR #3562, opened 2026-07-15
Default off origin tlsclientauth

### LINKED ISSUE #3565: Web Client requests ClientCertificate when Issuing request to Xrootd Origins
Turns off http.tlsclientauth in Xrootd Origin's by default.

Copy what we had from the Cache to the Origin to remove the annoying pop-ups when PROPFIND hits the Origin in the web client.

I tested locally by turning it off and on in the configuration and checking if Xrootd would request a certificate from the client.

CannonLock:
Turns off `http.tlsclientauth` in Xrootd Origin's by default.

Copy what we had from the Cache to the Origin to remove the annoying pop-ups when PROPFIND hits the Origin in the web client.

I tested locally by turning it off and on in the configuration and checking if Xrootd would request a certificate from the client.

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -1542,8 +1542,25 @@ description: |+
   present X.509 client credentials in order to authenticate.  The configuration
   of the authorization for these clients must be done by the admin; Pelican
   does not support automatic VOMS authorization configuration.
+
+  This relies on the client presenting a certificate during the TLS handshake,
+  so Origin.EnableTLSClientAuth must also be set to true for VOMS authentication
+  to function.
 type: bool
-default: true
+default: false
+components: ["origin"]
+---
+name: Origin.EnableTLSClientAuth
+description: |+
+  Turns client certificate authentication on or off in xrootd for the HTTPS protocol. When false (default) the
+  origin will never request a TLS client certificate. When true, the origin will always request
+  a client certificate from the client.
+
+  Note that X.509/VOMS-based authentication (see Origin.EnableVoms) relies on the client presenting
+  a certificate during the TLS handshake; if the origin should authenticate clients via VOMS,
+  this parameter must be set to true.
+type: bool
+default: false
 components: ["origin"]
 ---
 name: Origin.EnableDirListing
@@ -2567,6 +2584,10 @@ description: |+
   to present X.509 client credentials in order to authenticate.  The configuration
   of the authorization for these clients must be done by the admin; Pelican
   does not support automatic VOMS authorization configuration.
+
+  This relies on the client presenting a certificate during the TLS handshake,
+  so Cache.EnableTLSClientAuth must also be set to true for VOMS authentication
+  to function.
 type: bool
 default: false
 components: ["cache"]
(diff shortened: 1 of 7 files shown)


CannonLock:
I don't believe the existing test failures are my own doing.

h2zh:
@CannonLock Could you rebase the latest main? Those failed CIs should be fixed now

docs/parameters.yaml
```diff
@@ -1546,6 +1546,19 @@ type: bool
 default: true
 components: ["origin"]
 ---
+name: Origin.EnableTLSClientAuth
```
turetske, line 1553: Can you also switch Origin.EnableVoms to off? One should not be default to true while the other defaults to false.

And maybe add a check in the code for contradictory configs: a startup validation/warning when EnableVoms && !EnableTLSClientAuth, since that pair produces a non-functional setup.

turetske (changes requested):

docs/parameters.yaml
CannonLock: Done with both. I decided to hard error as this misconfiguration would likely cause a Origin to not perform in the expected way if done.

xrootd/xrootd_config.go
```diff
@@ -1312,6 +1313,16 @@ func ConfigXrootd(ctx context.Context, isOrigin bool) (string, error) {
 				}
 			}
 		}
+
+		// VOMS authentication relies on the client presenting an X.509 certificate
+		// during the TLS handshake, which only happens when TLS client auth is
+		// requested. Enabling VOMS without TLS client auth yields a non-functional
+		// setup: the origin never asks for a client certificate, so VOMS has nothing
+		// to extract.
+		if xrdConfig.Origin.EnableVoms && !xrdConfig.Origin.EnableTLSClientAuth {
```
turetske, line 1332: I realize I suggested we do this (and I agree with it), but it's different than what the cache requires. Can you also add this check for the cache so we have parity?

turetske (commented):
Can you add the following two tests?
1. EnableVoms=true, EnableTLSClientAuth=false → ConfigXrootd returns an error.
2. Rendered config contains http.tlsclientauth off by default and on when enabled.

xrootd/xrootd_config.go
CannonLock: > but it's different than what the cache requires

What does the cache require? @turetske

CannonLock: On the off chance they require the same thing I added that same check.

turetske: Yup, sorry. That's what I meant.

docs/parameters.yaml
```diff
@@ -1542,8 +1542,25 @@ description: |+
...
+  to function.
 type: bool
-default: true
+default: false
+components: ["origin"]
+---
+name: Origin.EnableTLSClientAuth
+description: |+
+  Turns client certificate authentication on or off in xrootd for the HTTPS protocol. When false (default) the
+  origin will never request a TLS client certificate. When true, the origin will always request
```
turetske:
```suggestion
  origin will never request a TLS client certificate. When true, the origin will sometimes request
```

turetske (approved):

PR #3569, opened 2026-07-17
Fixup - Serve Origin issuer from root when no Director present

### LINKED ISSUE #3570: Origin not advertising openid-configuration from root
Currently an Origin will advertise its namespace as being at the root but no longer provides it at that endpoint.

Introduced here, the registration of the endpoint was deleted.

https://github.com/PelicanPlatform/pelican/commit/63fcdc8e63ebbf7d29ebf7f264ecfbd4659c4e7d

CannonLock:
Currently an Origin will advertise its namespace as being at the root but does not have it there.

Introduced here:
https://github.com/PelicanPlatform/pelican/commit/63fcdc8e63ebbf7d29ebf7f264ecfbd4659c4e7d


server_utils/oidc_test.go
```diff
@@ -199,3 +199,95 @@ Server:
...
+		require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+		assert.Equal(t, issuer, resp.Issuer)
+		assert.Equal(t, issuer+"/.well-known/issuer.jwks", resp.JwksUri)
+
+		// ...but without the fix, the advertised root path is NOT served.
+		w = get(t, router, "/.well-known/openid-configuration")
+		assert.Equal(t, http.StatusNotFound, w.Code,
+			"pre-fix bug: advertised root discovery path 404s when only mounted under /api/v1.0/origin")
+	})
+}
```
h2zh: The docstring claims this "guards the OIDC mount decision in OriginServe," but the test can't reach `OriginServe` (import cycle from server_utils → launchers), so it hardcodes the branch's outcome: `RegisterOIDCAPI(router.Group("/"), false`, instead of exercising the if `modules.IsEnabled(DirectorType)` decision. It therefore tests RegisterOIDCAPI (never broken), not the branch selection (what broke in 63fcdc8e and what could regress). Reverting the fix leaves this test green. For now, I'll just remove this test and merge it into main to catch 7.27.

```suggestion
```

h2zh (approved):
Thanks for fixing this! The fix looks good and there's no duplicate-route. This has to be merged in before 7.27 otherwise the basic auth will fail for an origin.

PR #3575, opened 2026-07-21
Per-namespace IssuerJwks for the embedded OAuth2 issuer

### LINKED ISSUE #3574: Enhancement: Per-namespace `IssuerJwks` for the embedded OAuth2 issuer
## Current state of affairs

When `Origin.EnableIssuer` is `true` and `Origin.IssuerMode` is `embedded` (or unset), Pelican runs a built-in fosite-based OAuth2 issuer. For each export that requires authentication (i.e., not fully public-read with no writes), an embedded issuer is created and given a per-namespace issuer URL of the form `<Server.ExternalWebUrl>/api/v1.0/issuer/ns/<federationPrefix>`. An OAuth2 discovery document is served at `<issuer-url>/.well-known/openid-configuration`.

However, the `jwks_uri` in every per-namespace discovery document points to the **same** server-level JWKS endpoint (`/.well-known/issuer.jwks`). That endpoint is backed by `config.GetIssuerPublicJWKS()`, which combines keys from `IssuerKeysDirectory` / `IssuerKey` with any additional pre-generated public keys listed in `Server.IssuerJwks`.

The `Server.IssuerJwks` parameter applies globally — there is no way to augment the public-key set for just one namespace without affecting every other namespace on the same origin.

## Desired enhancement

Each per-namespace issuer should host its own JWKS endpoint (e.g., `/api/v1.0/issuer/ns/<federationPrefix>/.well-known/issuer.jwks`), and the `jwks_uri` in each per-namespace discovery document should point to that namespace-scoped endpoint rather than the shared server-level one.

Additionally, a new `IssuerJwks` field should be added to each entry in `Origin.Exports`, mirroring the semantics of `Server.IssuerJwks`. When set, the additional public keys from the specified JWKS file would be merged into that namespace's JWKS endpoint alongside the server's dynamic keys. Namespaces without `IssuerJwks` configured would still serve a per-namespace JWKS endpoint, just without any supplemental keys.

This makes it possible to register different external or pre-provisioned public keys for different namespaces on the same origin — a useful capability when, for example, independent teams operate separate namespaces on a shared origin and each team manages its own signing keys. Note that this is purely about which public keys are *published* per namespace; it is not a request for per-namespace private signing keys for Pelican-issued tokens.

brianaydemir:
Fixes #3574.

## Summary

This PR adds a new configuration parameter, `Origin.Exports[].IssuerJwks`, which allows a public keys to be added to specific namespaces. These keys are merged with the server's public keys.

The linked issue contains additional background and motivation.

## Notable implementation decisions

Public keys are now validated before being published: No duplicate `kid`s. No symmetric keys. This affects _all_ services. (There should be no symmetric keys in the wild.)

When disallowed keys are present, degradation is "graceful": the service makes a best-effort at continuing to serve _something_. Errors are only logged. The assumption is that in practice, this feature serves niche use cases, and the operator/admin can reasonably be expected to check that their service correctly starts up and serves the expected public keys (and that they won't later mangle keys out from under a running service).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2915,6 +2915,10 @@ func ResetConfig() {
 	// Clear out director's ad cache
 	ClearServerAds()
 
+	// Clear cached JWKS file contents so a test that writes a new file at a
+	// path an earlier test used does not see the earlier projection
+	ResetJWKSFileCache()
+
 	// Reset federation metadata
 	resetFedDiscoveryOnce()
 	globalFedInfo.Store(&pelican_url.FederationDiscovery{})
--- a/config/configtest/jwks.go
+++ b/config/configtest/jwks.go
@@ -1,0 +1,46 @@
+/***************************************************************
+ *
+ * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License.  You may
+ * obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ***************************************************************/
+
+// Package configtest provides shared test helpers for packages that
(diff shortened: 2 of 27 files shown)


bbockelm:
@brianaydemir - I'm grabbing this for a bit.  Instead of playing PR ping-pong, I just have a few things that I spotted.  Will rebase the PR, fix merge conflicts, and propose a few things.

bbockelm (approved):
LGTM.

(But then again, all the missing functionality I complained about I pushed fixes for!)

PR #3581, opened 2026-07-28
Remove stale pre-split CLI doc trees that break the MDX build

### LINKED ISSUE #3583: Garbage collect previously generated documentation
There is a deeper root cause:
When doc generation was reorganized into per-binary trees (commit https://github.com/PelicanPlatform/pelican/commit/792e324970a9080d4ddcf85cd1b035d986bae743), output moved to commands-reference/pelican/ and commands-reference/pelican-server/ (see the go:generate directives at cmd/main.go:41-42), but the old flat directories commands-reference/cache/ and commands-reference/object/ trees were never deleted. I reproduced this error and noticed it is actually happens on a obsolete file (see below). So the best solution would be completely remove those two old flat directories above.

Build Error
Expected a closing tag for `<hash>` (19:48-19:54) before the end of `paragraph`
./app/commands-reference/cache/introspect/metadata/page.mdx?metadata
Expected a closing tag for `<hash>` (19:48-19:54) before the end of `paragraph`

CannonLock:
Fixes invalid markdown.


CannonLock:
Screenshot 2026-07-28 at 11 42 04 AM (3)

Proof the escape works and results in correct docs.

cmd/cache_introspect.go
```diff
@@ -90,7 +90,7 @@ a particular version, or --instance to specify by instance hash.
 Example:
   pelican cache introspect metadata pelican://my-federation/data/file.dat
   pelican cache introspect metadata /data/file.dat --etag="abc123"
-  pelican cache introspect metadata --instance=<hash>`,
+  pelican cache introspect metadata --instance=\<hash>`,
```
h2zh: This addition is unnecessary. The escaping already exists at another correct layer: escapeMdxAngleBrackets `cmd/doc_gen.go:377` rewrites `<hash>` → `&lt;hash&gt;`

h2zh (changes requested):
Thanks for catching out this subtle problem! After my investigation, I think there is a deeper root cause:
When doc generation was reorganized into per-binary trees (commit 792e3249), output moved to `commands-reference/pelican/` and `commands-reference/pelican-server/` (see the `go:generate directives` at `cmd/main.go:41-42`), but the old flat directories `commands-reference/cache/` and `commands-reference/object/` trees were never deleted. I reproduced this error and noticed it is actually happens on a obsolete file (see below). So the best solution would be completely remove those two old flat directories above.

```
Build Error
Expected a closing tag for `<hash>` (19:48-19:54) before the end of `paragraph`
./app/commands-reference/cache/introspect/metadata/page.mdx?metadata
Expected a closing tag for `<hash>` (19:48-19:54) before the end of `paragraph`
```

CannonLock:
@h2zh I updated by deleting the directory then rerunning the build so that only the existing docs are added in.

h2zh (approved):
LGTM. Thanks for the iterations!

PR #3586, opened 2026-07-29
bject evict: fix exit codes, add --cache targeting, and stop the director shadowing cache evict/prestage

### LINKED ISSUE #3585: pelican object evict has no way to target a specific cache
`pelican object evict` doesn't allow the user to specify a cache to evict from. This seems counterintuitive since generally there is a bad object in a specific cache a use will want to target.

turetske:
This PR addresses several issues found while analyzing `pelican object evict`: a
latent exit-code bug shared across the object commands, the inability to target a
specific cache for eviction, and a director routing bug that shadowed the cache's
evict/prestage endpoints when the two are co-located. It also adds test coverage
for the evict path, which previously had none.

closes #3585

### DIFF
--- a/client/evict.go
+++ b/client/evict.go
@@ -123,33 +123,85 @@ func DoEvict(ctx context.Context, remoteObject string, immediate bool, options .
 		bearerToken, _ = tokenGen.Get()
 	}
 
-	// Determine the cache endpoint to contact. Use the first cache URL
-	// from the director response (the same cache that would serve a get).
-	if len(dirResp.ObjectServers) == 0 {
+	// Determine which cache(s) to evict from.
+	//
+	//   - With preferred caches (the --cache flag), evict from each specified
+	//     cache.  generateSortedObjServers returns exactly those caches (plus
+	//     the director's list when the "+" fallback sentinel is present), so a
+	//     user can target one specific cache, several, or "cache,+".
+	//   - Without --cache, evict from the director's top-choice cache only.
+	//     We deliberately do NOT fan out over every entry in ObjectServers by
+	//     default: for direct-read namespaces that list can include origins,
+	//     which have no evict endpoint, and blindly evicting against them would
+	//     fail.  Targeting the full federation is done by listing the caches
+	//     explicitly (or with "+").
+	sortedServers, nPreferred, sortErr := generateSortedObjServers(dirResp, tc.prefObjServers)
+	if sortErr != nil {
+		return "", errors.Wrap(sortErr, "failed to determine which caches to evict from")
+	}
+	if len(sortedServers) == 0 {
 		return "", errors.New("director returned no cache servers for this path")
 	}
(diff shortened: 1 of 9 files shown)


cmd/object_evict.go
```diff
@@ -67,6 +77,9 @@ func init() {
 	flagSet := objectEvictCmd.Flags()
 	flagSet.BoolP("immediate", "i", false, "Delete objects immediately instead of marking them for priority eviction")
 	flagSet.StringP("token", "t", "", "Token file to use for authorization")
+	flagSet.StringP("cache", "c", "", `A comma-separated list of caches to evict from, where a "+" in the list indicates
+the client should also fall back to the director-provided caches.  If omitted, the
```
h2zh, line 81: I think "fall back to the director-provided caches" might be kind of inaccurate here: every listed cache is contacted _regardless of success_. Also, remember to regenerate the command docs after the change.

client/evict.go
```diff
@@ -123,44 +123,92 @@ func DoEvict(ctx context.Context, remoteObject string, immediate bool, options .
...
+	//     which have no evict endpoint, and blindly evicting against them would
+	//     fail.  Targeting the full federation is done by listing the caches
+	//     explicitly (or with "+").
+	sortedServers, nPreferred, sortErr := generateSortedObjServers(dirResp, tc.prefObjServers)
+	if sortErr != nil {
+		return "", errors.Wrap(sortErr, "failed to determine which caches to evict from")
+	}
+	if len(sortedServers) == 0 {
 		return "", errors.New("director returned no cache servers for this path")
 	}
```
h2zh, line 144: What if `len(sortedServers) == 1`: on a zero-cache namespace, Director's top choice is an origin, and the evict hits it.

client/evict.go
```diff
@@ -123,44 +123,92 @@ func DoEvict(ctx context.Context, remoteObject string, immediate bool, options .
...
-	// Determine the cache endpoint to contact. Use the first cache URL
-	// from the director response (the same cache that would serve a get).
-	if len(dirResp.ObjectServers) == 0 {
+	// Determine which cache(s) to evict from.
+	//
+	//   - With preferred caches (the --cache flag), evict from each specified
+	//     cache.  generateSortedObjServers returns exactly those caches (plus
+	//     the director's list when the "+" fallback sentinel is present), so a
+	//     user can target one specific cache, several, or "cache,+".
+	//   - Without --cache, evict from the director's top-choice cache only.
```
h2zh, line 132: "evict from the director's top-choice cache only" contains ambiguity based on the context. On a zero-cache namespace, Director's top choice is an origin, and the evict hits it and _fails_. The evict action actually happens, but just fails.

h2zh (changes requested):
The features in this PR generally look good. Just a few narratives need to clarify:

A rebase is required since the main changed `DoEvict` too.

Also it took me a while to figure out what's "bject" in the title... probably just a typo of "Object".

PR #3587, opened 2026-07-29
Make devcontainer's hostname static

melody0123:
Add `runArgs` to devcontainer file to inject `--hostname pelican-dev` CLI argument for `docker run`.

In this way, container will have a static hostname `pelican-dev`, instead of a dynamic one (e.g., ade3467fac78), which eliminates some repetitive work like changing hostname in `/etc/pelican/pelican.yaml` every time a container is torn down and rebuilt.

Remember that some configurable parameters like `Federation.DirectorUrl` contains the hostname of the container.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,6 +4,7 @@
 	"name": "Pelican",
 	"image": "hub.opensciencegrid.org/pelican_platform/pelican-dev:latest-itb",
 	"workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+	"runArgs": [ "--hostname", "pelican-dev" ],
 	"forwardPorts": [
 		8444,
 		8443,


brianaydemir:
As someone who uses fixed hostnames for testing (https://github.com/brianaydemir/pelicanplatform-test-framework), this strikes me as a perfectly reasonable improvement, but I'll let those who actually use VSCode (@turetske, @h2zh) make the final call.

bbockelm:
@melody0123 - thanks for the PR!

I like the idea, but can we consider the use case a bit?

It feels a bit like if changing the hostname is hard then something is funky in the Pelican configuration language / setup itself.

Are we fixing the environment where we should be fixing the software?

Can you enumerate a few of the "pain points" to see if we can spot the problem?

melody0123:
@bbockelm The only use case that I'm aware of is during the test. During the test, you will want to run Director, Registry, Origin, and Cache inside the same container. You will need to configure all of them via `/etc/pelican/pelican.yaml`. A typical `pelican.yaml` to run everything locally looks like this:

```
TLSSkipVerify: true
Debug: true

Cache:
  Port: 8442

Origin:
  EnableMacaroons: false
  Port: 8443

Server:
  EnableUI: false

Logging:
  Cache:
    Pss: debug

Federation:
  RegistryUrl: https://70fa1e4d6777:8446
  DirectorUrl: https://70fa1e4d6777:8445

Director:
  MaxMindKeyFile: /etc/pelican/maxmind

OIDC:
  ClientID: "someguyhere"
  ClientSecretFile: /etc/pelican/pelican.yaml
```

Note that the container's name (also used as the hostname of the container), `70fa1e4d6777`, is used by `RegistryUrl` and `DirectorUrl`. Every time a container is torn down and brought up again, this value will change, which means we need to edit `pelican.yaml` file for the new container every time.

I tried to change the hostname to `127.0.0.1`, `localhost`, and `172.27.0.2` (this one is resolved from the hostname), but none of them worked. The error message thrown by the Director is `connection refused`. Only the hostname of the container will work. **However, this does not make sense to me, because TLS check is skipped so using an IP address should not result in any error.** This may suggest potential issue inside the codebase itself.

Considering all above, I injected a CLI argument into the dev container file to fix the hostname, and it works as intended.

As for your question, 

> Are we fixing the environment where we should be fixing the software?

I think the answer is no, because using the hostname in `RegistryUrl` and `DirectorUrl` is perfectly reasonable. It's sysadmin's responsibility to accurately fill in those values. A server may have multiple hostnames at the same time, and Pelican should not decide which one to use automatically on behalf of the sysadmin. Therefore, the burden of configuration should fall on the environment instead of the software.

bbockelm:
Let's dig into the registry/director URL items a bit.  Since that config is set to run on ports 8442 / 8443, where does 8445 and 8446 come from?

melody0123:
@bbockelm 8442 is port used by origin and 8443 is used by cache, while 8445 and 8446 are used by director and registry respectively. During development and testing, you will want to run all these 4 entities in one container so they should be 4 different ports. More details can be found in the Pelican doc (https://docs.pelicanplatform.org/parameters)

PR #3590, opened 2026-07-30
Fix db token init panic

### LINKED ISSUE #3589: Director panics during API token verification.
It appears that the API token subsystem makes a copy of the DB pointer at startup.  However, this copy occurs in parallel with another thread that initializes the database.

If initialization is too slow, then the API token subsystem copies a nil pointer and it stays that way until restart.

Explanation seems reasonable, matches stack trace.

turetske:
closes #3589 

Fix nil-database panic in API token verification

The api_token package held its own `ServerDatabase *gorm.DB`, copied
from database.ServerDatabase by web_ui.ConfigureServerWebAPI. That
call sits at launchers/launcher.go:97, ahead of every
database.InitServerDatabase call site -- the director's is in
DirectorServe, the registry's in RegistryServe, and so on. The copy
was therefore nil, and since it was a value rather than a live read,
it stayed nil for the life of the process.

This was not a startup race. Both statements run on the same
goroutine, in one function, in fixed order, so the handle was nil on
every server, every time. VerifyApiKey then dereferenced it and
panicked inside gorm's getInstance(). gin.Recovery() caught the panic
and returned 500, which is why it surfaced as stack traces in director
logs rather than a crashed process.

The panic required an actual API token to reach the check: issuers are
tried in order and APITokenIssuer is never first, and api_token's
format regex plus hex decode reject JWTs and garbage before the
database is consulted. That is why only some federations saw it.

Introduced by 482348da9, which replaced VerifyApiKey's explicit
`db *gorm.DB` parameter with the startup copy. First shipped in
v7.26.0; the v7.25.x line is unaffected.

Replace the copied handle with a `DB func() *gorm.DB` hook that web_ui
wires in its init(), mirroring the EffectiveScopesForUser hook already
in that file. Resolving through a function on every call fixes two
things at once: ordering no longer matters, and the handle cannot go
stale when InitServerDatabase reassigns the global in a process
running several modules. api_token still does not import database, so
the client binary keeps its independence from gorm and SQLite, and
api_token's init() still self-registers the verifier.

Guard both failure modes -- hook unwired, hook returning nil -- with
an error instead of a dereference, so a future ordering mistake
becomes a 403 and a log line rather than a panic.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/api_token/api_token.go
+++ b/api_token/api_token.go
@@ -44,6 +44,12 @@ var (
 	// API token format: <5-char ID>.<64-char secret>, total length = 70, alphanumeric
 	ApiTokenRegex = regexp.MustCompile(`^[a-zA-Z0-9]{5}\.[a-zA-Z0-9]{64}$`)
 
+	// DB returns the live server database handle. The web_ui layer
+	// wires this at init time; this package deliberately does not
+	// import database, so that the client binary never pulls in gorm
+	// and SQLite.
+	DB func() *gorm.DB = nil
+
 	// EffectiveScopesForUser is a hook the web_ui layer wires up at
 	// init time to expose the effective scope set for a user (the
 	// union of DB user_scopes / group_scopes plus config-derived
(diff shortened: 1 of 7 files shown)


bbockelm (approved):
LGTM.

Because the assignment to `DB` occurs in `init()`, I can't spot any potential race conditions.

Fix matches the symptoms.  I think we've got it.

PR #3597, opened 2026-08-01
Standalone mode for the Origin

### LINKED ISSUE #3596: Allow origin to run standalone
Just like the cache runs standalone, I can see value in letting the origin run standalone.

That is, we're now providing some pretty cool data sharing functionality (upload / download via the web!) even separate from the federation.  Let's enable folks to use this.

bbockelm:
Adds a standalone mode to the Origin: `Origin.EnableStandaloneMode`.

A standalone Origin belongs to no federation. It does not register at a Registry, does not advertise to a Director, and contacts no other federation service. It is for administrators who want Pelican's managed authorization over a storage backend without operating — or belonging to — a federation.

This is the Origin counterpart of `Cache.EnableSiteLocalMode`, but stricter: a site-local Cache still queries a Director as a client, whereas a standalone Origin makes no federation calls at all.

### A federation of one

Clients never address Origins directly — they ask a federation's discovery endpoint for a Director, then ask that Director which server holds the object. A standalone Origin answers the first question by serving `/.well-known/pelican-configuration` itself, and answers the second by making it unnecessary: the document names **no** Director, and a client that finds none asks the discovery host about the object directly.

That client-side path is `queryObjectServerDirectly`. It HEADs the object at the discovery host and reads the namespace metadata — prefix, whether a token is required, which issuers mint one — off the `X-Pelican-*` headers the Origin already attaches to every response. A HEAD is enough, costs nothing extra, and works whether the object turns out to be public (200), protected (401/403), or absent (404): the headers describe the namespace, not the object.

So there is no redirect and no Director shim. Objects stay where a native Origin already serves them, at their `FederationPrefix`, and the only new server-side surface is the discovery document.

### Scope

- Restricted to the native "v2" backends (`posixv2`, `s3v2`, `httpsv2`, `globusv2`, `ssh`); the XRootD-fronted backends build their configuration from federation metadata. `InitServer` also rejects co-locating a Director/Registry/Cache/Broker module, rejects `Origin.DisableDirectClients` (under which a standalone Origin would refuse every request), rejects a root `FederationPrefix` (which would collide with the Origin's own API routes), and rejects `Federation.DiscoveryUrl` (a standalone Origin belongs to no federation, so naming one contradicts the document it serves).
- The web UI drops the affordances only a Director or Registry could act on (registration status, Downtime, Federation Overview) via a new `standaloneOrigin` field on `GET /api/v1.0/servers`.

### Commits

Seven, each independently reviewable:

1. **`server_structs`: share the Director's client-facing header helpers.** Extracts the `X-Pelican-*` header writers and the longest-namespace-prefix match out of the Director. Pure delegation, no behavior change. Cherry-picks onto `main` on its own and may be worth landing separately, since #3510 (anycast support for the Cache) needs the same helpers.
2. **`client`: resolve and authorize against a federation with no director.** The directorless resolution path and the token-hint mechanism, shaped to reconcile with #3510; the anycast plumbing stays in that PR. Also fixes a pre-existing `DirRespCache` collision between federations sharing a namespace path.
3. **`origin`: add `Origin.EnableStandaloneMode`.** The feature and its startup validation.
4. **`origin`: measure disk usage from the backend.** A pre-existing bug, unrelated to standalone mode except that it removed the last limitation — see below.
5. **`web_ui`**, 6. **`docs`**, 7. **tests.**

### A pre-existing bug this turned up

Disk usage for non-POSIX backends was measured by pointing the Pelican client at the origin's own namespace — asking a Director where to find data the process was already serving. That path needs `config.InitClient`, which **no serve path calls**, so every walk failed with `client has not been initialized`, logged a warning, and reported the export as zero bytes. `TestDiskUsageSuccessPelican` passes only because `fed_test_utils` initializes the client, which is what kept it hidden.

Commit 4 walks the backend directly for the native backends. The XRootD-fronted ones keep the old path, since XRootD holds their storage and there is nothing local to walk.

### Note for reviewers: what a token hint is allowed to do

The client change touches credential acquisition, so the gate on it is worth reading closely. A hint that is honored can send the client to an issuer to register an OAuth client, post a stored refresh token, or raise an interactive login prompt. `canApplyTokenHint` requires two things before any of that:

1. **No Director vouched for the transfer.** Once a Director has named the issuers for a namespace, an object server claiming different ones is either misconfigured or hostile. Provenance is derived from the URL — a federation that publishes no Director endpoint has no Director to do the vouching — so it cannot be forgotten at a call site.
2. **The hint came from the host that published the federation's discovery document.** In a federation of one that host *is* the federation: it already tells clients where the JWKS lives and which issuers to trust, so a claim it makes about its own namespaces is exactly as authoritative as the document they already accepted. Any other endpoint — a cache named with `-c`, a redirect target — is merely somewhere the bytes live, and its hints are ignored. Scheme is compared along with host, so a plaintext endpoint cannot answer for an https discovery host.

A caller who named a credential (`WithToken`, or an external token provider) keeps it; a hint never displaces a chosen credential. Acquisition runs on a generator derived from the job's, so the hint is never written into shared state, and it is coalesced through the job's singleflight under a key naming the hint — a recursive transfer registers one OAuth client, not one per file. A credential the endpoint accepts is published back to the job's generator so sibling files and the checksum request that follows reuse it. A hint that yields the credential just rejected ends the attempt rather than retrying into the same 403.

`queryObjectServerDirectly` is hardened to match: it refuses to follow a redirect off the host it resolved against (the reply chooses the issuer, and Go attaches the caller's bearer token to same-origin hops), and it rejects a reply carrying no namespace metadata rather than reading it as "no token required".

Relatedly, `origin_serve` now returns 403 (rather than 401) when a presented token does not authorize, but **only on a standalone Origin** — federated Origins keep 401, which is what tells rclone and similar clients to re-run their `bearer_token_command`.

### Testing

- `TestStandaloneOriginCliEndToEnd` runs `pelican origin serve` and `pelican object {put,get,ls}` / `token create` as **separate processes** from the built binary, in an environment stripped of `PELICAN_*`, and checks uploaded bytes against the backing directory. `token create` runs with no `--issuer`, so the CLI must discover it from the Origin's discovery document and JWKS.
- `TestStandaloneOriginCliFederatedControl` is the negative control: the same config with standalone mode off must fail to start for want of a discovery endpoint.
- `TestStandaloneOriginServe` covers the wire contract in-process (discovery document, namespace metadata on the object's own response, token hints, and a real client transfer), plus unit tests for each config rejection.
- Two invariants that previously had no test: a **federated** Origin emits no `X-Pelican-*` hint headers (the regression that would leak Director-owned metadata from every federated Origin), and `LongestNSMatch` resolves nested, exact, trailing-slash and root prefixes the way the Director does — the function the "resolves namespaces identically" claim rests on.
- The token-hint tests are built so that removing the retry, removing the Director check, or removing the discovery-host check each makes a test fail: `TestTokenHintAcquiresAndRetries` is the positive case, and `TestTokenHintIgnoredWhenDirectorSupplied` / `TestTokenHintIgnoredFromNonDiscoveryHost` share its fixture so the difference in outcome isolates one half of the gate each.

### DIFF
--- a/client/acquire_token.go
+++ b/client/acquire_token.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -80,7 +81,12 @@ type (
 	// Thread-safe and will auto-renew the token throughout the lifetime
 	// of the process.  Satisfies the TokenProvider interface.
 	tokenGenerator struct {
-		DirResp                 *server_structs.DirectorResponse
+		DirResp *server_structs.DirectorResponse
+		// tokenIsExplicit reports whether the caller named the credential to use
+		// (WithToken, or a token option on the transfer).  A caller who picked a
+		// credential does not want it silently replaced by one acquired from an
+		// issuer some object server named.
+		tokenIsExplicit         bool
 		Destination             *pelican_url.PelicanURL
 		TokenLocation           string
 		TokenName               string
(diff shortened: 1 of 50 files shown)


bbockelm:
Reworked so that standalone mode involves **no redirects at all** — the previous revision had the Origin running a mini-director and relocating its data endpoint to `/api/v1.0/origin/data`, which was more machinery than the deployment needs.

The Origin now serves objects at their federation prefix, and the client reaches them directly:

- `/.well-known/pelican-configuration` names **no Director**. That absence is the signal: a client discovering a federation without one resolves objects against the discovery host itself.
- The Origin attaches `X-Pelican-{Namespace,Authorization,Token-Generation}` to every response it serves, not just rejections — with no Director, the object's own response is the only place a client can learn its namespace, whether a token is required, and which issuer mints one.
- `getDirectorInfoForPath`, given a federation with no Director, HEADs the object against the discovery host and builds the `DirectorResponse` from those headers. It's the choke point every caller already goes through, so `object get/put/ls/stat`, `token create` without `--issuer`, `AcquireToken`, and evict all work unchanged.

`director/standalone_origin.go` is gone; only the discovery document survives, in `origin/standalone.go`, which needs nothing from the `director` package — so `launchers` no longer imports `director` on the origin path.

The token hints keep a role: the probe answers "which issuer" up front, and a 403 mid-transfer still lets a client with a stale credential learn the same thing without a second round trip.

The CLI end-to-end test (`pelican origin serve` and `pelican object …` as separate processes) passes unchanged apart from the discovery-document assertion, and the in-process test now asserts a direct `200` with a non-redirect-following client, so a redirect would fail the test rather than be silently followed.

h2zh:
Add a milestone in case we lost it in the 26.0 release note. A reminder for the release manager @turetske

Issue #3593, opened 2026-07-31
Token-acquisition errors are silently discarded during recursive listing / download retries (`bearer_auth.Authorize`, `handle_http.go` retry loop)


melody0123:
### Found at commit

`6867fe3abedc3b6064b3269cb70fad252e9a51ca` (`main`). All line numbers and permalinks in this issue are pinned to that exact commit — if the surrounding code has since changed, re-check against this SHA specifically before assuming the finding is stale or already fixed.

### Status: UNVERIFIED — found via code-path audit, not yet reproduced with a failing test

This was found by static tracing of the code, prompted by a review discussion (see "How this was found" below). **No one has actually reproduced the failure with a live repro case yet.** Before treating this as confirmed, someone should write an integration/unit test that:
1. Starts a recursive `pelican object ls` (or a multi-file download) against a namespace requiring a token.
2. Lets the cached token expire *mid-operation* (after the first request has already succeeded).
3. Forces re-acquisition to fail in a way that would normally produce a specific, actionable error (e.g. issuer doesn't support the device-code flow, or `NonInteractive` is set with no usable cached token).
4. Confirms the user only ever sees a generic `"authorization failed with status 401..."` / `"authentication failed for N consecutive times..."` message, never the real underlying reason.

If that repro doesn't actually manifest as described, close this out or correct the description below.

### How this was found

While writing tests for an unrelated fix ([Issue #3584](https://github.com/PelicanPlatform/pelican/issues/3584), downgrading some expired-token log lines from `Warningln` to `Debugln` in `client/acquire_token.go`, so that routine expired-token/refresh-failure noise doesn't scare users during normal `--debug`-off operation), a reviewer @bbockelm asked a good question: *"What if the last-resort interactive device-code auth also fails — e.g. the issuer doesn't support it — and somehow no error message is produced? Would the whole operation just silently fail with no output at all?"*

That's a legitimate concern to raise before shipping a "make errors quieter" change, so it was investigated directly rather than assumed away. A static analysis on the code performed by Claude confirmed:

- The log-level change itself is **not** the risk: a log statement's severity (`Debugln` vs `Warningln`) and a function's returned `error` value are completely independent mechanisms in Go. Downgrading a log call cannot make an error vanish.
[...]

PR #3600, opened 2026-08-04
Downgrade expired-token warnings to debug-level logging

### LINKED ISSUE #3584: Reclassify token warning messages as debug statements
A user was seeing scary looking warning messages in a situation where their token on disk expired:

```
$ pelican object ls osdf:///foo/bar
WARNING[2026-05-11T16:24:03-07:00] Failed to parse token: "exp" not satisfied
WARNING[2026-05-11T16:24:04-07:00] Failed to renew an expired token: oauth2: "invalid_grant" "invalid refresh token"
To approve credentials for this operation, please navigate to the following URL and approve the request:

https://foo-bar-origin.osg-htc.org:8444/api/v1.0/issuer/device?user_code=BLAH_BLAH
```

Expired access or refresh tokens are par for the course so we should turn these into messages that show up in `--debug` output

melody0123:
Expired access/refresh tokens are common, but tokenIsValid and AcquireToken's refresh-retry path previously logged them via Warningln, which reads as alarming in normal (non-debug) output. Route the exp-related failure in tokenIsValid's jwt.Validate check, and the refresh failure in AcquireToken, to Debugln instead, while keeping other validation failures (malformed tokens, non-expiration errors) at Warning.

Also documents via comments that the equivalent
errors.Is(err, jwt.ErrTokenExpired()) branches in tokenIsAcceptable and tokenIsValid's initial UnsafeParseClaims error check are currently unreachable, since UnsafeParseClaims disables claim validation and never returns ErrTokenExpired -- kept only as a defensive guard in case that helper's behavior changes.

Adds test coverage for the reachable logging paths in tokenIsValid, tokenIsAcceptable, and refreshTokenEntry.

Side note: @bbockelm asked a really good question: What if the interactive approval message also disappears due to something else wrong, e.g., Issuer does not support device code method. If warning messages are buried into debug messages, and interactive approval messages are also missing due to some other errors, user will see no output at all, which may confuse them. Considering this, should we still convert the warning messages into debug messages?

My answer is yes. There are two plausible defenses to the change:
1. This is a philosophical question instead of an engineering question. The (non-technical) user's request is that the warning messages are too scary to see. They are happy if the warning messages will not appear. In this sense, users want [blissful ignorance](https://en.wikipedia.org/wiki/Blissful_ignorance_effect). The necessary tradeoff for this bliss is by having less information.
2. This is still an engineering question. Warning messages and missing interactive approval messages are two orthogonal problems. The warning messages that this PR tackles is a logging issue. The missing interactive approval messages is an error handling issue. Logging should not be used to handle errors. The only acceptable method to handle errors is either throwing it to the caller one level up, or directly handle it in the same level. If the interactive approval messages somehow are missing, it means we our error handling chain is problematic (e.g., an error is silently swallowed somewhere, e.g., [Issue 3593](https://github.com/PelicanPlatform/pelican/issues/3593)).

### DIFF
--- a/client/acquire_token.go
+++ b/client/acquire_token.go
@@ -508,7 +508,16 @@ func (tg *tokenGenerator) Get() (token string, err error) {
 func tokenIsAcceptable(jwtSerialized string, objectName string, dirResp server_structs.DirectorResponse, opts config.TokenGenerationOpts) bool {
 	tok, err := token.UnsafeParseClaims(jwtSerialized)
 	if err != nil {
-		log.Warningln("Failed to parse token:", err)
+		// token.UnsafeParseClaims disables claim validation (jwt.WithValidate(false)),
+		// so in practice err here is never jwt.ErrTokenExpired -- only a structurally
+		// malformed token lands in this branch. The expired-token check is kept anyway
+		// as a defensive guard in case UnsafeParseClaims's validation behavior ever
+		// changes; today it's unreachable, so this always logs at Warning.
+		if errors.Is(err, jwt.ErrTokenExpired()) {
+			log.Debugln("Failed to parse token:", err)
+		} else {
+			log.Warningln("Failed to parse token:", err)
+		}
 		return false
 	}
 
@@ -722,7 +731,17 @@ func isValidSciScope(authz string, operation config.TokenOperation) bool {
 func tokenIsValid(jwtSerialized string) (valid bool, expiry time.Time) {
 	tok, err := token.UnsafeParseClaims(jwtSerialized)
 	if err != nil {
-		log.Warningln("Failed to parse token:", err)
+		// token.UnsafeParseClaims disables claim validation (jwt.WithValidate(false)),
+		// so err here is never jwt.ErrTokenExpired in practice -- only a structurally
+		// malformed token reaches this branch (the real exp check happens below, via
+		// jwt.Validate). Kept as a defensive guard in case UnsafeParseClaims's
+		// validation behavior ever changes; today it's unreachable, so this always
+		// logs at Warning.
+		if errors.Is(err, jwt.ErrTokenExpired()) {
+			log.Debugln("Failed to parse token:", err) // Only convert to debug message if it's indeed an expired token
+		} else {
+			log.Warningln("Failed to parse token:", err)
+		}
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


melody0123:
This PR is an attempt to fix the issue: https://github.com/PelicanPlatform/pelican/issues/3584

turetske (changes requested):
@melody0123 I'm pushing back against your answers to @bbockelm's question. Specifically, the user was complaining that warnings that aren't real warnings are scary to see and confuse the users. He was NOT referring to actual problems with the token. We should not suppress real warnings. The task is suppress the warnings that are actually not applicable and will scare users.

So this isn't an orthogonal problem, it specifically is the task to lower the warnings that are unnecessary (and only those warnings).

melody0123:
@turetske Thanks for the review! I believe @bbockelm 's question is "if we convert these 2 specific warnings into debug messages, AND interactive approval somehow silently fails, user gets no output at all, is it a new user experience issue caused by converting these 2 specific warnings?" My answer is that logging issue (i.e., deciding if a message should be logged or shown to the user) and error handling issue (i.e., interactive approval silently fails) are two different questions and should not be mixed together (i.e., orthogonal). Because no matter how we adjust logging levels for a specific message, it does not change the error propagation / handling logic. If error propagation chain is incomplete (i.e., there is an error silently swallowed inside a function, no handling and no error returned to the caller), it is still incomplete no matter how we change the logging level.

With this insight, we should convert these 2 warnings to debug messages, and not worry about the silent failure (if any) happened elsewhere. Silent failure is an error handling issue and should be considered separately.

Some math statements on error handling asserted by me (you are very welcomed to come up with counter examples!): 
1. Error handling is considered complete if and only if an error is neither handled anywhere on the call chain nor thrown to the caller. Equivalent statement is: Error handling is not complete if and only if an error is either handled somewhere on the call chain or thrown to the caller (we assume that if an error is thrown to the main function, it has to be handled there).
2. Error handling is considered sound if and only if error is correctly handled somewhere.
3. Error handling is considered correct (i.e., not problematic) if and only if it's sound and correct.

If 1-3 are correct, we can argue that silent failure is an error handling issue because silent failure happens if and only if error handling is not complete.

turetske:
@melody0123 Ah, okay! I understand the difference now. I see that there is also already an issue for this as well.

turetske (approved):

PR #3610, opened 2026-08-06
Make Claude Code for VS Code's transcript persistent

melody0123:
Currently, if we rebuild the container image and start a new container, we will lose all Claude Code's transcript and login credentials, because they are stored in `/home/vscode/.claude`, which will be wiped out across image rebuild. We make `.claude` become a named volume managed by the docker. In this way, everything under `.claude` folder become persistent and can be shared among multiple containers.

Note: Transcripts are saved locally and not synced with the remote.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -37,7 +37,7 @@
 			]
 		}
 	},
-	"postCreateCommand": "go install github.com/go-delve/delve/cmd/dlv@latest",
+	"postCreateCommand": "sudo chown -R $(id -u):$(id -g) $HOME/.claude && go install github.com/go-delve/delve/cmd/dlv@latest",
 	"remoteEnv": {
 		"GOPATH": "/home/vscode/go",
 		"PATH": "/usr/local/go/bin:/home/vscode/go/bin:${containerEnv:PATH}"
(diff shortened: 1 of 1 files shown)


melody0123:
Don't review it yet. I found that this method is not stable.

h2zh (changes requested):
As someone who also uses the Claude Code extension in my IDE, I appreciate this change! That said, I think our internal docs (https://github.com/CHTC/pelican-docs) would be a better home for this config than the shared devcontainer. People may have different preferred AI agent providers, this creates the claude-code-config volume and runs the chown for every devcontainer user whether or not they use Claude Code. I think it's important to maintain the neutrality principle and we should keep vendor-specific personal tooling out of the shared dev environment. 

Two notes worth carrying over to the docs version:
- Claude Code also keeps state in `~/.claude.json` (onboarding, project trust, MCP servers), which is a top-level file outside the mounted directory, so rebuilds still reset that even though credentials under `~/.claude/` survive. 
- Chaining the chown with `&&` means that if it ever fails, the delve install is skipped too. We'd better keep it as a separate command.

I'm happy to review the docs write-up if you move it over!

melody0123:
@h2zh Fair enough! I will post my devcontainer.json there. This issue will be closed

PR #3618, opened 2026-08-13
Make usernames a truly global authorization handle

### LINKED ISSUE #3608: `Server.UIAdminUsers` breaking change in 7.27: fix docs and add startup validation
## Context

7.27 is the last release before the v26.0 LTS, and it intentionally lands the auth/identity breaking changes from #3299 (new user/group model) so the LTS starts on the new model.

As part of #3299, admin authorization now keys on the **server-managed username** only. Previously, `Server.UIAdminUsers` matched a user's username, internal ID, or OIDC subject; as of 7.27 it matches the username only. Entries containing a CILogon Identifier (e.g. `http://cilogon.org/serverA/users/12345`) silently stop granting admin access — an operator-facing lockout with no signal.

`Server.AdminGroups` is unaffected and is the recommended path going forward.

## To-do

- [ ] Mark this as a breaking change in the 7.27 release notes.
- [ ] Correct the `Server.UIAdminUsers` description in `docs/parameters.yaml` — it still tells operators to use the CILogon "subject" claim. (`Server.UserAdminUsers` and `Server.CollectionAdminUsers` descriptions carry the same error: the code matches username only.)
- [ ] Add a startup validation: log a loud error if `Server.UIAdminUsers` contains an entry that can't be a valid username (URL/oidc_sub-shaped values), naming the offending entries and pointing operators at the Pelican username (usually like `bob@wisc.edu`, confirmed from the user's profile page after first login) or `Server.AdminGroups`.

Per the OSDF Ops meeting decision, the validation should warn loudly, not fail startup.

### LINKED ISSUE #3620: Patch 7.27 breaking change for the user model
## Summary

Server.UIAdminUsers matching changed in 7.27: pre-existing oidc_sub entries 
silently lose admin, and usernames are not unique across issuers

## Context

As of v7.27 (#3299), `Server.UIAdminUsers` / `Server.UserAdminUsers` /
`Server.CollectionAdminUsers` entries are matched against the server-managed
**username**, no longer against the OIDC sub. Reported in [OPS-525](https://opensciencegrid.atlassian.net/browse/OPS-525) during 7.27.0
testing: an operator whose config listed CILogon User Identifiers (e.g.
`http://cilogon.org/serverE/users/112233`) logged in and silently received no
admin scope. The parameter docs still recommend the old oidc_sub format.

## Problems

1. **Silent privilege loss.** Pre-7.27 entries (oidc_sub URLs, internal IDs)
   can never match a valid Pelican username, but nothing warns the operator;
   admin access just disappears after upgrade.
2. **Ambiguous match target.** Admin matching compares usernames
   issuer-blind (`web_ui/scopes.go`), but the schema only enforced
   `UNIQUE(username, issuer)`. One username could therefore back two live
   accounts under different issuers (e.g. a local password account and an
   OIDC account), and both would satisfy a single admin entry.
3. **Usernames are hard to know ahead of time.** Per the discussion in
   [OPS-525 comment](https://opensciencegrid.atlassian.net/browse/OPS-525?focusedCommentId=417130):
   asking admins to read each user's profile page after first login requires
   admin/user coordination; operators need predictable username values (or at
   least a loud signal when an entry is wrong) to migrate their config.

## Proposed fix

- Error loudly at startup for admin-list entries that cannot be a valid
  Pelican username, naming each bad entry and pointing at
  `Server.AdminGroups` as the alternative.
- Make live usernames globally unique (partial unique index, soft-deleted
  rows excluded) so a username-based admin entry maps to at most one account.

Addressed by #3618.

h2zh:
Fixes #3620. Related: #3608.

This PR address the issues in 7.27 testing, see Jira ticket OPS-525.

Admin matching (`Server.UIAdminUsers` / `Server.AdminGroups` / the built-in
"admin" grant) compares on username alone and ignores the issuer, and the
design doc calls the username "the unique machine-readable name". The schema
only enforced uniqueness per issuer, so one username could exist under two
issuers (e.g. a local password account and an OIDC account) and both would
satisfy a single admin entry. This PR closes that gap in 2 steps:

- **958ddebf9** warns loudly at startup when `Server.UIAdminUsers`,
  `Server.UserAdminUsers`, or `Server.CollectionAdminUsers` contain entries
  that can never be a valid Pelican username (pre-7.27 OIDC subjects such as
  CILogon User Identifiers), since those entries silently grant nothing.
- **04e224073** replaces the per-issuer `UNIQUE(username, issuer)` index with a
  global partial unique index on `username` (live rows only, so soft-deleted
  accounts don't reserve names). `UNIQUE(sub, issuer)` is unchanged.
 - ~~**464efb744** makes the built-in `admin` a single global identity, which fixes
  the startup/login failure described at the bottom (when multiple Pelican services
  shared the same `Server.DbLocation`, e.g. fed-in-a-box, dev container). The
  startup bootstrap and the htpasswd/init-code login paths adopt the live
  local-shaped `admin` row (`sub == username`) a sibling service created,
  instead of minting a per-issuer duplicate the new index rejects. Adoption is
  restricted to the literal `admin` account: every other username stays keyed on
  `(sub, issuer)`, so a cross-issuer collision fails rather than silently linking
  two same-named accounts. OIDC-linked rows (`sub != username`) are never
  adopted, and the bootstrap refuses to backfill group ownership onto an
  OIDC-linked `admin`.~~

### Note to admin

- If your existing database holds live cross-issuer duplicate usernames
  (legal under the old constraint), the migration fails at startup with
  `UNIQUE constraint failed: users.username`. Rename or remove the duplicate
  accounts, then restart; the migration does not pick a winner for you.
- After upgrade, check startup logs for the new error about
  `Server.UIAdminUsers` entries that are not valid usernames, and replace
  them with the server-managed username (visible on the user's profile page)
  or grant access via `Server.AdminGroups`.
- ~~In deployments where several services share one `Server.DbLocation` (dev
  container / e2e), the built-in `admin` is now a single global account: the
  first service to start creates the row, the rest adopt it, and logging in as
  admin on any of them resolves to the same user. Only `admin` is shared this
  way; non-admin local (htpasswd) accounts stay keyed per issuer, so do not rely
  on shared non-admin htpasswd identities across services (use OIDC or
  `Server.AdminGroups`, or give the services separate databases).~~

~~Background on the startup/login failure the third commit fixes:~~

> ~~Each service records the admin account in the shared database tagged with its own web address (ExternalWebURL). So when you log in on service B, the password checks out, but B looks for an "admin" row tagged with B's address, doesn't find one (service A already created the row tagged with A's address), and tries to create its own. The new rule that a username can only exist once in the database rejects that duplicate, the login handler hits that unexpected error, and the user sees a 500.~~

### DIFF
--- a/database/collection.go
+++ b/database/collection.go
@@ -329,9 +329,9 @@ const (
 // for the correct first-login / return-visit flow.
 type User struct {
 	ID          string     `gorm:"primaryKey" json:"id"`
-	Username    string     `gorm:"not null;uniqueIndex:idx_user_issuer" json:"username"`
+	Username    string     `gorm:"not null;uniqueIndex:idx_user_username_live" json:"username"`
 	Sub         string     `gorm:"not null;uniqueIndex:idx_user_sub_issuer" json:"sub"`
-	Issuer      string     `gorm:"not null;uniqueIndex:idx_user_issuer;uniqueIndex:idx_user_sub_issuer" json:"issuer"`
+	Issuer      string     `gorm:"not null;uniqueIndex:idx_user_sub_issuer" json:"issuer"`
 	Status      UserStatus `gorm:"not null;default:active" json:"status"`
 	LastLoginAt *time.Time `json:"lastLoginAt"`
 	DisplayName string     `gorm:"not null;default:''" json:"displayName"`
--- a/database/collection.go
+++ b/database/collection.go
@@ -268,6 +268,14 @@ const (
 	UserStatusInactive UserStatus = "inactive"
 )
 
+// BuiltinAdminUsername is the reserved username of the built-in
+// administrator that the htpasswd / init-code bootstrap creates and
+// that CheckAdmin in the web layer keys its bypass off of. It is the
+// ONLY account eligible for issuer-blind adoption (see
+// adoptBuiltinAdminByUsername); every other username is strictly keyed
+// on (sub, issuer).
+const BuiltinAdminUsername = "admin"
+
 // User is the canonical user record. Four concepts live on this row and
 // they are intentionally distinct — code that conflates them is a bug.
 //
@@ -1688,12 +1696,30 @@ func GetOrCreateUser(db *gorm.DB, username string, sub string, issuer string, cr
 		return nil, err
 	}
 
+	// No (sub, issuer) linkage yet. For the built-in admin ONLY, the
+	// username may already have a live row that a sibling service
+	// created: several services sharing one server DB (the dev-container
+	// / e2e layout) each call this with their own ExternalWebUrl as
+	// issuer, and usernames are globally unique among live rows, so a
+	// second service must adopt the first's "admin" row instead of
+	// minting a per-issuer duplicate the unique index now forbids.
+	// Adoption is deliberately limited to "admin": for any other
+	// username we do NOT infer that two local-shaped rows are the same
+	// person, and the create below simply fails the collision (see
+	// adoptBuiltinAdminByUsername).
+	if adopted, ok, adoptErr := adoptBuiltinAdminByUsername(db, username, sub); adoptErr != nil {
+		return nil, adoptErr
+	} else if ok {
+		return adopted, nil
+	}
+
 	// User not found, create one.
 	created, createErr := CreateUser(db, username, sub, issuer, creator)
 	if createErr == nil {
 		return created, nil
 	}
-	// A concurrent request may have created the same (sub, issuer) user between
+	// A concurrent request may have created the same (sub, issuer) user
+	// (or, the built-in "admin" user under a sibling service's issuer) between
 	// our SELECT above and this INSERT, tripping a UNIQUE constraint. For a
 	// get-or-create that is not a failure: if the row now exists, return it so
 	// concurrent first-time authentications don't spuriously fail (a 500 on the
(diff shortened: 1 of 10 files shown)


bbockelm (changes requested):
I'm happy with the uniqueness piece, that gets to a better state.

I'm not sure I'm happy with the `adoptLocalUserByUsername`.

That feels a bit "magic" - the behavior differs based on whether there happens to be a pre-existing user of a certain type, right?

I'd feel more comfortable with just failing that case.  Why should we assume an existing local user should be linked to one based on `sub`?

Also, I think we have code to generate a username based on incoming user info (which may be the `sub` -- but may not!).  Having the username be the sub in some cases and not for others feels dicey.

h2zh:
@bbockelm Thanks for the review! I realized that `adoptLocalUserByUsername` is a too risky approach to handle this trivial case (only bothering dev container - when multiple Pelican services share the same DB), so I remove that commit shortly after creating this PR. I also updated the PR writeup to highlight the known limitation of making `username` globally unique, which is contained in dev container.

bbockelm:
@h2zh - I think I understand your problem now.

Basically, you have multiple servers all sharing the same database backend, right?  I'm surprised this generally works!

Latest commit has a heuristic for a "correctly shaped" token to be able to take over the `admin` account.  It seems dangerous to allow admin takeover - even narrowly scoped - to enable state database sharing across multiple processes, which I'm not sure something we want to generally support.

While the heuristic of "is it OK to take over admin?" looks currently correct, I'm afraid of the potential for a code refactor to accidentally open a hole.

@brianaydemir - thoughts?

If we do decide that we don't want the state database to be shared across processes, then we should add a lock to detect this at startup.

h2zh:
Update: this design is not used. Please skip this comment.

@bbockelm Right, besides the dev-container, fed-in-a-box (used by Pelican e2e tests) is also an example of "multiple servers all sharing the same database". Thus, I think we can't sacrifice fed-in-a-box for the global unique username.

Regarding your security concerns, I use Claude to analyze the attack surface. It turns out the "admin" user adoption is safe (I turn the following text into a diagram):

Gemini_Generated_Image_nljo5enljo5enljo

1. Adoption applies only to the literal built-in admin username (a single named constant), and only when both sides are local-shaped: the requester asserts `sub == username`, and the existing live row has `row.Sub == row.Username`. An OIDC-linked admin row (`sub != username`) is never adopted, so a password login can't take over an IdP-backed account that merely shares the name, and bootstrap explicitly refuses to backfill onto one. Every non-admin collision now just fails on the unique index.
2. Reachability. The htpasswd and init-code paths verify the credential before adoption runs. The one non-password path that resolves identities this way is the transfer bearer middleware, so the question is: which accepted issuer could sign a token with `sub=admin`? The local and origin issuers both sign with this server's own key, so minting one requires the server key itself. The only externally-signed issuer accepted is the federation, and that path is doubly closed: nothing in the director or registry mints the required `pelican.transfer` scope (only the local OAuth2 issuer does, gated by `transferAccessAllowed`), and the federation signing key is a trusted root held by OSG. Reaching local admin through adoption therefore requires federation-root compromise, which is already fatal in our trust model for bigger reasons.

bbockelm:
> fed-in-a-box (used by Pelican e2e tests) is also an example of "multiple servers all sharing the same database"

No ... that's one server with multiple services running, right?  They'll all have the same issuer URL and hence no problem with previously bootstrapped name.

And this feels *fragile*.  For example, there's the following assumption that:

> row.Sub == row.Username

implies it's actually a "locally issued" token and not a "real" issuer.  That's highly biased by CILogon-style subjects but I have no clue how broadly this might be true.

And again, just because it's right now doesn't mean it'll be right after a future refactor.

Having magic takeover of the `admin` account depending on heuristics feels like it'll backfire.

h2zh:
This PR is ready to review @bbockelm @turetske . I reverted the admin adoption commit entirely, and found the way to comply with the new unique username constraint in the fed tests.

bbockelm (approved):
Ok, I'm happy with it now!

PR #3627, opened 2026-08-16
client: list a plain object instead of claiming listings are unsupported

### LINKED ISSUE #3629: `pelican object ls <object>` returned empty against POSIXv2
As it says in the title - `pelican object ls <object>` returns empty against a POSIXv2 origin.

Appears to be an issue in the error handling: there's significant "work around XRootD" code that fails against a more WebDAV compliant origin.

bbockelm:
`pelican object ls <object>` against a POSIX v2 origin returns no listing. It reports:

```
Specification Error: Error code 5000: 405: object listings are not supported by the discovered origin
```

which is doubly wrong — the origin lists collections happily, and the path in question is simply an object rather than a collection.

### What happens

Listing a plain object PROPFINDs it. An origin that answers correctly returns 207 with a single, non-collection response describing the object itself. gowebdav's `ReadDir` treats the first response element as "self", and on finding self is not a collection it synthesizes a `PathError` carrying **405**. `classifyReadDirErr` read that as the server refusing to list.

The existing escape hatch only covered XRootD. XRootD answers `ReadDir` on a plain object with a **500**, which reaches the `Stat` fallback that surfaces a single `FileInfo` — hence listing an object works there. An origin whose PROPFIND is standards-correct never reaches that branch. Same shape as PelicanPlatform/pelican#3626 and PelicanPlatform/xrdcl-pelican#123: the client compensates for an XRootD-specific behavior, and POSIX v2's more correct answer falls through the gap.

### The change

Treat 405 the way 500 is already treated: stat the path first. If it is an object, emit it. Report unsupported listings only when the stat shows a collection, or cannot be performed at all.

An origin that genuinely refuses listings still produces the original error, because the collection it was asked to list stats as a directory and falls through — verified below, not just asserted.

### Testing

New `e2e_fed_tests/list_object_test.go` lists a plain object through a POSIX v2 federation. Confirmed **red before** the change with the exact error above, **green after**.

Regressions checked with the genuine-refusal and collection paths, all passing: `TestPosixv2ListingsCapabilityEnforcement` (origin without the Listings capability), `TestPosixv2OriginListingWithDirector`, `TestFailureOnOriginDisablingListings`, `TestObjectList`, `TestPelicanFS_ReadDir`.

Note for anyone running these locally on macOS: the federation-backed tests need the XRootD client plugin on the library path (`DYLD_FALLBACK_LIBRARY_PATH` pointing at an xrdcl-curl build, plus `XRD_PLUGINCONFDIR`). Without it they fail with ~30s timeouts that look environmental but are just the missing plugin.

### DIFF
--- a/client/handle_http.go
+++ b/client/handle_http.go
@@ -5769,9 +5769,16 @@ func classifyReadDirErr(cli *gowebdav.Client, remotePath string, err error) (fs.
 	if gowebdav.IsErrNotFound(err) {
 		return nil, error_codes.NewSpecification_FileNotFoundError(errors.New("404: object not found"))
 	}
-	if gowebdav.IsErrCode(err, http.StatusInternalServerError) {
-		// The origin returns 500 when asked to ReadDir on a plain object;
-		// fall back to Stat so we can still surface a single FileInfo for it.
+	// Two different errors can mean "this path is an object, not a
+	// collection". An XRootD origin answers ReadDir on a plain object with a
+	// 500. An origin that answers the PROPFIND correctly -- 207 with a single
+	// non-collection response for the object itself -- instead trips
+	// gowebdav's own check that "self" must be a collection, which surfaces as
+	// a synthesized 405. Neither is a refusal to list, and a Stat settles
+	// which case we are in before we report one.
+	isServerErr := gowebdav.IsErrCode(err, http.StatusInternalServerError)
+	isMethodNotAllowed := gowebdav.IsErrCode(err, http.StatusMethodNotAllowed)
+	if isServerErr || isMethodNotAllowed {
 		var info fs.FileInfo
 		statErr := retryWebDavOperation("Stat", func() error {
 			var err error
@@ -5779,13 +5786,16 @@ func classifyReadDirErr(cli *gowebdav.Client, remotePath string, err error) (fs.
 			return err
 		})
 		if statErr != nil {
-			return nil, errors.Wrap(statErr, "failed to stat remote path")
-		}
-		if !info.IsDir() {
+			if isServerErr {
+				return nil, errors.Wrap(statErr, "failed to stat remote path")
+			}
+			// For a 405 a failed stat only means we cannot show the path is an
+			// object; fall through and report that listings are unsupported.
+		} else if !info.IsDir() {
 			return info, nil
 		}
 	}
-	if gowebdav.IsErrCode(err, http.StatusMethodNotAllowed) {
+	if isMethodNotAllowed {
 		// We replace the error from gowebdav with our own because gowebdav
 		// returns "ReadDir /prefix/different-path/: 405" which is not very
 		// user friendly.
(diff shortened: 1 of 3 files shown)


bbockelm:
Pushed a second commit adding `e2e_fed_tests/backend_conformance_test.go`, and it changed my understanding of the scope.

**This was never posixv2-specific.** Sweeping the same matrix across backends, listing a plain object fails identically on **httpsv2**:

| backend | put | stat | list-collection | list-object | get |
|---|---|---|---|---|---|
| posixv2 | ✅ | ✅ | ✅ | ❌ → ✅ | ✅ |
| httpsv2 | ✅ | ✅ | ✅ | ❌ → ✅ | ✅ |

Both fail with the same `405: object listings are not supported by the discovered origin`, and every other cell passes. Verified by reverting the `classifyReadDirErr` change and re-running: exactly the two `list-object` cells go red.

That also explains a loose end — `client/fed_test.go:924-930` is a commented-out assertion, written against an **HTTPS** origin, expecting that 405 message verbatim. It looks like this was hit before on a second backend, encoded as expected behavior, and shelved.

**Why a matrix rather than another one-off test.** Every `DoList` call in the e2e suite targeted a collection; not one listed an object, on any backend. The per-backend files each cover whatever their author needed, so a common operation went untested everywhere at once. Sweeping one matrix means a new backend inherits the whole thing instead of starting from zero.

Measured gaps still open, for whoever wants them (not in this PR):

- **delete** — no `DoDelete` coverage in the posixv2, s3v2, httpsv2 or globusv2 e2e tests
- **cache × backend** — every cache test, v2 and XRootD, runs against posix/posixv2; s3v2 and globusv2 tests contain no cache references at all
- **stat** — no coverage in httpsv2 or globusv2

s3v2 (minio) and globusv2 (mock APIs) both stand their services up locally already, so they slot into the table by adding an entry; I stopped at the two that needed no new fixtures.

bbockelm:
Added **pstore** and **s3v2** to the conformance table. Both stand their dependencies up locally (pstore needs only a directory; s3v2 runs against minio and skips its row when minio is absent), so they cost an entry each.

With four backends in the table the blast radius is measured rather than guessed. Reverting the `classifyReadDirErr` fix and re-running:

| backend | put | stat | list-collection | list-object | get |
|---|---|---|---|---|---|
| posixv2 | ✅ | ✅ | ✅ | ❌ | ✅ |
| pstore | ✅ | ✅ | ✅ | ❌ | ✅ |
| s3v2 | ✅ | ✅ | ✅ | ❌ | ✅ |
| httpsv2 | ✅ | ✅ | ✅ | ❌ | ✅ |

**Listing a plain object was broken on every backend the suite can reach** — one column red, everything else green. With the fix, all four rows pass.

So `pelican object ls <object>` has been failing for every non-XRootD origin backend, not just POSIX v2. XRootD escapes only because it answers `ReadDir` on an object with a 500, which happens to hit the existing `Stat` fallback.

PR #3633, opened 2026-08-18
Move cookie to same-site lax

### LINKED ISSUE #3632: [Pelican v7.27.0] OIDC login loop for /view/cache due to Same-Site=Strict for Cilogon OAuth flow
When trying to do an OIDC login to a v7.27.0-rc.[01] Cache, after the redirect from a successful CILogon Authentication, the user ends up back at the Login page. If the user manually navigates to one of the Cache pages things work as normal.

Notes from Claude response: 

> SameSite=Strict on the login cookie causing it to not be sent to be sent with the /view/cache request, since it's coming back from CIlogon (cross-site)
> 
> [v7.26.0 L352](https://github.com/PelicanPlatform/pelican/blob/v7.26.0/web_ui/authentication.go#L352-L354) → [v7.27.0-rc.1 L476](https://github.com/PelicanPlatform/pelican/blob/v7.27.0-rc.1/web_ui/authentication.go#L476-L482) · [full diff](https://github.com/PelicanPlatform/pelican/compare/v7.26.0...v7.27.0-rc.1)
> 
> Additional note on why only the Cache is affected:
>
> ```go
> // ui.go:81  (was {"config", "origin", "cache"} in v7.26)
> adminAccessPages = []string{"config", "cache"}
> ```
> 
> The comment above it explains `origin` was pulled precisely because middleware-level gating "broke post-login flows whose returnURL pointed at the origin home." A Cache is now the only server whose *home page* is still middleware-gated, so it's the only one where a missing cookie turns into a redirect instead of a page that self-corrects via a same-site `/whoami` XHR. Also note the `pelican-session` cookie ([session.go:45](https://github.com/PelicanPlatform/pelican/blob/v7.27.0-rc.1/web_ui/session.go#L45-L46)) uses gin-contrib defaults (no SameSite → Lax), which is why the OIDC state survives and the login genuinely succeeds.

Rest of Cluade response: [OIDC_Loop_v7.27.0.md](https://github.com/user-attachments/files/31159136/OIDC_Loop_v7.27.0.md)

Related v7.27.0 Testing Jira sub-issue: https://opensciencegrid.atlassian.net/browse/OPS-524

CannonLock:
Fixes the authentication and maintains secure practices.

Needed to be fixed because a AI drive-by determined that we should have samesite strict which broke the authentication. We do not need samesite strict.

Does increase the possibility of security issues should someone add a mutating GET. It is standard practice to not make GETs change state though so I think we are fine.

### DIFF
--- a/web_ui/authentication.go
+++ b/web_ui/authentication.go
@@ -473,12 +473,28 @@ func setLoginCookie(ctx *gin.Context, userRecord *database.User, groups []string
 	}
 
 	// One cookie should be used for all path.
-	// SetSameSite must be called BEFORE SetCookie: gin reads the
-	// configured SameSite mode at SetCookie time, so setting it
-	// afterwards would emit the login cookie with no SameSite attribute
-	// (falling back to the browser default) instead of the intended
-	// Strict.
-	ctx.SetSameSite(http.SameSiteStrictMode)
+	//
+	// SetSameSite must be called BEFORE SetCookie: gin reads the mode at
+	// SetCookie time, so setting it afterwards emits no SameSite
+	// attribute at all and the browser default applies.
+	//
+	// Lax, NOT Strict — load-bearing for OIDC. This cookie is issued by
+	// the OAuth callback, reached via a redirect chain starting at the
+	// identity provider. Browsers compute SameSite context across the
+	// whole chain, so a cross-site hop withholds Strict cookies on every
+	// later hop (a server-side redirect can't launder it): the cookie is
+	// set but not sent on the navigation into the post-login page, which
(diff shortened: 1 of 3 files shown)


CannonLock:
Introduced in #3299 by accident when it fixed things to work as intended. Unfortunately the intended state was broken.

Fixes #3632

h2zh (approved):
LGTM. Manual tests look good. Failed CIs are irrelevant.

Outside the scope of this PR: I think auth middlewares other than `AuthHandler` should also guard with `IsSameOriginRequest`. It's not an emergency because modern browsers already withhold cookie with `SameSite=Lax` on cross-site POSTs.

PR #3502, opened 2026-06-03
Fix various docs issues on main

jhiemstrawisc:
It looks like I missed updating a build tag argument for our code generation check CI. This restores the functionality and commits the drift in generated code (as well as fixes a missing nextra import)

Note that 7.27 docs will not build without this patch. Separate fixes have been opened against 7.26 (#3501) and 7.25 (#3500)

PR #3639, opened 2026-08-19
Image: Install the requirements for the dev image from the osg-testing repos to maintain consistency with where we install the requirements for the prod images from

### LINKED ISSUE #3638: Install dev requirements from osg-testing repo
We install requirements for the prod images from the osg-testing repos, but for the dev images, we install them from the osg [production] repo. This inconsistency can lead to errors such as

> 34.73 Error: Transaction test error:
> 34.73   file /usr/bin/format-json-schema from install of json-schema-validator-devel-2.3.0-2.osg25.el9.aarch64 conflicts with file from package json-schema-validator-2.3.0-3.osg25.el9.aarch64

Install the dev requirements from the osg-testing repos as well.

Also, stop using the osg-contrib repo. We needed that a version of xrootd with a purge plugin patch, before xrootd 5.8.0. There's nothing we need from there now.

matyasselmeci:
Also stop unnecessarily using the osg-contrib repo.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -39,6 +39,7 @@
 ARG ALMALINUX_TAG=9
 ARG BASE_OS=el9
 ARG OSG_SERIES=25
+ARG OSG_REPO=osg-testing
 
 # For compiling custom/one-off releases of Pelican.
 # (We cannot supply a default value because GoReleaser will use it as-is.)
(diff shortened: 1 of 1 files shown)


brianaydemir (changes requested):
@matyasselmeci When I resolved a merge conflict against `main`, I apparently turned this PR into a no-op. All the images seem to build, though.

Is enabling `osg-testing` still the right thing to do in the long term?

matyasselmeci:
Yeah, the problem with using osg-testing for the library dependencies in the prod images but just osg for the header dependencies in the dev images is that it results in an unexpected downgrade (and, in this case, there was a downgrade-breaking organizational change to json-schema-validator-devel, making it obvious).

Another way to solve the problem - perhaps making the reasoning more obvious - is to add an `ARG OSG_REPO=osg-testing` and use that instead of hand-specifying the repos. Would that be better?

brianaydemir:
@matyasselmeci Yeah, some sort of `ARG` to define the repo(s) once sounds good to me.

h2zh (commented):
I noticed that the "Check Go Generate Has Been Run" CI was failed so took an extra look. It turns out that it's caused by another PR #3502 merged in main earlier today.

h2zh:
I will create another PR to fix the "Check Go Generate Has Been Run" CI failure. Thus, it should not be a blocker for this PR.

brianaydemir (approved):
LGTM.

brianaydemir:
@h2zh @matyasselmeci Do you know how far back this needs to go? v7.27 only?

matyasselmeci:
7.27 is fine.

PR #3650, opened 2026-08-21
Upgrade to Go 1.26

### LINKED ISSUE #3649: Upgrade to Go 1.26
Go 1.27 was released on August 19. Per Go's release policy, this meant the end of support for 1.25.

In #3641 (only on the `v7.26.x` branch), we hacked around this by enabling CI/CD to download unsupported versions of the Go toolchain.

For v7.27.x and going forward, we should address this properly.

brianaydemir:
Closes #3649. 

## Summary

The Go language version for Pelican, specified in `go.mod`, has been updated to 1.26.

We do **not** explicitly specify a toolchain version. The directive requires pinning to a specific release, and it is an intentional design choice that we want CI/CD to automatically pick up the most recent version of the toolchain. CI/CD _already_ works like this; I state it here only for emphasis.

While not strictly necessary, it is worth keeping the version of golangci-lint we use up-to-date with respect to the Go version, so that's been updated to v2.12.2 here. (v2.13 is technically available, but it's been out only for a few days. I'm opting to let other projects find any serious bugs in it.)

## Details

The release notes for Go 1.26 are at https://go.dev/doc/go1.26.

Between me and Claude, it would seem that no code in Pelican needs updating for 1.26 specifically. Deprecations in [`crypto/ecdsa`](https://go.dev/pkg/crypto/ecdsa/) are worth addressing, but in separate issues and PRs.

As such, most of the actual changes on this PR relate to golangci-lint:

- The upgrade itself caught deprecated uses of `reflect.Ptr`.

- It dawned upon me to check whether the configuration for golangci-lint accounted for the client/server split (#3343). It did not, so that lead to a round of configuration updates and minor tweaks to code (mostly cleaning up imports).

- The pre-commit hook's configuration has been updated to check the same set of files as the GitHub Action — namely, the entire repository. Previously, it checked only the currently staged changes.

It also turns out that the one function in `local_cache/cache_size_windows.go` had some bugs. Some tool half-fixed them; I opted to fix the rest (that I saw).

## Testing

Aside from the usual smoke tests, we'll need to keep an eye on memory usage in production-like scenarios because the Go runtime [switched to a new garbage collector](https://go.dev/doc/go1.26#new-garbage-collector).


h2zh:
Hi Brian! I'd like to confirm the Go upgrade strategy here.

Go 1.25 support ended on Aug 19, 2026, so `go.dev/dl/?mode=json` no longer lists a 1.25 toolchain and [Build and Test / Cache amd64 layers](https://github.com/PelicanPlatform/pelican/actions/runs/32423166384/job/96601866834?pr=3648) fails on every PR opened against main since then.

Just to confirm the intended split: #3641 (include=all) fixes the 7.26 release branches that stay on 1.25, and this PR fixes main by moving to 1.26? If so, I'll prioritize reviewing this to unblock main CI.

brianaydemir:
@h2zh At the moment, I think we do **not** want to backport this PR to v7.26. That decision might change if we have to do significantly more development on the branch, or if a Go-related security issue pops up.

As far as backporting this PR to v7.27 goes, only the first commit is required. The remainder of the commits are to avoid accumulating technical debt.

You might find it easier to review this PR commit by commit. I tried to structure them so that each one takes care of one, self-contained issue. Reading them all at once makes the 22 changed files look somewhat random.

local_cache/cache_size_windows.go
```diff
@@ -14,8 +15,11 @@ func getCacheSize(_ string, _ *CacheDB, _ StorageID) (cacheSize uint64, err erro
...
 		sizeStr = "10GB"
 	}
-	if signedCacheSize, err := units.ParseStrictBytes(param.LocalCache_Size.GetString()); err == nil {
-		cacheSize = uint64(signedCacheSize)
+	var signedCacheSize int64
+	signedCacheSize, err = units.ParseStrictBytes(sizeStr)
+	if err != nil {
+		return
 	}
+	cacheSize = uint64(signedCacheSize)
```
h2zh, line 23: It would be nice to add an one-line guard `if signedCacheSize < 0`

h2zh (approved):
LGTM after small tweaks. Thanks for orchestrating this PR!

Minor typo in the commit message: in commit a6d05802e3feb797c6e261d1b280d51138258e08, "becomes" should be "because".

local_cache/cache_size_windows.go
brianaydemir: This a good thought, but the non-Windows (i.e., Linux and macOS) version has the same fault:

https://github.com/PelicanPlatform/pelican/blob/fa12d642df07bf9b2fce0f3059fcb24001eacbb4/local_cache/cache_size_unix.go#L45-L50

At this point, I'm willing to say that if an admin configures a negative size, they should be more careful.

PR #3651, opened 2026-08-21
Director Streamling - Rework Logging to reduce blocking states

### LINKED ISSUE #3674: Director Response Speed is Slow
Director is getting enough requests that it is getting bogged down which is resulting in some transfers getting non 307 responses.

Lets investigate why this is happening and fix it.

CannonLock:
On a local director with simulated prod ads it appears the two biggest areas of improvement to speeding up the director is:

**Bug**: We paid to generate and format all logs, even if we threw them away.

The initial logging gate was hardcoded to the lowest level so that the consumers could see all the logs and do their own gating. This resulted in expensive locks on the global mutex for logs that were never consumed by consumers that had their own gates set to higher log levels.

The fix acknowledges that we know the log levels that are set for the consumers at the beginning so we can set the initial gate to the lowest log level of the consumers preventing unneeded locks for logs that are lower.

Current logs in Director redirect hot path that we paid for but didn't use:

Suppressed log call | Location | Count / redirect | Mutex acquisitions | ~Cost each
-- | -- | -- | -- | --
"Skipping '%s' server — filtered/downtime" | sort.go:166 (now ad_index.go:136) | K (per filtered server; 5–10 typical) | 3K | ~4µs
"Request %s requires features %v" | sort.go:377 | 1 | 3 | ~4µs
"will skip origin stats" | stat.go:642 | 2 (fn runs twice) | 6 | ~4µs
"will skip cache stats" | stat.go:647 | 2 (fn runs twice) | 6 | ~4µs
"Sorting caches for request %s" | sort.go:503 | 1 | 3 | ~4µs
"Grabbing coordinate from rand-assignment cache" / "Overriding Geolocation" | sort_utilities.go:506 / 478 | 0–1 | 0–3 | ~4µs
Total wasted |   | ~7 + K ≈ 12–17 | ~36–51 | ~50–70µs

### DIFF
--- a/config/log_buffer.go
+++ b/config/log_buffer.go
@@ -233,6 +233,11 @@ func StartLogRingBuffer(ctx context.Context) {
 	// install cannot be dropped by a rebuild running alongside it.
 	globalTransformMu.Lock()
 	log.AddHook(&logRingBufferHook{buf: buf})
+	// The ring documents an info-and-above capture floor (see shouldBuffer),
+	// so logrus's level gate must admit info entries while the ring is
+	// installed even when the operator configured a quieter level. The
+	// buffer was stored above, so the re-derivation sees it.
+	syncLogrusLevelLocked()
 	globalTransformMu.Unlock()
 }
 
@@ -249,6 +254,12 @@ func StopLogRingBuffer() {
 	// new entries are delivered.
 	buf.closed.Store(true)
 	removeLogRingBufferHook()
+	// With the buffer gone (swapped to nil above), drop logrus's level back
+	// to whatever the remaining consumers need -- the info floor the ring
+	// imposed no longer applies.
+	globalTransformMu.Lock()
+	syncLogrusLevelLocked()
+	globalTransformMu.Unlock()
 	// Stop the compression worker. Cancelling the worker context is sufficient
 	// on its own: compressLoop selects on workerCtx.Done(). We deliberately do
 	// NOT close compressQueue -- Fire performs a non-blocking send on it under
@@ -329,8 +340,14 @@ func (h *logRingBufferHook) Fire(entry *log.Entry) error {
 //
 // The effective level comes from GetEffectiveLogLevel (which knows about
 // the hook-based filtering initFilterLogging installs); logrus's own
-// GetLevel is often pinned to TraceLevel so hooks see everything even when
-// the operator asked for info-only output.
+// GetLevel can sit above the operator's configured level while a registered
+// RegexpFilter needs to observe more verbose entries (see logrusLevelFor),
+// so it is not a reliable statement of what the operator asked for.
+//
+// The "always buffered" tier is only deliverable because logrusLevelFor
+// floors logrus's level gate at info while the ring is installed -- info
+// entries on a warn/error-configured server would otherwise be rejected
+// before any hook fires. If you change that floor, change this contract.
 func shouldBuffer(level log.Level) bool {
 	if level <= log.InfoLevel {
 		return true
--- a/config/log_buffer.go
+++ b/config/log_buffer.go
@@ -234,9 +234,11 @@ func StartLogRingBuffer(ctx context.Context) {
 	globalTransformMu.Lock()
 	log.AddHook(&logRingBufferHook{buf: buf})
 	// The ring documents an info-and-above capture floor (see shouldBuffer),
-	// so logrus's level gate must admit info entries while the ring is
-	// installed even when the operator configured a quieter level. The
-	// buffer was stored above, so the re-derivation sees it.
+	// so register an Info demand: logrus's level gate must admit info entries
+	// while the ring is installed even when the operator configured a quieter
+	// level. It is a floor (not hookOnly), so it holds during the InitServer
+	// window before hook-based filtering is active.
+	registerLevelDemandLocked("log-ring-buffer", log.InfoLevel, false)
 	syncLogrusLevelLocked()
 	globalTransformMu.Unlock()
 }
@@ -254,10 +256,10 @@ func StopLogRingBuffer() {
 	// new entries are delivered.
 	buf.closed.Store(true)
 	removeLogRingBufferHook()
-	// With the buffer gone (swapped to nil above), drop logrus's level back
-	// to whatever the remaining consumers need -- the info floor the ring
-	// imposed no longer applies.
+	// Withdraw the info floor and re-derive: logrus drops back to whatever the
+	// remaining consumers need.
 	globalTransformMu.Lock()
+	releaseLevelDemandLocked("log-ring-buffer")
 	syncLogrusLevelLocked()
 	globalTransformMu.Unlock()
 	// Stop the compression worker. Cancelling the worker context is sufficient
@@ -341,11 +343,12 @@ func (h *logRingBufferHook) Fire(entry *log.Entry) error {
 // The effective level comes from GetEffectiveLogLevel (which knows about
 // the hook-based filtering initFilterLogging installs); logrus's own
 // GetLevel can sit above the operator's configured level while a registered
-// RegexpFilter needs to observe more verbose entries (see logrusLevelFor),
+// RegexpFilter needs to observe more verbose entries (see deriveLevelLocked),
 // so it is not a reliable statement of what the operator asked for.
 //
-// The "always buffered" tier is only deliverable because logrusLevelFor
-// floors logrus's level gate at info while the ring is installed -- info
+// The "always buffered" tier is only deliverable because StartLogRingBuffer
+// registers an info-level demand (see RegisterLevelDemand / deriveLevelLocked)
+// that floors logrus's level gate at info while the ring is installed -- info
 // entries on a warn/error-configured server would otherwise be rejected
 // before any hook fires. If you change that floor, change this contract.
 func shouldBuffer(level log.Level) bool {
--- a/config/logging.go
+++ b/config/logging.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"os"
 	"regexp"
+	"slices"
 	"sync"
 	"sync/atomic"
 
@@ -43,6 +44,13 @@ type (
 	RegexpFilter struct {
 		Regexp *regexp.Regexp
 		Name   string
+		// Levels declares which log levels this filter observes; Fire is only
+		// invoked for entries at these levels. The declaration also feeds
+		// logrusLevelFor, which raises logrus's internal level just far enough
+		// for the filter to see what it asked for. Leaving Levels empty means
+		// "observe everything" -- at the cost of pinning logrus to TraceLevel
+		// (full entry construction for every suppressed log call in the
+		// process) for as long as the filter is registered.
 		Levels []log.Level
 		Fire   func(*log.Entry) error
 	}
@@ -84,12 +92,20 @@ var (
 	// shouldBuffer reads it on every incoming log line, so answering the
 	// query must stay a single atomic load: it cannot take
 	// globalTransformMu or walk the transform hook's LogLevels slice, which
-	// is what the level would otherwise have to be derived from. Keeping
-	// that true is the job of every level-changing site -- SetLogging
-	// (called by the param-callback registered via RegisterLoggingCallback),
-	// initFilterLogging, and ResetGlobalLoggingHooks -- each of which writes
-	// the new level here.
+	// is what the level would otherwise have to be derived from. It is the
+	// authoritative record of the operator-configured level: seeded at
+	// package init and written ONLY by SetLogging (called by the
+	// param-callback registered via RegisterLoggingCallback). Every other
+	// level-changing site derives from it; none may write it.
 	effectiveLogLevel atomic.Uint32
+
+	// testLogFloor is a level floor declared by a test harness whose log hook
+	// forwards entries to t.Log for failure diagnostics (see
+	// test_utils.SetupTestLogging). Zero means "no floor". It participates in
+	// logrusLevelFor the same way the ring buffer and registered filters do,
+	// which is what lets the harness's verbosity survive a mid-test
+	// InitClient/InitServer re-deriving the level from the configured value.
+	testLogFloor atomic.Uint32
 )
 
 func (sw *syncWriter) Write(p []byte) (n int, err error) {
@@ -161,6 +177,12 @@ func (rt *regexpTransformHook) Levels() []log.Level {
 func (fh *RegexpFilterHook) Fire(entry *log.Entry) (err error) {
 	filters := fh.filters.Load()
 	for _, filter := range *filters {
+		// A filter only observes the levels it declared (empty means all).
+		// logrusLevelFor uses the same declaration to decide how far logrus's
+		// level gate must open, so declaration and delivery stay in agreement.
+		if len(filter.Levels) > 0 && !slices.Contains(filter.Levels, entry.Level) {
+			continue
+		}
 		if filter.Regexp.MatchString(entry.Message) {
 			curErr := filter.Fire(entry)
 			if curErr != nil && err == nil {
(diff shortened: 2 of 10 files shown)


CannonLock:
<h2>Benchmarking: how it was done and what it showed</h2>
<p><strong>Symptom:</strong> the OSDF director degrades under load (~100k redirects/min, ≈1,667/s).</p>
<h3>Setup</h3>
<p>A benchmark harness (<code>director/bench_redirect_test.go</code>, seeded from <code>director/testdata/osdf_servers_snapshot.json</code>) drives the real <code>redirectToCache</code> handler end-to-end through gin — recovery, access-log middleware, CORS, handler — under <code>b.RunParallel</code> so lock contention is visible. To make it representative rather than synthetic:</p>
<ul>
<li><strong>Ads:</strong> a captured snapshot of the live OSDF director's 124 advertisements (45 caches, 79 origins) from <code>/api/v1.0/director_ui/servers</code>, including real namespace lists (~30–100 prefixes per cache) and real filtered/downtime state — both turned out to be load-bearing for the findings.</li>
<li><strong>Config:</strong> mirrors production (adaptive sort, presence checks off, the 48 GeoIP overrides, <code>Logging.Level: Debug</code>).</li>
<li><strong>Clients:</strong> 4,096 distinct source /24s, caches pre-warmed, every response asserted to be a 307.</li>
</ul>
<h3>Method</h3>
<ol>
<li><strong>Ablation before profiling.</strong> The same workload runs under logging variants — <code>pinned-debug</code> (production v7.24 behavior: logrus internally pinned to Trace), <code>fixed-debug</code> (level unpinned), and <code>silent</code> (floor) — so the logging pipeline's share of each request is established by subtraction, independent of any profiler.</li>
<li><strong>Attribution with three instruments:</strong> CPU profile (<code>-cpuprofile</code>, read with <code>-cum</code>/<code>-peek</code>), <strong>mutex profile</strong> (<code>-mutexprofile</code> — CPU profiles can't see lock <em>waiting</em>, which is what collapses under load), and allocation accounting (<code>-benchmem</code>).</li>
<li><strong>Fix-as-probe:</strong> each candidate fix was measured against its own control, ×3 runs.</li>
</ol>
<h3>What it revealed</h3>
<ol>
<li><strong>Logging was ~50–70% of every redirect.</strong> The Trace pin (introduced in #3062 to let filter hooks observe all levels) disabled logrus's front level-gate, so every suppressed <code>Tracef</code>/<code>Debugf</code> paid full entry construction: string formatting plus <strong>3 acquisitions of the process-global logger mutex</strong>, one held across a formatting pass written to <code>io.Discard</code>. A redirect makes ~7+K such calls (K = servers in downtime). The mutex profile attributed <strong>99.5% of all lock delay to this</strong>, 59% from a single per-ad Trace line. Changing <code>Logging.Level</code> did nothing (pinned-info ≈ pinned-debug) — the cost was structural. Core-scaling confirmed the serialization: 2→10 cores bought the pinned pipeline <strong>+4%</strong> throughput.</li>
<li><strong><code>LongestNSMatch</code> heap-copied every namespace of every ad per request</strong> — before the prefix test — ≈8,300 allocations (~550KB) per redirect, surfacing as GC consuming <strong>36% of CPU</strong>. Found via <code>-benchmem</code> (8,722 allocs/op for a redirect), not visible in the CPU flat profile because GC cost is billed to the runtime, not its causer.</li>
<li><strong>Ruled out</strong> (measured cheap, contrary to initial code-reading): the ad-snapshot copy/topo-sort (~5% of the scan), geolocation incl. the 48-override walk (625ns), Prometheus counters, config reads.</li>
</ol>
<h3>Results (Apple M5, 10 threads, live ad set)</h3>

State | µs/redirect | allocs/req | Max redirects/s
-- | -- | -- | --
v7.24 behavior (baseline) | ~245 | 8,722 | ~4,100
+ LongestNSMatch fix (a7b0694) | ~176 | 925 | ~5,700
+ logging unpin (780494b) | ~50–57 | 578 | ~18,000+


<p>Combined: <strong>~4.5× throughput</strong>, 15× fewer allocations, with <code>Logging.Level: Debug</code> and all output behavior unchanged. A follow-up (not in this PR) that caches the namespace→ads index takes <code>getAdsForPath</code> from ~24µs to ~103ns/0 allocs and the request to ~32µs.</p>

server_structs/director.go
```diff
@@ -601,33 +601,43 @@ func SetXNamespaceHeaderWithCollections(hdr http.Header, collUrl string, bestNSA
 // LongestNSMatch returns the namespace ad whose path is the longest logical prefix of reqPath.
 // For example, for path `/foo/bar/baz` and namespace ads `/foo` & `/foo/bar`, it returns the
 // ad for `/foo/bar`.  Returns nil if no ad matches.
+//
+// The returned pointer aliases an element of the provided namespaceAds slice
+// (no copy is made): treat it as read-only, and copy the struct before any
+// mutation, or the change will write through to the caller's slice -- which,
+// on the director, is live TTL-cache memory read concurrently by other
+// goroutines.
```
brianaydemir: The aliasing should be removed, not documented around. It introduces fragility for no real gain.

server_structs/director.go
```diff
@@ -601,37 +601,50 @@ func SetXNamespaceHeaderWithCollections(hdr http.Header, collUrl string, bestNSA
...
+		nsPath := namespaceAds[idx].Path
+		nsLen := len(nsPath)
+		if strings.HasSuffix(nsPath, "/") {
+			if !strings.HasPrefix(reqPath, nsPath) {
+				continue
+			}
+		} else {
+			if len(reqPath) <= nsLen || reqPath[nsLen] != '/' || reqPath[:nsLen] != nsPath {
+				continue
+			}
```
brianaydemir: **What:** The non-trailing-slash branch of `LongestNSMatch` open-codes its prefix test as `len(reqPath) <= nsLen || reqPath[nsLen] != '/' || reqPath[:nsLen] != nsPath` (`server_structs/director.go:632`), while the trailing-slash branch four lines above uses `strings.HasPrefix`. Go string slicing allocates nothing, so `strings.HasPrefix(reqPath, nsPath)` would cost exactly the same as the slice comparison it replaces.

**Why it matters:** The two branches express the same idea in two different idioms, and the hand-rolled one puts an index expression and a slice comparison in front of the reader where a named predicate would do -- the kind of asymmetry that invites an off-by-one when this function on the director's redirect path is next edited.

**Fix:** Write the branch as `if !strings.HasPrefix(reqPath, nsPath) || len(reqPath) <= nsLen || reqPath[nsLen] != '/' { continue }` so both branches read alike while keeping the explicit bounds check.

server_structs/director.go
```diff
@@ -601,37 +601,50 @@ func SetXNamespaceHeaderWithCollections(hdr http.Header, collUrl string, bestNSA
 // LongestNSMatch returns the namespace ad whose path is the longest logical prefix of reqPath.
 // For example, for path `/foo/bar/baz` and namespace ads `/foo` & `/foo/bar`, it returns the
 // ad for `/foo/bar`.  Returns nil if no ad matches.
+//
+// The returned ad is a copy, detached from the provided slice, so callers may
+// hold or mutate it freely -- important on the director, where namespaceAds
+// is live TTL-cache memory read concurrently by other goroutines. Only the
+// single best match is copied; the scan itself does not allocate.
```
brianaydemir: Claude: Stop trying to add a comment for behavior that didn't change on this PR, not least of all because your comment is misleading, if not simply wrong.

The returned ad is a **shallow** copy. To tell the caller that they may "mutate it freely" invites mutation of the returned `NamespaceAd`'s embedded slices, with all the consequences that entails.

config/log_buffer.go
```diff
@@ -329,8 +329,9 @@ func (h *logRingBufferHook) Fire(entry *log.Entry) error {
 //
 // The effective level comes from GetEffectiveLogLevel (which knows about
 // the hook-based filtering initFilterLogging installs); logrus's own
-// GetLevel is often pinned to TraceLevel so hooks see everything even when
-// the operator asked for info-only output.
+// GetLevel can sit above the operator's configured level while a registered
+// RegexpFilter needs to observe more verbose entries (see logrusLevelFor),
+// so it is not a reliable statement of what the operator asked for.
 func shouldBuffer(level log.Level) bool {
```
brianaydemir: **What:** `shouldBuffer` in `config/log_buffer.go:336` still promises that `info, warn, error, fatal, panic` are "always buffered" into the in-memory log ring, but `config/logging.go:342` (and `:460`) stopped pinning logrus to `TraceLevel` and now runs logrus at the operator's configured level. logrus gates entry construction on its own level (`Logger.IsLevelEnabled`), so with `Logging.Level: warn` (or `error`/`fatal`/`panic`) an `log.Info(...)` call never produces an entry and `logRingBufferHook.Fire` is never invoked -- the `level <= log.InfoLevel` branch becomes unreachable.

**Why it matters:** The log viewer and `GET /logs/download` silently lose all info-level history on such servers, including XRootD's forwarded stdout/stderr (`daemon.ForwardCommandToLogger` logs every daemon line at `Info`). An operator triaging an incident sees a buffer that looks complete but is missing the lines the code says it retains, and the two-tier design (info+ always retained, debug/trace only when enabled) collapses to "whatever the file already has". The capture also becomes intermittent rather than absent: while `xrootd.LaunchDaemons`'s startup filters are registered, `logrusLevelFor` raises logrus to `Info`, so info lines are buffered for that window and then stop.

**Possible fix:** Have `StartLogRingBuffer` participate in the level derivation the way `logrusLevelFor` does for filters (raise logrus to at least `Info` while the buffer is installed), or drop the unconditional info tier from `shouldBuffer` and its comment so the buffer is documented as tracking the configured level.

test_utils/utils.go
```diff
@@ -562,6 +562,12 @@ func SetupTestLogging(t testing.TB) func() {
 	logrus.SetOutput(io.Discard)
 	logrus.StandardLogger().ReplaceHooks(make(logrus.LevelHooks))
 	logrus.SetReportCaller(true)
+	// The test hook wants to observe every entry regardless of the configured
+	// level: tests assert on debug/trace lines. Production code no longer pins
+	// logrus to TraceLevel (see config.initFilterLogging), so the test harness
+	// must request the verbosity itself.
+	originalLevel := logrus.GetLevel()
+	logrus.SetLevel(logrus.TraceLevel)
```
brianaydemir: **What:** `SetupTestLogging` now raises logrus to `TraceLevel` (`test_utils/utils.go:570`) so its per-test hook still observes debug and trace entries. That raise is silently undone by any test that then initializes a server: `config.InitServer` calls `SetLogging` with the configured level (default `info` for servers), which calls `log.SetLevel`. Each affected test therefore has to re-raise the level after setup, and the tree now does that three different ways -- raw `logrus.SetLevel(logrus.DebugLevel)` (`web_ui/ui_unix_test.go:57`), `config.SetLogging(logrus.DebugLevel)` (`server_utils/server_utils_test.go:49`, `config/env_test.go:37`), and the harness's own Trace raise. Nothing documents which is correct, or that setup order decides whether the raise survives.

**Why it matters:** The failure mode is a test that quietly stops exercising what it asserts -- an assertion on the absence of a log line, or a count of hook entries, passes vacuously once the entries are never constructed -- so the harness can lose coverage without any test turning red. The raw `logrus.SetLevel` variant also leaves logrus's level above `config`'s cached effective level with no filter registered, a state the production invariants in `config/logging.go` describe as filter-only.

**Possible fix:** Add a `test_utils` helper that raises both logrus's level and config's effective level, intended to be called after server initialization, and have the tests that need debug output use only that; note in `SetupTestLogging`'s comment that server initialization resets the level.

h2zh (changes requested):
This PR generally looks good in my manual run, the unpin is clearly the right call for the hot path, and the benchmark methodology is great. I haven't worked on logging before, so I use Claude to grill this PR really hard. The following is a problem that I think it needs a deliberate decision rather than an accidental one:

**The log ring buffer can no longer deliver its documented contract.** `shouldBuffer` (`config/log_buffer.go`) promises:

> info, warn, error, fatal, panic: **always** buffered — debug, trace: only when they are actually enabled

i.e. the `pelican.log_read` triage tail (`/api/v1.0/logs/tail`, `/logs/download`) was designed to keep recent info context even on servers running `Logging.Level: warn|error`. That guarantee was only deliverable while the Trace pin forced logrus to construct every entry. With the pin removed, entries **below** the configured level are rejected before any hook fires, so the floor silently became unreachable.

Note: Logrus Levels from low to high https://github.com/sirupsen/logrus/blob/6d6a132bc03324d4ceb78e1b927f995d014cda20/logrus.go#L103

Reproduced on a live director built from this branch: at `Logging.Level: error`, `/api/v1.0/logs/tail` returned only the error lines; every info/warn line gone. At the `info` default the tail is complete, so most deployments won't notice.

There's also a subtler wrinkle: on an origin/cache configured at `error`, the xrootd startup filters (`Levels: [Info]`) hold logrus's gate at Info during `LaunchDaemons`, so info lines **do** get buffered for that window and then stop once the filters are removed. Time-varying buffer content is an artifact, not a design.

Two ways to resolve it:

1. Floor `logrusLevelFor` at `InfoLevel` while the ring buffer is running (preserves the triage feature; the floor only bites configs quieter than info).
2. Make the narrowing deliberate: the configured level governs the buffer like every other output.

(cross off after BrianA's comment below)
~~We'd argue for option 2 — "the level you configured is the level you get, everywhere" is the cleaner contract, and it also fixes the filter-window inconsistency above. The functional change is three lines:~~

```diff
 func shouldBuffer(level log.Level) bool {
-	if level <= log.InfoLevel {
-		return true
-	}
 	return level <= GetEffectiveLogLevel()
 }
```

…plus rewriting the now-false comments on `shouldBuffer`/`logRingBufferHook.Levels` (the gate's remaining job is rejecting entries that reach the hook only because a registered `RegexpFilter` holds logrus's level above the configured one), and updating the two tests: `TestLogBuffer_LevelGating` now asserts 5/5 entries buffered at `info` and 3/5 at `error` — the `error` case drives entries into `Fire` directly, which is exactly the filter-window scenario — and `TestShouldBufferReadsEffectiveLevel` gains the error-level assertions. Full `config` suite passes with `-race`.

brianaydemir:
@h2zh Regarding the ring buffer: I'd argue Option 1: It is not accidental that someone designed the buffer to always include everything up to Info level (someone went out of their way to write that conditional).

It is what makes the feature useful in the first place: an admin doesn't have to reconfigure the service to get an idea of general activity; they simply log into the web UI. (Keep in mind changing the log level for an origin or cache requires a restart, including waiting for in-flight transfers to clear, and all that.)

config/logging.go
```diff
@@ -257,24 +271,76 @@ func redactEntryCopy(entry *log.Entry) *log.Entry {
...
-	// Our filters may want to see every log message, even those that
-	// are not otherwise printed.  Have the log levels printed via a hook
-	// (instead of the typical output mechanism) so we can crank up the
-	// global log.
+	// Our filters may want to see log messages that are not otherwise
+	// printed. Printing happens via a hook (instead of the typical output
+	// mechanism), so logrus's own level only controls which entries the
+	// hooks get to see -- raise it no further than the filters require.
 	filters := make([]*RegexpFilter, 0)
 	globalFilters.filters.Store(&filters)
```
brianaydemir: The locking discipline in this function should match that of `AddFilter` and `RemoveFilter`. Take the mutex at the top, `defer` the release, and do all the work with the lock held. `initFilterLogging` is the only function (in this file, anyway) that modifies `globalFilters` or reads `addedGlobalFilters` without the lock.

If we're going to add more locking/unlocking in this function, we might as well do it correctly.

test_utils/utils.go
brianaydemir: Claude: Did you even try to think this one through? Your own edit to `web_ui/ui_unix_text.go` should have clued you into the fact that this isn't quite so simple.

There are 2400+ test functions in this code base. ~610 call `SetupTestLogging` directly or indirectly. ~400 lose the `Trace` setting partway through because they call something like `InitClient` or `InitServer` _after_ `SetupTestLogging`. 

You accounted for test assertions. You utterly failed to account for the actual diagnostic output that reaches a developer when a unit test fails. The blast radius of tests that stop emitting diagnostics (via `TestLogHook` forwarding entries to `t.Log`) is significant. If you want a simple example to work with:

```
go test ./client -run '^TestQueryDirector$/^Basic' -v -count=1
```

Find a better solution, or update the PR's description to justify why this change in behavior is acceptable.

brianaydemir (changes requested):

web_ui/ui_unix_test.go
```diff
@@ -51,6 +51,13 @@ func TestDoReload(t *testing.T) {
 	require.NoError(t, param.Server_UIPasswordFile.Set(passwordFile))
 	hook := test.NewGlobal()
 
+	// This test asserts on Debug-level entries; logrus no longer runs above
+	// the configured level (see config.initFilterLogging), so request the
+	// verbosity explicitly.
+	prevLevel := logrus.GetLevel()
+	logrus.SetLevel(logrus.DebugLevel)
+	t.Cleanup(func() { logrus.SetLevel(prevLevel) })
```
h2zh, line 59: We should not use the lorus raw functions anymore since we already have `SetTestLogFloor`.

h2zh (approved):
My comment on ring-buffer was addressed and I can't find any other issues! Btw I think the PR title should be renamed to something more specific, like "Stop paying for log lines nobody reads".

h2zh:
Yesterday, when I was doing some re-review and pre-merge check on this PR, I noticed it was built on top of another large [PR #3512](https://github.com/PelicanPlatform/pelican/pull/3512), which was merged to main after v7.27.x branched off. It means it is not feasible to be cherry-picked into v7.27. I can think of two options:

- Leave this PR for the incoming v26.0 and rollout v7.27.0-rc.3 immediately
- Have the author develop a standalone patch for v7.27 from the scratch.

I'm leaning towards Option 1, given our pacing and man-hour. Then @CannonLock took a look at this matter this morning and agreed that this should place it in 26.0 (scheduled to branch off next week) instead of 7.27.

Takeaway:
A patch PR should branch off the version branch at the first place, not main.

PR #3679, opened 2026-08-26
Initialize config before gin engine build

### LINKED ISSUE #3678: `Server.TrustedProxies` is non-functional
## Summary

#3341 made it so that HTTP proxy headers are no longer trusted by default. The `Server.TrustedProxies` configuration parameter was added so that operators can explicitly set the proxies that they trust.

Unfortunately, the parameter is non-functional. For a director sitting behind a proxy, this breaks its ability to geo-locate clients.

## Details

We initialize the Gin engine before we read any config:

https://github.com/PelicanPlatform/pelican/blob/c0926cdd2367002441c24a2cb3fa5af3945d7433/launchers/launcher.go#L73-L82

As a result, when we read the parameter's value in `GetEngine`…

https://github.com/PelicanPlatform/pelican/blob/c0926cdd2367002441c24a2cb3fa5af3945d7433/web_ui/ui.go#L1125-L1135

…it's empty.

The fix is simply to swap the ordering of `InitServer` and `GetEngine`.

Ironically, we order these calls correctly in unit tests.

h2zh:
Closes #3678 

This should be backported to 7.27, and 7.26 @turetske

### DIFF
--- a/launchers/launcher.go
+++ b/launchers/launcher.go
@@ -70,14 +70,17 @@ func LaunchModules(ctx context.Context, modules server_structs.ServerType) (serv
 
 	config.LogPelicanVersion()
 
-	var engine *gin.Engine
-	engine, err = web_ui.GetEngine()
-	if err != nil {
+	// Config must be initialized before the gin engine is built: GetEngine
+	// reads parameters (e.g. Server.TrustedProxies) that are baked into the
+	// engine at construction time and never re-applied.
+	if err = config.InitServer(ctx, modules); err != nil {
+		err = errors.Wrap(err, "Failure when configuring the server")
 		return
 	}
 
-	if err = config.InitServer(ctx, modules); err != nil {
-		err = errors.Wrap(err, "Failure when configuring the server")
+	var engine *gin.Engine
+	engine, err = web_ui.GetEngine()
+	if err != nil {
 		return
 	}
 
--- a/launchers/launcher_test.go
+++ b/launchers/launcher_test.go
@@ -1,4 +1,4 @@
-//go:build !windows
+//go:build server && !windows
 
 /***************************************************************
  *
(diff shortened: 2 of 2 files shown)


brianaydemir (approved):
LGTM.

@h2zh @turetske This fix needs to be patched back to 7.26. **Cherry pick only the fix commit.** In the interest of time, I'm not holding up this PR over questions about the new unit test; the fix itself is sound.

brianaydemir:
@h2zh Feedback about the new unit test file to consider:

- The build tag should be `server && !windows`. It has nothing to do with the client.

- The closure passed to `t.Cleanup` can exit early due to `require.NoError` failing. Naïvely, I would expect that errors here would simply be ignored or caught somewhere else, because the goal is to run as much clean-up as possible. 

- Use `t.TempDir()` consistently. The call to `os.MkdirTemp` stands apart.

And directly from Claude, because I don't have the time at the moment to dig into the unit test idioms:

- **`viper.Set("config", …)` at :70 bypasses the param layer.** The established idiom is `param.SetRaw("config", path)` (`fed_test_utils/fed.go:310`); this is the only test that imports `spf13/viper` for it. Functionally equivalent — `config.go:1444` reads `viper.GetString("config")` — but off-convention.

- **The logrus hook at :103 is never removed.** It attaches to the global standard logger for the rest of the test binary; `ResetTestState()` in cleanup doesn't remove hooks. With `Logging.Level: debug` and the director/registry background goroutines still logging, it accumulates every entry unboundedly, and it will see — and be seen by — any second test added to this package. The repo idiom is `test_utils.SetupTestLogging(t)`, which saves and restores hook state; it doesn't compose here (it calls `ReplaceHooks`), but a `t.Cleanup` that restores the prior `logrus.StandardLogger().Hooks` would.

h2zh:
The failed tests are not caused by this PR, so I'm gonna merge it. I will investigate the `go generate` issue caused by #3502 seperately today.

PR #3686, opened 2026-08-27
Fail a third-party copy whose stream ends without a success or failure verdict

### LINKED ISSUE #3687: TPC client bugfix - require a terminal state
While working on #3663, I noticed that if the server fails to send a terminal message (success / failure) and just closes the connection, it is assumed to be a success.

We should always make the server declare one way or the other instead of assuming.

Not observed in the wild; don't have any known server behavior that would trigger this.  Just want defense-in-depth.

bbockelm:
Found while answering a question on #3685 about whether any of that PR's fixes changed an exit code. They did not — but this does.

## The bug

The WLCG TPC protocol ends the COPY response body with exactly one verdict line, `success:` or `failure:`. Perf markers are optional; the verdict is not.

`monitorTPC` only treated `failure:` as an error. A stream that ended without *either* line fell out of the scan loop, `scanner.Err()` returned nil at clean EOF, and it emitted `tpcStatus{err: nil, done: true}` — which the consumer's `if msg.err != nil || msg.done` reads exactly like success. `pelican object copy` printed no error and **exited 0**.

The byte count made it worse. On success `transferResults.TransferredBytes` is set from `totalSize`, which is the `Content-Length` of the HEAD against the *source*. A copy that moved nothing reported the full expected size.

Nothing else catches it. The client is not in the data path, so the verdict line is the only report of an outcome that ever arrives. Checksum verification is not a backstop either: it runs only when the source served `Digest` headers, a failure to fetch the destination's checksums is a warning, and a mismatch is fatal only under `requireChecksum`.

## The fix

`monitorTPC` tracks whether a verdict arrived and synthesizes an error when the stream ends without one.

A body truncated mid-chunk does **not** reach that check — the transport surfaces it as an unexpected-EOF read error and `scanner.Err()` already reports it. The case this closes is a clean close with no verdict: a destination whose TPC handler dies without writing one (a panic through gin's recovery, say), or an intermediary that ends the stream early but properly.

## Blast radius

Both server implementations I can check always write a verdict, so no working copy changes behavior:

- XRootD's TPC module.
- Pelican's own `origin_serve/tpc_handler.go` — every terminal path emits one, including the slow-transfer cancel, the size-mismatch check, and the commit-failure path.

## Deliberately not changed

- **`TransferredBytes` reporting the source's expected size.** With a verdict now required, `err == nil` means the destination explicitly affirmed success, and perf markers are optional — a server that emits none would report 0 bytes if this preferred the marker count. Moot for the bug above, and changing it risks worse statistics for a real case.
- **The checksum-verification holes** (soft dest-checksum fetch, mismatch fatal only under `requireChecksum`). Those are policy questions about when a mismatch should sink a transfer, and deserve their own issue and review rather than riding along here.

## Testing

`TestMonitorTPC/EmptyBody` asserted the old reading — that a destination saying nothing at all is a success. It now asserts the opposite. Two subtests added:

- `PerfMarkersThenSilence` — the dangerous shape: real progress markers, so the transfer looks alive and the byte counts land, and then the stream simply stops.
- `SuccessWithoutPerfMarkers` — pins that it is the verdict, not the markers, that decides the outcome.

`./client/ -run TestMonitorTPC` and `origin_serve` `TestTPCWithPOSIXv2` / `TestTPCWithXRootD` pass locally.

## Note for the reviewer

This branch is off `main` and does not depend on #3685, though both touch `client/handle_http_copy.go` in different places. No issue is linked yet, so the second `validate-pr` gate will fail until one is — happy to file one, or link it to whatever you'd prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/client/handle_http_copy.go
+++ b/client/handle_http_copy.go
@@ -406,6 +406,11 @@ func monitorTPC(ctx context.Context, messages chan tpcStatus, body io.Reader) er
 	curStripe := 0
 	curStripeBytes := uint64(0)
 	var err error
+	// The WLCG spec ends the stream with exactly one verdict line, "success:"
+	// or "failure:".  Perf markers are optional; the verdict is not.  Its
+	// absence is tracked rather than assumed benign -- see the check after the
+	// loop for why.
+	sawVerdict := false
 Listener:
 	for scanner.Scan() {
 		text := scanner.Text()
@@ -440,8 +445,10 @@ Listener:
 			switch key {
 			case "failure":
 				err = errors.Errorf("TPC copy failed: %s", value)
+				sawVerdict = true
 				break Listener
 			case "success":
+				sawVerdict = true
 				break Listener
 			case "Stripe Index":
 				idx, pErr := strconv.Atoi(value)
@@ -468,6 +475,23 @@ Listener:
 	}
 	if err == nil {
 		err = scanner.Err()
+	}
+	// Falling out of the loop with no verdict and no read error means the
+	// destination closed the stream cleanly without ever saying how the copy
+	// went.  That is not a success: the client is not in the data path, so the
+	// verdict line is the only thing that ever reports one.  Treating silence
+	// as success reported a completed transfer -- with the source's full byte
+	// count attached, since TransferredBytes comes from the source HEAD -- for
+	// a copy that may have moved nothing at all, and exited 0.
+	//
+	// A body truncated mid-chunk does not land here: the transport surfaces
+	// that as an unexpected-EOF read error, which scanner.Err() reports above.
+	// This is the clean-close case -- a destination whose TPC handler exits
+	// without writing a verdict, or an intermediary that ends the stream
+	// early but properly.
+	if err == nil && !sawVerdict {
+		err = errors.New("the destination ended the third-party-copy stream without reporting success or failure; " +
+			"the copy cannot be assumed to have completed")
 	}
 	select {
 	case messages <- tpcStatus{
--- a/client/http_copy_test.go
+++ b/client/http_copy_test.go
@@ -139,8 +139,51 @@ func TestMonitorTPC(t *testing.T) {
 		}
 	})
 
+	// A destination that says nothing at all has not reported a successful
+	// copy.  This used to be read as success -- the caller got a completed
+	// transfer, carrying the source's byte count, and exited 0.
 	t.Run("EmptyBody", func(t *testing.T) {
 		body := strings.NewReader("")
+
+		messages := make(chan tpcStatus, 1)
+		err := monitorTPC(context.Background(), messages, body)
+		require.NoError(t, err)
+
+		msg := <-messages
+		assert.True(t, msg.done)
+		require.Error(t, msg.err, "a stream carrying no verdict must not be reported as a successful copy")
+		assert.Contains(t, msg.err.Error(), "without reporting success or failure")
+	})
+
+	// The dangerous shape: real progress markers, so the transfer looks alive
+	// and the byte counts land, and then the stream simply stops.
+	t.Run("PerfMarkersThenSilence", func(t *testing.T) {
+		body := strings.NewReader(
+			"Perf Marker\n" +
+				"Stripe Index: 0\n" +
+				"Stripe Bytes Transferred: 1024\n" +
+				"Total Stripe Count: 1\n" +
+				"End\n",
+		)
+
+		messages := make(chan tpcStatus, 2)
+		err := monitorTPC(context.Background(), messages, body)
+		require.NoError(t, err)
+
+		msg1 := <-messages
+		assert.Equal(t, uint64(1024), msg1.xferred)
+		assert.False(t, msg1.done)
+
+		msg2 := <-messages
+		assert.True(t, msg2.done)
+		require.Error(t, msg2.err, "progress markers are not a verdict; only \"success:\" is")
+		assert.Contains(t, msg2.err.Error(), "without reporting success or failure")
+	})
+
+	// The verdict is what decides the outcome, not the presence of markers:
+	// a bare "success:" with no perf markers at all is still a success.
+	t.Run("SuccessWithoutPerfMarkers", func(t *testing.T) {
+		body := strings.NewReader("success: Created\n")
 
 		messages := make(chan tpcStatus, 1)
 		err := monitorTPC(context.Background(), messages, body)


melody0123:
TL;DR version: Please prompt coding agent to "follow Occam's razor principle, and write PR comment in brief and to the point in plain language. No self-invented terms are allowed. If new terms are necessary, define first then use."

---

Long version: An advice on AI technical writing style: The first comment in this thread is too verbose to approach. Coding agent will bury the fact that can be described in simple language in a long essay. Human study new knowledge with focus. There is no focus point in this write-up. A similar analogy is a hard-working student highlight the entire textbook, which is the same as highlighting nothing.

Coding agent also invents terms without proper definition. For example:

> ### Blast radius
> Both server implementations I can check always write a verdict, so no working copy changes behavior:
> - XRootD's TPC module.
> - Pelican's own origin_serve/tpc_handler.go — every terminal path emits one, including the slow-transfer cancel, the size-mismatch check, and the commit-failure path.

What is "blast radius", "verdict", "terminal path", "slow-transfer cancel" and "commit-failure path"? Coding agent lossfully compressed too much info in these self-invented terms. A maintainer who never touched this part of the codebase will get lost when reading these words piled up inorganically by the agent.

melody0123:
Suggested PR comment:

## Terms

- **Third-party copy (TPC):** the client asks a destination server to fetch a file directly from a source server. The file's bytes go source → destination. The client never sees them.
- **Status stream:** the text the destination sends back to the client during a TPC. It may contain progress lines. It must end with exactly one final line: `success: ...` or `failure: ...`. This final line is the only report the client receives of whether the copy worked.

## Bug

If the status stream ended without a final `success:` or `failure:` line, the client treated the copy as successful. `pelican object copy` printed no error and exited 0, even if no bytes were copied.

Cause: `monitorTPC` in `client/handle_http_copy.go` only checked for `failure:`. When the stream closed cleanly with neither line, the read loop ended with no error, and the caller read that as success.

## Fix

`monitorTPC` now records whether a `success:` or `failure:` line arrived. If the stream ends without one, it returns an error and the copy is reported as failed.

## Risk

Copies whose status stream ends without a final line used to pass. They now fail.

Both servers Pelican copies with always write the final line before closing the stream, so no working copy should change behavior:

- Pelican's own server: `origin_serve/tpc_handler.go`. Every exit path writes `success:` or `failure:` first.
- XRootD: [link to spec or source needed]

## Not changed

- The reported byte count on success still comes from the source file's size, not from the progress lines. Progress lines are optional, so using them could report 0 bytes for a real copy.
- Checksum comparison after a copy is unchanged. Its rules for when a mismatch fails a transfer belong in a separate issue.

## Tests

In `client/http_copy_test.go`, `TestMonitorTPC`:

- `EmptyBody`: an empty stream. Previously asserted success. Now asserts failure.
- `PerfMarkersThenSilence` (new): progress lines, then the stream ends. Asserts failure.
- `SuccessWithoutPerfMarkers` (new): only `success: Created`. Asserts success.

Run: `go test ./client/ -run TestMonitorTPC -count=1 -v`

## For the reviewer

No issue is linked yet, so the `validate-pr` check will fail until one is. This branch does not depend on #3685, though both edit `client/handle_http_copy.go`.

Two things I left as gaps on purpose, because I would be inventing evidence otherwise:

1. The XRootD link. The original PR asserts it without a source. Ask the author for one.
2. The "both servers" claim. I wrote "both servers Pelican copies with." Verify there is no third. Grep the repo for other TPC server implementations before accepting that word.

Next: post the draft as a suggestion in the PR conversation, prefixed with "Proposed rewrite for a new-maintainer audience."

client/handle_http_copy.go
```diff
@@ -469,6 +476,23 @@ Listener:
 	if err == nil {
 		err = scanner.Err()
 	}
+	// Falling out of the loop with no verdict and no read error means the
+	// destination closed the stream cleanly without ever saying how the copy
+	// went.  That is not a success: the client is not in the data path, so the
+	// verdict line is the only thing that ever reports one.  Treating silence
+	// as success reported a completed transfer -- with the source's full byte
+	// count attached, since TransferredBytes comes from the source HEAD -- for
+	// a copy that may have moved nothing at all, and exited 0.
```
melody0123, line 485: Starting from "the client is not in the data path", the comment reads like a rambling. No clear purpose is presented to the reader. Consider rewrite this part.

Suggested format, one label per line, one fact per label:

  // When we reach this line: <the exact state, in terms the next `if` tests>
  // What we do and why: <the decision first, then the single fact that makes it correct>
  // Consequence of not doing it: <the wrong path, followed forward in time>

Rules:
- One idea per sentence. No dashes or parentheses splicing in a second idea.
- Describe what a variable holds instead of naming it. If you have to refer to a variable, use back tick to quote it. Hard-encoded string like "success" are fine, but should be quoted with quotation marks. 
- Define jargon inline or drop it.
- No bug history. Describe the code as it is now.

Suggested text:

  // When we reach this line: the destination closed the progress stream
  //   cleanly, no read error occurred, and no "success" or "failure" line
  //   was ever received.
  // What we do and why: report this as a failed copy. The bytes flow directly
  //   from source to destination, never through this client, so the verdict
  //   line is the only evidence that the copy finished. No verdict means
  //   no evidence.
  // Consequence of not doing it: the copy would be reported as successful,
  //   with the full source size counted as bytes transferred, and the command
  //   would exit 0, even if nothing was copied.

client/handle_http_copy.go
melody0123, line 504: Although your PR does not touch this part, but it is still confusing: Why is the `err` not `return`'d, but instead sent by channel? It's worth documentation for this confusion

melody0123 (changes requested):
TL;DR version: The bug is real and the patch is correct. Please address the doc issue described in my comments, and fill in "Projects" + "Milestone" section on the right-side panel of this page.

---

Long version: I can verify the bug is a real and the patch correctly fixes the bug. The only issue with this PR is documentation. PR description/comment is the first place where reviewer will go, so it's important to keep them brief and to the point. As a reviewer, I spent a lot of time just to understand the PR description. Writing docs with coding agent assistance is fine but needs more effort in prompt engineering (i.e., how to prompt coding agent to produce docs understandable and manageable for an engineer new to this part of the codebase). I left my advice on this issue in my comments. Also, "Projects" and "Milestone" section on the right-side panel of this page is not filled in. We are good to go once they are addressed.

PR #3691, opened 2026-08-27
Generate tag-gated CLI reference docs to fix check-go-generate

### LINKED ISSUE #3692: Fix failed check-go-generate CI
PR #3502 extended the "Check Go Generate Has Been Run" workflow to also run `go generate -tags client ./cmd/...` and `go generate -tags server ./cmd/...`. The plain `go generate ./...` invocation never reaches the cmd packages (build constraints exclude their files without those tags), so the generated command-reference pages under docs/app/commands-reference/ were never verified by CI and had silently drifted behind the code. Since #3502 merged, the check fails on main and on every open PR, regardless of the PR's content.

h2zh:
PR #3502 extended the "Check Go Generate Has Been Run" workflow to also run `go generate -tags client ./cmd/...` and `go generate -tags server ./cmd/...`. The plain `go generate ./...` invocation never reaches the cmd packages (build constraints exclude their files without those tags), so the generated command-reference pages under docs/app/commands-reference/ were never verified by CI and had silently drifted behind the code. Since #3502 merged, the check fails on main and on every open PR, regardless of the PR's content.

This commit contains only the output of the three generate commands the workflow now runs — no hand-written changes. Regenerating picks up:

- New pages and index entries for the `pelican-server origin introspect`, `object-metadata`, and `pstore` command trees
- The new `--metadata-body`, `--metadata-content-type`, and `--metadata-file` flags on `pelican object put`
- Corrected `pelican-server` command prefixes in the `cache introspect` usage examples

After this lands, re-running the three generate commands produces an empty diff, so the gate goes green again on main and on rebased PRs.

### DIFF
--- a/docs/app/commands-reference/pelican-server/cache/introspect/consistency/page.mdx
+++ b/docs/app/commands-reference/pelican-server/cache/introspect/consistency/page.mdx
@@ -17,10 +17,10 @@ Use --metadata-only or --data-only to run only one scan type.
 The data scan is expensive as it reads and checksums every cached file.
 
 Example:
-  pelican cache introspect consistency
-  pelican cache introspect consistency --metadata-only
-  pelican cache introspect consistency --data-only
-  pelican cache introspect consistency --json
+  pelican-server cache introspect consistency
+  pelican-server cache introspect consistency --metadata-only
+  pelican-server cache introspect consistency --data-only
+  pelican-server cache introspect consistency --json
 
 ```
 pelican-server cache introspect consistency [flags]
--- a/docs/app/commands-reference/pelican-server/cache/introspect/disk-usage/page.mdx
+++ b/docs/app/commands-reference/pelican-server/cache/introspect/disk-usage/page.mdx
@@ -14,8 +14,8 @@ This is an EXPENSIVE operation that stats every file in the cache's
 storage directories. For large caches this may take significant time.
 
 Example:
-  pelican cache introspect disk-usage
-  pelican cache introspect disk-usage --json
+  pelican-server cache introspect disk-usage
+  pelican-server cache introspect disk-usage --json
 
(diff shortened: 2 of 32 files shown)


brianaydemir (commented):
@h2zh @bbockelm Has anyone checked the commands being documented vs. features that are in the code base but shouldn't be so prominent?

h2zh:
@brianaydemir I didn't check yet, all changes in this PR are auto-generated to pass the "Check Go Generate Has Been Run" CI. I'm also a bit concerned about the unfamiliar commands generated by `go generate -tags client ./cmd/...` and `go generate -tags server ./cmd/...`. That's why I open this seperated PR for these changes. Maybe.. we should not have these two commands in the CI until the `./cmd` gets tidied up?

brianaydemir:
@h2zh Having the commands in the CI is fine: The CI is only _checking_ that `go generate`, in its various incarnations, has been run. It isn't committing any new files; the developer was supposed to do that.

My main concern is that the CLI reference docs generation has been broken since April, and we've added a lot of new, experimental functionality since then, so it's not clear to me what should be publicly documented and what shouldn't.

Or to put this question in terms of code, has anyone reviewed this list recently:

https://github.com/PelicanPlatform/pelican/blob/1540a17528c70436dcac968002cb6db0bac06d55/cmd/doc_gen.go#L40-L60

bbockelm:
> My main concern is that the CLI reference docs generation has been broken since April, and we've added a lot of new, experimental functionality since then, so it's not clear to me what should be publicly documented and what shouldn't.

If it shows up in the CLI's `--help`, it should show up here.  Scanning through the diffs here, I don't see anything I'd object to.

brianaydemir (approved):
LGTM.

PR #3693, opened 2026-08-27
Also enable osg-testing for the dev containers, and add comment explaining why

### LINKED ISSUE #3638: Install dev requirements from osg-testing repo

matyasselmeci:
Fixes a landmine introduced in #3639. osg-development only has the latest version of each package, so if we are pinned to an older version, then only having osg-development enabled will result in an error.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -543,7 +543,10 @@ RUN --mount=type=bind,from=xrootd-build,source=/xrootd-build,target=/xrootd-buil
     else
       # Non-release builds always install from osg-development -
       # these are the "hot-off-the-press" builds of software.
-      dnf install -y --enablerepo=epel-testing --enablerepo=osg-development ${REPO}-${!VER}
+      # We _also_ need to enable osg-testing, because osg-development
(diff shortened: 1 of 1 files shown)


h2zh (approved):
LGTM. The image build works.

bbockelm:
@h2zh - please don't merge anything with failing CI!!  Even if it's clearly not the fault of this PR, it is a bad habit and where bugs can hide.

PR #3695, opened 2026-08-28
Updated the error messages human readable

### LINKED ISSUE #1895: Plain English messages for 403, 404 errors.

turetske:
Updated the 404 and 403 messages to be more
comprehensable to the human user.

close #1895 

Pinging @aowen-uwmad to review the messages themselves.

### DIFF
--- a/client/errorAccum.go
+++ b/client/errorAccum.go
@@ -39,9 +39,15 @@ type PermissionDeniedError struct {
 	noToken bool // true when no token was sent (e.g., public namespace via stale cache)
 }
 
+// credentialRefused is what a 403 actually means, for the reader who does not
+// know the status codes. It is only ever a credential problem: an authorized
+// request for an object that is not there gets a 404, and a namespace no origin
+// serves never reaches a server at all, because the director answers 404 first.
+const credentialRefused = "your credential does not grant access to this object"
+
 func (e *PermissionDeniedError) Error() string {
 	if e.message == "" {
-		return "permission denied"
+		return "permission denied: " + credentialRefused
 	}
 	return e.message
 }
--- a/client/errorAccum.go
+++ b/client/errorAccum.go
@@ -39,10 +39,16 @@ type PermissionDeniedError struct {
 	noToken bool // true when no token was sent (e.g., public namespace via stale cache)
 }
 
-// credentialRefused is what a 403 actually means, for the reader who does not
-// know the status codes. It is only ever a credential problem: an authorized
-// request for an object that is not there gets a 404, and a namespace no origin
-// serves never reaches a server at all, because the director answers 404 first.
+// credentialRefused is what a 403 on a read means, for the reader who does not
+// know the status codes: an authorized request for an object that is not there
+// gets a 404, and a namespace no origin serves never reaches a server at all,
+// because the director answers 404 first, so what is left is the credential.
+//
+// It covers reads only. A 403 on an upload usually means the opposite — the
+// object already exists, and an upload's credential covers creation but not
+// replacement unless Client.EnableOverwrites is set — so PermissionDeniedError
+// (built solely on the download paths) states this as the cause, while
+// StatusCodeError, which both operations reach, offers it as one of two.
 const credentialRefused = "your credential does not grant access to this object"
 
 func (e *PermissionDeniedError) Error() string {
--- a/client/errorAccum.go
+++ b/client/errorAccum.go
@@ -39,16 +39,9 @@ type PermissionDeniedError struct {
 	noToken bool // true when no token was sent (e.g., public namespace via stale cache)
 }
 
-// credentialRefused is what a 403 on a read means, for the reader who does not
-// know the status codes: an authorized request for an object that is not there
-// gets a 404, and a namespace no origin serves never reaches a server at all,
-// because the director answers 404 first, so what is left is the credential.
-//
-// It covers reads only. A 403 on an upload usually means the opposite — the
-// object already exists, and an upload's credential covers creation but not
-// replacement unless Client.EnableOverwrites is set — so PermissionDeniedError
-// (built solely on the download paths) states this as the cause, while
-// StatusCodeError, which both operations reach, offers it as one of two.
+// credentialRefused spells out the 403 above for a reader who does not know
+// the status codes. It describes a read; a write's 403 has other causes, so
+// callers on a write path should say what a refusal means there instead.
 const credentialRefused = "your credential does not grant access to this object"
 
 func (e *PermissionDeniedError) Error() string {
--- a/client/error_helpers.go
+++ b/client/error_helpers.go
@@ -329,8 +329,19 @@ func (e *ConnectionSetupError) Is(target error) bool {
 // StatusCodeError methods
 
 func (e *StatusCodeError) Error() string {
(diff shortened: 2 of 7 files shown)


client/error_helpers.go
```diff
@@ -722,7 +733,7 @@ func wrapDownloadError(err error, transferEndpointURL string, tokenContents stri
 				parts = append(parts, "token expired at "+expiration.Format(time.RFC3339))
 				pde.expired = true
 			} else {
-				parts = append(parts, "token appears valid but was rejected by the server")
+				parts = append(parts, "token has not expired, so it likely does not cover this object's path")
```
aowen-uwmad: I like the addition of `token has not expired`.

I think I understand the second half, but I'm not sure. Also, is there another reason this error would get triggered? The previous message was not helpful because it didn't provide information, but now I worry that `does not cover this object's path` could be a red herring if there is another reason it can be triggered. 

Some comment on line 749.

client/error_helpers.go
```diff
@@ -735,22 +746,22 @@ func wrapDownloadError(err error, transferEndpointURL string, tokenContents stri
...
-				parts = append(parts, "federation token appears valid but was rejected by the server")
+				parts = append(parts, "federation token has not expired, so it likely does not cover this object's path")
 			}
 		}
 		var tokenDetail string
 		if len(parts) == 0 {
 			// No token was sent (e.g., public namespace accessed via a cache that doesn't know about it yet).
 			// Mark as retryable so the client can attempt at a different cache that isn't stale.
-			tokenDetail = "no token was provided"
+			tokenDetail = "no token was sent with the request"
```
aowen-uwmad: This change doesn't appear to impart any new meaning.

If this is just a slight rewording, I guess it's fine.

If the change is supposed to clarify the meaning or cause, then it needs more work.

client/error_helpers.go
turetske: I think I cleared up the ambiguousness by listing all possible reasons

turetske:
@aowen-uwmad Addressed your concerns

client/error_helpers.go
```diff
@@ -722,7 +733,10 @@ func wrapDownloadError(err error, transferEndpointURL string, tokenContents stri
 				parts = append(parts, "token expired at "+expiration.Format(time.RFC3339))
 				pde.expired = true
 			} else {
-				parts = append(parts, "token appears valid but was rejected by the server")
+				// Expiry is the one cause we can rule out from the token alone.
+				// The server does not tell us which of the remaining checks
+				// failed, so name them rather than asserting a single cause.
+				parts = append(parts, "token has not expired, so the cause is one of the server's other checks: its scopes may not cover this object's path or the requested operation, or the server may not accept its issuer or audience")
```
aowen-uwmad:
```suggestion
				parts = append(parts, "token has not expired, so either its scopes do not cover this object's path or the requested operation, or the server does not accept its issuer or audience")
```

client/error_helpers.go
```diff
@@ -735,22 +749,22 @@ func wrapDownloadError(err error, transferEndpointURL string, tokenContents stri
 					pde.expired = true
 				}
 			} else {
-				parts = append(parts, "federation token appears valid but was rejected by the server")
+				parts = append(parts, "federation token has not expired, so the cause is one of the server's other checks: its scopes may not cover this object's path or the requested operation, or the server may not accept its issuer or audience")
```
aowen-uwmad:
```suggestion
				parts = append(parts, "federation token has not expired, so either its scopes do not cover this object's path or the requested operation, or the server does not accept its issuer or audience")
```

aowen-uwmad (approved):
Small tweak to wording.

client/error_helpers.go
```diff
@@ -329,8 +329,19 @@ func (e *ConnectionSetupError) Is(target error) bool {
...
 func (e *StatusCodeError) Error() string {
-	if int(*e) == http.StatusGatewayTimeout {
+	switch int(*e) {
+	case http.StatusGatewayTimeout:
 		return "cache timed out waiting on origin"
+	case http.StatusUnauthorized:
+		return fmt.Sprintf("server returned %d %s: no valid credential was presented for this object",
+			int(*e), http.StatusText(int(*e)))
+	case http.StatusForbidden:
+		return fmt.Sprintf("server returned %d %s: %s", int(*e), http.StatusText(int(*e)), credentialRefused)
```
h2zh: StatusCodeError(403) now says "your credential does not grant access to this object", which looks contradicted by the sibling code in [object PUT](https://github.com/PelicanPlatform/pelican/blob/990fba66d80235bad9f842dd4e1ef04ac0cce33e/client/handle_http.go#L5493): "403 Forbidden, object likely already exists".

client/errorAccum.go
```diff
@@ -39,9 +39,15 @@ type PermissionDeniedError struct {
 	noToken bool // true when no token was sent (e.g., public namespace via stale cache)
 }
 
+// credentialRefused is what a 403 actually means, for the reader who does not
+// know the status codes. It is only ever a credential problem: an authorized
+// request for an object that is not there gets a 404, and a namespace no origin
+// serves never reaches a server at all, because the director answers 404 first.
```
h2zh: same as above. either theisnew error string or the old debug log is wrong.

h2zh (approved):
This PR looks good, I just got one question as below

h2zh (approved):

client/errorAccum.go
```diff
@@ -39,10 +39,16 @@ type PermissionDeniedError struct {
...
+// credentialRefused is what a 403 on a read means, for the reader who does not
+// know the status codes: an authorized request for an object that is not there
+// gets a 404, and a namespace no origin serves never reaches a server at all,
+// because the director answers 404 first, so what is left is the credential.
+//
+// It covers reads only. A 403 on an upload usually means the opposite — the
+// object already exists, and an upload's credential covers creation but not
+// replacement unless Client.EnableOverwrites is set — so PermissionDeniedError
+// (built solely on the download paths) states this as the cause, while
+// StatusCodeError, which both operations reach, offers it as one of two.
```
h2zh: One note inspired by BrianA: I think we should keep Claude’s comments scoped to the code immediately around them. Otherwise, if the referenced feature changes in the future, the comment could become stale and potentially misleading.

PR #3703, opened 2026-08-31
Bind Registration with Pelican User

### LINKED ISSUE #3677: Improve Origin/Cache Registration Workflow
Design Docs
https://docs.google.com/document/d/1vrUL4Sj0PP5jcm_yY7gGsUgxeUuTvTLInP8qU7BETdQ/edit?usp=sharing

h2zh:
Closes #3677

image

This PR comes with the docs change, see `docs/app/federating-your-data/registration/page.mdx`.

## What changes for admins

Today an origin or cache auto-registers as `Pending` with no owner, and the registrant can only find it by hunting through the registry UI. With this branch:

1. Auto-registrations start as `Incomplete` (key material + site name only).
2. The service **logs a completion link** (and writes it to `Server.RegistrationCompletionLinkFile`) every 10 min while incomplete; the link carries a 15-min token signed with the service's own key. If a server admin doesn't bind their account to their Pelican user yet, this will pop up FOREVER.
3. Opening the link, after login, shows a **claim page**; accepting records the logged-in **Pelican user ID** as the owner and drops the user into the registration form. A complete form moves the row to `Pending`.
4. Owners can **transfer ownership** with a single-use, 7-day invite link, and can **edit and resubmit** a `Denied` registration back into the queue.
5. **[For Registry admin]** Legacy rows still owned by a raw CILogon `sub` are **replaced** by the matching Pelican user ID at registry start (see this gated [Google Doc](https://docs.google.com/document/d/1vrUL4Sj0PP5jcm_yY7gGsUgxeUuTvTLInP8qU7BETdQ/edit?usp=sharing)).

Approval stays the trust boundary: the Director only serves approved registrations, and approved rows can only be changed by an admin.

### DIFF
--- a/database/registration_owner_migration.go
+++ b/database/registration_owner_migration.go
@@ -1,0 +1,154 @@
+/***************************************************************
+ *
+ * Copyright (C) 2026, Pelican Project, Morgridge Institute for Research
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License.  You may
+ * obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ***************************************************************/
+
+package database
+
+import (
+	"database/sql"
+	"encoding/json"
+	"regexp"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+// This file serves as the db migration for v26.0 and could be removed later.
+
+// cilogonSubjectRegexp matches the owner encoding the registry used before the
+// 7.27 user model: the raw CILogon OIDC `sub` claim of the user who registered,
+// e.g. http://cilogon.org/serverA/users/12345.
+var cilogonSubjectRegexp = regexp.MustCompile(`^https?://cilogon\.org/server[A-Za-z0-9]+/users/[0-9]+$`)
+
+// migrateLegacyRegistrationOwners re-homes registrations whose
+// admin_metadata.user_id still holds a raw CILogon subject onto the Pelican
+// user ID of the account carrying that subject, so the ownership checks
+// (which compare Pelican user IDs only) recognise the original registrant.
+//
+// Only registrations whose owner value looks like a CILogon subject are
+// considered; rows with an empty owner, a Pelican user ID, or any other
+// value are left untouched (those are handled through other paths, e.g.
+// self claiming link or an admin-minted ownership-transfer invite). A
+// subject is only resolved when exactly one live user carries it; an
+// ambiguous subject present under several issuers is skipped with a warning.
+func migrateLegacyRegistrationOwners(db *gorm.DB) error {
+	type candidate struct {
+		id     int
+		prefix string
+		sub    string
+	}
+
+	// Read the raw JSON: legacy rows may carry an empty or malformed
+	// admin_metadata column that the GORM JSON serializer would reject.
+	var candidates []candidate
+	subjects := make(map[string]struct{})
+	if err := func() error {
+		rows, err := db.Raw("SELECT id, prefix, admin_metadata FROM registrations").Rows()
+		if err != nil {
+			return errors.Wrap(err, "failed to list registrations")
+		}
+		defer rows.Close()
+		for rows.Next() {
+			var (
+				id     int
+				prefix string
+				meta   sql.NullString
+			)
+			if err := rows.Scan(&id, &prefix, &meta); err != nil {
+				return errors.Wrap(err, "failed to scan registration row")
+			}
+			if !meta.Valid || meta.String == "" {
+				continue
+			}
+			var parsed struct {
+				UserID string `json:"user_id"`
+			}
+			if err := json.Unmarshal([]byte(meta.String), &parsed); err != nil {
+				log.Debugf("Skipping registration %d (%s): admin_metadata is not valid JSON: %v", id, prefix, err)
+				continue
+			}
+			if parsed.UserID == "" || !cilogonSubjectRegexp.MatchString(parsed.UserID) {
+				continue
+			}
+			candidates = append(candidates, candidate{id: id, prefix: prefix, sub: parsed.UserID})
+			subjects[parsed.UserID] = struct{}{}
+		}
+		return rows.Err()
+	}(); err != nil {
+		return err
+	}
+	if len(candidates) == 0 {
+		log.Debug("No registrations owned by a legacy CILogon subject; nothing to migrate")
+		return nil
+	}
+
+	// Resolve subjects to Pelican user IDs. The users table is unique on
+	// (sub, issuer), so one subject can exist under several issuers; such
+	// subjects cannot be attributed to a single account and are skipped.
+	// Soft-deleted users are excluded by GORM automatically.
+	subjectList := make([]string, 0, len(subjects))
+	for sub := range subjects {
+		subjectList = append(subjectList, sub)
+	}
+	var users []User
+	if err := db.Select("id", "sub", "issuer").Where("sub IN ?", subjectList).Find(&users).Error; err != nil {
+		return errors.Wrap(err, "failed to look up users by subject")
+	}
+	subToUserID := make(map[string]string, len(users))
+	ambiguous := make(map[string]bool)
+	for _, u := range users {
+		if _, seen := subToUserID[u.Sub]; seen {
+			ambiguous[u.Sub] = true
+			continue
+		}
+		subToUserID[u.Sub] = u.ID
+	}
+
+	rewritten, unresolved, skippedAmbiguous := 0, 0, 0
+	if err := db.Transaction(func(tx *gorm.DB) error {
+		for _, c := range candidates {
+			if ambiguous[c.sub] {
+				skippedAmbiguous++
+				log.Warnf("Not re-homing registration %d (%s): legacy owner %s matches more than one user account", c.id, c.prefix, c.sub)
+				continue
+			}
+			userID, ok := subToUserID[c.sub]
+			if !ok {
+				unresolved++
+				continue
+			}
+			// json_set rewrites only the user_id key and leaves the rest of the
+			// stored metadata byte-for-byte as it was.
+			if err := tx.Exec(
+				"UPDATE registrations SET admin_metadata = json_set(admin_metadata, '$.user_id', ?) WHERE id = ?",
+				userID, c.id,
+			).Error; err != nil {
+				return errors.Wrapf(err, "failed to re-home registration %d (%s)", c.id, c.prefix)
+			}
+			rewritten++
+			log.Infof("Re-homed registration %d (%s) from legacy owner %s to Pelican user %s", c.id, c.prefix, c.sub, userID)
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	log.Infof("Legacy registration owner migration: %d re-homed, %d waiting for the owner's first login, %d ambiguous",
+		rewritten, unresolved, skippedAmbiguous)
+	return nil
+}
--- a/database/universal_migrations/20260704120000_invite_links_registration_ownership.sql
+++ b/database/universal_migrations/20260704120000_invite_links_registration_ownership.sql
@@ -1,0 +1,22 @@
+-- +goose Up
+-- +goose StatementBegin
+
+-- New invite-link kind: 'registration_ownership'. The link, when
+-- redeemed by an authenticated user, transfers ownership of a registry
+-- registration (registrations.admin_metadata user_id) to the redeemer.
+-- Single-use is forced at the application layer, mirroring
+-- collection-ownership invites.
+--
+-- RegistrationID parallels the existing GroupID / TargetUserID /
+-- CollectionID pattern: per-row, defaulted to 0 so older rows pass the
+-- NOT NULL constraint without a backfill. (Registration IDs are the
+-- integer primary keys of the registrations table.)
+ALTER TABLE group_invite_links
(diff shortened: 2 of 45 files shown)


h2zh:
The diff seems large, though many of them are tests/swagger files. Could @turetske review the backend; and @CannonLock review the web UI part?

server_structs/registry.go
```diff
@@ -45,8 +45,8 @@ type RegistrationStatus string
 // endpoint to tell the UI if a field is required. For other validator tags,
 // visit: https://pkg.go.dev/github.com/go-playground/validator/v10
 type AdminMetadata struct {
-	UserID                string             `json:"user_id" post:"exclude"` // "sub" claim of user JWT who requested registration
-	Description           string             `json:"description"`
+	UserID                string             `json:"user_id" post:"exclude" validate:"required"` // "sub" claim of user JWT who requested registration
+	Description           string             `json:"description" validate:"required"`
```
turetske, line 49: What happens to old registrations in the database once "validate: required"? is added

database/universal_migrations/20260704120000_invite_links_registration_ownership.sql
```diff
@@ -0,0 +1,22 @@
+-- +goose Up
```
turetske, line 1: Claude is telling me there is an issue with this being lower than the already shipped 'global_unique_username.sql' in the same directory?

registry/registry.go
turetske, line 1220: Won't this cause all old registrations (which don't have the validated users) to now be incomplete? So previously approved registrations without that will fail. Maybe that's intentional by design, but I thought I would point it out.

database/registration_owner_migration.go
```diff
@@ -0,0 +1,154 @@
...
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+// This file serves as the db migration for v26.0 and could be removed later.
+
+// cilogonSubjectRegexp matches the owner encoding the registry used before the
+// 7.27 user model: the raw CILogon OIDC `sub` claim of the user who registered,
+// e.g. http://cilogon.org/serverA/users/12345.
+var cilogonSubjectRegexp = regexp.MustCompile(`^https?://cilogon\.org/server[A-Za-z0-9]+/users/[0-9]+$`)
```
turetske, line 36: I think that the fact that "--with-identity" is still left in makes this not quite yet "before 7.27", that bit still needs to be removed?

registry/registry_ui.go
turetske, line 682: This is no longer dead code, but allows an isAdmin bypass. Is that still okay? If so, then do you want to add the same bypass at around line 504?

launcher_utils/register_namespace.go
```diff
@@ -287,6 +291,122 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
...
+				log.Warningf("Failed to check registration completeness for %s: %v", prefix, err)
+			} else if result, ok := res.Results[prefix]; !ok {
+				log.Warningf("Registry response does not contain the registration status for %s", prefix)
+			} else if result.Completed {
+				log.Infof("Registration for %s is complete", prefix)
+				if _, path, wErr := updateRegCompletionLinkFile(prefix, ""); wErr != nil {
+					log.Warningf("Failed to update the registration completion link file %s: %v", path, wErr)
+				}
+				return nil
+			} else if result.EditUrl != "" {
```
turetske, line 371: I think there still needs to be a case for result.Completed == False and and empty result.EditUrl?

launcher_utils/register_namespace_test.go
```diff
@@ -531,3 +531,77 @@ func TestReconcileKeysWhenAlreadyRegistered(t *testing.T) {
 	sort.Strings(expected)
 	require.Equal(t, expected, fetchRegistryKids(t, svr.URL, prefix))
 }
+
+func TestUpdateRegCompletionLinkFile(t *testing.T) {
```
turetske, line 535: The various different paths of watchRegistrationCompletion should also be tested

web_ui/groups.go
```diff
@@ -1709,6 +1709,67 @@ func handleRedeemCollectionOwnershipInvite(ctx *gin.Context) {
...
+			Msg:    "Failed to identify user",
+		})
+		return
+	}
+	registrationID, prefix, err := database.RedeemRegistrationOwnershipInviteLink(database.ServerDatabase, req.Token, userId)
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			ctx.JSON(http.StatusNotFound, server_structs.SimpleApiResp{
+				Status: server_structs.RespFailed,
+				Msg:    "invite link not found",
```
turetske, line 1754: Is this possibly too broad? Can the ErrRecordNotFound error occur if something (not the invite link) is missing?

registry/registry_ui.go
```diff
@@ -712,7 +772,7 @@ func getNamespace(ctx *gin.Context) {
 	belongsTo := false
 
 	if !isAdmin { // Not admin, need to check if the namespace belongs to the user
-		belongsTo, err = registrationBelongsToUserId(id, user)
+		belongsTo, err = registrationBelongsToUser(id, userId)
```
turetske, line 775: This line is called multiple times in different if branch, is there reason it can't be done outside the branch only once?

registry/registry_db.go
```diff
@@ -153,15 +153,64 @@ func registrationExistsById(id int) (bool, error) {
...
+// the registration already carries a non-empty owner. Handlers translate it
+// to an HTTP conflict (or forbidden, on the tokenized-edit leg).
+var errRegistrationAlreadyOwned = errors.New("registration already has an owner")
+
+// errRegistrationNotFound is the typed not-found error for registration
+// lookups so handlers can map it to a 404 without string matching.
+var errRegistrationNotFound = errors.New("registration not found in database")
+
+// claimRegistration atomically binds an unowned registration to owner (a
+// Pelican user ID). Ownership is written in two places: here and in an
```
turetske, line 179: I don't like that "Ownership is written in two places" but with two different functions. Specifically the issue that in changing ownership, both functions need to separately call RevokeRegistrationOwnershipInviteLinks. For maintainability of what is a security issue, have both this'd RedeemRegistationOwnershipInviteLink call a separate function which changes the ownership and calls RevokeRegistrationOwnershipInviteLinks, keeping the two functions in line with each other.

turetske (changes requested):

server_structs/registry.go
h2zh: Nothing happens to existing registrations. The `validate` tags are only evaluated in three places, all on request paths: the web-UI create handler, the web-UI update handler, and the completeness check that origins/caches poll (`checkNamespaceComplete`). 

The thrid path determines this registration is "complete" or not. If not, it will throw a claim link through logs/stdout, file, and webui to prompt the server admin to complete the registration. Note that "completed" and "approved" are two different independent checks! For those existing approved registrations with empty UserID, the validator will mark them as "incomplete" but origin still advertises them.

database/universal_migrations/20260704120000_invite_links_registration_ownership.sql
h2zh: good catch!

registry/registry.go
h2zh: Yes, and it's intentional. Same as the [previous comment](https://github.com/PelicanPlatform/pelican/pull/3703#discussion_r3983691809), Completed and Approved are independent checks: the completeness endpoint only feeds the claim-link prompt (log line, link file, origin UI card). Approval is untouched, so the director still serves those namespaces. The prompt is the mechanism that gets our existing registrations bound to a Pelican account.

CannonLock (approved):
UI/UX Looks good.

General UI thought clicking around to apply in the future.

Allow a view to register their Namespace and Origin in one go. It is annoying I have to fill out the same form with the same interface twice. It would be great to go to a page on the registry with a list of updates I need to make then go through them all at once.

database/registration_owner_migration.go
h2zh: I double checked the prod DB and found that there's no non-empty `identity` column, which means no namespace is registered with Pelican CLI "--with-identity" option. But I think it's too aggressive to remove an existing feature, so I just made the patch for it.

registry/registry_ui.go
h2zh: I remove this skip for the admin. Even when an admin create/update a namespace, it should have a owner (non-empty user id)

launcher_utils/register_namespace.go
h2zh: Added. Thanks for catching it!

web_ui/groups.go
h2zh: Right, `RedeemRegistrationOwnershipInviteLink` surfaces a 404 "invite link not found" for

1. a registration deleted after the invite was minted
2. a stale user ID
3. a missing invite link, literally

Now fix the first two misses with return descriptive errors 

1. "the registration ... no longer exists"
2. "redeemer user ... does not exist"

registry/registry_db.go
```diff
@@ -153,15 +153,100 @@ func registrationExistsById(id int) (bool, error) {
...
+
+// ownerIdFromOidcIdentity maps an OIDC identity presented at registration
+// time (the CLI's --with-identity device flow) onto the Pelican user ID that
+// carries it, so the registration is owned by the same account the person
+// would use in the web UI. When the issuer is known the (sub, issuer) pair is
+// looked up exactly, across primary and linked identities; otherwise (or on
+// an issuer miss) the subject alone is used, but only when it resolves to a
+// single account. Returns "" when no (or no unique) account matches, leaving
+// the registration unowned for the regular claim flow.
+func ownerIdFromOidcIdentity(sub string, iss string) string {
```
turetske, line 173: This seems like a dupiclate of database.GetUserByIdentity in collection.go, let's not duplicate code. It's also has bugs that function doesn't have.:

This path checks the users table only. The other path also checks user_identities. So a linked identity resolves when iss is present and silently fails when it is not, and the docs line "matched to the Pelican account carrying it" is only true half the time. Also neither path checks User.Status.

registry/registry_ui.go
```diff
@@ -660,6 +708,24 @@ func createUpdateNamespace(ctx *gin.Context, isUpdate bool) {
 				Msg:    "Fail to update namespace"})
 			return
 		}
+
+		// A fully-valid submission moves an auto-registered (Incomplete)
+		// registration into the admin review queue. Partially-filled updates
+		// (e.g. an admin correcting an unclaimed registration) keep it Incomplete.
```
turetske: Partially filled updates now return 400 before this block is reached. This is an outdated comment.

launcher_utils/register_namespace.go
```diff
@@ -287,6 +291,132 @@ func registerNamespaceImpl(key jwk.Key, prefix string, siteName string, registra
...
+			} else {
+				// Incomplete with no link to offer: the registry could not
+				// resolve the registration (e.g. the prefix is no longer
+				// registered, or a lookup failed on its side). Surface the
+				// registry's explanation instead of silently retrying.
+				msg := strings.TrimSpace(result.Msg)
+				if msg == "" {
+					msg = "the registry returned neither a completion link nor an explanation"
+				}
+				log.Warningf("Registration for %s is incomplete and no completion link is available: %s", prefix, msg)
```
turetske, line 413: This should still call `updateRegcompletionLinkFile(prefix, "")` It logs a stale link, but isn't retiring it. The test for this code missed it because it never wrote the link first and then checked it it was removed

web_ui/groups.go
```diff
@@ -1709,6 +1709,67 @@ func handleRedeemCollectionOwnershipInvite(ctx *gin.Context) {
...
+			})
+		} else {
+			log.Warningf("Failed to redeem registration ownership invite link: %v", err)
+			ctx.JSON(http.StatusBadRequest, server_structs.SimpleApiResp{
+				Status: server_structs.RespFailed,
+				Msg:    "Failed to redeem invite link",
```
turetske: Can we make this message stronger? It's not descriptive and doesn't help the user figure out what the issue might be.

launcher_utils/register_namespace_test.go
```diff
@@ -531,3 +538,289 @@ func TestReconcileKeysWhenAlreadyRegistered(t *testing.T) {
...
+	)
+	setResponse := func(fn func(w http.ResponseWriter)) {
+		respondMu.Lock()
+		defer respondMu.Unlock()
+		respond = fn
+	}
+	statusResponse := func(res server_structs.CheckNamespaceCompleteRes) func(w http.ResponseWriter) {
+		return func(w http.ResponseWriter) {
+			w.Header().Set("Content-Type", "application/json")
+			require.NoError(t, json.NewEncoder(w).Encode(res))
```
turetske: Require in http test is unsafe. This is one place where we should use assert.

launcher_utils/register_namespace_test.go
```diff
@@ -531,3 +538,289 @@ func TestReconcileKeysWhenAlreadyRegistered(t *testing.T) {
...
+		requests.Add(1)
+		respondMu.Lock()
+		fn := respond
+		respondMu.Unlock()
+		fn(w)
+	})
+	svr := httptest.NewServer(mux)
+	defer svr.Close()
+	require.NoError(t, param.Set(param.Federation_RegistryUrl, svr.URL))
+	require.NoError(t, config.InitServer(ctx, server_structs.OriginType))
```
turetske, line 682: This is being called twice in the function. Why? InitServer should only be called once.

turetske (changes requested):

registry/registry_db.go
h2zh: Updated `ownerIdFromOidcIdentity` by calling `GetUserByIdentity` for de-duplication, and adding the check for User.Status. Also, throw a warning if either sub or issuer is missing.

registry/registry_ui.go
h2zh: I bring back the admin bypass (I paste it below) so they can edit pending registrations with empty UserID. Thus, the re-validation here at this promotion code block is not dead any more.
```
				if err.Field() == "UserID" && err.Tag() == "required" && isAdmin {
					continue
				}
```

launcher_utils/register_namespace.go
h2zh: Done

web_ui/groups.go
h2zh: Done

turetske (approved):

PR #3706, opened 2026-09-01
Restructure docs

CannonLock:
I binned things into three users. 

**Accessing your Data**

This is for client users that just want to Read/Write their data. 

**Sharing your Data** ( via an Origin )

This has a split audience that can get narrowed down depending on Pelican's direction. Right now it considers: PI's that manage data access, and Origin administrators.

I feel like the target is to have the Origin so easy to spin up that these are one and the same so I combined them. @bbockelm might want to weigh in.

**Federation Operators** ( Director, Registry, and Cache )

This is for William and Fabio types. Focused on system adminstrators.

I did not update any guides. I think that should be guided by @aowen-uwmad. Reading through things the following docs need updates:

- Delete `Accessing Data`, this is the same as `CLI docs`
- Make group orientated content for the `Start Here` sections.
- `Getting Started` is very general and sticks out in this user focused doc system.

The cli bits are a mess. I need to fix them outside of this PR.


CannonLock:
Screenshot 2026-09-01 at 12 08 50 PM (3)

PR #3707, opened 2026-09-01
Add Server.ConfigReadOnly Knob

### LINKED ISSUE #3647: Make a knob for a Read Only Configuration page
Lets give admins a knob to make the configuration Read Only.

The goal is a limit the security risks posed by a web accessible server as well as reduce the likelihood of someone changing something outside of version control.

CannonLock:
- Add knob to prevent mutating state requests from hitting the config endpoint
- Updates UIs to acknowledge if this is true

### DIFF
--- a/docs/parameters.yaml
+++ b/docs/parameters.yaml
@@ -4340,6 +4340,24 @@ type: filename
 default: ${ConfigBase}/session-secret
 components: ["cache", "director", "origin", "registry"]
 ---
+name: Server.ConfigReadOnly
+description: |+
+  When enabled, the configuration API refuses to change the server's configuration:
+  state changing requests to `/api/v1.0/config` (PATCH, POST, PUT, DELETE) are rejected
+  and the web UI renders its configuration page as a read-only view.
+
+  Use this on servers whose configuration is managed outside of Pelican -- by Puppet,
+  Ansible, a container image, etc. -- so that a web UI administrator cannot introduce
+  drift by writing to `Server.WebConfigFile`.
+
+  Unlike `Server.WebReadOnly`, this affects only the configuration API; every other web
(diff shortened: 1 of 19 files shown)


docs/parameters.yaml
```diff
@@ -4340,6 +4340,24 @@ type: filename
...
+  Use this on servers whose configuration is managed outside of Pelican -- by Puppet,
+  Ansible, a container image, etc. -- so that a web UI administrator cannot introduce
+  drift by writing to `Server.WebConfigFile`.
+
+  Unlike `Server.WebReadOnly`, this affects only the configuration API; every other web
+  UI route keeps accepting state changing methods. Since the configuration API is the
+  only way to change this value from the web UI, enabling it means it can only be
+  disabled again by editing a configuration file and restarting the server.
+type: bool
+default: false
```
h2zh, line 4358: Should we set the default to true? (maybe this is an ops question?)

h2zh (changes requested):
The feature tests good! Though I think the current UX is kinda annoying to the user - on the config tab of web ui, user changes a config field and click "save", then the rejection error pops up. Would it be easy to just block the editing for all config fields when `Server.ConfigReadOnly` is true?

CannonLock:
@h2zh Is there a chance you didn't rebuild the UI when you tested? I did exactly what you suggested, curious if there is something i missed that made the read only interface not trigger.

h2zh (approved):
LGTM. You're right, I didn't include the webUI build in the goreleaser rebuild. One nitpick: in config read only mode, the unknown network ip geolocation input should've been disabled. It is not a big deal cuz the middleware eventually blocks the PUT request, also only the federation admin uses this feature. 
image

PR #3722, opened 2026-09-08
Fix the crash loop for embedded issuer with all-public namespaces

### LINKED ISSUE #3719: [Pelican v7.27.0] Embedded Issuer + All-Public Origin crash loop
It seems like that when the Issuer is set to embedded, there’s a check for a first auth namespace, which fails over to the check for the oa4mp check for /api/v1.0/issuer/.well-known/openid-configuration if the Origin doesn't have a non-public namespace. That check seems to fail with a 404 (due to not serving it?).

```
Error: url https://pelicanplatform2-origin.osdf-dev.chtc.io/api/v1.0/issuer/.well-known/openid-configuration didn't respond with the expected status code 200 within the timeout of 30s (Server.StartupTimeout): received bad status code from https://pelicanplatform2-origin.osdf-dev.chtc.io/api/v1.0/issuer/.well-known/openid-configuration: 404 (expected 200), body: 404 page not found
```

Current v7.27.0 Issuer check: https://github.com/PelicanPlatform/pelican/blob/2df58ebbdfc1443e8347bd3ca40099c9cc339447/launchers/launcher.go#L258-L282 fails for Public Origin w/ embedded issuer.

Current Main Issuer check: https://github.com/PelicanPlatform/pelican/blob/f81f2fbcd525a09d2eb01c47fc6bf216364c2002/launchers/launcher.go#L300-L333 claims to fix this.

We might need to port the launcher change from https://github.com/PelicanPlatform/pelican/commit/4ac260c2e57c517d5d2ce561daefab654280afb5   to v7.27.0? I don’t know if that commit is build on top anything else from main.

This is relevant for the Web Client for Origins since that requires the embedded issuer

Paired Jira issue : https://opensciencegrid.atlassian.net/browse/OPS-545

h2zh:
Skip issuer health check on an all-public embedded-issuer origin

This PR only needs to merge into v7.27.x, see the reason below.

An origin with the embedded issuer enabled but only public-read exports crash-looped at startup on v7.27.0 (issue #3719). The issuer startup health check probes the first auth-requiring export's namespace discovery document; when no export requires authentication it fell back to the namespace-less OA4MP path
(/api/v1.0/issuer/.well-known/openid-configuration), which the embedded issuer never serves. WaitUntilWorking then 404'd until Server.StartupTimeout and LaunchModules returned an error.

Main fixes this by health-checking the server's local issuer (4ac260c2e), but that relies on the local issuer being registered whenever the embedded issuer is enabled, which is not on v7.27.x. As a stopgap for the release branch, skip the issuer health check entirely when the embedded issuer is in use and no export requires authentication: there is no registered provider to probe. OA4MP mode and the export-lookup-failure path keep the existing fallback.

Adds TestAllPublicOriginStartsWithEmbeddedIssuer, which brings up an all-public embedded-issuer origin through the real launcher and asserts it starts and serves no issuer provider. Verified the test fails without the fix (reproducing the startup timeout from the issue) and passes with it.

Closes #3719

### DIFF
--- a/e2e_fed_tests/origin_all_public_issuer_e2e_test.go
+++ b/e2e_fed_tests/origin_all_public_issuer_e2e_test.go
@@ -1,4 +1,4 @@
-//go:build server && !windows
+//go:build !windows
 
 /***************************************************************
  *
@@ -22,7 +22,6 @@ package fed_tests
 
 import (
 	"fmt"
-	"io"
 	"net/http"
 	"testing"
 
@@ -52,28 +51,11 @@ Origin:
 // TestAllPublicOriginStartsWithEmbeddedIssuer verifies that an origin with the
 // embedded issuer enabled but only public-read exports starts cleanly
 // (issue #3719).
-//
-// The embedded issuer creates a provider per auth-requiring export, so an
-// all-public origin registers none and serves no issuer discovery endpoint.
-// The launcher's issuer startup health check used to fall back to the
-// namespace-less OA4MP discovery path in that case, which the embedded issuer
-// never serves; the probe 404'd until Server.StartupTimeout and the whole
-// launch failed, crash-looping the origin. The launcher now skips the issuer
-// health check when no export requires authentication.
-//
-// Note: on main the same configuration instead health-checks the server's
-// local issuer, which is registered whenever the embedded issuer is enabled
-// (commit 4ac260c2e). That feature is not on v7.27.x, so this test also pins
-// that no issuer provider exists on an all-public origin — if the local
-// issuer is ever backported, the second assertion flips and the launcher
-// skip should be replaced with the health check from 4ac260c2e.
 func TestAllPublicOriginStartsWithEmbeddedIssuer(t *testing.T) {
 	t.Cleanup(test_utils.SetupTestLogging(t))
 	server_utils.ResetTestState()
 	t.Cleanup(func() { server_utils.ResetTestState() })
 
-	// Without the launcher fix the fed never finishes starting: the issuer
-	// health check probes a URL nothing serves and LaunchModules errors out.
 	originConfig := fmt.Sprintf(allPublicEmbeddedIssuerConfig, t.TempDir())
 	ft := fed_test_utils.NewFedTest(t, originConfig)
 	require.NotNil(t, ft, "an all-public embedded-issuer origin must start cleanly")
@@ -81,19 +63,8 @@ func TestAllPublicOriginStartsWithEmbeddedIssuer(t *testing.T) {
 	serverURL := param.Server_ExternalWebUrl.GetString()
 	httpClient := &http.Client{Transport: config.GetTransport()}
 
-	// The server is up and healthy — the same endpoint the launcher gates on.
 	resp, err := httpClient.Get(serverURL + "/api/v1.0/health")
 	require.NoError(t, err)
 	defer resp.Body.Close()
 	require.Equal(t, http.StatusOK, resp.StatusCode, "the origin's health endpoint must respond after startup")
-
-	// No issuer provider is registered for the public export: its discovery
-	// document is not served. This is why the launcher must skip the issuer
-	// health check rather than probe anything.
-	discResp, err := httpClient.Get(serverURL + "/api/v1.0/issuer/ns/public-data/.well-known/openid-configuration")
-	require.NoError(t, err)
-	defer discResp.Body.Close()
-	body, _ := io.ReadAll(discResp.Body)
-	require.Equal(t, http.StatusNotFound, discResp.StatusCode,
-		"an all-public origin must register no issuer provider (body: %s)", string(body))
 }
--- a/e2e_fed_tests/origin_all_public_issuer_e2e_test.go
+++ b/e2e_fed_tests/origin_all_public_issuer_e2e_test.go
@@ -1,0 +1,99 @@
+//go:build server && !windows
+
+/***************************************************************
+ *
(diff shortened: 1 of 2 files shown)


h2zh:
The failed Linter CI is because this PR is branching off `v7.27.x` branch, which is not a blocker.

> Error: cmd/token.go:571:1: File is not properly formatted (gofmt)

e2e_fed_tests/origin_all_public_issuer_e2e_test.go
```diff
@@ -0,0 +1,99 @@
+//go:build server && !windows
```
brianaydemir: The `server` build tag here is not necessary and, arguably, also incorrect.

The only places in the repo that specify either the `client` or `server` build tag are the `./cmd/` package. These tags control what goes into the final binary; they do _not_ label a file's functional area. And indeed, nothing else in `./e2e_fed_tests/` specifies either of these build tags, despite there being other e2e tests that test server-only functionality.

e2e_fed_tests/origin_all_public_issuer_e2e_test.go
```diff
@@ -0,0 +1,99 @@
...
+// never serves; the probe 404'd until Server.StartupTimeout and the whole
+// launch failed, crash-looping the origin. The launcher now skips the issuer
+// health check when no export requires authentication.
+//
+// Note: on main the same configuration instead health-checks the server's
+// local issuer, which is registered whenever the embedded issuer is enabled
+// (commit 4ac260c2e). That feature is not on v7.27.x, so this test also pins
+// that no issuer provider exists on an all-public origin — if the local
+// issuer is ever backported, the second assertion flips and the launcher
+// skip should be replaced with the health check from 4ac260c2e.
```
brianaydemir: These two paragraphs are exactly why I don't like Claude (or whatever model's) comments:

They refer to code and functionality far outside of the the thing being commented (in this case, the test), which means they potentially become stale or inaccurate whenever those faraway places are edited (`main` is an especially bad reference: it's always in a constant state of flux). That is a maintenance nightmare and exactly the sort of thing that both people and A.I.s tend to forget about.

The first paragraph already explains the behavior that the test is trying to pin down, and that behavior is self-evidently good and reasonable.

e2e_fed_tests/origin_all_public_issuer_e2e_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// No issuer provider is registered for the public export: its discovery
+	// document is not served. This is why the launcher must skip the issuer
+	// health check rather than probe anything.
+	discResp, err := httpClient.Get(serverURL + "/api/v1.0/issuer/ns/public-data/.well-known/openid-configuration")
+	require.NoError(t, err)
+	defer discResp.Body.Close()
+	body, _ := io.ReadAll(discResp.Body)
+	require.Equal(t, http.StatusNotFound, discResp.StatusCode,
+		"an all-public origin must register no issuer provider (body: %s)", string(body))
```
brianaydemir: > That feature is not on v7.27.x, so this test also pins that no issuer provider exists on an all-public origin — if the local issuer is ever backported, **the second assertion flips** and the launcher

If the observation is, "you might need to change this test later", is it _actually_ testing something important?

brianaydemir (approved):
Approving.

The comment about the build tag isn't critical. The rest is me pontificating.

e2e_fed_tests/origin_all_public_issuer_e2e_test.go
h2zh: Good point — removed in bc277479e. You're right that these tags only gate what lands in the final binaries via ./cmd/, and no other file in e2e_fed_tests/ uses them.

e2e_fed_tests/origin_all_public_issuer_e2e_test.go
h2zh: Fair — trimmed the doc comment down to the first paragraph in bc277479e. The mechanism/backport context lives in the PR description and issue #3719 where it won't go stale in the code.

e2e_fed_tests/origin_all_public_issuer_e2e_test.go
h2zh: Agreed — the assertion was pinning an implementation detail rather than behavior anyone depends on. Removed it in bc277479e; the test now only asserts the origin starts and its health endpoint responds, which is the actual regression from #3719.

e2e_fed_tests/origin_all_public_issuer_e2e_test.go
h2zh: I totally agree. I'm thinking of is there any principles we can enforce to avoid the verbose comment generated by AI. Maybe we can discuss this in the next Pelican (dev) meeting?

brianaydemir: My two cents: The problem is that when Claude adds comments and documentation for a change it's making, it documents _the change and why_. That works well for a PR description or a Git commit message. It works less well for comments in code for two reasons:

1. It often leads to comments referring to things "faraway" from the thing being commented on. That's distasteful for the same reason global variables are often distasteful: it interferes with one's ability to think and work locally.

2. A comment generally doesn't need to explain how things were in the past; the version control system (Git) already answers that. Instead, a comment should simply focus on how things _currently are_. The reason to document alternative implementations is when someone or something might consider them to be a "reasonable improvement" and to dissuade them from pursing those alternatives.

PR #3732, opened 2026-09-09
Fix error hierarchical matching

### LINKED ISSUE #3694: Error.Is not properly matching
The CLI never reports the exit codes documented in `docs/error_codes.yaml`. Every
classified failure exits 1 (or 11 when retryable) instead.

Four subcommands — `object get`, `sync`, `prestage`, `evict` — guard the exit-code
branch with:

    var pe error_codes.PelicanError
    if errors.Is(err, &pe) {
        os.Exit(pe.ExitCode())
    }

`pe` is a freshly-declared zero value, not a sentinel, and `PelicanError` defines no
`Is` method. `errors.Is` therefore falls back to comparing each error in the chain
against that brand-new pointer, which nothing can equal. The branch is dead and the
`else` fallback always wins.

The remaining four — `object put`, `ls`, `copy`, `stat` — have no exit-code branch at
all, so they exit 1 even when the error was classified correctly.

There is a related defect in the cache's error handling. `local_cache`'s `handleError`
uses `errors.As` to reach the classification, which stops at the *first* PelicanError
in tree order. For a transfer accumulator that is whichever object server happened to
answer first, so when one origin was throttled and another answered "not found", the
ordering — not the meaning — decides whether the client sees 404 or 500.

Exit 7, the `clientExitCode` documented for `Authorization` (code 4000).

Actual: exit 1. The error is classified correctly and the message even says
`Authorization Error: Error code 4000` — only the exit code is lost.

The underlying gap is that the taxonomy has no sentinels to compare against, so
callers can only ask type-shaped questions (`errors.As`) when what they mean is
"is this that error".

Suggested fix:

- Generate one sentinel per entry in `docs/error_codes.yaml` (`ErrAuthorization`,
  `ErrSpecification_FileNotFound`, …) alongside the existing constructors.
- Give `*PelicanError` an `Is` method matching on the dotted taxonomy, so
  `errors.Is(err, ErrAuthorization)` is also true of an `Authorization.TokenNotFound`.
- Add an exit-code lookup driven by those sentinels. The exit code is a property of
  the classification, not of the individual error, so a matched sentinel supplies it
  without touching the instance.

This also fixes the cache case: `errors.Is` walks the whole error tree, including the
`Unwrap() []error` branches an accumulator produces, so a definitive answer is found
wherever it sits rather than wherever it happened to land first.

turetske:
closes #3694 

Four failure paths in the client reached users with the wrong error family, or
none at all, and the exit-code taxonomy in `docs/error_codes.yaml` never
reached anyone using the CLI. This fixes both, and makes exit code 11 mean
exactly one thing: the failure is worth retrying. The exit code 11 meaning retry 
is a requirement by condor.


melody0123:
Reviewer comment:

The PR description uses two terms invented by either the author or the
author's coding agent ("error family" and "exit-code taxonomy"). Neither
is defined by Go or used elsewhere in Pelican, which degrades readability
for anyone who touches this part of the codebase for the first time or
long from now. Define such a term before first use, or do not invent one
at all. 

If a coding agent writes the PR description, consider this
prompt: Follow Occam's razor. Keep the explanation brief and to the
point. Do not use self-invented terms. If inventing a term is necessary
or beneficial, define it first, and do not use another invented term in
the definition.

Rewritten description:

closes #3694

Four failure paths in the client reached users with the wrong error
type, or none at all. The error type is the name Pelican gives an error,
such as Parameter.FileNotFound; docs/error_codes.yaml maps each type to
an exit code. The client never used that table when exiting, so users of
the CLI never saw those codes. This fixes both, and makes exit code 11
mean exactly one thing: the failure is worth retrying. HTCondor requires
that meaning, because it decides whether to resubmit a job from the exit
code alone.

docs/error_codes.yaml
```diff
@@ -34,6 +34,15 @@
 # This means that the whole client process will restart with the HTCondor job so the process could succeed
 # if it hits a different cache/site or hits the same cache/site when it is in a better state.
 
+# clientExitCode and retryable are not independent: exit code 11 is reserved for, and required by, every
```
melody0123: > exit code 11 is reserved for, and required by, every retryable error

I feel that this is an "if and only if" relationship:
- exit code 11 is reserved for retryable error: if an error has exit code 11, it has to be a retryable error.
- exit code 11 is required by retryable error: if an error is retryable, its exit code has to be 11.

If this is the case, I would suggest to use "if and only if" term to describe this relationship -- It's formal logical terminology for no ambiguity and better readability

docs/error_codes.yaml
```diff
@@ -34,6 +34,15 @@
...
+# retryable error, so a wrapper can decide whether to resubmit from the exit status alone without having
+# to consult this table. Non-retryable errors take a code in 1-10, which then says which *kind* of
+# permanent failure it was. Every retryable entry below therefore has clientExitCode 11, and no
+# non-retryable entry may claim it. An entry that breaks this is rejected when this file is processed
+# rather than left to review, since the mistake is invisible until a job either fails to retry or
+# retries forever. Codes 1-3 stay unassigned: 1 remains the exit for an error that reached the client
```
melody0123, line 43: > An entry that breaks this is rejected when this file is processed
> rather than left to review, since the mistake is invisible until a job either fails to retry or
> retries forever.

I'm still confused about when this file is processed. By saying "processed", do you mean CI pipeline checks it, or code generator reads and generates sentinels?

Also, what does the "review" refer to? Does it mean human developers review PR?

docs/error_codes.yaml
```diff
@@ -34,6 +34,15 @@
 # This means that the whole client process will restart with the HTCondor job so the process could succeed
 # if it hits a different cache/site or hits the same cache/site when it is in a better state.
 
+# clientExitCode and retryable are not independent: exit code 11 is reserved for, and required by, every
+# retryable error, so a wrapper can decide whether to resubmit from the exit status alone without having
```
melody0123, line 38: > exit code 11 is reserved for, and required by, every retryable error

I feel that this is an "if and only if" relationship:
- exit code 11 is reserved for retryable error: if an error has exit code 11, it has to be a retryable error.
- exit code 11 is required by retryable error: if an error is retryable, its exit code has to be 11.

If this is the case, I would suggest to use "if and only if" term to describe this relationship -- It's formal logical terminology for no ambiguity and better readability

docs/error_codes.yaml
```diff
@@ -124,15 +133,15 @@ retryable: false
 ---
 type: Contact.ConnectionReset
 code: 3005
-clientExitCode: 6
+clientExitCode: 11
 description: >-
   The client attempted to contact a server but the connection was reset by the remote peer.
   This is often a transient network error that can be resolved by retrying.
 retryable: true
```
melody0123, line 140: I assume the `ConnectionReset` here means RST bit = 1 in TCP packet. Whether it is retryable based in the context. If this happens during connection establishment, this is usually a permanent error (e.g., SSH aborts immediately if RST during connection establishment).

Edit: After reading the `ConnectSetup` entry below, I think `ConnectionReset` happens after TCP connection is established? If that is the case, retry is justified because this usually means server suddenly crashes or a middle man on the path goes wrong (and route may change later to avoid that broken node in the network).

docs/error_codes.yaml
```diff
@@ -124,15 +133,15 @@ retryable: false
...
 ---
 type: Contact.ConnectionSetup
 code: 3006
-clientExitCode: 6
+clientExitCode: 11
 description: >-
   The client attempted to establish a connection to the server but failed before the request could be completed.
 retryable: true
```
melody0123, line 147: Same problem as above. I think TCP RST = 1 happens during connection establishment is a stronger case where the error is not retryable. Same example: SSH client receives TCP RST = 1 during handshake.

melody0123 (changes requested):
I reviewed docs based on my understanding of the problem: some retryable errors are previously assigned with wrong exit code other than 11. HTCondor relies on exit code 11 to decide if a failed pelican task should be retried or not, so wrongly assigned exit code will not trigger HTCondor retry. This patch re-assigned correct exit code 11 to those retryable errors.

Due to the knowledge gap (lack of Go basics and understanding of the entire codebase), I just reviewed docs instead of judging the correctness of the code, including test code. This is a job for someone who knows both better than me.

docs/error_codes.yaml
turetske: Pushing back on this:

Contact.ConnectionSetup is a catch-all. It can be: isContextDeadlineError || isDNSError || (isTLSError && !isTLSCertificateValidationError) || isDialError. Marking it non-retryable would stop DNS failures and connection timeouts, which can definitely be transient. ,from being retried. 

TLS certificate validation failures are wrapped as Specification errors precisely because those aren't retryable, so the case you are worried about isn't here.

It's possible the process could succeed if it hits a different cache/site or hits the same cache/site when it is in a better state. Retry means job resubmission, possibly to a different cache, not necessarily to the exact same host. So this is not always the non-retryable.

docs/error_codes.yaml
turetske: Yes. This fires on syscall.ECONNReset or sys call.EPIPE

docs/error_codes.yaml
melody0123: OK, verified in `client/error_helpers.go`. This makes sense. `Contact.ConnectionSetup` is retryable.

error_codes/matching.go
```diff
@@ -0,0 +1,95 @@
...
+	"strings"
+)
+
+// ErrPelican matches any classified Pelican error, whatever its type. It is the
+// sentinel to use when the question is "did anything classify this at all",
+// which is what the CLI exit paths ask before choosing an exit code.
+var ErrPelican = &PelicanError{}
+
+func (e *PelicanError) Is(target error) bool {
+	if exact, ok := target.(exactMatch); ok {
```
melody0123: I cannot understand the embedded struct used here, and unfortunately, it seems that this is the core of this fix.

melody0123:
I gaved this PR one last shot. I checked `error_codes.go` and it looks correct to me. However, when reading `matching.go`, the file with core mechanism in this PR, I was blocked by the custom `Is()` method, and cannot understand the purpose of inventing the type `exactMatch`. 

To the best my knowledge in Go, sentinel errors are defined as package-level values, are never returned, and are just used to compare in the `errors.Is()` function. But in our case, they are not directly compared, but are wrapped by `exactMatch`, and then compared. I cannot understand this trick. `exactMatch` type seems to be a simple wrapper around `*PelicanError` without any additional data members or methods, and I'm confusing by the purpose of having it. 

Due to my existing gap in Go and understanding the core mechanism of this patch, I would pass this PR to @bbockelm for further review.

turetske:
@melody0123 

> I gaved this PR one last shot. I checked `error_codes.go` and it looks correct to me. However, when reading `matching.go`, the file with core mechanism in this PR, I was blocked by the custom `Is()` method, and cannot understand the purpose of inventing the type `exactMatch`.
> 
> To the best my knowledge in Go, sentinel errors are defined as package-level values, are never returned, and are just used to compare in the `errors.Is()` function. But in our case, they are not directly compared, but are wrapped by `exactMatch`, and then compared. I cannot understand this trick. `exactMatch` type seems to be a simple wrapper around `*PelicanError` without any additional data members or methods, and I'm confusing by the purpose of having it.
> 
> Due to my existing gap in Go and understanding the core mechanism of this patch, I would pass this PR to @bbockelm for further review.

So, errors.Is(err, target) calls err.Is(target) with the target untouched. PelicanError.Is matches hierarchically by default. However, the bare Specification code shouldn't be matched hierarchically, but not its FileNotCreated/FileAlreadyExists children. The only way to ask "Is" for a different kind of match is a different type of target like exact match.

I've cleaned it up by replaced Exactly() with IsExactly(err, sentinel) bool, so the marker is an internal detail with a comment explaining it, and Is has a doc comment again.

PR #3737, opened 2026-09-10
Soft Remove OA4MP issuer 

### LINKED ISSUE #3550: Switch the default built-in issuer to "embedded"
We decided to flip default for `Origin.IssuerMode` from "oa4mp" to "embedded".

Before implementing this flip, make sure the "embedded" issuer has been enabled in a few production origins and past the canary testing.

CannonLock:
Origin.IssuerMode now defaults to "embedded", and setting it to "oa4mp" is a fatal error at origin startup rather than a supported mode.

The embedded fosite-based issuer has been the intended replacement since v7.25.0; this makes it the only supported implementation. Rather than silently swapping an admin's issuer out from under them, InitServer refuses to start when "oa4mp" is set: the two implementations do not share state, and the embedded issuer's identity and discovery URLs are per-namespace (<external web url>/api/v1.0/issuer/ns<prefix>), so tokens issued before the switch stop validating, and any
Origin.Exports[].IssuerUrls pinned to the old issuer URL must be cleared or updated to match.

The check runs for the origin module regardless of Origin.EnableIssuer, so an inert "oa4mp" setting is reported rather than ignored.

The OA4MP proxy, launcher, and oa4mp package are left in place; this commit only closes the configuration path that reaches them.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2335,6 +2335,20 @@ func InitServer(ctx context.Context, currentServers server_structs.ServerType) e
 	}
 
 	if currentServers.IsEnabled(server_structs.OriginType) {
+		// The external Java-based OA4MP issuer is deprecated; the embedded issuer
+		// is now the default and only supported implementation.  Refuse to start
+		// rather than silently swap an admin's issuer out from under them: the two
+		// implementations do not share state, so registered clients and grants do
+		// not carry over.
+		if param.Origin_IssuerMode.GetString() == "oa4mp" {
+			return errors.Errorf("%s is set to \"oa4mp\", but the OA4MP issuer is deprecated and no longer "+
+				"supported; set %s to \"embedded\" to use the built-in issuer. The embedded issuer's identity "+
+				"and discovery URLs are per-namespace (<external web url>/api/v1.0/issuer/ns<prefix>): tokens "+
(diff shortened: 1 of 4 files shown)


brianaydemir (changes requested):
> Origin.IssuerMode now defaults to "embedded", and setting it to "oa4mp" is a fatal error at origin startup rather than a supported mode.

That is not what this PR does.

In its current shape, it **deletes** `Origin.IssuerMode` as a configuration parameter, causing every existing reference to `param.Origin_IssuerMode` to fail to compile — which is the cause of every action and linter failure (or close to it).

If the idea is to not muck around with all the code, keep the parameter, set it to `hidden`, and change its `default`.

---

Also, if OA4MP is intended to be unavailable, full stop, no hidden backdoor to it, we might as well update `images/Dockerfile` to stop installing it. Last I checked, the Java runtime and OA4MP is literally half of the origin container image's size.

CannonLock:
@brianaydemir Bah, thats what I get for updating code myself. 

I updated to make it as minimal change as possible. We can remove the JAVA stuff when time isn't as short in the future.

brianaydemir (approved):
LGTM.

I went ahead and removed the OA4MP install from the container images (unlike cleaning up the Go code, this is a straightforward "delete some lines from the `Dockerfile`").

PR #3742, opened 2026-09-11
Fixup the Makefile

CannonLock:
- Add web-build step as preface to the container build
- Cache more items locally to reduce container size
- Update the web container to node 26


matyasselmeci (approved):
LGTM. Thanks!

PR #3743, opened 2026-09-12
Remove support for deprecated environment variable prefixes

### LINKED ISSUE #2700: Remove OSDF env vars and deprecation notice
@matyasselmeci pointed out a poorly-worded warning:
```
Environment variables with OSDF prefix will be deprecated in the next feature release. Please use PELICAN prefix instead.
```

Rather than fix the warning message, which we've been printing for over a year, let's do what we promised to eons ago -- remove them!

brianaydemir:
Closes #2700.

## Summary

**`PELICAN_` is now the only prefix Pelican reads.**

The [`OSDF_`/`STASH_` rewriter](https://github.com/PelicanPlatform/pelican/blame/e6e6eda8e36b7b251e7ed448316e4e4ce0e40b86/config/env.go#L42) has warned that these prefixes would go away "in the next feature release" since v7.8.0. 

Variables that are no longer read are reported once per startup rather than dropped silently.

The `PELICAN_`-prefixed aliases that do not match their canonical parameter name are deliberately kept.

Binary-name prefix detection is not an environment variable and stays.


config/config.go
```diff
@@ -647,20 +638,6 @@ func SetPreferredPrefix(newPref ConfigPrefix) (oldPref ConfigPrefix, err error)
 	return oldPrefix, nil
 }
 
-// Get the list of valid prefixes for this binary.  Given there's been so
-// many renames of the project (stash -> osdf -> pelican), we allow multiple
-// prefixes when searching through environment variables.
-func GetAllPrefixes() []ConfigPrefix {
```
turetske, line 653: I'm confused, isn't this code just getting rid of the StashPrefix? Won't this still need to append the PelicanPrefix?

config/config_test.go
```diff
@@ -1198,9 +1187,11 @@ func TestDiscoverFederationImpl(t *testing.T) {
 	}
 }
 
-// TestEnvVarConfigMapping tests that environment variables are correctly mapped into
-// Pelican config structs for both client and server initialization paths.
-// This addresses issue #2819 where PELICAN_* env vars were ignored in certain modes.
+// TestEnvVarConfigMapping tests that environment variables are
+// correctly mapped into Pelican config structs for both client and server
+// initialization paths. This addresses:
+//   - Issue #2819, where PELICAN_* env vars were ignored in certain modes
```
turetske: I feel like labeling explicit issues in code comments isn't a good practice. I know Claude likes to do it, but it requires referencing a GitHub issue. Code comments should stand alone.

config/config_validation.go
```diff
@@ -56,8 +56,7 @@ func validateConfigKeys() []string {
 	envs := os.Environ()
 	for _, env := range envs {
 		parts := strings.SplitN(env, "=", 2)
-		// Until we fully deprecate OSDF and STASH prefixes, we'll check for them here
-		if strings.HasPrefix(parts[0], "PELICAN_") || strings.HasPrefix(parts[0], "OSDF_") || strings.HasPrefix(parts[0], "STASH_") {
+		if strings.HasPrefix(parts[0], "PELICAN_") {
```
turetske, line 59: And not "OSDF_"? Earlier it looks like we are still keeping the OSDF prefix, which is why I'm confused.

config/env_test.go
```diff
@@ -33,128 +33,152 @@ import (
...
+	// Under the pelican/pelican-server binary names these prefixes were never
+	// read, so there is nothing to warn about; the variable may belong to
+	// entirely unrelated software.
+	t.Run("osdf-prefixes-ignored-under-pelican-binary", func(t *testing.T) {
 		ResetConfig()
 		hook.Reset()
-		testingPreferredPrefix = OsdfPrefix
+		require.Equal(t, PelicanPrefix, GetPreferredPrefix())
+		t.Setenv("OSDF_FEDERATION_DIRECTORURL", "https://director.example.com")
+		t.Setenv("STASH_FEDERATION_DIRECTORURL", "https://director.example.com")
```
turetske, line 93: Wait, aren't we removing "STASH_" prefixes? I'm confused as to why it's in the test at all?

turetske (changes requested):
I have a few questions that need answering.

config/config.go
brianaydemir: This function was only ever used to determine which environment variables to go looking for. The answer now is always the same, regardless of the binary's name: `PELICAN_*`.

config/config_validation.go
brianaydemir: Previously, the binary name's prefix determined:

1. Which env var prefixes to go looking for (and in which order)
2. Whether to use OSDF-specific default config values.

The answer to (1) is now always `PELICAN_`.

The answer to (2) still varies.

config/env_test.go
brianaydemir: This sub-test is testing that `pelican`/`pelican-server` don't spuriously emit a warning about environment variables they never actually read (but that other binary names did).

config/config_test.go
brianaydemir: Fixed.

Although, I will point out that the original reference to #2819 started [with Patrick](https://github.com/PelicanPlatform/pelican/blame/main/config/config_test.go#L1201-L1203). 😛

brianhlin:
@biozit @williamnswanson @matyasselmeci do we have anything still using `OSDF_` or `STASH_` env vars to configure Pelican services?

matyasselmeci:
STASH? Nope. For Kubernetes deployments where we set `OSDF_*`, we also set the corresponding `PELICAN_*`. For the Condor FT plugin (named... condor_stash_plugin 😳  ), we do use and tell people to use the OSDF prefix for preferred caches because the PELICAN prefix wasn't working. In my tests, I've been setting both. We should probably explicitly mention this in the documentation.

brianaydemir:
@matyasselmeci Is it `condor_stash_plugin` or `stash_plugin`? I thought it was the latter.

matyasselmeci:
You're right, it is `stash_plugin`.

biozit:
To remove the OSDF_ vars, I first need to review all deployments to make sure nothing breaks.

config/env_test.go
```diff
@@ -33,128 +33,152 @@ import (
...
+		t.Setenv("PELICAN_DIRECTOR_URL", "https://director.example.com")
+
+		bindLegacyClientEnv()
+
+		assert.Equal(t, "https://director.example.com", viper.GetString(param.Federation_DirectorUrl.GetName()))
+	})
+
+	t.Run("osdf-director-url-is-ignored", func(t *testing.T) {
+		ResetConfig()
+		t.Setenv("OSDF_DIRECTOR_URL", "https://director.example.com")
```
turetske, line 168: The old code never read OSDF_ and STACP_ names with a pelican binary, which is what's running this test. So this test doesn't actually check the removal. This test passes in main currently as well, for example. You need to add `testingPreferredPrefix = OsdfPrefix
		t.Cleanup(func() { testingPreferredPrefix = "" })` like you do in other tests

config/env.go
```diff
@@ -30,115 +32,111 @@ import (
...
 		}
 	}
+	if len(found) == 0 {
+		return
+	}
+	warnRemovedEnvOnce.Do(func() {
+		// Sort so that the message is deterministic regardless of environment order.
+		slices.Sort(found)
+		log.Warningf("Ignoring environment variable(s) %s: Pelican no longer reads configuration "+
+			"from these variables. Please use the equivalent PELICAN_ variable instead.",
```
turetske, line 87: But aren't PELICAN_ variables now deprecated as well? Specifically later you have "log.Warningf("You are using a legacy/deprecated parameter 'PELICAN_NEAREST_CACHE' to indicate preferred caches. Please use %s instead", param.Client_PreferredCaches.GetName())"

config/env_test.go
```diff
@@ -33,128 +33,152 @@ import (
...
+		assert.Contains(t, hook.LastEntry().Message, "OSG_DISABLE_PROXY_FALLBACK")
+		assert.Contains(t, hook.LastEntry().Message, "PELICANCP_MINIMUM_DOWNLOAD_SPEED")
 	})
 
-	t.Run("pelican-env-still-works", func(t *testing.T) {
+	t.Run("removed-var-does-not-reach-viper", func(t *testing.T) {
 		ResetConfig()
 		hook.Reset()
+		// The override is load-bearing twice over: it is what makes OSDF_
+		// eligible for the warning at all, and what the old code keyed on.
```
turetske, line 120: 'old code' will be ambiguous for later readers when they can't see it.

turetske (changes requested):

config/env.go
brianhlin: > But aren't PELICAN_ variables now deprecated as well?

Uhhh, I hope not! It doesn't look like `PELICAN_` env vars were in any deprecation notice so I would consider it an unacceptable breakage.

brianaydemir: Technically, we _have_ been warning that `PELICAN_NEAREST_CACHE` has been deprecated.

However, nothing in this PR should be removing support for it, or other deprecated `PELICAN_` env vars. The removal of `OSDF_`, `STASH_`, `OSG_`, and who knows what else is probably enough.

biozit:
I removed all the OSDF_vars from the NRP deployments.

williamnswanson:
I believe we're all good for Tiger things, when we set these we're also setting `PELICAN_*`, so it shouldn't change anything for us. (🤞)

biozit:
Let's keep this in mind if any odd behavior shows up.

PR #3751, opened 2026-09-16
Change the default of `Registry.Require{Origin,Cache}Approval` to `true`

### LINKED ISSUE #3191: Registry.Require{Cache,Origin}Approva should be True by default
The defaults should be secure

melody0123:
After this change, by default, admins need to approve new origin and cache manually, in order to make directors redirect object requests to them. This is a good default value because it's secure.

Notes:
- Doc is sync'd to reflect this change.
- Misleading/Ambiguous comment in `registry_test.go` is clarified.
- `osdf_default` for these config parameters is removed from `parameters.yaml`, because the default value is the same as `osdf_default`.


brianaydemir (changes requested):
@melody0123 Flipping the defaults broke some of the unit tests. For example, regarding `broker/`:

> The PR's new comment in `registry/registry_test.go` attributes `RequireOriginApproval == false` to `viper.Reset()` clearing defaults. That reasoning doesn't hold wherever `config.InitServer` runs afterward, since it re-applies `SetParameterDefaults` — which is precisely why the broker tests now see `true`.

melody0123:
The broken tests are fixed by turning off admin approval requirement for the tests. The 2 remaining failed CI checks are irrelevant: one is because of a non-responsive OSDF director, the other is because of a missing `pelican` package in the base image. Reviewers can ignore failed CI tests and continue.

PR #3759, opened 2026-09-17
Key collection ownership and ACLs on IDs, not names

### LINKED ISSUE #3752: Renamed or deleted group/user names can be reclaimed to inherit their collection ACL grants
Summary

Collection ACLs are keyed by the name of the granted principal, stored as plain text in collection_acls.group_id with no foreign key. Both group names and user names are mutable, and neither rename path rewrites or removes the dependent ACL rows. Because group creation is open to any authenticated user, and soft-deleted usernames are released for reuse, a vacated name can be reclaimed by a different principal who then inherits every grant the original held. The embedded issuer mints wlcg.storage.* tokens from those grants, so this is an authenticated privilege escalation to object read (and write, where the original held write) on affected namespaces.

Exploit

Group stomping

1. Group alpha-writers holds read/write ACLs on one or more collections.
2. An admin renames it to alpha-editors via UpdateGroup. The ACL rows still say alpha-writers.
3. Any authenticated user POSTs a new group named alpha-writers. CreateGroup only rejects the user- prefix; the old row no longer occupies the name.
4. The creator is owner and adds themselves as a memb
5. ExpandCallerACLGroups joins group_members → groups.name, producing alpha-writers, which matches the orphaned ACL rows.
6. A token request through the OA4MP proxy calls GetU emits wlcg.storage.read (and modify/create for writeroles) on each affected collection's namespace. The hijacked name also lands in the token's wlcg.groups claim.

The auth_template_eligible bit does not help: it filters only Issuer.AuthorizationTemplates and Server.*AdminGroups matching.
Collection ACLs are explicitly a surface non-eligible

Username stomping

The personal group user-<username> is never a row in from the caller's username at check time, and ACL rowsgranting it hold the literal string.

1. User alice holds personal grants (user-alice) on collections.
2. Either an admin renames her via RenameUser, or an ser is a soft delete; the partial unique index(20260503120000) releases (username, issuer) for reuse.
3. A new account claims the username alice. For OIDC ootstrapUser takes the username from configured IdPclaims, so a user at a federated IdP who controls their preferred_username (or equivalent) can pick it. No admin action is needed
   for the claim.
4. The new account's requests synthesize user-alice and match the orphaned rows; the token path mints storage scopes as above.

Additionally, sessions open before a rename keep authorizing under the old name until the login cookie expires, since the username is
read from the cookie's sub rather than the DB.

Root cause

- collection_acls.group_id TEXT NOT NULL with no FK (ons/20250729143942_collections.sql:40). Name keying isintentional: IdP-asserted groups from the wlcg.groups claim have no local row, so ACLs must match on name.
- UpdateGroup (database/collection.go:2310) writes gr
- RenameUser (database/collection.go:1903) writes users.username (and sub for local-issuer accounts) only.
- DeleteGroup (:2733) and DeleteUser (:2767) do cleanent name, so rows orphaned by an earlier rename survive deletion too.
- CreateGroup (:2095) reserves only the user- prefix; a name still appears in collection_acls.
- The design doc (docs/user-group-design.md) chooses names as the authorization handle and makes them admin-editable, but says nothing
  about migrating grants on rename. This is an unaddrocumented trade-off.

Precursors

| Piece | Commit | Date | First tag |
|---|---|---|---|
| Name-keyed ACL table + user-<name> synthesis | ac5ef4f16, 6b90cd00a, b47c3c162 | 2025-07-29 to 2025-08-05 | v7.19.0 |
| User and group renames (no ACL rewrite) | e2ae9ea01 | 2026-01-27 | v7.25.0-experimental.0 |
| RenameUser, user soft delete | df4c9449c | 2026-04-25 | v7.27.0-rc.0 |
| Group creation opened to all authenticated users | 0fd37ceb2 | 2026-05-06 | v7.27.0-rc.0 |
| Soft-deleted usernames released for reuse | c7324863d | 2026-07-02 | v7.27.0-rc.0 |

The rename-orphans-grants bug exists in all 7.26.x re there requires admin privilege to create thereplacement group. The full unprivileged exploit (open creation, username reuse) exists only in the 7.27 release candidates.

Affected code

- database/collection.go:151 — ExpandCallerACLGroups (synthesizes user-<name>, joins group names)
- database/collection.go:901 — GrantCollectionAcl (pe
- database/collection.go:1507 — EffectiveCollectionRole (same synthesis)
- database/collection.go:1903 — RenameUser
- database/collection.go:2095 — CreateGroup (prefix check only)
- database/collection.go:2310 — UpdateGroup
- database/collection.go:2733, :2767 — DeleteGroup / DeleteUser ACL cleanup by current name
- web_ui/groups.go:1025 — admin rename handler
- oa4mp/proxy.go:353 — GetUserCollectionScopes (mints storage scopes from ACL rows)
- database/universal_migrations/20250729143942_collec
- database/universal_migrations/20260503120000_partial_user_unique_indexes.sql — username reuse

Proposed fix

Keep the name-keyed model (required for IdP-asserted groups) and make renames carry their grants:

1. In UpdateGroup, when name changes, inside the existing transaction:
   UPDATE collection_acls SET group_id = :new WHERE g
2. In RenameUser, same with 'user-' || :old → 'user-' || :new.
3. Handle the PK collision where both names already hame collection (upsert / delete-then-insert), mirroring the ON CONFLICT clause in GrantCollectionAcl.
4. Defense in depth: have CreateGroup refuse a name tction_acls.group_id, and have tryCreateUser /RenameUser refuse a username whose user-<name> still appears there. This closes inheritance regardless of how an orphan arose
   (including pre-fix data).
5. Optionally, a one-time cleanup migration that deletes ACL rows whose group_id matches neither a live group name, a live
   user-<username>, nor @authenticated — but only if  expected to persist as ACL targets without a localrow. If they are, skip this step.
6. Add a line to docs/user-group-design.md: renaming ll name-keyed grants.

Severity

Moderate. Authenticated privilege escalation to data-namespace where a vacated name held a grant. Requires aprior admin rename (group case) or admin rename/delete (user case); the reclaim step itself needs no privilege in 7.27-rc. Recommend fixing before 7.27.0 final.

### LINKED ISSUE #3753: Username Fallback enables extended access to transferred namespace
CallerIsCollectionOwner at database/collection.go:1604 grants full owner authority when the caller's username equals collection.Owner. The same fallback appears in the role resolver at lines 1476 and 1547, and the listing queries at lines 748, 756, 1184 and 1192 use owner = ?. 

Three consequences:                                                                                                        
                                                                                                                                        
  - PATCH ownership transfer leaves the old owner in control. UpdateCollection writes owner_id only. The previous owner's username stays
    in owner, so they still pass the owner check and can delete or re-transfer the collection. The invite-redemption path at line 3033  
    does update both columns, so only the direct PATCH is affected.                                                                     
  - Deleted users' collections are inheritable. DeleteUser soft-deletes and does not touch collections.owner. The partial unique index  
    from 20260503120000 frees the username for reuse. A new account that claims the name becomes owner of every collection the old      
    account left behind.                                                                                                                
  - Renamed users leave a claimable owner string. RenameUser does not update owner. The renamed user keeps access via owner_id, but     
    whoever next holds the old username also gets owner authority.                                                                      
                                                                                                                                        
  Fix: drop the username fallback from the three owner checks and the listing queries, or have RenameUser, DeleteUser and               
  UpdateCollection keep the column in sync. Removing the fallback is cleaner and matches what the migration comment already claims.

bbockelm:
Users and groups each carry two handles: an immutable ID, and a name that is both **mutable** (admins rename) and **reusable** (a soft-deleted user releases its username). The collections tables mixed the two, so authority followed the name rather than the principal.

Fixes #3753
Fixes #3752

## The two reported bugs

`collections.owner` held a username and was consulted as an ownership fallback beside `owner_id`. A PATCH transfer wrote only `owner_id`, so the previous owner kept full control — including transferring it back — and a renamed or deleted account left a claimable username that the next holder inherited.

`collection_acls.group_id` held a group name, the synthesized `user-<username>`, or the `@authenticated` sentinel, with no foreign key. Renaming a group or user orphaned every grant under the vacated name, and since group creation is open to any authenticated user, that name could be reclaimed to inherit the grants.

## What changed

**Everything stored for an authorization decision is an ID.** `collections.owner` is dropped; `owner_id` is the sole ownership handle. `collection_acls` is re-keyed to `(subject_type, subject_id)` — a `users.id`, a `groups.id`, or empty for the all-authenticated sentinel. `granted_by` and `added_by` move from usernames to user IDs. `user-<username>` and `@authenticated` survive as API *presentation* forms; they never reach a column. Granting to a name this server has no record of is now an error rather than a stored string.

**Group names had two unreconciled sources** — records Pelican created, and whatever the provider named by `Issuer.GroupSource` asserted. Pelican now records a group the first time a provider asserts its name, owned by the built-in admin and stamped with that provider in a new `groups.source` column (`pelican` / `oidc` / `file` / `github`, plus `unknown` for migration-backfilled records). That gives the name a stable ID for ACLs to key on, and holds it against an unprivileged user creating a group that shadows it. Opt out with `Issuer.DisableGroupAutoCreation`.

**Which identifier space a value belongs to is carried by its type**, never inferred from how it looks. `ACLSubjectRef` is the name space (a group name, `user-<username>`, `@authenticated`); `ACLSubject` is the ID space (a kind plus an ID); over HTTP the same split is `groupId` versus `subjectType` + `subjectId`, chosen by which fields the caller populates. Neither resolver consults the other's table.

A first draft of this decided by the *shape* of the string, which was wrong: shape is exactly what an attacker controls, and group creation is open to any authenticated user, so a group **named** after another principal's **ID** was enough to intercept grants addressed to that ID. A guess at a security boundary is a vulnerability waiting for the input that fools it. `ValidateIdentifier` still refuses locally-created ID-shaped names, demoted to what it actually is — hygiene, with nothing depending on it.

One wire consequence: `groupId` no longer accepts a group slug, and a bare user ID is not one of its spellings. That is the price of having no guess.

**Membership is mirrored too.** Recording that a group exists is not enough: membership still lives at the provider, which works for the caller in front of us and fails for every decision made *about* someone who isn't — the share-token clamp asks what a share's **owner** can do on the parent collection at mint time with no session for them, and `IsSystemAdminUserID` asks whether a target account is an admin. Mirrored memberships land in `group_members` alongside Pelican-created ones, distinguished by `source`.

A mirrored membership is a **cached authorization fact**, which is the same shape as the bug this PR removes. One rule keeps it honest:

> **Freshness gates granting, not existence.**

A mirrored row grants only while the provider asserted it within `Issuer.AssertedGroupMembershipTTL` (7d). Past that it is *kept*, not deleted, because a check asking whether an account might **hold** a privilege must keep seeing it — for `IsSystemAdminUserID`, "we last saw this account in an admin group a month ago" has to mean *refuse*, and treating a stale copy as absence is exactly what opens that guard. Granting paths filter on freshness; restricting ones deliberately don't. The caller's own live assertion is never gated by the TTL. Rows go away only when the provider retracts.

`file` is the only source Pelican can refresh without the user present, so `Issuer.GroupFileRefreshInterval` re-reads it and reconciles every known account; `oidc`/`github` refresh at login, which is what the TTL bounds.
(Updated by Howard): an admin who has not signed in since the upgrade has no rows in the database at all. 

**"We have not checked" is now a state, not a silence.** `Server.AdminGroups` confers `server.admin` by group *name*, so when membership comes from a provider there may be nothing on this server to evaluate — and a guard that asks "is this account an administrator?" and answers *no* when it cannot tell opens on exactly the accounts it exists to protect. `users.group_admin_status` latches the answer (`unknown` → refuse, `possible` → refuse and never downgrade, `ruled-out` → allow), so the question becomes "have we proved it is **not** one". The predicate is a pair, because which direction is safe depends on what a true answer does: `MustTreatAsSystemAdmin` refuses on uncertainty, `IsConfirmedSystemAdmin` declines on it — `transfer.registerOAuthClient` needs the latter, since a true answer there marks a client *shared*.

**Two more reusable handles**, found by reviewing the surrounding code for the same problem class:

- `api_keys.created_by` was read as a user ID everywhere (the revocation contract re-intersects a key's scopes against that user's *current* effective scopes) but written as a username. A released username revived a dead key's dormant management scopes for whoever still held the secret; a rename bricked every key the user minted.
- `Group.ID` could come back: `DeleteGroup` hard-deleted and `generateSlug` is 32 bits with no uniqueness check, so a later group could be minted with a dead one's ID and inherit whatever still referenced it. Groups now soft-delete like users — the ID is spent permanently, the *name* is released. `DeleteGroup` also clears what it used to leave dangling: group scopes, and any collection or group naming it as administrator.

## Are the tests load-bearing?

These are pinning tests, not smoke tests. Every property was ported to the pre-fix API and run on the parent commit, where all six fail and reproduce the reported behavior (previous owner still passes the owner check and can delete; a reclaimed username owns the deleted account's collections; a renamed user *loses* their grants while the next holder of the name *gains* them; likewise for groups; an unresolvable ACL target is stored verbatim; a grant survives its group's deletion). The mapping is in the header of `database/collection_id_keyed_authz_test.go`.

One trap worth knowing about: the test helpers built their schema from GORM struct tags, which can only emit **full** unique indexes — so a soft-deleted row kept reserving its name and the reuse paths were unreachable. Any test of them passed vacuously. `ApplyPartialIndexesForTests` now recreates the production partial shapes, and both the `database` and `web_ui` helpers call it.

## Known gaps, stated plainly

- **A `possible` admin latch is never cleared.** An account observed holding an administrative group stays un-manageable by a `server.user_admin` even if the provider later retracts the membership, and even if `Server.AdminGroups` is later reconfigured. Deliberate — the recourse is a full `server.admin`, who bypasses these guards — but it is a one-way door and worth knowing about.
- **After upgrade, every account is `unknown`**, so a `server.user_admin` can act on none of them until each has signed in once. Scoped: where `Server.AdminGroups` is unset, no group can confer admin and nothing changes.
- **A restart does not reset the TTL clock.** An outage longer than the TTL leaves every mirrored membership stale; `file` self-heals via the startup refresh pass, `oidc`/`github` recover at each user's next login. Fails closed, so it is silent breakage rather than exposure.
- **`LookupOrBootstrapUser` matches only `users.sub/issuer`, never `user_identities`**, so one human can end up with two accounts. Deliberately untouched — a separate PR covers that area.

## Reviewing

Eight commits, each self-contained and readable in order. The migrations (`20260916120000`, `20260917120000`, `20260918120000`, `20260919120000`) carry the reasoning for each data change inline, and `docs/user-group-design.md` has the resulting contract under "Identifiers in the database", "Group sources", "Mirrored memberships", and "Establishing that an account is *not* an administrator".

Two judgment calls I would especially like a second opinion on: asserted groups are created **auth-template eligible** (a name with no record already matched templates, so creating one with the bit clear would silently revoke authority), and `EnsureAssertedGroups` deliberately skips `ValidateIdentifier` so WLCG-style `/cms/production` gets a record — those are exactly the groups an operator wants to ACL.

Updated by Howard:
image
A diagram for the DB schema changes in this PR


h2zh:
Could you take a look at the linter failure and the [Windows test failure](https://github.com/PelicanPlatform/pelican/actions/runs/35229138136/job/105228310687?pr=3759)? Other two failures seem irrelevant to this PR.

database/universal_migrations/20260916120000_id_keyed_collection_authz.sql
```diff
@@ -0,0 +1,235 @@
+-- +goose Up
+-- +goose StatementBegin
+
+-- Re-key every collection authorization decision onto immutable IDs.
+--
+-- Background (issues #3752, #3753). Two collections tables mixed
```
h2zh, line 6: One non-blocking thought on the four migration files: the comments are fairly long, and a good chunk of it restates things that already live in `docs/user-group-design.md`, the PR description, and the Go doc comments (the freshness rule and the latch semantics in particular).

Would it make sense to keep the migrations to a short "what this column holds and why this step is destructive/irreversible" note per step, and move the longer reasoning into the design doc / commit messages? My worry is mostly drift: multiple copies of the same rule tend to disagree eventually, and the design doc is the one people usually read.

web_ui/group_file_refresh.go
```diff
@@ -0,0 +1,127 @@
...
+// — which is exactly what Issuer.AssertedGroupMembershipTTL exists to
+// bound. Refreshing the file source keeps it out of that bind
+// altogether: its memberships stay continuously fresh, and an operator
+// removing someone from the file sees it take effect within one
+// interval rather than at that person's next login.
+//
+// A pass is a no-op when the file is unset, and errors are logged
+// rather than returned — a malformed group file must not take the
+// server down, and the next pass will pick up a corrected one.
+func LaunchPeriodicGroupFileRefresh(ctx context.Context, egrp *errgroup.Group) {
```
h2zh, line 49: Since the group file is treated as an always-on second source that layers onto whatever the primary source (oidc/github) is, we need to update its config parameter docs in `parameter.yaml` too.

h2zh: For an OIDC account, the group file usually has no entry. The refresher then asserts an empty list for that account, the intersection with `Server.AdminGroups` is empty, and `RecordGroupAdminObservation(false)` moves the latch from `unknown` to `ruled-out`. The account's real groups, which live at the IdP, were never consulted. I verified this bug with a CILogon account on `GroupSource: oidc` plus a group file.

web_ui/authentication.go
```diff
@@ -1080,6 +1195,12 @@ func loginHandler(ctx *gin.Context) {
 		log.Errorf("Failed to generate group info for user %s: %s", userRecord.Username, err)
 		groups = nil
```
h2zh, line 1196: With the new invocation of `RecordAssertedGroups()` below, where `MirrorAssertedGroupMemberships()` is called, setting groups to nil is a dangerous move.
This is because `MirrorAssertedGroupMemberships()` treats an empty list as a retraction:  with `assertedIDs` empty, the DELETE drops every `group_members` row for this user with
`source = 'file'`. The function cannot distinguish "the file lists no groups for this user" from "the file could not be read", because by the time it runs both arrive as an empty slice. Aa the result, a transient read error wipes the account's mirrored memberships.

database/universal_migrations/20260917120000_id_keyed_api_keys_and_group_soft_delete.sql
```diff
@@ -0,0 +1,118 @@
...
+     admin_type, updated_at, display_name, creator_auth_method,
+     creator_auth_method_id, created_for_collection_id,
+     auth_template_eligible, source)
+SELECT id, name, description, created_by, created_at, owner_id, admin_id,
+       admin_type, updated_at, display_name, creator_auth_method,
+       creator_auth_method_id, created_for_collection_id,
+       auth_template_eligible, source
+FROM groups;
+
+DROP TABLE groups;
```
h2zh, line 101: Dropping groups while group_members still declares `FOREIGN KEY … ON DELETE CASCADE` will delete everything in `group_members`.

database/collection.go
```diff
@@ -2138,6 +2600,415 @@ func CreateGroup(db *gorm.DB, name, displayName, description string, creator Cre
...
+		for _, groupID := range assertedIDs {
+			member := GroupMember{
+				GroupID:    groupID,
+				UserID:     userID,
+				AddedBy:    creatorOrUnknown(""),
+				Source:     source,
+				AssertedAt: &now,
+			}
+			res := tx.Model(&GroupMember{}).
+				Where("group_id = ? AND user_id = ? AND source <> ?", groupID, userID, GroupSourcePelican).
```
h2zh, line 2993: `source <> 'pelican'` lets any asserted provider claim a row another provider
wrote: the UPDATE re-stamps `source`, so whichever provider reconciled last owns
the membership. Since password logins record under `file` even on an `oidc`
server (authentication.go:1203, :1291), `file` and `oidc` do coexist here, and
the retraction above (`source = ?`) then cuts across them: a `file` retraction
deletes a membership the IdP still asserts, and a row re-stamped `file` is
invisible to the `oidc` retraction, so a revoked IdP membership survives until
the TTL. Suggest `source = ?` (this provider's own rows) so refresh and
retraction agree on scope. Note the PK `(group_id, user_id)` then means a second
provider's assertion silently no-ops on the DoNothing insert; if two providers
really are meant to coexist, the key needs `source` in it — otherwise gating the
group file on `GroupSource: file` removes the second provider entirely.

database/identifiers.go
```diff
@@ -70,6 +92,11 @@ func ValidateIdentifier(name string) error {
 	if strings.Contains(name, "..") {
 		return ErrInvalidIdentifier
 	}
+	// Keep locally-created names visually distinct from IDs. Hygiene
+	// only — see slugShapePattern for why nothing may depend on it.
+	if slugShapePattern.MatchString(name) {
```
h2zh, line 97: This shape check reaches usernames the IdP asserts, not just names people pick.

`SanitizeIdentifier` ends by calling `ValidateIdentifier` and returns `""` on
failure (identifiers.go:187), and `LookupOrBootstrapUser` drops every candidate
that sanitizes to `""` (collection.go:2439). So an IdP that hands out an 8-hex
handle as `preferred_username` loses every candidate, and the account is created
under the synthesized `user-<4 hex>` fallback (collection.go:2481). From there:

- `Server.UIAdminUsers` / `UserAdminUsers` / `CollectionAdminUsers` match on the
  exact username, so an entry for the real handle grants nothing, silently.
- An admin cannot fix it: `RenameUser` to the real value hits this same check.

That contradicts the design doc's own carve-out ("a name a provider asserts is
exempt"), which `EnsureAssertedGroups` honors for groups but nothing honors for
users.

Suggest moving the 8-hex rule out of `ValidateIdentifier` and applying it only
where a person chooses the name: `CreateGroup` and the admin rename path.
`SanitizeIdentifier` / bootstrap should accept whatever the provider says. Since
the comment above already says nothing may depend on this rule, narrowing it
costs nothing.

database/collection.go
```diff
@@ -2138,6 +2600,415 @@ func CreateGroup(db *gorm.DB, name, displayName, description string, creator Cre
...
+	// Resolve the asserted names to the group records
+	// EnsureAssertedGroups keeps. A name with no record contributes
+	// nothing — there is no ID to key a membership on, which is the same
+	// reason an ACL cannot name it either.
+	assertedIDs := []string{}
+	if len(assertedNames) > 0 {
+		var rows []struct{ ID string }
+		if err := db.Table("groups").
+			Select("id").
+			Where("name IN ? AND deleted_at IS NULL", assertedNames).
```
h2zh, line 2955: This lookup has no `source` filter, so a name `EnsureAssertedGroups` just
declined (a `pelican` group already holds it, :2757) still resolves here, and
the assertion lands a provider-sourced membership in the admin's local group.
That row carries the group's scopes, admin authority and ACL roles, and nobody
can remove it: `RemoveGroupMember` / `LeaveGroup` return `ErrMembershipNotLocal`
with no admin bypass, and "remove it at the provider" is not an option for a
group Pelican owns.

Suggested Fix: add `AND source <> 'pelican'` here, or better, have 
`EnsureAssertedGroups` return the IDs it accepted and mirror only those, so the
two functions can't disagree about which names count.

web_ui/authentication.go
```diff
@@ -811,22 +811,137 @@ func CheckCollectionAdmin(identity UserIdentity) (bool, string) {
...
+func IsConfirmedSystemAdmin(db *gorm.DB, userID string) bool {
 	user, err := database.GetUserByID(db, userID)
 	if err != nil {
 		return false
 	}
-	identity := UserIdentity{
+	// Errors are logged rather than folded into the verdict: the
+	// config-derived username paths still answer, which is the
+	// pre-mirroring behavior.
+	groups, err := database.GroupNamesForRestrictionCheck(db, userID)
```
h2zh, line 914: This is a granting decision that reads memberships without the freshness cutoff. `IsConfirmedSystemAdmin` decides that an OAuth client is shared, but reads `GroupNamesForRestrictionCheck`, which by design includes rows past `Issuer.AssertedGroupMembershipTTL`. `isGroupOwnerOrAdmin` counts group_members for `admin_id` with no `asserted_at` filter, so a stale mirrored row still confers group-admin authority.
Suggested Fix: route both through `grantingMembershipsFor`.

database/collection.go
```diff
@@ -125,72 +175,322 @@ func FilterAuthTemplateEligibleGroups(db *gorm.DB, names []string) []string {
...
+			for _, id := range ids {
+				add(id)
+			}
+		}
+	}
+	if len(groupNames) > 0 {
+		var rows []struct{ ID string }
+		if err := db.Table("groups").
+			Select("id").
+			Where("name IN ?", groupNames).
```
h2zh, line 313: Need to filter by `deleted_at` too. Same for line 157  (the query in FilterAuthTemplateEligibleGroups)

database/universal_migrations/20260917120000_id_keyed_api_keys_and_group_soft_delete.sql
```diff
@@ -0,0 +1,118 @@
...
+-- legacy username could itself be shaped like one. A value that matches
+-- neither becomes the empty string, which intersectWithUserScopes
+-- already treats as "cannot attribute this key to any current
+-- authority" and fails closed on every user-grantable scope. Data-plane
+-- and inter-server scopes are unaffected either way: those are bearer
+-- authority, not derived from a creator's role.
+UPDATE api_keys
+SET created_by = COALESCE(
+    (SELECT u.id FROM users u WHERE u.id = api_keys.created_by),
+    (SELECT u.id FROM users u WHERE u.username = api_keys.created_by AND u.deleted_at IS NULL),
```
h2zh, line 37: If a key's creator has been deleted and their username has since been taken by someone else, the key is rebound to the new account with the same name, which is precisely the "revival" this migration's own comment claims to close.

web_ui/group_file_refresh.go
```diff
@@ -0,0 +1,127 @@
...
+			log.Warnf("Failed to list users for the periodic group-file refresh: %v", err)
+		}
+		return
+	}
+	refreshed := 0
+	for _, user := range users {
+		if ctx.Err() != nil {
+			return
+		}
+		groups, err := generateGroupInfo(user.Username)
```
h2zh, line 112: Maybe this line should be moved out of the loop?

oa4mp/proxy.go
```diff
@@ -499,7 +508,7 @@ func GetUserCollectionScopes(db *gorm.DB, user, userID string, groupsList []stri
 				// Parent gone; emit no storage scopes for this share.
 				continue
 			}
-			ownerRole := database.EffectiveCollectionRole(db, parent, perm.ownerID, perm.ownerUsername)
+			ownerRole := database.EffectiveCollectionRole(db, parent, perm.ownerID, "")
```
h2zh, line 511: This clamp only counts memberships asserted within the last TTL. So if a share's owner has access to the parent collection only through a provider group, and has not logged in for 7 days, the clamp sees no role for them at all. Every share they created then gets tokens with no storage scopes, with no error logged or returned. It fixes itself the next time the owner logs in — but the provider still lists them as a member the whole time, so this is a false denial.

cmd/origin_collections.go
```diff
@@ -917,11 +939,11 @@ func init() {
 	originCollectionsUpdateCmd.Flags().String("description", "", "New collection description")
 	originCollectionsUpdateCmd.Flags().String("visibility", "", "New collection visibility (private|public)")
 
-	originCollectionsACLGrantCmd.Flags().String("group-id", "", "Group ID for the ACL (required)")
+	originCollectionsACLGrantCmd.Flags().String("group-id", "", "The principal to grant to: a group name or ID, user-<username>, or @authenticated (required)")
```
h2zh, line 942: We should check all comment/text related to `group-id` to remove the group name narrative. 

Other places I found

1. same file :946, 
2. swagger/pelican-swagger.yaml GrantACLRequest.groupId description (~:1007), 
3. web_ui/frontend/helpers/api/Collection/types.ts:101, 
4. the regenerated docs/app/commands-reference/pelican-server/origin/collection/acl/grant/page.mdx:17

web_ui/oauth2_client.go
```diff
@@ -504,6 +599,18 @@ func generateUserGroupInfo(userInfo map[string]interface{}, idToken map[string]i
...
+	// user's memberships in them. `internal` is reading these rows in
+	// the first place, so it has nothing to record.
+	switch groupSource {
+	case GroupSourceTypeOIDC:
+		RecordAssertedGroups(database.GroupSourceOIDC, userRecord.ID, username, groups)
+	case GroupSourceTypeFile:
+		RecordAssertedGroups(database.GroupSourceFile, userRecord.ID, username, groups)
+	case GroupSourceTypeGitHub:
+		RecordAssertedGroups(database.GroupSourceGitHub, userRecord.ID, username, groups)
+	}
```
h2zh, line 612: Could you add a default clause here so that in the future when another developer adds a new OIDC provider here, they are reminded to handle this place too.

web_ui/oauth2_client.go
```diff
@@ -234,7 +234,102 @@ func pickDisplayName(claims map[string]interface{}, fallback string) string {
...
+	eligible := database.FilterAuthTemplateEligibleGroups(database.ServerDatabase, groups)
+	for _, want := range configured {
+		for _, got := range eligible {
+			if want == got {
+				return true
+			}
+		}
+	}
+	return false
```
h2zh, line 321: Is it possible to simply this func by reusing `EffectiveScopesForIdentity`:
```suggestion
    // Same evaluator the session and API-key paths use. No ID and no
    // username, so only group-derived scopes can come back.
    scopes := EffectiveScopesForIdentity(UserIdentity{Groups: groups})
    return slices.Contains(scopes, token_scopes.Server_Admin)

```

docs/parameters.yaml
```diff
@@ -4888,6 +4888,89 @@ type: bool
 default: false
 components: ["origin"]
 ---
+name: Issuer.DisableGroupAutoCreation
+description: |+
+  Disable recording a group for every group name observed from the
+  provider named by `Issuer.GroupSource` (the OIDC group claim, the
+  group file, or GitHub organizations).
```
h2zh, line 4895: I think it's worth mentioning that this gate also stops mirroring and freezing every mirrored row.
Nitpicking: Maybe we don't need to iterate all options for `Issuer.GroupSource` here in a seperate config param, since there might be more options in the future.

web_ui/authentication.go
```diff
@@ -811,22 +811,137 @@ func CheckCollectionAdmin(identity UserIdentity) (bool, string) {
 	return false, "You don't have collection administrator permission"
 }
 
-// IsSystemAdminUserID checks whether the given user ID belongs to a system admin.
-// This is used to prevent user administrators from modifying system admin accounts.
-func IsSystemAdminUserID(db *gorm.DB, userID string) bool {
```
h2zh, line 816: This PR splits `IsSystemAdminUserID` into two new functions, but when I grep `IsSystemAdminUserID` in the codebase, there are still multiple matches. Please revise those occurances to avoid future confusion.

database/collection.go
```diff
@@ -923,42 +1367,56 @@ func GrantCollectionAcl(db *gorm.DB, id, user, userID string, groups []string, g
...
+	collection := &Collection{}
+	if result := db.Preload("ACLs").Where("id = ?", id).First(collection); result.Error != nil {
+		return result.Error
+	}
+
+	if !isAdmin {
+		if err := validateACL(db, collection, user, userID, groups, token_scopes.Collection_Delete); err != nil {
+			return err
+		}
+	}
```
h2zh, line 1391: This block repeats multiple times in the adjacent functions. Maybe it could be consolidate as a helper function

h2zh (changes requested):
It would be nice to have a “note for admin” to tell them what data is modified in the DB in the PR description, and eventually in the release notes. 

Tested this PR live on a four-process rig with password and CILogon users. The ID-keyed ownership and ACL model works as designed: grants survive renames, vacated names inherit nothing, transfers strip the old owner, deleted groups stay dead. See the following comments for the bug (or behavior) I found.

Btw, thanks @brianaydemir for giving me the SWAMP ™️ output for this PR!

cmd/origin_collections.go
bbockelm: Ok, fixed in the latest set of commits.  Scrubbed all I could find.

database/universal_migrations/20260917120000_group_soft_delete.sql
```diff
@@ -0,0 +1,118 @@
...
+     admin_type, updated_at, display_name, creator_auth_method,
+     creator_auth_method_id, created_for_collection_id,
+     auth_template_eligible, source)
+SELECT id, name, description, created_by, created_at, owner_id, admin_id,
+       admin_type, updated_at, display_name, creator_auth_method,
+       creator_auth_method_id, created_for_collection_id,
+       auth_template_eligible, source
+FROM groups;
+
+DROP TABLE groups;
```
bbockelm: Fixed in the latest.

database/collection.go
bbockelm: Ok - I think I got this factored out.

database/collection.go
bbockelm: Switched it so there's only one configured non-pelican source at a time (file and oidc can't be active at the same time).

I would eventually like to do better source tracking but, for now, I'm OK with saying the non-pelican source can "grab" the rows.

oa4mp/proxy.go
bbockelm: I think this is still a sticky point in the current PR -- but, for now, we allow stale membership only for shares (otherwise, mysterious failures would occur after just a few days).

A real solution would probably involve offline access / refresh... but the PR is already complex/long enough I don't want to add this on.

web_ui/authentication.go
bbockelm: Done!

web_ui/authentication.go
bbockelm: I closed out the case with `file`.  The case with `oidc` is purposeful: an empty (or missing) group list is considered equivalent to a statement of "user is a member of no groups"

web_ui/group_file_refresh.go
bbockelm: Did some large-ish refactors to make sure that only one source is active at a time.  I'd eventually like to have a setup that allows multiple sources -- but that is going to have to be left for a future follow-up.

web_ui/oauth2_client.go
bbockelm: Done in the latest commit!

database/collection.go
```diff
@@ -2694,7 +3851,18 @@ func LeaveGroup(db *gorm.DB, groupID, userID string) error {
 	if group.OwnerID == userID {
 		return ErrForbidden
 	}
-	res := db.Where("group_id = ? AND user_id = ?", groupID, userID).Delete(&GroupMember{})
+	// A mirrored membership is not ours to remove: the provider decides
```
h2zh, line 3854: Switching `Issuer.GroupSource` away from `file` (to `oidc` or `github`) can leave mirrored memberships that
  nothing will ever retract. Retraction runs only when the account itself logs in through the new source, so
  a password-only account keeps every `file` row forever.

  If that row is in a group that confers admin, the account is locked:

  - `MustTreatAsSystemAdmin` still sees the row and refuses a user-admin (`403`, "the target account holds
  administrator privileges");
  - the `possible` latch cannot be cleared (`groupAdminRuledOut` → `409`);
  - the row cannot be removed, because `RemoveGroupMember` treats it as provider-owned
  (`ErrMembershipNotLocal` → `409`);
  - the group cannot be deleted while `Server.*AdminGroups` names it (`409`).

Taking user `bob` as example: mirrored into `ops` under `GroupSource: file`, then the origin
  restarted with `GroupSource: oidc`. His `ops:file` row, his `possible` latch, and the `403` all
  persisted, and editing the group file no longer had any effect (the refresher is disabled when the file is
  not the configured source).

web_ui/oauth2_client.go
```diff
@@ -234,24 +227,182 @@ func pickDisplayName(claims map[string]interface{}, fallback string) string {
...
+// user, on a login path that has no token to read them from — the
+// password and init-code handlers.
+//
+// The second return says whether the configured source was actually
+// consulted. It is not the same as "groups is empty": a file with no
+// entry for a user is an authoritative statement that they are in no
+// groups, whereas an `oidc` server reached through a password login has
+// simply not been asked. Only the first may be recorded, which is what
+// keeps a non-authoritative path from ruling an account out.
+func GroupsForLogin(user *database.User) (groups []string, consulted bool, err error) {
```
h2zh, line 240: Let's say `charlie`: under `Issuer.GroupSource: file` the group file put him in `ops`, a Pelican-created group that carries `server.collection_admin`. He was never added to `ops` membership row in local DB. After the origin restarted with`GroupSource: oidc`, his next password login no longer read the group file, so he lost the collection-admin scope he had held through the asserted `ops` name. Had `ops` also been what made him an admin (listed in `Server.AdminGroups`, or granted `server.admin`), he would have lost `server.admin` the same way; authority tied to his username, such as `Server.UserAdminUsers`, is unaffected. This is a compatibility break against `main`, where the password-login handlers read the group file whenever `Issuer.GroupFile` was set, regardless of `GroupSource`. No mirrored row is involved: because `ops` is a Pelican-created group, the file's assertion was never mirrored into it.

database/universal_migrations/20260917120000_group_soft_delete.sql
```diff
@@ -0,0 +1,123 @@
...
+-- to release it would leave a name permanently unusable after a
+-- cleanup. That requires rebuilding the table, because the inline
+-- `name TEXT NOT NULL UNIQUE` from 20250729143942 is backed by an
+-- implicit sqlite_autoindex that DROP INDEX cannot touch. The column
+-- list below is the state after 20260916120000.
+--
+-- foreign_keys=OFF for the duration: see the NO TRANSACTION note at the
+-- top of this file. Without it the DROP TABLE below cascades through
+-- group_members.group_id and deletes every membership.
+PRAGMA foreign_keys = OFF;
```
h2zh, line 40: `NO TRANSACTION` is right for the rebuild, but it now also covers the `api_keys` UPDATE at :38, and a
  failure inside the rebuild leaves `groups_new` behind. I reproduced it by pre-creating `groups_new` on a
  seeded pre-PR DB: the block failed with `table groups_new already exists`, yet `api_keys.created_by` was
  already rewritten and goose stayed at `20260916120000`; a second start failed identically until I dropped
  the table by hand. Two small changes: move the `api_keys` step into its own transactional migration
  (nothing has shipped, renumbering is free), and start the rebuild with `DROP TABLE IF EXISTS groups_new;`.
  Also worth an assertion in the cascade test that `PRAGMA foreign_keys` reads `1` afterwards, since :113
  turns it back on by hand and that is not observable from outside the process. (database/id_keyed_authz_migration_test.go:290)

database/collection.go
```diff
@@ -397,6 +810,104 @@ const (
...
+	// Not a value this build knows. Almost always a typo in the
+	// configuration, but it is also what a new provider looks like if
+	// someone adds a GroupSourceType constant and forgets the table
+	// above. Say so rather than returning "" silently: "" means NO
+	// provider decides group membership here, which switches off
+	// mirroring, the group-file refresher, and the guard that stops a
+	// user-administrator acting on an unobserved account.
+	log.Warnf("Issuer.GroupSource is set to %q, which this server does not recognize; "+
+		"no provider will decide group membership", raw)
+	return ""
```
h2zh, line 899: Misconfiguration can start cleanly and are only noticed later here, which may lead to many unexpected behaviors. Can misconfigurations on `Issuer.GroupSource` become startup errors in `config/config.go` for every server type?

h2zh (changes requested):
Thanks for addressing the comments above! In this round of re-review, I checked the diff and ran manual tests. The first two comments below share a cause: the group source is now single-valued, and switching it (`file` → `oidc`) has consequences the PR doesn't yet account for. Do you think we should fix them in this PR, or just have some kind of stopgap?

  1. Password logins stop reading the group file, so accounts whose groups or admin authority came from it silently lose them at the next login. That is a behavior change from `main`.
  2. Memberships mirrored under the previous source are never retracted for accounts that don't log in through the new one, and there is no admin path to remove them; if the group confers admin, the account becomes permanently untouchable by user-admins.

database/collection.go
bbockelm: Fixed in 04ae47fbd88efd930918bbb87752af5c05eba3ba

database/universal_migrations/20260917120000_group_soft_delete.sql
bbockelm: Good catch; broken out in 04ae47fbd88efd930918bbb87752af5c05eba3ba.

web_ui/oauth2_client.go
bbockelm: Ok - in b6c105a6bf5d1f763f533b1bf7c9915119941df7, I added a background sweep for non-default GroupSource rows.

I considered just removing them all on startup but I was a bit reluctant to dump information, especially if a sysadmin maybe typo'd or was experimenting with things.

Now, stale rows for non-default group sources are flat-out removed.  This gets it into the eventually-clean state.

database/collection.go
bbockelm: Same as the below comment -- we now periodically refresh so this would age out.

There's some follow-up work to do here around group info refresh - but I want to avoid getting this PR any larger.

h2zh (approved):
Approving with known non-blocking behaviors/limitations:

1. Password logins on an `oidc`/`github` server no longer read `Issuer.GroupFile`. Accounts whose groups or admin authority came only from the file lose them at their next login (before this PR, the file was read regardless of `GroupSource`). Needs a release note; sites with local admin accounts in the file should move them to `Server.UIAdminUsers` or a group in local DB before upgrading.
 2. The orphan sweep frees the row in the group_members table, not the account in the users table. After a `GroupSource` switch, a password-only account (whose stale membership in an admin-conferring group is swept) keeps its `possible` latch. User-admins cannot manage it, but full admin can. Suggested follow-up: clear `possible` → `unknown` when the sweep removes the last admin-conferring row.

